PR #7, opened 2018-08-06
BCDA-22: User model, token auth

rnagle:
### Relates to [BCDA-22](https://jira.cms.gov/browse/BCDA-22)

Adds SQL for a user model to `db/api.sql`. Some details:

- A user relates to one ACO.
- An ACO can have many users.
- A job is associated with a user (for audit trail)

### Proposed changes

- Both `users` and `acos` tables now use a `UUID` as primary key
- Adds `make test` command to run linting and unit tests
- Adds `db/fixtures.sql` with some dummy data for local testing
- `Dockerfile.bcda` updated to generate RSA key pair to sign JWTs for local testing
- Adds `bcda/auth` module with:
    - Backend functions to facilitate JWT generation and 
    - Middleware to enforce JWT auth on on endpoints (currently: `/claims` and `/job/{jobId}`)
- All model files regenerated
- Dependencies for JWT auth saved to `vendor` dir

### Security Implications

The generated JWTs carry two pieces of information:

1. The subject (user) associated with the token and
2. The ACO the user belongs to

JWTs are signed using an RSA key pair. This signature is how we determine the validity of a token.

As a convenience, this PR adds a `/api/v1/token` endpoint when env var `DEBUG=true` (this is set in `docker-compose.yml`). Use the endpoint to retrieve a valid token for debugging other endpoints requiring auth.

### TODO

- [x] In order to revoke/invalidate a token, we must maintain a blacklist via db table (and/or a key,val store like redis)
- [x] The TTL for JWTs should be configurable, set to some sane default
- [x] Tests


bcda/router.go
```diff
@@ -17,8 +19,12 @@ func NewRouter() http.Handler {
 		}
 	})
 	r.Route("/api/v1", func(r chi.Router) {
-		r.Get("/claims/{acoId}", bulkRequest)
-		r.Get("/job/{jobId}", jobStatus)
+		r.With(auth.RequireTokenAuth).Post("/claims", bulkRequest)
+		r.With(auth.RequireTokenAuth).Get("/job/{jobId}", jobStatus)
```
embh: Would it make sense for this endpoint to be /jobs/{jobId}? I think plural is the usual recommendation and it would be consistent with /claims

PR #12, opened 2018-08-24
BCDA-147: Add UUID to access token claims

embh:
### Fixes [BCDA-147](https://jira.cms.gov/browse/BCDA-147)
Our current process of finding tokens by user ID when checking to see if they've been blacklisted could be slow if many tokens exist for a user.

### Proposed changes:
Add a UUID to the token claims and replace serial ID with application-generated UUID in the tokens table. Search on UUID when checking to see if a token is active.

### Change Details
* `GenerateToken()` adds a UUID to the JWT claims
* `RevokeToken()` saves the UUID to the database
* `IsBlacklisted()` queries by token UUID

### Security Implications
This impacts the way we revoke token access.

### Acceptance Validation
All auth backend tests still passing.

### Feedback Requested
General review - any improvements to be made?

### DIFF
--- a/bcda/auth/backend.go
+++ b/bcda/auth/backend.go
@@ -172,10 +172,10 @@ func (backend *JWTAuthenticationBackend) IsBlacklisted(token *jwt.Token) bool {
 
 	const sqlstr = `SELECT value ` +
 		`FROM public.tokens ` +
-		`WHERE user_id = $1 ` +
+		`WHERE uuid = $1 ` +
 		`AND active = false`
 
-	rows, err := db.Query(sqlstr, claims["sub"])
+	rows, err := db.Query(sqlstr, claims["id"])
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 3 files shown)


bcda/auth/backend.go
```diff
@@ -135,11 +138,24 @@ func (backend *JWTAuthenticationBackend) RevokeToken(tokenString string) error {
...
+		err = token.Update(db)
+	} else {
+
+		const sqlstr = `INSERT INTO public.tokens (` +
+			`uuid, user_id, value, active` +
+			`) VALUES (` +
+			`$1, $2, $3, $4` +
+			`)`
+
+		_, err = db.Exec(sqlstr, token.UUID, token.UserID, token.Value, false)
```
rnagle: Why not use `token.Insert(db)` here?

embh: I tried using that but got the "No rows in result set" error. I don't think Postgres returns anything on insert without `RETURNING...` in the statement, which isn't in the one xo generated.

PR #13, opened 2018-08-31
BCDA-193: Revise HTTP responses for existing endpoints to match existing FHIR bulk spec

embh:
### Fixes [BCDA-193](https://jira.cms.gov/browse/BCDA-193)

We have a simple Hello World-style app whose endpoints are different from those indicated by the FHIR spec.

### Proposed changes:
* Convert `POST /claims` to `GET /Patient/$export` and update responses
* For `GET /jobs/{jobId}`, add a delay to simulate data retrieval transaction and update responses

### Change Details
* `POST /claims` will return `202 Accepted` if successful, with the job status URL in the `Content-Location` header
   * Requires `Accept` header = `application/fhir+json` and `Prefer` = `respond-async`
* `GET /jobs/{jobId}` returns `202 Accepted` when the job is processing and `200 OK` when complete
   * Completed response includes transaction timestamp, original request URL, auth info (`requiresAccessToken: true`), an array of file items (currently one with a placeholder URL), and an array of errors (currently empty)

### Security Implications
Token auth remains the same. No PII.

### Acceptance Validation
screen shot 2018-08-31 at 9 15 11 am
screen shot 2018-08-31 at 9 15 03 am

### Feedback Requested
Anything we need in this PoC that was missed? (We'll be adding a real URL with synthetic data with BCDA-194.)


bcda/main.go
```diff
@@ -116,16 +159,43 @@ func jobStatus(w http.ResponseWriter, r *http.Request) {
...
+		fallthrough
+	case "In Progress":
+		w.Header().Set("X-Progress", job.Status)
+		w.WriteHeader(http.StatusAccepted)
+	case "Failed":
+		http.Error(w, http.StatusText(500), 500)
+	case "Completed":
+		w.Header().Set("Content-Type", "application/json")
+
+		fi := fileItem{
```
rnagle: I understand this `fileItem` is a placeholder right now, but do you have any thoughts on the `Type` that we'll ultimately need to use here? 🤔

bcda/main.go
```diff
@@ -55,20 +70,32 @@ func bulkRequest(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	err = validateBulkRequestHeaders(r.Header)
```
rnagle: Seems like a good candidate for a middleware to me. What do you think?

bcda/main.go
```diff
@@ -84,17 +111,33 @@ func bulkRequest(w http.ResponseWriter, r *http.Request) {
...
+}
+
+func validateBulkRequestHeaders(h http.Header) error {
+	acceptHeader := h.Get("Accept")
+	preferHeader := h.Get("Prefer")
+
+	var errMsgs = []string{}
+
+	if acceptHeader == "" {
+		errMsgs = append(errMsgs, "Accept header is required")
```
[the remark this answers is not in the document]
embh: Will do! I wasn't sure if it was better to do that or tell them all the errors upfront.

bcda/main.go
embh: Didn't have anything in mind. The specs say each file item has `type` and `url`, and optionally `count`: https://github.com/smart-on-fhir/fhir-bulk-data-docs/blob/master/export.md#response---complete-status Open to suggestions!

bcda/main.go
embh: Agree, that seems like a better place for it. I need to get a better sense of the typical Go application structure.

PR #14, opened 2018-09-06
BCDA-194: Create download endpoint for the FHIR bulk script

embh:
### Fixes [BCDA-194](https://jira.cms.gov/browse/BCDA-194)

We need to create a PoC of providing beneficiary data.

### Proposed changes:
* Serve a bundle of synthetic ExplanationOfBenefit data in an NDJSON file
* Provide file URL in job status response body when complete

### Change Details
* Adds static NDJSON file named with the UUID of one of our fixture ACOs
* Updates job status response to include URL of file
* Validates access token and confirms that token ACO matches file ACO
* Moves API functions from `main.go` into `api.go`

### Security Implications
This currently serves synthetic data. JWT ACO UUID check against file name in addition to token validation.

### Acceptance Validation
screen shot 2018-09-06 at 9 16 22 am
screen shot 2018-09-06 at 9 17 20 am

### Feedback Requested
* Does the new `api.go` file make sense as a location for the API router destination functions?


bcda/api.go
```diff
@@ -0,0 +1,159 @@
...
+			http.Error(w, http.StatusText(500), 500)
+			return
+		} else {
+			w.WriteHeader(http.StatusOK)
+		}
+	}
+}
+
+func serveData(w http.ResponseWriter, r *http.Request) {
+	http.ServeFile(w, r, "./static/DBBD1CE1-AE24-435C-807D-ED45953077D3.ndjson")
```
rnagle: This depends on the repo directory structure of the repo and the way that `fresh` builds `bcda`. Without the context of the repo, `bcda` won't be able to find this file. Not a big deal right now, but this will have an impact on the way we package the app for deployment later.

PR #15, opened 2018-09-08
BCDA-234: Containerize Unit Tests

msnook:
The intent of BCDA-234 is to:
1) Containerize the unit tests to ensure they are run in an environment that we know is stable through automation.  This removes the need to install extraneous dependencies on host machine.
2) Remove local dependency on `psql` (which is bloated) and replace with `usql` (which is lean)
3) Beef up README with more details on how to set up environment, how to run tests, and how to interact with the app

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,25 +1,22 @@
 models:
+	docker-compose up -d db
+	echo "Waiting for db to be ready..."
+	sleep 5	
 	PGSSLMODE=disable xo postgresql://postgres:toor@localhost:5432/bcda -o models
 
 test:
-	golangci-lint run
-	docker-compose up -d db && sleep 5
-	PGPASSWORD=toor psql -U postgres -h localhost -p 5432 -q -c 'drop database if exists bcda_test;'
-	PGPASSWORD=toor psql -U postgres -h localhost -p 5432 -q -c 'create database bcda_test;'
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -1,25 +1,22 @@
...
+	docker-compose up -d db
+	docker-compose -f docker-compose.test.yml up
 
 load-fixtures:
 	docker-compose up -d db
 	echo "Wait for db to be ready..."
 	sleep 5
-	psql "postgresql://postgres:toor@localhost:5432/bcda" -f db/fixtures.sql
-	docker-compose stop db
+	docker-compose run -v $(shell pwd)/db:/tmp/db db psql "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -f /tmp/db/fixtures.sql
```
rnagle: IIRC using `docker-compose up -d db` and `docker-compose run db` will run two distinct postgresql containers. I'd be concerned that this will result in changes to the writable layer for the wrong service container, though it's only a hunch.

An alternative approach: run the container with modified volume arg, then execute the `psql` command in said running container:

```
load-fixtures:
        docker-compose run -d -v $(shell pwd)/db:/tmp/db db
	echo "Wait for db to be ready..."
	sleep 5
        docker-compose exec db psql "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -f /tmp/db/fixtures.sql
```

PR #16, opened 2018-09-10
BCDA-211 Feature: UUID V4 instead of V1

knollfear:
Fixes [BCDA-211](https://jira.cms.gov/browse/BCDA-211)

Use UUID V4 because it is more secure than UUID V1
### Proposed changes:

Changed NewUUID() to NewRandom()

### Security Implications

No PII impacted, but issued tokens should be a bit more secure. 

### Acceptance Validation

Tests ran, that should be sufficient to make sure it works.


embh:
The UUID changes look good to me, but I still see @msnook 's commits showing up (even though under Files Changed it's just the UUID stuff). Can this be synced with master?

PR #17, opened 2018-09-11
BCDA-235: Generate test coverage metrics/reports 

msnook:
Test coverage metrics and reports are generated as part of `make test`, and the resulting reports are output in the `test_results` folder on the host (individual timestamped folders are created in `test_results` in order to separate the metrics by test run).  

The files that are generated are:
1) `testresults.out` -- Results of test case execution
2) `testcoverage.out` -- Generic test coverage file generated by `-coverprofile`
3) `testcov_byfunc.out` -- Test coverage metrics by function
4) `testcoverage.html` -- Highlighted test coverage analysis for each go file

See attached to see examples of the outputs.

[testcovfiles.zip](https://github.com/CMSgov/bcda-app/files/2371227/testcovfiles.zip)

Two additional things to note:
1) Empty test files were created in some of our packages which did not have test cases, to be used as placeholders so coverage could be generated for those particular go files.  We will need to create test cases for these packages (examples: logging, database).
2) Test metrics are reported at the unit level.  That means that if a test is executed and calls a function in another package under the covers, test coverage is not reported for that as we have gone outside of the unit level.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 **/.DS_Store
 bcda/bcda
 bcdaworker/bcdaworker
+test_results/*
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ models:
 
 test:
 	docker-compose up -d db
-	docker-compose -f docker-compose.test.yml up
+	docker-compose -f docker-compose.test.yml up --force-recreate
 
 load-fixtures:
 	docker-compose up -d db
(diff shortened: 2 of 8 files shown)


msnook:
Yup, I plan to create three tickets... (1) boost coverage where we have low number, (2) integrate with Travis and expose reports in GitHub dash, (3) integrate with Jenkins and expose reports on Jenkins dash.

PR #18, opened 2018-09-11
bcda-226 feature created mock aco/beneficiary data

tbellj:
### Problem:
Reasoning behind three synthetic ACOs: we can create small, medium and large ACOs (e.g., 2.5k, 7.5k, 20k synthetic beneficiaries respectively). What this means is when we start performing dynamic lookup of synthetic records, we'll have some variation in the file size that BCDA delivers.

### Proposed changes:
fixtures.sql - insert statement for 3 synthetic aco's
synthetic_beneficiaries - created a beneficiaries table and insert statements to populate
Makefile - added a load-beneficiaries step to add the synthetic_beneficiary data

### Change Details
modified db/fixtures.sql file to add 3 synthetic ACO's.
added db/synthetic_beneficiaries.sql file to create a beneficiaries table with patient id's and aco_ids that match to the 3 synthetic ACO's from earlier.
Modified the makefile for an extra load beneficiary step.

### Security Implications
No security concerns, all data is artificial

### Acceptance Validation
At a later date I believe all 30k will be loaded locally and queries can be performed

### Feedback Requested
All feedback


Makefile
```diff
@@ -14,6 +14,12 @@ load-fixtures:
 	sleep 5
 	docker-compose exec db psql "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -f /var/db/fixtures.sql
 
+load-beneficiaries:
```
msnook: `load-beneficiaries` target is not tied into the build.  I don't think we want to explicitly have to call it, do we?  I think the intention was to incorporate this data load as part of `load-fixtures`.  My recommendation is to remove the `load-beneficiaries` target completely, and just add this line to the end of `load-fixtures`:

`docker-compose exec db psql "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -f /var/db/synthetic_beneficiaries.sql
`

msnook:
Some other areas where the beneficiaries should probably be loaded:

- `unit_test.sh` has a line where the fixtures are being loaded.  Probably want to load the beneficiaries into our unit test DB as well. 
- `.travis.yml` also loads the fixtures into the test DB around line [44](https://github.com/CMSgov/bcda-app/blob/master/.travis.yml#L44).  Probably want to load beneficiaries there too.

Makefile
tbellj: Yep, I think that was the original intent but there was a conversation off line about adding a load-beneficiaries target to the makefile and then executing it later when we actually want to perform the db queries. Your suggestion absolutely makes sense if we want to just load it all now. It really all depends on what the team wants.

msnook: Thats cool.  If thats the case, then I would think we would at least want to update the Makefile so that `docker-bootstrap` calls `load-benficiaries`.  @rnagle what do you think?

tbellj: oh I see what you're saying, that makes sense to me

rnagle: Not opposed to adding to `make docker-boostrap`, we'll definitely want it there eventually. But, it could be troublesome/confusing until we actually start using the data in our app. For instance, with this data loaded, running `make models` will produce an unused `models/beneficiaries.xo.go` file.

PR #19, opened 2018-09-12
BCDA-182: Add log levels and write logs to files

embh:
### Fixes [BCDA-182](https://jira.cms.gov/browse/BCDA-182)

* Our Hello World app handles many problems with `log.Fatal(err)`, causing it to stop working even in the case of errors the application could recover from. 
* To use our logs with Splunk, they need to be written somewhere other than stdout.

### Proposed changes:
* Replace non-catastrophic calls to `log.Fatal(err)` with `log.Error(err)` and return an appropriate API response
* Write log entries to files

### Change Details
* Per FHIR spec, API error responses should be an OperationOutcome resource in JSON format. Currently, these are empty placeholders only, using models from https://github.com/eug48/fhir. Ticket to populate these responses is [BCDA-276](https://jira.cms.gov/browse/BCDA-276).
* Request and error log can be configured with environment variables `BCDA_REQUEST_LOG` and `BCDA_ERROR_LOG`. Added values for these in Dockerfile.bcda.
* Request and error log entries are structured (JSON-formatted); error logs may need more data ([BCDA-183](https://jira.cms.gov/browse/BCDA-183))

### Security Implications 
`IsBlacklisted()` now returns `(bool, error)` to allow the calling function to handle the error as appropriate. In the case of an error in determining whether a token has been blacklisted, the boolean value is `true` so a database error, for example, does not allow a blacklisted user/system to make requests.

### Acceptance Validation
* Logs
screen shot 2018-09-12 at 5 33 24 pm
screen shot 2018-09-14 at 5 22 36 pm

* OperationOutcome response
screen shot 2018-09-12 at 5 34 33 pm

### Feedback Requested
Having not used Splunk before, is there somewhere else the log entries should be written to? Happy to receive any other feedback as well.


bcda/logging/middleware.go
```diff
@@ -15,6 +16,12 @@ import (
 func NewStructuredLogger() func(next http.Handler) http.Handler {
 	logger := logrus.New()
 	logger.Formatter = &logrus.JSONFormatter{}
+	file, err := os.OpenFile("/var/log/bcda-request.log", os.O_CREATE|os.O_WRONLY, 0666)
```
rnagle: Please make the location of the file something we can pass by setting an env var (e.g., `BCDA_REQUEST_LOG`). If the env var is not set, fallback to `logger` default.

bcda/main.go
```diff
@@ -40,6 +41,15 @@ type bulkResponseBody struct {
 	Errors              []fileItem `json:"error"`
 }
 
+func init() {
+	file, err := os.OpenFile("/var/log/bcda.log", os.O_CREATE|os.O_WRONLY, 0666)
```
rnagle: Same deal here -- please allow setting the log location via env var like `BCDA_ERROR_LOG`.

models/operationoutcome.go
```diff
@@ -0,0 +1,27 @@
+package models
```
rnagle: Should we just `import "github.com/eug48/fhir/models"` for this stuff rather than duplicate?

PR #22, opened 2018-09-17
BCDA-195 Feature: returning FHIR bulk responses for errors

tbellj:
When a user request a job status response the state of that job should be accurate and represented in a format that meets the FHIR Bulk guidelines.

Proposed changes:
Added error response handling for the Failed status to match the output format in the spec. Taken directly from the spec. "The body MUST be a FHIR OperationOutcome in JSON format". All of the other responses were already handled (Pending, In Progress, Completed) (I think Emily did them, but I could be wrong). For expired status see ticket: (https://jira.cms.gov/browse/BCDA-294)

Change Details:
* Added a struct to match the operation outcome response to the bcda api classes. (main.go)
* Also added the error handling for request being made during the job status generation. Once we actually start processing jobs (anything that's more than serving up a static file), then the job "Failed" status can be added during processing. Added the creation of the operation outcome response in this class as well. (api.go)

Security Implications:
Currently there are no security implications but a conversation about how much data we're willing to output in our error responses should take place. see ticket (https://jira.cms.gov/browse/BCDA-293)

Acceptance Validation:
I was able to test the changes by manipulating a status to "Failed" in the db, then added some code to see the operation outcome response. Currently it's hard to test a failed status because nothing is really setting a status to "Failed" yet. But this will obviously change once we start file/data processing.

Feedback Requested
Anything

### DIFF
--- a/bcda/api.go
+++ b/bcda/api.go
@@ -123,18 +123,55 @@ func jobStatus(w http.ResponseWriter, r *http.Request) {
 
 		jsonData, err := json.Marshal(rb)
 		if err != nil {
+			errorStatus := createOperationOutcome("marshalling error")
+			_, err = w.Write([]byte(errorStatus))
+			if err != nil {
+				http.Error(w, http.StatusText(500), 500)
+				return
+			}
 			http.Error(w, http.StatusText(500), 500)
 			return
 		}
 
 		_, err = w.Write([]byte(jsonData))
 		if err != nil {
+			errorStatus := createOperationOutcome("writing error")
+			_, err = w.Write([]byte(errorStatus))
+			if err != nil {
+				http.Error(w, http.StatusText(500), 500)
+				return
+			}
 			http.Error(w, http.StatusText(500), 500)
 			return
 		} else {
 			w.WriteHeader(http.StatusOK)
 		}
 	}
+}
+
+func createOperationOutcome(outcome string) []byte {
+	is := issue {
+		Severity: "",
+		Code: "",
+		Details: codeableConcept{nil,""},
+		Diagnostics: "",
+		Location: nil,
+		Expression: nil,
+	}
+
+	oo := operationOutcomeBody{
+		ResourceType: "OperationOutcome",
+		Id: outcome,
+		Text: "",
+		Issue: []issue{is},
+	}
+
+	jsonData, err := json.Marshal(oo)
+	if err != nil {
+		log.Fatal(err)
+		return nil
+	}
+	return jsonData
 }
 
 func serveData(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


bcda/api.go
```diff
@@ -123,12 +123,24 @@ func jobStatus(w http.ResponseWriter, r *http.Request) {
 
 		jsonData, err := json.Marshal(rb)
 		if err != nil {
+			errorStatus := createOperationOutcome("marshalling error")
+			_, err = w.Write([]byte(errorStatus))
+			if err != nil {
```
msnook: Are lines 128-130 redundant?  Looks like we are doing the following regardless:
`http.Error(w,` http.StatusText(500), 500)
return`

bcda/api.go
```diff
@@ -123,12 +123,24 @@ func jobStatus(w http.ResponseWriter, r *http.Request) {
...
+			}
 			http.Error(w, http.StatusText(500), 500)
 			return
 		}
 
 		_, err = w.Write([]byte(jsonData))
 		if err != nil {
+			errorStatus := createOperationOutcome("writing error")
+			_, err = w.Write([]byte(errorStatus))
+			if err != nil {
```
msnook: Are lines 140-142 redundant?  Looks like we are doing the following regardless:
`http.Error(w,` http.StatusText(500), 500)
return`

msnook:
In the PR description, it was mentioned that it was hard to test because nothing is setting a status to Failed yet.  Can we write a unit test for this?

bcda/api.go
```diff
@@ -137,6 +149,31 @@ func jobStatus(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func createOperationOutcome(outcome string) []byte {
```
msnook: Should this function be developed such that all the fields can be set?  It looks like we are only supporting the ability to set `Id`, but would we ever want the ability to set severity, code, diagnostics, etc, etc, etc (rather than leave them as hardcoded empty strings)?

tbellj: True. We absolutely want to set the real details of the error response once we start processing for errors. Currently nothing really causes a failure for a job because we're just serving up a static file, but you are correct in the future we will want this info to be populated. This makes me think of a future design question: I think this method will change from passing function args to just handling of the db job object.

bcda/api.go
tbellj: This is also true, it is a bit redundant. So basically we marshal the data but if an error happens we have to write the operation outcome response. But the writing of that response error could also fail so I try to capture that as well. Kinda stuck in this error handling predicament but would love a new design here.

bcda/api.go
tbellj: Same deal as the above comment

PR #23, opened 2018-09-17
BCDA-282: GORM and bulk token revocation

knollfear:
### Fixes [BCDA-189](https://jira.cms.gov/browse/BCDA-189)

Remove XO and add in GORM
Improve Unit Test coverage
Add method for invalidating all tokens for a given user
add method for invalidating all tokens for an ACO
### Proposed changes:

GORM
Remove XO
Invalidating User TOkens
Invalidating ACO tokens
### Security Implications

Check out the tests, they handle token security which is important

### Feedback Requested

Pleas ask questions about the GORM process if it isn't clear.  It is a change from XO, but it is much simpler and easier to code.


msnook:
If we are deprecating `xo`, can we update this target in the Makefile to use `GORM`?  Maybe this process should be tested?  
https://github.com/CMSgov/bcda-app/blob/master/Makefile#L5

msnook:
The README references `xo`.  Can that be removed?

knollfear:
GORM is not a CLI program, but a dependency in the project.  It should not require anything except proper inclusion in the vendor folder on build.

Removed XO from readme.md
Removed go get from the docker files
Added initialize to main.go to get the DB set up correctly

msnook:
I was looking at line 5 of the Makefile. 
https://github.com/CMSgov/bcda-app/blob/master/Makefile#L5

Should this command line invocation of `xo` be replaced?

Makefile
```diff
@@ -1,9 +1,3 @@
-models:
-	docker-compose up -d db
-	echo "Waiting for db to be ready..."
-	sleep 5
-	PGSSLMODE=disable xo postgresql://postgres:toor@localhost:5432/bcda -o models
```
msnook: Now that `xo` has been completely removed from Makefile, does the `models` target need to be replaced with something that is GORM-specific?  Will we need to continue to execute the model generation in the future, or is this now completely obsolete?

knollfear: completely obsolete.  GORM doesn't write go code the way XO did.  It handles dynamic query execution at run time instead of prewriting it all.

Dockerfiles/Dockerfile.bcda
```diff
@@ -7,6 +7,7 @@ RUN go get github.com/tools/godep
 RUN go get github.com/pressly/fresh
 RUN go get -u github.com/derekparker/delve/cmd/dlv
 
+
```
rnagle: A couple extraneous newlines in the Dockerfiles.

bcda/auth/backend.go
```diff
@@ -261,22 +226,48 @@ func getPublicKey() *rsa.PublicKey {
...
 			return nil, fmt.Errorf("Unexpected signing method: %v", token.Header["alg"])
 		}
 
 		return backend.PublicKey, nil
 	})
+	return token, err
+}
 
+// Save a token to the DB for a user
+func (backend *JWTAuthenticationBackend) CreateToken(user bcdagorm.User) (bcdagorm.Token, error) {
```
rnagle: What's the plan for this function? Should this be called when running the `bcda create-token` command? I only see it invoked in test cases. I'm probably just missing something.

bcda/auth/backend.go
```diff
@@ -261,22 +226,48 @@ func getPublicKey() *rsa.PublicKey {
...
-		return nil
+	// Get the claims of the token to find the token ID that was created
+	claims := backend.GetJWTClaims(tokenString)
+	token := bcdagorm.Token{
+		UUID:   uuid.Parse(claims["id"].(string)),
+		UserID: user.UUID,
+		Value:  tokenString,
+		Active: true,
 	}
+	db.Create(&token)
```
rnagle: NB for the whole team: if we do start using this function as part of the `bcda create-token` command, we will be storing each generated token in our db.

bcdagorm/models.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+func Initialize() *gorm.DB {
+	db := database.GetGORMDbConnection()
+
+	defer db.Close()
+
+	// Migrate the schema
+	// Add your new models here
+	db.AutoMigrate(
```
rnagle: I'm a little worried about running migrations automatically when the app starts. Are these always safe to run? What happens when there are multiple instances of the app started across many servers?

Thinking it'd be nice to have this wrapped up in a `bcda migrate` command or similar to have control over the timing of migrations. What do you think?

I'm also wondering if there's a way to have gorm output the SQL it plans to apply for migrations. Maybe a [`bcda sqlmigrate`](https://docs.djangoproject.com/en/2.1/ref/django-admin/#django-admin-sqlmigrate) command?

bcda/auth/backend_test.go
```diff
@@ -33,43 +40,233 @@ func (s *BackendTestSuite) TestCreateACO() {
 }
 
 func (s *BackendTestSuite) TestCreateUser() {
-	userUUID, err := s.authBackend.CreateUser("First Last", "firstlast@example.com", uuid.Parse("DBBD1CE1-AE24-435C-807D-ED45953077D3"))
+	name, email, sampleUUID := "First Last", "firstlast@exaple.com", "DBBD1CE1-AE24-435C-807D-ED45953077D3"
```
rnagle: All these tests! 🍔

bcdagorm/models.go
knollfear: I think a sql migrate command makes sense.  I will change it.  It needs to be added to the jenkins deploy though.

Dockerfiles/Dockerfile.bcda
knollfear: removed in next commit.

bcda/auth/backend.go
knollfear: I don't know exactly when this will be run.  I feel like we are still waiting for approval on our auth-X schemes.

bcda/auth/backend.go
knollfear: It doesn't yet.  It can very easily though.

bcdagorm/models.go
knollfear: Added this command.

bcda/auth/backend.go
knollfear: I didn't want to change the function of this command.  I am happy to change it, but I wasn't sure if it was in scope or not.

Dockerfiles/Dockerfile.bcda
knollfear: removed

bcda/auth/backend.go
knollfear: I left this function functionally unchanged so it will NOT persist a token just yet.   It probably should soon though.

bcda/auth/backend.go
```diff
@@ -261,22 +226,48 @@ func getPublicKey() *rsa.PublicKey {
...
 	}
-
-	if !token.Valid {
-		return nil
+	// Get the claims of the token to find the token ID that was created
+	claims := backend.GetJWTClaims(tokenString)
+	token := bcdagorm.Token{
+		UUID:   uuid.Parse(claims["id"].(string)),
+		UserID: user.UUID,
+		Value:  tokenString,
```
rnagle: One last request for me: please make sure we're hashing token values anywhere we are inserting/updating them in the database. Check out: https://github.com/CMSgov/bcda-app/blob/master/bcda/auth/backend.go#L135

bcda/auth/backend.go
```diff
@@ -261,22 +226,51 @@ func getPublicKey() *rsa.PublicKey {
...
+	}
+
+	if !token.Valid {
+		return nil
+	}
+
+	return token.Claims.(jwt.MapClaims)
+}
+
+func (backend *JWTAuthenticationBackend) GetJWTToken(tokenString string) (*jwt.Token, error) {
```
embh: Not a big deal at all, but can we call this `GetJWT` or `GetToken`, since the T in JWT = Token?

knollfear: getToken isn't going to work because that is ambiguous about what kind of token, our token object in the db vs a JWT token.  I think getJWT is also unclear because that could just be the JWT package or some JWT context instead of an actual JWT token.  JWT is a clear part of our system and has other objects like claims (See line 229) that we use.  I think getJWToken might be fine, but it isn't jwt.oken, it is jwt.Token.

bcda/auth/backend.go
knollfear: Resolved in latest commit.  token.Value is now hashed on save IF it is a valid JWT token.  If it isn't then I assume it is already hashed and don't want to double hash it.

bcda/auth/backend.go
knollfear: Committed a change to GetJWToken

PR #25, opened 2018-09-19
BCDA-253: Develop tests for bcda/api.go

embh:
### Fixes [BCDA-253](https://jira.cms.gov/browse/BCDA-253)

We don't have any tests for `api.go`!

### Proposed changes:
* Add test for bad request to `bulkRequest()`
* Add jobs to fixtures with different statuses and matching `jobStatus()` tests
* Add `serveData()` test

### Change Details
This starts creating tests for `api.go`, but coverage is not even close to 100%. Will create a part 2 ticket.

### Security Implications
None

### Acceptance Validation
1) bcda/api_test.go is created and populated with unit tests for bcda/api.go ✔️ 
2) All unit tests pass ✔️ 
3) Code coverage of bcda/api.go is at 100% ❌ 41.2%, additional ticket to follow

### Feedback Requested
Is this a decent start?

### DIFF
--- a/bcda/api.go
+++ b/bcda/api.go
@@ -6,7 +6,6 @@ import (
 	"github.com/CMSgov/bcda-app/bcda/bcdaModels"
 	"net/http"
 	"strconv"
-	"time"
 
 	"github.com/CMSgov/bcda-app/bcda/auth"
 	"github.com/CMSgov/bcda-app/bcda/database"
@@ -123,7 +122,7 @@ func jobStatus(w http.ResponseWriter, r *http.Request) {
 		}
 
 		rb := bulkResponseBody{
-			TransactionTime:     time.Now(),
+			TransactionTime:     job.CreatedAt,
 			RequestURL:          job.RequestURL,
 			RequiresAccessToken: true,
 			Files:               []fileItem{fi},
--- a/bcda/api.go
+++ b/bcda/api.go
@@ -34,6 +34,9 @@ func bulkRequest(w http.ResponseWriter, r *http.Request) {
 			writeError(fhirmodels.OperationOutcome{}, w)
 			return
 		}
+	} else {
+		http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
+		return
 	}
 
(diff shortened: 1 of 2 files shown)


db/fixtures.sql
```diff
@@ -9,3 +9,8 @@ insert into users values ('EFE6E69A-CD6B-4335-A2F2-4DBEDCCD3E73', 'User Two', 'u
 
 insert into users values ('B6DFAD18-62A1-4EA3-B623-38F11D49E0F6', 'User Three', 'userthree@email.com', 'A40404F7-1EF2-485A-9B71-40FE7ACDCBC2', default, default);
 insert into users values ('1E270119-E503-4F13-A6AC-54EC3FB44478', 'User Four', 'userfour@email.com', 'A40404F7-1EF2-485A-9B71-40FE7ACDCBC2', default, default);
+
+insert into jobs values (1, 'DBBD1CE1-AE24-435C-807D-ED45953077D3', '82503A18-BF3B-436D-BA7B-BAE09B7FFD2F', '/api/v1/Patient/$export', 'Pending', '2018-09-18T20:35:31.128361Z');
```
msnook: I'm a little leary on creating a large database fixture which populates data randomly to be used by all our tests, and then our test results rely on that data being pristine over time.  The last thing we want to happen is flaky/unstable tests due to data issues, causing false alarms and maintenance nightmares.  This could become hard to maintain over time, as more and more tests (which depend upon data) are created.
I think we want a test to be able to execute independently, and ensure that the system is left in the same state after the test as it was before the test started. 
My suggestion is to create the mock data from within the test (as part of setup), and remove the mock data when the test completes (as part of teardown).  This way, we don't have a situation where tests which are executed down the line rely on data created for this test specifically.

embh: Makes sense. In that case, I think I'll have to revisit this after https://github.com/CMSgov/bcda-app/pull/23 is merged so I can use the new GORM commands to create test jobs.

msnook: Sounds good.  I think the test code looks great, just need to address the data issue and good to go!

PR #29, opened 2018-09-24
BCDA-256 Feature: Unit test for db connections

tbellj:
Fixes [BCDA-256](https://jira.cms.gov/browse/BCDA-256)

Problem:
Created unit test for our database connectivity. Previously there were none.

Proposed changes:
Added a test suite to connection_test.go (bcda/database/connection_test.go)

Security Implications:
None.

Acceptance Validation:
I wasn't able to fully get the unit test coverage to 100% due to the inability to properly test log statements. Not sure how to do that other than writing unit test for our logger itself which seems kinda excessive.
Take a look at this screenshot:
screen shot 2018-09-24 at 9 12 04 am

Feedback Requested:
Any feedback. How can we get our test coverage to 100% with this problem? It's not just this test but for all of our test.

### DIFF
--- a/bcda/database/connection.go
+++ b/bcda/database/connection.go
@@ -4,26 +4,30 @@ import (
 	"database/sql"
 	"log"
 	"os"
-
 	"github.com/jinzhu/gorm"
-	_ "github.com/jinzhu/gorm/dialects/postgres"
-	_ "github.com/lib/pq"
+        _ "github.com/jinzhu/gorm/dialects/postgres"
+        _ "github.com/lib/pq"
 )
+
+// Variable substitution to support testing.
+var logFatal = log.Fatal
 
 func GetDbConnection() *sql.DB {
 	databaseURL := os.Getenv("DATABASE_URL")
-	db, err := sql.Open("postgres", databaseURL)
-	if err != nil {
-		log.Fatal(err)
+	db, _ := sql.Open("postgres", databaseURL)
+	pingErr := db.Ping()
+	if pingErr != nil {
+		logFatal(pingErr)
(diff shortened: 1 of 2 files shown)


msnook:
Please see my latest commit: https://github.com/CMSgov/bcda-app/pull/29/commits/94d4f1f13e92172971a0e6032a5d0b40d919cc2d

This addresses a few concerns:

1) According to Go SQL doc, Open() will only return an error if there is a problem with the driver.  I have updated our error handling to react on the results of Ping() rather than Open().  https://golang.org/pkg/database/sql/#Open

2) Added variable substitution so that we can mock out log.Fatal()

3) Resetting the `DATABASE_URL` environment variable (a few times) so that we can properly test positive and negative test scenarios.

4) Code coverage is now 100% (see attached screenshot).
screen shot 2018-09-25 at 7 13 06 pm

5) I've written the following story as a follow-up so that we can update the system code to incorporate custom (and consistent) logging, which might allow us to mock the interface within the test code instead of using variable substitution. https://jira.cms.gov/browse/BCDA-335

bcda/database/connection.go
```diff
@@ -4,26 +4,30 @@ import (
 	"database/sql"
 	"log"
 	"os"
-
 	"github.com/jinzhu/gorm"
-	_ "github.com/jinzhu/gorm/dialects/postgres"
-	_ "github.com/lib/pq"
+        _ "github.com/jinzhu/gorm/dialects/postgres"
```
rnagle: Please run `go fmt` against `connection.go` and `connection_test.go`.

PR #30, opened 2018-09-24
Eh/bcda 301

embh:
### Fixes [BCDA-301](https://jira.cms.gov/browse/BCDA-301)
We need a way to interact with the Blue Button backend.

### Proposed changes:
* Add BlueButtonClient and tests
* Add `/bb-metadata` endpoint

### Change Details
* BlueButtonClient has basic functions for BB patient, coverage, and explanation of benefit data, as well as metadata
* The app passes the response from BB's metadata endpoint through our API to `/bb-metadata`
* BB URL and PEM file paths configurable with environment variables `BB_SERVER_LOCATION`, `BB_CLIENT_CERT_FILE`, `BB_CLIENT_KEY_FILE`

### Security Implications
None. Connects to Blue Button development server; no real data.

### Acceptance Validation
- [x] Able to receive data from authenticated endpoint at BB
- [x] Add a stub API endpoint that forwards BB 2.0 metadata endpoint
   - [x] Simply make a request to BB 2.0 for their metadata endpoint, return it in a payload presented at /api/v1/bb_metadata or similar in BCDA.
   - [x] Make this available when the app is run with DEBUG=true
- [x] Unit tests for discrete pieces of functionality

### Feedback Requested
All feedback welcome!


bcda/client/bluebutton.go
```diff
@@ -0,0 +1,110 @@
...
+
+	req, err := http.NewRequest("GET", bbServer+path, nil)
+	if err != nil {
+		return "", err
+	}
+
+	req.URL.RawQuery = params.Encode()
+	addRequestHeaders(req)
+
+	resp, err := bbc.httpClient.Do(req)
```
rnagle: It'd be great to log these requests. See how BB is doing this here: https://github.com/CMSgov/bluebutton-web-server/blob/master/apps/logging/serializers.py#L80-L101

bcda/client/bluebutton.go
```diff
@@ -0,0 +1,110 @@
...
+	return string(data), nil
+}
+
+func addRequestHeaders(req *http.Request) {
+	// https://github.com/CMSgov/bluebutton-web-server/blob/master/apps/fhir/bluebutton/utils.py#L224-L231
+	req.Header.Add("BlueButton-OriginalQueryTimestamp", time.Now().String())
+	req.Header.Add("BlueButton-OriginalQueryId", "1")
+	req.Header.Add("BlueButton-OriginalQueryCounter", "1")
+	req.Header.Add("BlueButton-BeneficiaryId", "")
+	req.Header.Add("BlueButton-UserId", "")
```
rnagle: We can probably safely drop `UserId`, `Application`, `ApplicationId`, `DeveloperId` and `Developer` headers. I think they're rather specific to BB frontend.

bcda/client/bluebutton.go
```diff
@@ -0,0 +1,110 @@
...
+		return "", err
+	}
+
+	return string(data), nil
+}
+
+func addRequestHeaders(req *http.Request) {
+	// https://github.com/CMSgov/bluebutton-web-server/blob/master/apps/fhir/bluebutton/utils.py#L224-L231
+	req.Header.Add("BlueButton-OriginalQueryTimestamp", time.Now().String())
+	req.Header.Add("BlueButton-OriginalQueryId", "1")
```
rnagle: The value for this should be a `UUID` for the request. Nice to have for tracing and debugging requests in logs on both sides.

bcda/client/bluebutton.go
```diff
@@ -0,0 +1,110 @@
...
+	req.Header.Add("BlueButton-OriginalQueryCounter", "1")
+	req.Header.Add("BlueButton-BeneficiaryId", "")
+	req.Header.Add("BlueButton-UserId", "")
+	req.Header.Add("BlueButton-Application", "")
+	req.Header.Add("BlueButton-ApplicationId", "")
+	req.Header.Add("BlueButton-DeveloperId", "")
+	req.Header.Add("BlueButton-Developer", "")
+	req.Header.Add("BlueButton-OriginatingIpAddress", "")
```
rnagle: Please create a ticket to figure out and implement appropriate values for the remainder of the headers. 👌

bcda/client/bluebutton.go
```diff
@@ -0,0 +1,110 @@
...
+	certFile := os.Getenv("BB_CLIENT_CERT_FILE")
+	keyFile := os.Getenv("BB_CLIENT_KEY_FILE")
+	cert, err := tls.LoadX509KeyPair(certFile, keyFile)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	tlsConfig := &tls.Config{
+		Certificates:       []tls.Certificate{cert},
+		InsecureSkipVerify: true,
```
rnagle: A thing I forgot -- we'll eventually need to provide a [CA cert](https://gist.github.com/michaljemala/d6f4e01c4834bf47a9c4#file-tls-client-go-L28-L39) here as well. Maybe we can check for `BB_CA_CERT_FILE` and modify the `tls.Config` if the file exists/can be loaded? If nothing else, we should create a ticket to make sure this requirement is tracked!

karlmdavis: Maybe I'm not following, but the BB2 backend doesn't have a CA cert; it's self-signed.
You could use a CA for your client cert(s) if you wanted, but that's not currently a requirement from our side, either.

Now I'm wondering if we should start using an internal CA for these service-to-service client certs. Thoughts? (My current thinking: No, self-signed is probably still okay, maybe even preferable, as I don't think centrally managing them really gets us anything other than hassle & complexity right now.)

rnagle: I probably have my terminology wrong. What I want to accomplish: make sure that the self-signed cert presented by the BB server is the one that BCDA expects/trusts.

I agree with you on client certs -- I don't think there's a compelling reason to manage centrally atm.

bcda/client/bluebutton.go
embh: ✔️ https://jira.cms.gov/browse/BCDA-334

bcda/client/bluebutton.go
embh: Added ticket: https://jira.cms.gov/browse/BCDA-336 Would probably benefit from more detail. :)

Dockerfiles/Dockerfile.bcda
```diff
@@ -13,6 +13,10 @@ RUN dep ensure
 
 ENV BCDA_ERROR_LOG /var/log/bcda-error.log
 ENV BCDA_REQUEST_LOG /var/log/bcda-request.log
+ENV BCDA_BB_LOG /var/log/bcda-bb-request.log
+ENV BB_CLIENT_CERT_FILE client/test-cert.pem
+ENV BB_CLIENT_KEY_FILE client/test-key.pem
+ENV BB_SERVER_LOCATION https://fhir.backend.bluebutton.hhsdevcloud.us
```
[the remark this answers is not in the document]
rnagle: Yep!

bcda/api.go
```diff
@@ -173,6 +174,23 @@ func getToken(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func blueButtonMetadata(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
rnagle: Here you go: https://jira.cms.gov/browse/BCDA-231

bcda/router.go
```diff
@@ -29,6 +29,7 @@ func NewRouter() http.Handler {
 
 		if os.Getenv("DEBUG") == "true" {
 			r.Get("/token", getToken)
+			r.Get("/bb_metadata", blueButtonMetadata)
```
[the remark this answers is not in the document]
rnagle: We'll drop this endpoint in favor of `/api/v1/metadata` when tackling https://jira.cms.gov/browse/BCDA-231. Only using this as a sanity check to make sure auth to the backend is working properly.

bcda/client/bluebutton.go
```diff
@@ -0,0 +1,135 @@
...
+	certFile := os.Getenv("BB_CLIENT_CERT_FILE")
+	keyFile := os.Getenv("BB_CLIENT_KEY_FILE")
+	cert, err := tls.LoadX509KeyPair(certFile, keyFile)
+	if err != nil {
+		logger.Fatal(err)
+	}
+
+	tlsConfig := &tls.Config{
+		Certificates:       []tls.Certificate{cert},
+		InsecureSkipVerify: true,
```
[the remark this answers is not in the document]
embh: It doesn't work without that: "x509: certificate signed by unknown authority" So it looks like we need the certificate checking with this ticket.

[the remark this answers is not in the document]
karlmdavis: From that error message and reading through https://golang.org/pkg/crypto/tls/ a bit, it sounds like the test server cert you're trying to use isn't marked as trusted.

If I'm reading correctly, the right way to do that is to set the sibling `RootCAs` config option to the server's public key.

It looks like, for the current backend sandbox, that cert is auto-(re-)generated on each deploy [1]. That's a jerk move on my part, so apologies: you'll have either set `InsecureSkipVerify` to `true` for now or add some test bootstrapping code that pulls that public key at the start of your tests. I've filed https://jira.cms.gov/browse/BLUEBUTTON-484 to fix that on our end.

What I'd probably do for now is add a hardcoded-ish hack like:

```
// FIXME Karl's a jerk: https://jira.cms.gov/browse/BLUEBUTTON-484
verifyServerCert = (os.Getenv("BB_SERVER_LOCATION") == "https://fhir.backend.bluebutton.hhsdevcloud.us")

...

tlsConfig := &tls.Config{
		Certificates:       []tls.Certificate{cert},
		InsecureSkipVerify: !verifyServerCert,
		...
}

[1] https://github.com/CMSgov/bluebutton-ansible-playbooks-data-sandbox/blob/master/fhir_server_ami.yml#L86

karlmdavis:
:tada: Great work!

PR #32, opened 2018-09-26
bcda-220 feature:Add Swager Documentation 

knollfear:
Fixes [BCDA-220](https://jira.cms.gov/browse/BCDA-220)

Adds Swagger documentation to the project.  Existing routes are documented and documentation is served up by the app when it is running.
### Proposed changes:
Implement swagger code comments
Serve code comments up from the app when it is running

### Change Details
added swagger
included doc gen to build/run process
serve swagger docs 

### Security Implications

PII should be unaffected as this is just documentation

### Acceptance Validation

Ensure docs exist when app is running at localhost:3000/api/v1/swagger

### Feedback Requested

Any comments on the specific words are welcomed.  You are also welcome to just change them as you see fit.


Dockerfiles/Dockerfile.documentation
```diff
@@ -0,0 +1,11 @@
+FROM golang:1.10.3
+
+RUN curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
```
msnook: Replace with: `go get -u github.com/golang/dep/cmd/dep`

bcda/swaggerui/swagger.json
```diff
@@ -0,0 +1,300 @@
+{
```
msnook: Do we really want to commit this to the repo?  I was thinking the documentation would be generated dynamically?

Dockerfiles/Dockerfile.documentation
knollfear: all other docker files install dep in this manner.  Should I refactor all of them?

msnook: Yeah, if its not too much of a lift.  The ops files use the `go get` approach.

bcda/swaggerui/swagger.json
knollfear: It is generated dynamically.  It will be created fresh on docker-compose up. I included it in case anyone wanted to look at it offline or locally.  I can add to .gitignore if you think that would be better.

knollfear:
Why exactly are the curl statements removed in favor of go get?

My guess is if someone at dep changed the installdep.sh to do rm -rf * instead of actually install dep we would be in a really bad spot.

tbellj:
alrighty then I can see your reasoning. I thought at one point @msnook  and I tried that approach and ran into some issues but if it works it works.

PR #38, opened 2018-10-02
BCDA-292 Feature: Added fhir spec error responses

tbellj:
Fixes [BCDA-292](https://jira.cms.gov/browse/BCDA-292)

**Problem:**
The FHIR spec requires error responses to be formatted a certain way (as an operation outcome response). Previously we weren't meeting that spec requirement.

**Proposed changes:**
bcda/api.go
bcda/middleware.go
bcda/auth/middleware.go
All files above now include a call to create an operation outcome and write out the response via http.ResponseWriter.

All files below are new.
bcda/responseutils/issuedetails.go - const values that we use for describing the details of the error. This is where we will control how much information we want out there (No data leak). And this gives us a repeatable set of errors to use throughout our application.
bcda/responseutils/issueseverity.go - const values that mirror the required value set of an operation outcome IssueSeverity.
bcda/responseutils/issuetype.go - const values that mirror the required value set of an operation outcome IssueType.
bcda/responseutils/writer.go - functions for creating/writing an operation outcome

**Security Implications:**
No security concerns. Would like reviewers to take a look at the issue details and see if that's enough information for diagnosing an issue without causing a data leak.

**Feedback Requested:**
Any


bcda/api.go
```diff
@@ -200,7 +210,7 @@ func jobStatus(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("X-Progress", job.Status)
 		w.WriteHeader(http.StatusAccepted)
 	case "Failed":
-		writeError(fhirmodels.OperationOutcome{}, w)
+		responseutils.WriteError(&fhirmodels.OperationOutcome{}, w, http.StatusOK)
```
embh: Should this be OK or an error code? Based on https://github.com/smart-on-fhir/fhir-bulk-data-docs/blob/master/export.md#response---error-status-1 it seems like an error response would be correct (different from Completed but with some errors which is 200)

bcda/auth/middleware.go
```diff
@@ -94,7 +99,8 @@ func RequireTokenACOMatch(next http.Handler) http.Handler {
 			if uuid.Equal(uuid.Parse(aco), uuid.Parse(string(urlUUID))) {
 				next.ServeHTTP(w, r)
 			} else {
-				http.Error(w, http.StatusText(http.StatusNotFound), http.StatusNotFound)
+				oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, "", responseutils.TokenErr)
+				responseutils.WriteError(oo, w, http.StatusUnauthorized)
```
embh: It's true that the user is unauthorized, but I'd lean toward keeping this as a 404. My concern is that a request from someone with the wrong credentials would show that the file does exist by giving a 401 rather than a 404. (Like how S3 handles it, if I remember correctly.) @rnagle or @msnook, what do you think?

bcda/api.go
tbellj: Yeah I thought about this for a while. Here's my reasoning, the response for the job status itself is still a successful response, we're just returning a "failed status" for that particular job. So should this really be an error response for the polling itself? They also don't mention anything about a "fail/failure" on the bulk data status page but mix in language of "in-progress, complete, and error" which in my opinion causes this type of confusion. If an actual error happens then that will be an operation outcome (before the switch case) and if a "job failed" happens then we will have to write out the operation outcome to file(s) and list them in the response.error array while still capturing the successful jobs.

tbellj: So basically an overall job can still succeed while failing on some of the file downloads. If we actually wrote out an error response then our users would have no way of accessing the "successful" files from the output array of a complete status, and we'll essentially be throwing away all of the good files if one doesn't make it through. 1 job can be x number of files to be downloaded. Does this make sense?

tbellj: good catch, will change this. My thinking was off

bcda/api.go
```diff
@@ -182,14 +190,16 @@ func jobStatus(w http.ResponseWriter, r *http.Request) {
...
+		responseutils.WriteError(oo, w, http.StatusBadRequest)
 		return
 	}
 	var job models.Job
 	err = db.First(&job, i).Error
 	if err != nil {
 		log.Print(err)
-		http.Error(w, http.StatusText(404), 404)
+		oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, "", responseutils.DbErr)
+		responseutils.WriteError(oo, w, http.StatusInternalServerError)
```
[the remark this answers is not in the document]
tbellj: Interesting. So I've always thought about 404 being like a missing web page due to a bad url link or something. But maybe it can also represent a missing job in our database?? Not opposed to changing this

bcda/auth/middleware.go
tbellj: I agree with this, 404 is safer

msnook: Agree.  I think we should keep this as 404.

PR #41, opened 2018-10-03
Testing Env Vars, Status Propagation, and Test Case Cleanup

msnook:
### Fixes Test Status Propagation Issues

Test status was not being propagated correctly.  See here: https://travis-ci.org/CMSgov/bcda-app/builds/436835705
Additionally, cleaning up some test cases and making sure our tests have the appropriate env vars.

### Proposed changes:

- Use `bash` instead of `sh` for better error handling control.
- Ensure appropriate environment variables are pre-set in unit test image.
- Fix some flaky test cases.

### Change Details

- Update Dockerfile to executed unit test script with `bash`
- Update unit test script to `set -o pipeline` to ensure the worst status code from a pipeline is propagated
- Updated Dockerfile with additional pre-set environment variables (as the system would have)
- Fixed some test cases

### Security Implications

None

### Acceptance Validation

Tests are passing in Travis.

### DIFF
--- a/bcda/main_test.go
+++ b/bcda/main_test.go
@@ -80,42 +80,42 @@ func (s *MainTestSuite) TestCreateACO() {
 	// Blank UUID
 	badUserUUID, err := createUser("", name, email)
 	assert.NotNil(s.T(), err)
-	assert.Equal(s.T(), badUserUUID, "")
+	assert.Equal(s.T(), "", badUserUUID)
 
 	// Blank UUID
 	badUserUUID, err = createUser(BADUUID, name, email)
 	assert.NotNil(s.T(), err)
-	assert.Equal(s.T(), badUserUUID, "")
+	assert.Equal(s.T(), "", badUserUUID)
 
 	// Blank Name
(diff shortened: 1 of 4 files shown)


bcda/main_test.go
```diff
@@ -80,42 +80,42 @@ func (s *MainTestSuite) TestCreateACO() {
 	// Blank UUID
 	badUserUUID, err := createUser("", name, email)
 	assert.NotNil(s.T(), err)
-	assert.Nil(s.T(), badUserUUID)
+	assert.Equal(s.T(), badUserUUID, "")
```
embh: Would it be possible to swap the order on these `assert.Equal()` calls to have the expected value come before actual? https://github.com/stretchr/testify/blob/master/assert/assertions.go#L328

msnook: Done!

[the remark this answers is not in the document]
knollfear: sry, this was last night.  I thought I sent it in but apparently not

PR #42, opened 2018-10-04
BCDA-231 Feature: Implemented metadata endpoint for capability statement

tbellj:
Fixes [BCDA-231](https://jira.cms.gov/browse/BCDA-231)

Problem:
We needed a capability statement (FHIR compliant) as to what our API can do. And now we have one!!

Proposed changes:
router.go - added metadata endpoint
api.go - added http handler function to call for the creation of the capability statement
writer.go - added function for creating a capability statement and for writing the response.

Security Implications
None

Acceptance Validation
Try to hit the "/api/v1/metadata" and see if a capability statement is generated.

Feedback Requested
Any

### DIFF
--- a/bcda/api.go
+++ b/bcda/api.go
@@ -307,3 +307,8 @@ func blueButtonMetadata(w http.ResponseWriter, r *http.Request) {
 
 	w.WriteHeader(http.StatusOK)
 }
+
+func metadata(w http.ResponseWriter, r *http.Request) {
+	statement := responseutils.CreateCapabilityStatement()
+	responseutils.WriteCapabilityStatement(statement, w)
+}
--- a/bcda/api.go
+++ b/bcda/api.go
@@ -40,6 +40,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"time"
 
 	"github.com/CMSgov/bcda-app/bcda/auth"
 	"github.com/CMSgov/bcda-app/bcda/client"
@@ -309,6 +310,7 @@ func blueButtonMetadata(w http.ResponseWriter, r *http.Request) {
 }
 
 func metadata(w http.ResponseWriter, r *http.Request) {
-	statement := responseutils.CreateCapabilityStatement()
+	dt := time.Now()
+	statement := responseutils.CreateCapabilityStatement(dt, "0.1")
 	responseutils.WriteCapabilityStatement(statement, w)
 }
--- a/bcda/api.go
+++ b/bcda/api.go
@@ -311,6 +311,12 @@ func blueButtonMetadata(w http.ResponseWriter, r *http.Request) {
 
 func metadata(w http.ResponseWriter, r *http.Request) {
 	dt := time.Now()
-	statement := responseutils.CreateCapabilityStatement(dt, "0.1")
+
+	scheme := "http"
+	if r.TLS != nil {
+		scheme = "https"
+	}
+	host := fmt.Sprintf("%s://%s", scheme, r.Host)
+	statement := responseutils.CreateCapabilityStatement(dt, "0.1", host)
 	responseutils.WriteCapabilityStatement(statement, w)
 }
--- a/bcda/responseutils/writer.go
+++ b/bcda/responseutils/writer.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	fhirmodels "github.com/eug48/fhir/models"
 	"net/http"
+	"time"
 )
 
 func CreateOpOutcome(severity, code, detailsCode, detailsDisplay string) *fhirmodels.OperationOutcome {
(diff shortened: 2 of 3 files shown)


bcda/router.go
```diff
@@ -32,6 +32,7 @@ func NewRouter() http.Handler {
 		if os.Getenv("DEBUG") == "true" {
```
msnook: Is this only supposed to be enabled in debug mode?

bcda/responseutils/writer.go
```diff
@@ -22,4 +23,99 @@ func WriteError(outcome *fhirmodels.OperationOutcome, w http.ResponseWriter, cod
...
+				},
+				Interaction: []fhirmodels.CapabilityStatementSystemInteractionComponent{
+					{
+						Code: "batch",
+					},
+					{
+						Code: "search-system",
+					},
+				},
+				Operation: []fhirmodels.CapabilityStatementRestOperationComponent{
```
msnook: Maybe redundant, but should we also list this endpoint in the Capability Statement?  (`/metadata`)

bcda/responseutils/writer.go
```diff
@@ -22,4 +23,99 @@ func WriteError(outcome *fhirmodels.OperationOutcome, w http.ResponseWriter, cod
...
+	baseurl := "<host>"
+	statement := &fhirmodels.CapabilityStatement{
+		Status:       "active",
+		Date:         &fhirmodels.FHIRDateTime{Time: dt, Precision: fhirmodels.Date},
+		Publisher:    "Centers for Medicare & Medicaid Services",
+		Kind:         "capability",
+		Instantiates: []string{"https://api.bluebutton.cms.gov/v1/fhir/metadata"},
+		Software: &fhirmodels.CapabilityStatementSoftwareComponent{
+			Name:        "Beneficiary Claims Data API",
+			Version:     "0.1",
```
msnook: Should version number be configurable for dynamic updating?

tbellj: I like it! I will change this as well

bcda/responseutils/writer.go
tbellj: good point! I'll add this.

bcda/router.go
tbellj: Is it?? I'm not too sure either way.

embh: It should be accessible in all environments.

tbellj: Will do

bcda/responseutils/writer.go
```diff
@@ -22,4 +23,105 @@ func WriteError(outcome *fhirmodels.OperationOutcome, w http.ResponseWriter, cod
...
-func WriteCapabilityStatement() {}
+func CreateCapabilityStatement(reldate time.Time, relversion string) *fhirmodels.CapabilityStatement {
+	usecoors := true
+	baseurl := "<host>"
+	statement := &fhirmodels.CapabilityStatement{
+		Status:       "active",
+		Date:         &fhirmodels.FHIRDateTime{Time: reldate, Precision: fhirmodels.Date},
+		Publisher:    "Centers for Medicare & Medicaid Services",
+		Kind:         "capability",
+		Instantiates: []string{"https://api.bluebutton.cms.gov/v1/fhir/metadata"},
```
rnagle: This url is for the BB 2.0 frontend, but we are interacting with the BB 2.0 backend. Can you either drop the `Instantiates` key or have it use the value of the env var [`BB_SERVER_LOCATION`](https://github.com/CMSgov/bcda-app/blob/master/Dockerfiles/Dockerfile.bcda#L19)?

bcda/responseutils/writer.go
```diff
@@ -22,4 +23,105 @@ func WriteError(outcome *fhirmodels.OperationOutcome, w http.ResponseWriter, cod
 	}
 }
 
-func WriteCapabilityStatement() {}
+func CreateCapabilityStatement(reldate time.Time, relversion string) *fhirmodels.CapabilityStatement {
+	usecoors := true
+	baseurl := "<host>"
```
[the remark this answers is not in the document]
tbellj: nice, will do this

bcda/responseutils/writer.go
```diff
@@ -22,4 +23,105 @@ func WriteError(outcome *fhirmodels.OperationOutcome, w http.ResponseWriter, cod
...
+					},
+					{
+						Name: "jobs",
+						Definition: &fhirmodels.Reference{
+							Reference: baseurl + "/api/v1/jobs/{jobId}",
+							Type:      "Endpoint",
+						},
+					},
+					{
+						Name: "token",
```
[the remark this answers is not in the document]
tbellj: good catch.

bcda/responseutils/writer.go
tbellj: cool, will do

bcda/router.go
```diff
@@ -29,6 +29,7 @@ func NewRouter() http.Handler {
 	r.Route("/api/v1", func(r chi.Router) {
 		r.With(auth.RequireTokenAuth).With(ValidateBulkRequestHeaders).Get("/Patient/$export", bulkRequest)
 		r.With(auth.RequireTokenAuth).Get("/jobs/{jobId}", jobStatus)
+		r.Get("/metadata", metadata)
```
msnook: Probably a question for @rnagle ... is it OK to expose this endpoint without requiring a token to access?

bcda/responseutils/writer.go
```diff
@@ -22,4 +24,91 @@ func WriteError(outcome *fhirmodels.OperationOutcome, w http.ResponseWriter, cod
...
+		Software: &fhirmodels.CapabilityStatementSoftwareComponent{
+			Name:        "Beneficiary Claims Data API",
+			Version:     relversion,
+			ReleaseDate: &fhirmodels.FHIRDateTime{Time: reldate, Precision: fhirmodels.Date},
+		},
+		Implementation: &fhirmodels.CapabilityStatementImplementationComponent{
+			Description: "",
+			Url:         baseurl,
+		},
+		FhirVersion:   "3.0.1",
```
msnook: If we are going to hardcode this `FhirVersion`, maybe we need to create a constant that can be referenced by other apps?  Do we envision this to change frequently?  If so, we might need this to be an env var.

bcda/api.go
```diff
@@ -307,3 +308,15 @@ func blueButtonMetadata(w http.ResponseWriter, r *http.Request) {
...
+
+func metadata(w http.ResponseWriter, r *http.Request) {
+	dt := time.Now()
+
+	scheme := "http"
+	if r.TLS != nil {
+		scheme = "https"
+	}
+	host := fmt.Sprintf("%s://%s", scheme, r.Host)
+	statement := responseutils.CreateCapabilityStatement(dt, "0.1", host)
```
msnook: Hardcoding version 0.1 ?

tbellj: well its a parameter value that can be changed to whatever we want in the future. its really a placeholder at the moment

bcda/responseutils/writer.go
tbellj: Not sure about how often the fhir version will change, would like @rnagle feedback on this

bcda/api.go
msnook: Yeah, I think the concern is that it is hardcoded in the code, which means we have to change it whenever the version might change.  We'll probably run into this in other tasks as well (see BCDA-317).

bcda/responseutils/writer.go
rnagle: This value is not likely to change often. We can shield ourselves from providing the wrong or outdated value by using the corresponding value from the BB 2.0 backend metadata endpoint.

bcda/api.go
rnagle: It's possible to provide a version string at compile time so that we can keep this up to date.

For example:

```
go build -ldflags "-X main.version=$(VERSION_STRING)"
```

And in `package main`, define:

```
var version string
```

Then use `version` where appropriate.

As long as there are no objections, I say we leave this as-is for now. I will add this info to BCDA-317 and note that, once this is incorporated, we should also update the `CapabilityStatement` to use the value of `version`.

msnook: Works for me!

bcda/responseutils/writer.go
tbellj: yep that's exactly where I took it from

bcda/api.go
tbellj: yep yep

bcda/responseutils/writer.go
rnagle: I guess what I mean is, if we fetch the backend's capability statement here, we can use its `FhirVersion` and not have to worry about updating this manually.

tbellj: true, but imho I think that's overkill for this especially since relying on bb metadata endpoint for data might not be the most accurate thing to do. Some of the data in their own capability statement is "shaky" like broken url's for example. But I'm not opposed to doing this if that's what we want to do.

karlmdavis: @tbellj Did you document the issues you found? If not, could you please? Ideally in a JIRA ticket.

PR #47, opened 2018-10-12
BCDA-313: Develop unit tests for bcda/api.go, part 2

embh:
### Fixes [BCDA-313](https://jira.cms.gov/browse/BCDA-xxx)
More unit tests for api.go.

### Proposed changes:
Several new or expanded tests in api_test.go

### Change Details
* Tests:
   * `bulkRequest()`
      * success response
      * no beneficiaries in the ACO
      * user ID not found in database
      * queue not created
   * `jobStatus()`
      * invalid job ID (string)
      * job ID valid but does not exist
      * completed with errors
   * `getToken()` success
   * `getBlueButtonMetadata()`
      * client error
   * `metadata()` success
* Added 50 beneficiaries to test database creation
* Added queue to test database

### Security Implications
All synthetic data.

### Acceptance Validation
Reached 66.9%, which is short of the goal of 80%. A lot of edge cases for errors are not done, many of which would be tricky to force an error for. I think this is something we can continue to iterate on.

screen shot 2018-10-12 at 5 11 06 pm

### Feedback Requested
Any, though someone else may need to make needed changes due to my being mostly unavailable next week


bcda/api_test.go
```diff
@@ -31,14 +39,225 @@ func (s *APITestSuite) SetupTest() {
...
+		s.T().Error(err)
+	}
+
+	token := jwt.New(jwt.SigningMethodRS512)
+	token.Claims = jwt.MapClaims{
+		"sub": user.UUID.String(),
+		"aco": acoID,
+		"id":  uuid.NewRandom().String(),
+	}
+	token.Valid = true
```
knollfear: This is a weird usage.  I think this is normally set internally when the token is parsed/verified

bcda/api_test.go
```diff
@@ -31,14 +39,225 @@ func (s *APITestSuite) SetupTest() {
...
+	defer pgxpool.Close()
+
+	qc = que.NewClient(pgxpool)
+
+	handler := http.HandlerFunc(bulkRequest)
+	handler.ServeHTTP(s.rr, req)
+
+	assert.Equal(s.T(), http.StatusAccepted, s.rr.Code)
+	s.db.Where("user_id = ?", user.UUID).Delete(models.Job{})
+	s.db.Where("uuid = ?", user.UUID).Delete(auth.User{})
```
knollfear: why delete these?  We should be able to continue with these rows in the DB.

bcda/api_test.go
embh: I thought so too, but it didn't seem to get set when creating it manually

bcda/api_test.go
embh: Just for the sake of keeping the test database clean-- having extra data could impact other tests

knollfear: I understand.  I just feel like if randomly created data does impact other tests we should figure that out now because we will have multiple users and jobs when this goes live and if things are colliding that will be really bad.

msnook: Standard test practices are that test cases should execute independently.  We do not want to go down a path where test cases executed after this one start relying on test data that was created as part of this test case.  Test suites (or even better, test cases) should be responsible for setting up and tearing down their own data.
Knoll, for your concern regarding test data affecting multiple users/jobs... this is a use case for integration testing.  For now, all we care about in the unit test exercise is whether or not the module/function is working properly.

knollfear: I get that future tests should not rely on this data and I am not trying to make them do so.  However, we will almost certainly NOT be deleting job records in production. If this was data that would regularly be pruned deleting makes sense but we will likely never delete job records for any reason. Therefore, it is important that having multiple unrelated jobs in the DB have no impact or else some other unit is not working as expected in a way that our testing process ignores by design.  With our focus on unit test coverage and cleaning data we will miss problems of a specific type.  We aren't currently doing any integration testing and it isn't on the PI7 board in any meaningful way so I don't think relying on them is feasible.

PR #53, opened 2018-10-22
BCDA-409 Feature Format BB data to NDJson

knollfear:
Refactoring how data provider works for blue button client to allow for it to provide static files off the file system.  

Fixing up data processing methods to properly create NDJson data

Fixes [BCDA-409](https://jira.cms.gov/browse/BCDA-409)

Files from Blue Button need to be formatted into proper NDJSON according to the FHIR spec 
### Proposed changes:

Create different interface for getData method to pull from some fixed data off of the file system
Format the JSON data to NDJson by unmarshalling it and only writing out the entries in the file.

### Security Implications

mostly affects test data or data that was otherwise exposed to clients.

### Acceptance Validation

FIle produced by writeEOBdataToFile should be in actual ndjson format


bcda/client/bluebutton.go
```diff
@@ -73,34 +73,34 @@ func NewBlueButtonClient() (*BlueButtonClient, error) {
 	return &BlueButtonClient{*client}, nil
 }
 
-func (bbc *BlueButtonClient) GetPatientData(patientID string) (string, error) {
+func GetPatientData(patientID string, bbc APIClient) (string, error) {
```
embh: Does this have to be a parameter instead of keeping the `BlueButtonClient` receiver? I thought that types implementing the interface could have additional functions, so `BlueButtonClient` would have to have `getData()` from `APIClient` but could have these other functions as well. (As a Go newbie I could be totally wrong here.)

bcdaworker/main.go
```diff
@@ -162,6 +158,33 @@ func appendErrorToFile(acoID, code, detailsCode, detailsDisplay string) {
...
+	err := json.Unmarshal([]byte(jsonData), &jsonOBJ)
+	if err != nil {
+		log.Error(err)
+		appendErrorToFile(acoID, responseutils.Exception, responseutils.InternalErr, fmt.Sprintf("Error UnMarshaling %s from data for beneficiary %s in ACO %s", jsonType, beneficiaryID, acoID))
+		return
+	}
+
+	entries := jsonOBJ["entry"]
+	for _, entry := range entries.([]interface{}) {
+		entryJSON, err := json.Marshal(entry)
```
karlmdavis: This is probably me just not knowing Go (at all!) but it doesn't look like you're actually doing anything with the data you've marshalled here?

shared_files/baseDstu3/Coverage
```diff
@@ -0,0 +1 @@
..."part-b-20000000000001","extension":[{"url":"https://bluebutton.cms.gov/resources/variables/ms_cd","valueCoding":{"system":"https://bluebutton.cms.gov/resources/variables/ms_cd","code":"10","display":"Aged without end-stage renal disease (ESRD)"}},{"url":"https://bluebutton.cms.gov/resources/variables/b_trm_cd","valueCoding":{"system":"https://bluebutton.cms.gov/resources/variables/b_trm_cd","code":"0","display":"Not Terminated"}}],"status":"active","type":{"coding":[{"system":"Medicare","code":"Part B"}]},"beneficiary":{"reference":"Patient/20000000000001"},"grouping":{"subGroup":"Medicare","subPlan":"Part B"}}},{"fullUrl":"https://fhir.backend.bluebutton.hhsdevcloud.us/baseDstu3/Coverage/part-d-20000000000001","resource":{"resourceType":"Coverage","id":"part-d-20000000000001","extension":[{"url":"https://bluebutton.cms.gov/resources/variables/ms_cd","valueCoding":{"system":"https://bluebutton.cms.gov/resources/variables/ms_cd","code":"10","display":"Aged without end-stage renal disease (ESRD)"}}],"status":"active","type":{"coding":[{"system":"Medicare","code":"Part D"}]},"beneficiary":{"reference":"Patient/20000000000001"},"grouping":{"subGroup":"Medicare","subPlan":"Part D"}}}]}
```
karlmdavis: All three of these files should be clearly marked as sample data (with no PII/PHI). I'd suggest:

1. Adding a `README.md` to the same directory that explains where they came from (and how to regenerate them, if needed).
2. Renaming the directory "`sample_backend_synthetic_results`" (or some hopefully-less-wordy version of that).

bcdaworker/main_test.go
```diff
@@ -137,11 +137,28 @@ func TestAppendErrorToFile(t *testing.T) {
...
-	return eobData, nil
+// Returns copy of a static json file (From Blue Button Sandbox originally) after replacing the patient ID of 20000000000001 with the requested identifier
+
+func (bbc *MockBlueButtonClient) GetData(path string, params url.Values) (string, error) {
+	var identifier string
+	if strings.Contains(path, "Coverage") {
+		identifier = "beneficiary"
+	} else if strings.Contains(path, "Patient") {
+		identifier = "_id"
+	} else if strings.Contains(path, "ExplanationOfBenefits") {
```
embh: Should this be singular ("ExplanationOfBenefit")?

bcdaworker/main.go
```diff
@@ -162,6 +158,33 @@ func appendErrorToFile(acoID, code, detailsCode, detailsDisplay string) {
 	}
 }
 
+func jsonToNDJSON(w *bufio.Writer, jsonData, jsonType, beneficiaryID, acoID string) {
```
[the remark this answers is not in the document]
knollfear: I can change it.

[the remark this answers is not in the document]
knollfear: This function will always write something to the file and doesn't return anything including errors.  AppendErrorToFile writes a json formatted error message where the record would be. My feeling is that this is one of a group of records so if we have one or two bad ones we should log that and keep trucking and not blow up the process.  If it blows up in the unmarshalling process (163) it returns and never tries to write any entries. This is where I think it most likely to fail.

I think the likelihood of it failing on 172 or 186 is exceedingly low, but the methods require that I handle the error.

bcdaworker/main.go
knollfear: entries is the in memory map of just the values within entry.  I have to marshal that back to JSON before I write it out to the file(line 179).  I can't just do string(entry) because it wouldn't be in the right format.

bcdaworker/main_test.go
knollfear: yes.  FIxing.

shared_files/baseDstu3/Coverage
knollfear: The file paths are mirroring the Blue Button url pattern so I can use the path variable to find the files.  I will add a readme.md to this dir explaining what is going on.

bcda/client/bluebutton.go
knollfear: Interfaces can't have methods so unless I want to rewrite each of the GetPatientData, GetCoverageData .... methods for my mock client I have to pass it in.  I may also just be doing inheritance incorrectly in Go though.

embh: Maybe a few months from now when we're more Go-knowledgable we'll look back at this and say, "Oh, we totally could've used ____!" 😄 But this makes sense for being able to use the mock client. 👍

knollfear: probably.  I look forward to that day.

shared_files/baseDstu3/Coverage
knollfear: I added the read me and also made new copies of these for the new BB url schema.

bcdaworker/main.go
karlmdavis: Is there a streaming JSON processor available for Go that could do this without marshalling the whole thing to RAM? Some of these are going to be... big.

If not, I think this section of the code needs a comment explaining that this harmless-looking line could eat tons of RAM (especially when run concurrently across bunches of goroutines!):

    json.Unmarshal([]byte(jsonData), &jsonOBJ)

(Suggestion: please follow up and ask something like, "Backend peeps, do we know how big we expect those NDJSON files to be in the average, 90% and worst cases?" Preliminary answer that's framed in terms of number of EOBs rather than bytes: "Claim Counts" on <https://confluence.cms.gov/display/BB/Blue+Button+Numbers>)

bcdaworker/main.go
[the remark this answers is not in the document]
karlmdavis: > My feeling is that this is one of a group of records so if we have one or two bad ones we should log that and keep trucking and not blow up the process.

**Strong** disagree here. We need to be very very careful to always return correct **and complete** results to ACOs here. It's much better for us to return only an error than for us to return incomplete results.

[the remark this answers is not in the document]
knollfear: I think we definitely need and have agreed to some retry options for the get Data process.  However, for the scope of this ticket I need to just expect that I am getting valid data that can be collapsed effectively.  I think especially as we operate in alpha with only synthetic data returning something is actually preferred to completeness even if that isn't the ultimate answer in production.  We shouldn't be holding up our end users process unnecessarily if we have an issue internally.  By returning data and alerting us we can have the best of both worlds in our sandbox.

bcdaworker/main.go
```diff
@@ -158,6 +154,38 @@ func appendErrorToFile(acoID, code, detailsCode, detailsDisplay string) {
...
+	// There might be no entries.  If this happens we can't iterate over them.
+	if entries != nil {
+
+		for _, entry := range entries.([]interface{}) {
+			entryJSON, err := json.Marshal(entry)
+			// This is unlikely to happen because we just unmarshalled this data a few lines above.
+			if err != nil {
+				log.Error(err)
+				appendErrorToFile(acoID, responseutils.Exception, responseutils.InternalErr, fmt.Sprintf("Error Marshaling %s to Json for beneficiary %s in ACO %s", jsonType, beneficiaryID, acoID))
+				continue
```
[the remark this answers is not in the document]
knollfear: Yeah, the code requires that I handle it.  I could ignore the error with _ but that is generally a sign that code has been written poorly.

PR #58, opened 2018-10-25
BCDA-415 Create Access Tokens for Alpha Partners

DeirdreHolub:
### Proposed changes:

- Adds a 'create-alpha-token' command to the CLI
- Implements that command in auth/backend.go

The plan is that a human will use the CLI to create tokens for our alpha participants in the alpha sandbox. For more details and discussion, see the Jira Ticket.

### Change Details

- Creates an ACO entry with a fixed name format like 'Alpha ACO #'.
- Creates a User entry with a fixed name format like 'Alpha User#' and email like alpha.user.#@nosuchdomain.com
- Copies the 50 beneficiaries associated with 'ACO dev' and associates those copies with the new ACO
- Creates an authentication token
- Displays the ACO name and user name in plain text, and the token in base64 encoded string form

### Security Implications

We chose to generate synthetic ACO and User names and email addresses to avoid any issues with PII.

Also, since we generate the token and sign it with our key, we can know for sure that the key has not been changed since we issued it.

### Acceptance Validation

I have run the whole process from end-to-end locally. Here's the new command shown in the CLI's help message:

screen shot 2018-10-24 at 10 08 37 pm

I am a huge fan of CLIs for services; they make operationalizing them so much easier! Kudos team! 

Ran the CLI command locally with this docker command:
`docker exec -it bcda-app_api_1 bash -c 'tmp/bcda create-alpha-token'`

The command output looks like this:
screen shot 2018-10-24 at 11 05 22 pm

The token is the long string on the third line in 3 parts separated by '.'

I then used Postman to submit requests to 
- http://localhost:3000/api/v1/ExplanationOfBenefit/$export
- http://localhost:3000/api/v1/jobs/1
- http://localhost:3000/data/ff273da5-9881-4a60-a2f0-83d04bd296e0.ndjson
  - I stopped scrolling through at line 379193
- http://localhost:3000/data/ff273da5-9881-4a60-a2f0-83d04bd296e0-error.ndjson

### Feedback Requested

@epfrankle @michaelrobertsutton  we will need to correlate between the synthetic names / tokens and the actual ACO that is using each one. Do we need any more information?

@rnagle @embh @knollfear @tbellj thoughts / comments / questions?

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,10 +15,11 @@ To get started, install some dependencies:
 
 Build the images and start the containers:
 
-1. Build the images
+1. Build the images and load with fixture data
 ```sh
 make docker-bootstrap
 ```
+
 2. Start the containers
 ```sh
 docker-compose up
@@ -36,6 +37,7 @@ make test
 ### Use the application
 
 See: [API documentation](https://github.com/CMSgov/bcda-app/blob/master/API.md)
+
 
 ### Environment variables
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -65,3 +67,22 @@ BB_SERVER_LOCATION <url>
 FHIR_PAYLOAD_DIR <directory_path>
 BB_TIMEOUT_MS <integer>
 ```
+
+### Other things you can do
+
+Use docker to look at the api database with psql:
+```sh
+docker run --rm --network bcda-app_default -it postgres psql -h bcda-app_db_1 -U postgres bcda
```
embh: Thank you for adding this database info! I've had to do it a hundred times and never thought to document it.

PR #63, opened 2018-11-02
BCDA-485 Feature: version and data added to metadata endpoint

tbellj:
Fixes [BCDA-485](https://jira.cms.gov/browse/BCDA-485)

Problem:
When querying the metadata endpoint the version and data endpoints were not being included

Proposed changes:
bcda/responseutils/writer.go - added two more endpoints to our capability statement

Security Implications:
None

Acceptance Validation:
hit the api/v1/metadata endpoint and see the updated response

Feedback Requested
Any


bcda/responseutils/writer.go
```diff
@@ -95,6 +96,20 @@ func CreateCapabilityStatement(reldate time.Time, relversion, baseurl string) *f
...
+						Name: "version",
+						Definition: &fhirmodels.Reference{
+							Reference: baseurl + "/_version",
+							Type:      "Endpoint",
+						},
+					},
+					{
+						Name: "data",
+						Definition: &fhirmodels.Reference{
+							Reference: baseurl + "/data/{jobID}/{acoID}.ndjson",
```
rnagle: I couldn't find a reference in the FHIR docs explaining how to indicate variables in operation urls. Do you have a link you can share?

tbellj: Good question! I'm sure I saw it somewhere in the FHIR spec. The base spot for everything is here https://www.hl7.org/fhir/capabilitystatement.html. Meanwhile I'll try to find a specific example

rnagle: Found this, but I have no idea what it means: https://www.hl7.org/fhir/references-definitions.html#Reference.reference

tbellj: I think this covers us. 
_**" Absolute URLs do not need to point to a FHIR RESTful server, though this is the preferred approach. If the URL conforms to the structure "/[type]/[id]" then it should be assumed that the reference is to a FHIR RESTful server."**_

embh: I can't find anything definitive on this, but I'm wondering if we should at least switch it to brackets around the URL params instead of braces, because that's the format they use in various places in the documentation.

rnagle: I'm wondering the same.

tbellj: ahh yes I agree with this. For some reason I read it and didn't realize the difference between {} and []

PR #66, opened 2018-11-06
BCDA-441: Implement expiration status for jobs after 24 hours

tbellj:
Fixes [BCDA-441](https://jira.cms.gov/browse/BCDA-441)

Problem:
Before we had no way of setting expired statuses for our jobs, or moving files once they became expired.

Proposed changes:
Dockerfiles/Dockerfile.bcdaworker - added a directory and a time interval for running the expired updates.
bcdaworker/main.go - added the cron job and the update logic for setting statuses. 
vendor files - added cron library.

Security Implications:
No PII, but files should/need to be moved to an inaccessible location once expired. This ticket should implement that in our AWS env.

Acceptance Validation:
Yes

Feedback Requested:
Any


bcdaworker/main.go
```diff
@@ -293,5 +295,61 @@ func main() {
...
+
+			err = os.Rename(jobDir, expDir)
+			if err != nil {
+				log.Error(err)
+			}
+		}
+	}
+}
+
+func expiredScheduler() {
```
rnagle: We could very easily use `crontab` handle scheduling and execution of this routine. That would mean we could drop the `github.com/robfig/cron` dependency. It would require refactoring this to be a CLI command instead.

bcdaworker/main.go
```diff
@@ -293,5 +295,61 @@ func main() {
 	fmt.Println("Starting bcdaworker...")
 	workerPool := setupQueue()
 	defer workerPool.Close()
+	expiredScheduler()
 	waitForSig()
 }
+
+func updateStatus(hrThreshold int) {
```
rnagle: Consider that there will be multiple instances of this app running across many hosts. What happens if this routine runs at the same time across said hosts? Will this introduce a race condition?

msnook: Yes I think, at the very least, lines 313, 325, and 334 are susceptible to a race condition between workers.  It may be best to leverage a system-wide cron instead of an app-specific cron.

msnook:
A couple items that we might still need to address:

1. System-level cron vs app-specific cron to avoid race condition
2. Acceptance criteria for BCDA-441 mentions "Timestamp added to the job table for the initial request time".  I didn't see that implemented here.  Is that already in place?  Without it, how will we know which jobs have expired?
3. How do we intend to move the files to an inaccessible location?

bcdaworker/main.go
```diff
@@ -293,5 +295,61 @@ func main() {
...
+			log.Error(err)
+		}
+	}
+
+	for _, j := range jobs {
+		t := j.CreatedAt
+		elapsed := time.Since(t).Hours()
+		if int(elapsed) >= hrThreshold {
+			j.Status = "Expired"
+			err = db.Save(j).Error
```
karlmdavis: This isn't atomic or robust in the face of failures: If this code dies halfway through, you will have marked the job as expired in the DB but left the files orphaned on the disk.

tbellj: really good point! im thinking about how to refactor this to be atomic but if you have any suggestions let me know.

bcdaworker/main.go
tbellj: Okay cool, I'm unfamiliar but I'll definitely look into cron tab

msnook: Its something we would configure on the instance and not within the app.  When it triggers on the system, it would invoke the app to do the file moving.

bcdaworker/main.go
tbellj: @rnagle:
Good point, definitely something I should have thought about. 

@msnook, to the list below:
1. Yes!! A system level cron is where it's at. Using Ryan's approach for the cron tab in combination with a system level cron you suggest would handle the race condition issue.

2. The timestamp part was already implemented so we're good there. Implemented lines 321 - 323
3. That's in the code, line 331 - 334.

bcdaworker/main.go
msnook: https://docs.ansible.com/ansible/2.5/modules/cron_module.html

msnook:
Terrance and I discussed... current plan is to configure a cron job via the bcda-api playbook: https://github.com/CMSgov/bcda-ops/blob/master/playbooks/api/main.yml

The cron will be configured in such a way that when it triggers, it calls `bcda remove-expired-jobs`, which is wired to execute Terrance's logic.

Feedback requested on this design decision...

bcda/main.go
```diff
@@ -379,3 +401,49 @@ func getEnvInt(varName string, defaultVal int) int {
...
+			log.Error(err)
+		}
+	}
+
+	for _, j := range jobs {
+		t := j.CreatedAt
+		elapsed := time.Since(t).Hours()
+		if int(elapsed) >= hrThreshold {
+			j.Status = "Expired"
+			err = db.Save(j).Error
```
knollfear: I would just move this way down to the bottom of the code and then make sure the work succeeded before marking it as having succeeded.  Worst case there is you move the files, but leave the job as completed.

bcda/main.go
```diff
@@ -379,3 +401,49 @@ func getEnvInt(varName string, defaultVal int) int {
...
+			if err != nil {
+				log.Error(err)
+			}
+
+			id := strconv.Itoa(int(j.ID))
```
knollfear: this seems complicated.  Why not just do string(j.ID)?  or jobDir := fmt.Sprintf("%s/%d", os.Getenv("FHIR_PAYLOAD_DIR"), j.ID) works as well.

bcdaworker/main.go
knollfear: +1 for system level cron.  As long as we run it on only one machine it seems fine to me.

bcda/main.go
tbellj: Yes! I'll try this, hopefully this will satisfy the atomic operation problem.

bcda/main.go
tbellj: Neither one of those options will work.

bcda/main.go
```diff
@@ -379,3 +390,45 @@ func getEnvInt(varName string, defaultVal int) int {
...
+		elapsed := time.Since(t).Hours()
+		if int(elapsed) >= hrThreshold {
+
+			id := int(j.ID)
+			jobDir := fmt.Sprintf("%s/%d", os.Getenv("FHIR_PAYLOAD_DIR"), id)
+			expDir = fmt.Sprintf("%s/%d", os.Getenv("FHIR_EXPIRED_DIR"), id)
+
+			err = os.Rename(jobDir, expDir)
+			if err != nil {
+				log.Error(err)
```
karlmdavis: In order for this to actually be robust in the case of failures, I think you can't proceed past here in the case of errors, because you're still marking the cleanup as done in the DB even when the rename failed. Two suggestions:

1. Do whatever the Golang equivalent of `continue;` is after you log the error to skip the rest of this iteration.
2. After fixing that, if there's something up with the filesystem that will always prevent a `rename` from succeeding, then some things will stay marked as ready-to-expire in the DB forever.
    * You should add monitoring to ensure that things actually are being cleaned up in the time period we expect them to be. Something like a CloudWatch alarm or somesuch, so that folks are actually prompted to go investigate it.
        * If it's not already covered by something else, e.g. an alarm that fires for every error log entry.
        * Could be done as part of a separate ticket, but you should reference that ticket here or in JIRA.
        * (I didn't check JIRA, so apologies if that's already there.)

(And as always, apologies if I'm just missing something obvious.)

tbellj: 1) Yep!! I see your point about the continue.
2) Let's say something does happen where we fail on this, the next time this code runs it will cleanup the "ready-to-expire" jobs on the next go-round assuming no real outstanding issues. But to your point if we do have a **real issue** and there's something up with the filesystem, (then we obviously having larger problems), and something like a CloudWatch alarm could definitely be implemented. I'll search for a corresponding ticket to link to this or create one if I can't find it.

bcda/main.go
```diff
@@ -379,3 +390,46 @@ func getEnvInt(varName string, defaultVal int) int {
...
+	}
+
+	for _, j := range jobs {
+		t := j.CreatedAt
+		elapsed := time.Since(t).Hours()
+		if int(elapsed) >= hrThreshold {
+
+			id := int(j.ID)
+			jobDir := fmt.Sprintf("%s/%d", os.Getenv("FHIR_PAYLOAD_DIR"), id)
+			expDir = fmt.Sprintf("%s/%d", os.Getenv("FHIR_EXPIRED_DIR"), id)
```
rnagle: Are these actually directory names or rather file names?

bcda/main.go
rnagle: @tbellj I don't think there's an existing ticket for monitoring this. Would you mind creating one to instrument this process so that we can be notified if it is not working properly?

bcda/main.go
tbellj: Good question! They're directory names, the files themselves will be in the job directory and they get moved to the new expDir location. folder --> to folder.

PR #67, opened 2018-11-07
BCDA-349: Introduce integration/smoke test into the pipeline

msnook:
### Fixes [BCDA-349](https://jira.cms.gov/browse/BCDA-349)

Introduce an integration test utility that can be leveraged in local environment.

### Proposed changes:

Baseline the BCDA client demo and transform it into a smoke test utility.  Additionally, ensure that it can easily be leveraged by the development team for local testing.

### Change Details

- Baselined the BCDA client.  To see differences from the original version (pulled from Ryan's gist) and the current version, please see this diff (scroll to the bottom and only necessary to focus on the diff for the `bcda_client.go` file: https://github.com/CMSgov/bcda-app/compare/14407f6...d617d9e 
- Added a `Dockerfile` so that the smoke tests can be executed in local development environment.
- Update the `Makefile` to include a smoke test target.
- Prepare the travis configuration for future enabling of the smoke test execution in travis CI.  It is currently disabled due to this bug: https://jira.cms.gov/browse/BCDA-500
- Ops specific updates for BCDA-349 can be found in this PR: https://github.com/CMSgov/bcda-ops/pull/41

### Security Implications

- None.

### Acceptance Validation

This has been tied into the CI.  Here is the Jenkins job: https://bcda-ci.adhocteam.us/job/BCDA%20-%20Test%20-%20Integration%20Smoke/

### Feedback Requested

Please review.


Dockerfiles/Dockerfile.smoke_test
```diff
@@ -0,0 +1,23 @@
+FROM golang:1.10.3
+
+RUN openssl genrsa -out /var/local/private.pem 2048
```
rnagle: I don't think the client needs this key pair. Can be removed.

Dockerfiles/Dockerfile.smoke_test
```diff
@@ -0,0 +1,23 @@
...
+RUN openssl rsa -in /var/local/private.pem -outform PEM -pubout -out /var/local/public.pem
+
+RUN go get -u github.com/golang/dep/cmd/dep
+
+WORKDIR /go/src/github.com/CMSgov/bcda-app
+COPY . .
+
+RUN dep ensure
+
+ENV BCDA_ERROR_LOG /var/log/bcda-error.log
```
rnagle: Again, don't think the client needs any of these `ENV` vars defined. Can probably remove.

Dockerfiles/Dockerfile.smoke_test
msnook: You are right.  Removed.

Dockerfiles/Dockerfile.smoke_test
msnook: Removed.

bcda/api.go
```diff
@@ -297,7 +297,7 @@ func getToken(w http.ResponseWriter, r *http.Request) {
 	// Generates a token for fake user and ACO combination
 	token, err := authBackend.GenerateTokenString(
 		"82503A18-BF3B-436D-BA7B-BAE09B7FFD2F",
-		"3461c774-b48f-11e8-96f8-529269fb1459",
+		"0C527D2E-2E8A-4808-B11D-0FA06BAF8254",
```
tbellj: yes finally!

PR #71, opened 2018-11-13
Set connection: close response header

rnagle:
Since we are routing requests to two different instances of `http.Server`, we can not support persistent connections and must set a `Connection: close` header in our responses.

The consequence of not setting this header is that, when a client makes back-to-back requests for endpoints handled by separate `http.Server` instances, the app will be unable to route the request properly and send back a false `404` error.

The [HTTP/1.1 whitepaper](https://tools.ietf.org/html/rfc2616#section-14.10) indicates setting the `Connection` header is acceptable (and, in fact, required) when the application does not support persistent connections.

### Proposed changes:

- Adds a `ConnectionClose` middleware and uses it with the api and data server routers

### Change details

- Note: since we are deploying the app behind AWS ELB, the "client" in our case is _technically_ the load balancer. Even though we send `Connection: close` on the backend, the load balancer will still use a keep-alive connections on the frontend (between the load balancer and end users), meaning there is ultimately no performance/resource implication for them.

### DIFF
--- a/bcda/api.go
+++ b/bcda/api.go
@@ -318,11 +318,29 @@ func getToken(w http.ResponseWriter, r *http.Request) {
 
 	authBackend := auth.InitAuthBackend()
 
+	db := database.GetGORMDbConnection()
+	defer db.Close()
+
+	var user auth.User
+	err := db.First(&user, "name = ?", "User One").Error
+	if err != nil {
+		log.Print(err)
+		oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, "", responseutils.DbErr)
+		responseutils.WriteError(oo, w, http.StatusInternalServerError)
+		return
+	}
+
+	var aco auth.ACO
+	err = db.First(&aco, "name = ?", "ACO Dev").Error
+	if err != nil {
+		log.Print(err)
+		oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, "", responseutils.DbErr)
+		responseutils.WriteError(oo, w, http.StatusInternalServerError)
+		return
+	}
+
 	// Generates a token for fake user and ACO combination
-	token, err := authBackend.GenerateTokenString(
-		"82503A18-BF3B-436D-BA7B-BAE09B7FFD2F",
-		"0C527D2E-2E8A-4808-B11D-0FA06BAF8254",
-	)
+	token, err := authBackend.GenerateTokenString(user.UUID.String(), aco.UUID.String())
 	if err != nil {
 		log.Error(err)
 		oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, "", responseutils.TokenErr)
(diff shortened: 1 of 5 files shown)


bcda/api.go
```diff
@@ -318,11 +318,29 @@ func getToken(w http.ResponseWriter, r *http.Request) {
 
 	authBackend := auth.InitAuthBackend()
 
+	db := database.GetGORMDbConnection()
+	defer db.Close()
+
+	var user auth.User
+	err := db.First(&user, "name = ?", "User One").Error
+	if err != nil {
+		log.Print(err)
```
msnook: Nit:  `log.Error(err)` instead of `log.Print(err)` ??

bcda/api.go
```diff
@@ -318,11 +318,29 @@ func getToken(w http.ResponseWriter, r *http.Request) {
...
+		oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, "", responseutils.DbErr)
+		responseutils.WriteError(oo, w, http.StatusInternalServerError)
+		return
+	}
+
+	var aco auth.ACO
+	err = db.First(&aco, "name = ?", "ACO Dev").Error
+	if err != nil {
+		log.Print(err)
```
msnook: Nit: log.Error(err) instead of log.Print(err) ??

msnook:
This is excellent... Prior to these changes, the smoke test failed every time it was executed in Travis.

Now take a look at the Travis output:

```
$ make smoke-test
docker-compose up -d 
bcdaapp_db_1 is up-to-date
Starting bcdaapp_documentation_1 ... 
Starting bcdaapp_documentation_1
Creating bcdaapp_worker_1 ... 
Creating bcdaapp_worker_1
Creating bcdaapp_api_1 ... 
Creating bcdaapp_api_1
docker-compose -f docker-compose.test.yml up --force-recreate --exit-code-from smoke_test smoke_test
WARNING: using --exit-code-from implies --abort-on-container-exit
WARNING: Found orphan containers (bcdaapp_api_1, bcdaapp_worker_1, bcdaapp_queue_1, bcdaapp_db_1) for this project. If you removed or renamed this service in your compose file, you can run this command with the --remove-orphans flag to clean it up.
Creating bcdaapp_smoke_test_1 ... 
Creating bcdaapp_smoke_test_1
oke_test_1
smoke_test_1  | Access Token not supplied.  Retrieving one to use.
smoke_test_1  | making request to start data aggregation job
smoke_test_1  | checking job status...
smoke_test_1  |   => job is still pending. waiting...
smoke_test_1  | checking job status...
smoke_test_1  | file is ready for download...
smoke_test_1  | fetching: http://api:3000/data/1/0c527d2e-2e8a-4808-b11d-0fa06baf8254.ndjson
smoke_test_1  | writing download to disk: /tmp/download.json
smoke_test_1  | validating file...
smoke_test_1  | done.
bcdaapp_smoke_test_1 exited with code 0
Aborting on container exit...
The command "make smoke-test" exited with 0.
```

PR #74, opened 2018-11-15
Add a (temporary) way for clients to request file encryption

rnagle:
### Overview/background

This change adds a check for `?encrypt=true` appended to `/api/v1/ExplanationOfBenefit/$export` and executes the file encryption routine when it is present.

This is necessary since we do not plan to share and discuss file encryption functionality with our pilot users in research sessions. We'll hold off on sharing it until after the first round of research is done and we have documentation explaining how they're meant to interact with the encrypted files (see: https://jira.cms.gov/browse/BCDA-512).

At the same time, we need a way to allow security auditors to test/validate that our file encryption is present and works properly. This should help accomplish that goal.

### Proposed changes:

As stated above, this adds a check for `?encrypt=true` in requests for bulk data. When this query parameter is present, the aggregation job that is created is marked for encryption. When the worker process picks up a job marked for encryption, the file encryption routine runs.

When the `encrypt` query param is omitted (or is any value other than `true`), file encryption is bypassed.

### Security Implications

- Since this is only intended for use in our sandbox environment where no PHI/PII is stored, bypassing file encryption presents no inherent risk.
- There are several `TODO` items marked in the code where we will need to make changes when we are ready to make file encryption the default.

### DIFF
--- a/bcda/api.go
+++ b/bcda/api.go
@@ -37,6 +37,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/CMSgov/bcda-app/bcda/auth"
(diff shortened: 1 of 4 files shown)


michaelrobertsutton:
Big fan; I like this approach is a simplistic and temporary approach to enabling both Alpha Partners to engage in R1 without encryption, and providing a method for the SCA/ACT auditing team to validate the planned encryption process of our architecture. 

If we move forward we'll run manual positive and negative integration tests with the release to verify the split functionality. 

Tagging @StewGoin FYSA, input, and opinions as well.

michaelrobertsutton:
Chatting on the phone with @rnagle and @knollfear 
> rule 34: Make it hard to do the wrong thing: the default thing, the easy thing, and the right thing should all be the same thing. “Build your opponent a golden bridge to retreat across.” – Sun Tzu
@StewGoin had a great quote yesterday -^

While this use-case, goes against the philosophy; motioning for the current implementation as it would reduce the confusion of R1 for Alpha Partners, is within the capabilities of the ACT/SCA auditors, and is only temporary solution for the ACT/SCA.

StewGoin:
It's understandable for the R1 phase, as long as we make the more permanent implementation in keeping with the rule.

...and not end up complying with the OTHER Rule 34 (which I got a chuckle about this morning thanks to a witty Dev)!

PR #75, opened 2018-11-16
Knollfear/bcda 214 Implement GOsec

knollfear:
### Fixes [BCDA-214](https://jira.cms.gov/browse/BCDA-214)

Implement Gosec static code inspection tool to proactively check for bad code we might write
### Proposed changes:

Add Gosec code inspection tool to unit test execution.

### Change Details

Changed unit_test script to include Gosec execution
Changed file permissions of 0666 to 0600 which is MORE restrictive
Added /*#nosec*/ comment to a few places where it wasn't flagging a real issue
Excluded unhandled error test - Remediate with BCDA-554

### Security Implications

This is a ststic code inspection tool.  It can only improve security.

### Acceptance Validation

Unit tests should run and pass.  
Results should be found in test results folder


unit_test.sh
```diff
@@ -9,6 +9,13 @@ set -o pipefail
 echo "Running linter..."
 golangci-lint run
 
+timestamp=`date +%Y-%m-%d_%H-%M-%S`
+mkdir -p test_results/${timestamp}
+mkdir -p test_results/latest
+
```
msnook: Since you are creating `timestamp` and the two directories here, can you remove these lines:

https://github.com/CMSgov/bcda-app/pull/75/files#diff-12c47e9f6d43cdddab176851f23f31dbR31
https://github.com/CMSgov/bcda-app/pull/75/files#diff-12c47e9f6d43cdddab176851f23f31dbR33
https://github.com/CMSgov/bcda-app/pull/75/files#diff-12c47e9f6d43cdddab176851f23f31dbR34

unit_test.sh
```diff
@@ -9,6 +9,13 @@ set -o pipefail
...
+echo "Running gosec..."
+gosec -fmt=junit-xml -exclude=G104 -out=test_results/latest/gosec_results.xml ./...
```
msnook: For rules we are excluding in the command line invocation... should we have follow-up stories to address them?

knollfear: this one is BCDA-554 as described in the PR.

msnook: Got it!

bcda/client/bluebutton.go
```diff
@@ -51,6 +51,7 @@ func NewBlueButtonClient() (*BlueButtonClient, error) {
 	// TODO Fix when Blue Button has a static cert: https://jira.cms.gov/browse/BLUEBUTTON-484
 	if os.Getenv("BB_SERVER_LOCATION") != "https://fhir.backend.bluebutton.hhsdevcloud.us" {
 		caFile := os.Getenv("BB_CLIENT_CA_FILE")
+		/* #nosec */
```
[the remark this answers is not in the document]
knollfear: I think this one is just unavoidable.  We need to use an environment var for this or else we will have to refactor how we set run time vars.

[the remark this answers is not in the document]
msnook: OK, all good.  Just wondering.  No worries.

msnook:
In looking at the output, is there any way to push the error messages to the console?
Right now, it appears as though 1 failure case is being hit but it isn't be spit out to the console.  It is instead being spit out to the XML file only.  And now that we've been able to shift this process to the left and execute it as part of `make test`... we probably don't even need to spit it out to XML there is no reason to have it be digested by Jenkins.  I'm thinking that if we run it in the same manner as `golint`, we'll be fine.
Something as simple as this would probably work:
```
gosec -exclude=G104 ./...
```

PR #78, opened 2018-11-20
BCDA 442: Cleanup files for expired jobs

DeirdreHolub:
### Fixes [BCDA-442](https://jira.cms.gov/browse/BCDA-442)

When a job is expiring due to age, its files are moved to an inaccessible location. See [BCDA-441](https://jira.cms.gov/browse/BCDA-442) for more about this. The work of this ticket is to cleanup that inaccessible location by removing files for jobs that have not been touched for a certain period of time.

### Proposed changes:

Add a CLI command to cleanup the archive location. To run locally in docker:

docker exec -it bcda-app_api_1 bash -c 'tmp/bcda cleanup-archive --threshold 30'

--threshold is required, because I don't think it's appropriate for this argument to have a default.

### Change Details

1. Renamed the original status associated with moving files to the archive location from 'Expired' to 'Archived'

2. When a job has been in 'Archived' state for `threshold` hours, remove the files from the archive and set the job status to 'Expired'

### Security Implications

Can only be run through CLI.

### Acceptance Validation

Tests included.

Logging message
{"files_removed":"2018-11-20T07:41:19.5631881Z","job_began":"2018-11-16T16:42:00Z","job_id":1,"level":"info","msg":"Files cleaned from archive and job status set to Expired","time":"2018-11-20T07:41:19Z"}

### DIFF
--- a/bcda/main.go
+++ b/bcda/main.go
@@ -521,6 +521,7 @@ func cleanupArchive(hrThreshold int) error {
 			log.WithFields(log.Fields{
 				"job_began":     job.CreatedAt,
 				"files_removed": time.Now(),
+				"job_id":        job.ID,
 			}).Info("Files cleaned from archive and job status set to Expired")
 		}
 	}
--- a/bcda/main.go
+++ b/bcda/main.go
@@ -275,12 +275,12 @@ func setUpApp() *cli.App {
 			},
 		},
 		{
-			Name:     "remove-expired-jobs",
-			Category: "Remove expired jobs",
+			Name:     "archive-job-files",
+			Category: "Archive files for jobs that are expiring",
 			Usage:    "Updates job statuses and moves files to an inaccessible location",
 			Action: func(c *cli.Context) error {
-				threshold := getEnvInt("EXPIRED_THRESHOLD_HR", 24)
-				removeExpired(threshold)
+				threshold := getEnvInt("ARCHIVE_THRESHOLD_HR", 24)
+				archiveExpiring(threshold)
 				return nil
 			},
 		},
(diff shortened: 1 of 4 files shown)


bcda/main.go
```diff
@@ -439,19 +458,73 @@ func removeExpired(hrThreshold int) {
 
 			id := int(j.ID)
 			jobDir := fmt.Sprintf("%s/%d", os.Getenv("FHIR_PAYLOAD_DIR"), id)
-			expDir = fmt.Sprintf("%s/%d", os.Getenv("FHIR_EXPIRED_DIR"), id)
+			expDir = fmt.Sprintf("%s/%d", os.Getenv("FHIR_ARCHIVE_DIR"), id)
```
tbellj: not a big deal but this name should probably change too.

bcda/main.go
```diff
@@ -439,19 +458,73 @@ func removeExpired(hrThreshold int) {
...
+			}
+
+			job.Status = "Expired"
+			err = db.Save(job).Error
+			if err != nil {
+				return err
+			}
+
+			log.WithFields(log.Fields{
+				"job_began":     job.CreatedAt,
```
tbellj: Also not sure what the exact logging info we want to capture here, but this seems okay to me. Might wanna get additional eyes on this part before merging.

bcda/main.go
DeirdreHolub: I left it there because I wasn't sure if it had been used anywhere. I plan to delete it after I'm sure no clean up needs to happen (kind of like a database migration, but in reverse).

bcda/main.go
DeirdreHolub: I ran this by Ryan yesterday. I think logging is something we should expect to evolve atm, and it's probably better to get something in the logs to work with.

bcda/main.go
tbellj: makes sense

bcda/main.go
tbellj: agreed, we can always evolve later

PR #81, opened 2018-11-27
Ensure Splunk can see log files

msnook:
### Fixes Splunk Bug

Splunk is unable to see files with permissions of 600.  

### Proposed changes:

Update necessary log files to have permissions of 640.

### Change Details

- Update log file permissions to be 640 instead of 600
- Identify in code commentary why we are circumventing a `gosec` rule 

### Security Implications

Group now has `read` privileges.

### DIFF
--- a/bcdaworker/main.go
+++ b/bcdaworker/main.go
@@ -39,7 +39,9 @@ type jobEnqueueArgs struct {
 func init() {
 	log.SetFormatter(&log.JSONFormatter{})
 	filePath := os.Getenv("BCDA_WORKER_ERROR_LOG")
-	file, err := os.OpenFile(filePath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0600)
+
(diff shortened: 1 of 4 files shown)


bcdaworker/main.go
```diff
@@ -189,7 +191,10 @@ func appendErrorToFile(acoID, code, detailsCode, detailsDisplay string, jobID st
 
 	dataDir := os.Getenv("FHIR_STAGING_DIR")
 	fileName := fmt.Sprintf("%s/%s/%s-error.ndjson", dataDir, jobID, acoID)
-	f, err := os.OpenFile(fileName, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0600)
+
+	/* #nosec -- 0640 permissions required for Splunk ingestion */
+	f, err := os.OpenFile(fileName, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0640)
```
rnagle: Don't think this change is necessary since this is not a file that is ingested by Splunk.

PR #83, opened 2018-11-28
bcda-521: log jobID for BlueButton ExplanationOfBenefit requests

alexhova:
### Fixes [BCDA-521](https://jira.cms.gov/browse/BCDA-521)

We want to log the job id for workers in our app that request ExplanationOfBenefit data from the BlueButton api

### Proposed changes:

- Added jobID to the instantiation of the blue button client
- Added jobID to methods that get ExplanationOfBenefit data
- Added jobID to method that logs requests

### Security Implications

None really

### Acceptance Validation

Looks good.  Here is an example log message
```
worker_1         | {"bb_query_id":"0e8189df-9d53-4362-bbb5-25f308340f1f","bb_query_ts":"2018-11-28 18:09:10.026879745 +0000 UTC m=+118.259473555","bb_uri":"https://fhir.backend.bluebutton.hhsdevcloud.us/v1/fhir/ExplanationOfBenefit/?_format=application%2Ffhir%2Bjson\u0026patient=19990000002915","job_id":"25","level":"info","msg":"Blue Button request","time":"2018-11-28T18:09:10Z"}
```

### Feedback Requested

This works but I think it would be nicer if we didn't have to pass an empty string when there is no job associated with the request.  Thoughts?  Also, do you think we should be logging this id in the request (as I am here), the response, or both?


alexhova:
Yeah, why not!  Now logging job id in request and response logs

PR #92, opened 2018-12-03
BCDA-613: Error handling in archive-job-files command 

msnook:
### Fixes [BCDA-613](https://jira.cms.gov/browse/BCDA-613)

[This Jenkins job](https://bcda-ci.adhocteam.us/job/BCDA%20-%20Archive%20and%20Clean%20Job%20Data/) was passing, even when the `archive-job-files` command would fail.  Basically, it came down to `archive-job-files` not returning a valid error code.

### Proposed changes:

Update `archive-job-files` to return a valid error code.

### Change Details

- `archive-job-files` returns and exits if initial DB connection fails or if the `FHIR_ARCHIVE_DIR` is not set.
- `archive-job-files` returns the last job error if any of the iterative job processing fails (all jobs are still processed though, even if a failure is encountered)

### Security Implications

None.

### Acceptance Validation

Tested locally, positive and negative scenarios, and ensured that `$?` was set appropriately.

### Feedback Requested

Please review.


bcda/main.go
```diff
@@ -466,16 +468,20 @@ func archiveExpiring(hrThreshold int) {
 			err = os.Rename(jobDir, expDir)
 			if err != nil {
 				log.Error(err)
+				lastJobError = err
 				continue
```
DeirdreHolub: If we're going to fail on an error, I think we should just fail and not continue. Otherwise, the status will be updated on the job when it really shouldn't be

rnagle: I could be misunderstanding the flow control here, but I think breaking this loop with a `continue` will skip the portion of the code that updates job status.

msnook: @DeirdreHolub The `continue` stems from [Karl's comments](https://github.com/CMSgov/bcda-app/pull/66#discussion_r232318503) in the original PR.

I don't think the status will get updated on this job if the file manipulation fails, because `continue` simply continues with the next iteration in the `for` loop instead of executing the rest of the code in the block.  In this case, if the file manipulation fails, the job status will not be updated.

PR #93, opened 2018-12-03
BCDA-624: Use os.O_APPEND mode for error/request logs

rnagle:
### Fixes [BCDA-624](https://jira.cms.gov/browse/BCDA-624)

Makes `bcda` use "append" mode when opening log files to prevent the issue described here: https://serverfault.com/questions/642286/logrotate-continuous-write-issue

### Proposed changes:

- Update locations where `bcda` opens `BB_ERROR_LOG`, `BCDA_REQUEST_LOG` and `BCDA_ERROR_LOG` to use `os.O_APPEND` mode.

### DIFF
--- a/bcda/client/bluebutton.go
+++ b/bcda/client/bluebutton.go
@@ -33,7 +33,7 @@ func init() {
 	filePath := os.Getenv("BCDA_BB_LOG")
 
 	/* #nosec -- 0640 permissions required for Splunk ingestion */
-	file, err := os.OpenFile(filePath, os.O_CREATE|os.O_WRONLY, 0640)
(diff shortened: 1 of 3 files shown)


msnook:
I have verified this implementation and seen it working when setting `logrotate` configuration to initiate the log rotation based on file size (to force the log rotation).  However, to be certain, I'd like to see our  daily `logrotate` configuration execute.  The daily rotation occurs at ~3:25AM according to the `cron.daily` documentation (this time coincides with all the historical logs on `test` and `sbx`).  
If tomorrow's new version of `request.log` exists with no binary data after the daily log rotation, then I will be satisfied and approve.

PR #94, opened 2018-12-04
Knollfear/bcda 520 clean up unit test results

knollfear:
### Fixes [BCDA-520](https://jira.cms.gov/browse/BCDA-520)

Unit test results are a hot mess.  We need something that cuts the clutter and produces junit.xml files for jenkins
### Proposed changes:

Added gotestsum to unit test docker image
changed test execution to use gotestsum

### Change Details

used a different test runner
modified test run command

### Security Implications

no pii is affected.  This only affects how test are executed

### Acceptance Validation

Checkout branch and run make test
Break tests and make sure the output is useful
Confirm existence of junit.xml in test result directories when test succeed or fail

test results with intentional failure
image

Successful test results

image

### Feedback Requested

If this isn't helpful output lmk and I will keep searching for a more useful format.


bcda/api.go
```diff
@@ -3,13 +3,11 @@
 
  The Beneficiary Claims Data API (BCDA) allows downloading of claims data in accordance with the FHIR Bulk Data Export specification.
 
- Terms Of Service:
+ If you have a token you can use this page to explore the API.  To do this click the green "Authorize" button below and enter "Bearer {YOUR_TOKEN}"
```
msnook: Wasn't expecting this file to change as part of BCDA-520.  Can you re-sync with `master`?

bcda/models/SwaggerStructs.go
```diff
@@ -106,9 +106,12 @@ type FileParam struct {
 // swagger:parameters bulkRequest
 type BulkRequestHeaders struct {
 	// required: true
-	// pattern: application/fhir\+json
+	// in: header
```
msnook: Same with this file.  I wasn't expecting this file to change as part of BCDA-520.  Can you re-sync with `master`?

bcda/api.go
knollfear: I will.  I am just closing my other pr's right now.

PR #95, opened 2018-12-04
BCDA-395: Add /Patient endpoint for bulk export of patient data

embh:
### Fixes [BCDA-395](https://jira.cms.gov/browse/BCDA-395)
We need a way for ACOs to retrieve patient data as found in CCLF-8.

### Proposed changes:
* Add endpoint for starting a patient data export job

### Change Details
* Adds `/Patient/$export` route
* Adds resource type parameter to `bulkRequest()`
* Uses environment variable `ENABLE_PATIENT_EXPORT` to determine whether endpoint is accessible. Enabled only if that value is true.

### Security Implications
Retrieves PII from Blue Button and returns it in the generated data file.

### Acceptance Validation
Includes smoke and integration tests. Does *not* include Swagger docs because we can't seem to control that with an env var. There is a separate ticket for narrative documentation, which we can roll this into.

### Feedback Requested
Any

### DIFF
--- a/Dockerfiles/Dockerfile.integration_test
+++ b/Dockerfiles/Dockerfile.integration_test
@@ -25,4 +25,4 @@ WORKDIR /go/src/github.com/CMSgov/bcda-app/test/integration_test
 
 # Use simple tester command once the fork gets merged to main repo
 # CMD ["tester", "-var", "PROTOCOL=http", "-var", "HOST=api:3000", "version.toml", "swagger.toml", "metadata.toml", "jobs.toml", "eob.toml"]
-CMD ["go", "run", "/go/src/github.com/whytheplatypus/tester/main.go", "-var", "PROTOCOL=http", "-var", "HOST=api:3000", "version.toml", "swagger.toml", "metadata.toml", "jobs.toml", "eob.toml"]
+CMD ["go", "run", "/go/src/github.com/whytheplatypus/tester/main.go", "-var", "PROTOCOL=http", "-var", "HOST=api:3000", "version.toml", "swagger.toml", "metadata.toml", "jobs.toml", "eob.toml", "patient.toml"]
--- a/bcda/api.go
+++ b/bcda/api.go
@@ -83,26 +83,6 @@ func bulkEOBRequest(w http.ResponseWriter, r *http.Request) {
 	bulkRequest("ExplanationOfBenefit", w, r)
 }
 
-/*
-	swagger:route GET /Patient/$export bulkData bulkPatientRequest
-
-	Start patient data export
-
-	Initiates a job to collect data from the Blue Button API for your ACO.
-
-	Produces:
-	- application/fhir+json
-
-	Schemes: http, https
-
-	Security:
-		api_key
-
-	Responses:
-		202:BulkRequestResponse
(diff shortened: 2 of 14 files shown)


bcda/api.go
```diff
@@ -55,7 +56,11 @@ import (
 )
 
 /*
+<<<<<<< HEAD
```
tbellj: I think a merge conflict messed this up right here.

docker-compose.yml
```diff
@@ -39,6 +39,7 @@ services:
       - ATO_PUBLIC_KEY_FILE=../../shared_files/ATO_public.pem
       - ATO_PRIVATE_KEY_FILE=../../shared_files/ATO_private.pem
       - HTTP_ONLY=true
+      - ENABLE_PATIENT_EXPORT=true
```
tbellj: do we wanna leave this here?

bcda/api.go
embh: Thanks for catching that!

docker-compose.yml
embh: Without it, the endpoint isn't available for the integration tests. Talked to @msnook about it briefly yesterday, and I think we can override it by environment.

msnook: @tbellj Yup, I think it is better to place in `docker-compose` then in `Dockerfile`.  Placing in `Dockerfile` would mean that everyone's images are now out of sync.  Much more flexible to place these vars in `docker-compose`.

Dockerfiles/Dockerfile.integration_test
```diff
@@ -25,4 +25,4 @@ WORKDIR /go/src/github.com/CMSgov/bcda-app/test/integration_test
 
 # Use simple tester command once the fork gets merged to main repo
 # CMD ["tester", "-var", "PROTOCOL=http", "-var", "HOST=api:3000", "version.toml", "swagger.toml", "metadata.toml", "jobs.toml", "eob.toml"]
-CMD ["go", "run", "/go/src/github.com/whytheplatypus/tester/main.go", "-var", "PROTOCOL=http", "-var", "HOST=api:3000", "version.toml", "swagger.toml", "metadata.toml", "jobs.toml", "eob.toml"]
+CMD ["go", "run", "/go/src/github.com/whytheplatypus/tester/main.go", "-var", "PROTOCOL=http", "-var", "HOST=api:3000", "version.toml", "swagger.toml", "metadata.toml", "jobs.toml", "eob.toml", "patient.toml"]
```
msnook: Yeessssss!

test/smoke_test/bcda_client.go
```diff
@@ -129,64 +129,66 @@ func writeFile(resp *http.Response) {
...
-			if status.StatusCode == 200 {
-				fmt.Println("file is ready for download...")
-
-				defer status.Body.Close()
-
-				var objmap map[string]*json.RawMessage
-				err := json.NewDecoder(status.Body).Decode(&objmap)
-				if err != nil {
-					panic(err)
+	for _, t := range [2]string{"ExplanationOfBenefit", "Patient"} {
```
msnook: This is awesome.

PR #99, opened 2018-12-06
bcda-602 add new relic monitoring to worker job processing

alexhova:
### Fixes [BCDA-602](https://jira.cms.gov/browse/BCDA-602)
We want be able to monitor workers in new relic.

### Proposed changes:
- import our monitoring package into bcdaworker/main.go
- record processJob as a transaction and its nested functions as segments in new relic

### Security Implications
None really.  High security mode is turned on in new relic

### Acceptance Validation
you can see what is being reported in new relic here: https://rpm.newrelic.com/accounts/2057345/applications/128793549/transactions?type=other#id=5b224f746865725472616e73616374696f6e2f476f2f70726f636573734a6f62222c22225d

### DIFF
--- a/bcdaworker/main.go
+++ b/bcdaworker/main.go
@@ -12,20 +12,22 @@ import (
 	"strconv"
 	"syscall"
 
-	"github.com/CMSgov/bcda-app/bcda/encryption"
-
 	"github.com/jackc/pgx"
+	newrelic "github.com/newrelic/go-agent"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/CMSgov/bcda-app/bcda/client"
 	"github.com/CMSgov/bcda-app/bcda/database"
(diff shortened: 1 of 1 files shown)


bcdaworker/main.go
```diff
@@ -12,20 +12,22 @@ import (
...
 	"github.com/CMSgov/bcda-app/bcda/models"
+	"github.com/CMSgov/bcda-app/bcda/monitoring"
 	"github.com/CMSgov/bcda-app/bcda/responseutils"
 	"github.com/bgentry/que-go"
 )
 
 var (
-	qc *que.Client
+	qc  *que.Client
+	txn newrelic.Transaction
```
tbellj: yeah I see what your saying but this part can be included in the monitoring file also. But I'm not married to the idea of keeping all of our new relic references/core implementation in one place. Just a thought

PR #102, opened 2018-12-10
BCDA-609 Feature: Updated capability statement with Patient endpoint

tbellj:
Fixes [BCDA-609](https://jira.cms.gov/browse/BCDA-609)

Problem:
Capability statement was out of date with our latest code changes, specifically the Patient endpoint that was recently added.

Proposed changes:
This will adjust our capability statement to correctly reflect all of our endpoints (in their respective environments)

Change Details:
bcda/responseutils/writer.go -- updated capability statement with the "Patient" endpoint

Security Implications:
No PII/PHI but this statement should only show the Patient endpoint when that environment is configured to expose that resource.

Acceptance Validation:
Be sure capability statement shows the correct metadata in the correct environments.

Feedback Requested
Any

### DIFF
--- a/bcda/responseutils/writer.go
+++ b/bcda/responseutils/writer.go
@@ -117,11 +117,11 @@ func CreateCapabilityStatement(reldate time.Time, relversion, baseurl string) *f
 			},
 		},
 	}
-	checkEnv(statement, baseurl)
+	addPatientEndpointToStatement(statement, baseurl)
 	return statement
 }
 
-func checkEnv(statement *fhirmodels.CapabilityStatement, baseUrl string) {
+func addPatientEndpointToStatement(statement *fhirmodels.CapabilityStatement, baseUrl string) {
 	if os.Getenv("ENABLE_PATIENT_EXPORT") == "true" {
 		restComponent := statement.Rest[0].Operation
 
--- a/bcda/responseutils/writer.go
+++ b/bcda/responseutils/writer.go
@@ -117,8 +117,24 @@ func CreateCapabilityStatement(reldate time.Time, relversion, baseurl string) *f
 			},
 		},
 	}
-
+	checkEnv(statement, baseurl)
 	return statement
+}
+
+func checkEnv(statement *fhirmodels.CapabilityStatement, baseUrl string) {
+	if os.Getenv("ENABLE_PATIENT_EXPORT") == "true" {
+		restComponent := statement.Rest[0].Operation
+
+		element := fhirmodels.CapabilityStatementRestOperationComponent{
+			Name: "export",
+			Definition: &fhirmodels.Reference{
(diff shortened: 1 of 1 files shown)


bcda/responseutils/writer.go
```diff
@@ -117,10 +117,26 @@ func CreateCapabilityStatement(reldate time.Time, relversion, baseurl string) *f
...
+	checkEnv(statement, baseurl)
 	return statement
 }
 
+func checkEnv(statement *fhirmodels.CapabilityStatement, baseUrl string) {
+	if os.Getenv("ENABLE_PATIENT_EXPORT") == "true" {
+		restComponent := statement.Rest[0].Operation
+
+		element := fhirmodels.CapabilityStatementRestOperationComponent{
+			Name: "export",
```
embh: Is the `export` name defined by FHIR spec? Wondering if it would make sense to be more specific so they are different, e.g., `patient-export` and `eob-export`.

knollfear: I don't see any testing for this.  If we don't have any tests for the capability statement already I would love to see a ticket for adding them.

tbellj: Yeah good question. I searched the fhir spec and used this as a reference: "The name of the operation or query. For an operation, this is the name prefixed with $ and used in the URL"

I agree with your idea, I just don't know how flexible we can be here and still follow the standard.

embh: Aha, that's what I was looking for. Thanks for the info!

bcda/responseutils/writer.go
```diff
@@ -117,10 +117,26 @@ func CreateCapabilityStatement(reldate time.Time, relversion, baseurl string) *f
 			},
 		},
 	}
-
+	checkEnv(statement, baseurl)
 	return statement
 }
 
+func checkEnv(statement *fhirmodels.CapabilityStatement, baseUrl string) {
```
msnook: Any chance we can rename this function to something like `addPatientEndpointToStatement`?  The main purpose of the function is to add the patient endpoint to the capability statement; checking the env var is just something it does under the covers.

bcda/responseutils/writer.go
tbellj: Ticket created for our unit tests capability statement here: https://jira.cms.gov/browse/BCDA-658. But I do think this is outside the scope of this particular PR since it's covering the entire metadata endpoint and not just this additional piece.

bcda/responseutils/writer.go
tbellj: Yeah I see what you're saying but I avoided calling it those names because theoretically this method is only temporary. Once we "release" the Patient endpoint to all of our environments this function won't even exist and it'll be added the same way just like our other endpoints.

PR #110, opened 2018-12-14
knollfear/bcda-684nfixing enum to allow correct usage of swagger for bulk request endpoint

knollfear:
### Fixes [BCDA-684](https://jira.cms.gov/browse/BCDA-684)

Swagger is not providing correctly formatted header values for accept and prefer because they are wrapped in []
### Proposed changes:

Remove the [] from the comments used to generate the 

### Change Details
just fixed up the comments

### Security Implications

NO, this is just documentation

### Acceptance Validation

run app locally
generate a token
add bearer {YOUR_TOKEN} to the auth dialog in swagger
make a bulk export request and confirm that the result is successful

### DIFF
--- a/bcda/models/SwaggerStructs.go
+++ b/bcda/models/SwaggerStructs.go
@@ -107,11 +107,10 @@ type FileParam struct {
 type BulkRequestHeaders struct {
 	// required: true
 	// in: header
-	// enum: [application/fhir+json]
+	// enum: application/fhir+json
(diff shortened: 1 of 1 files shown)


embh:
I still get an error for this, but the cURL command works for me. Are you able to get a 202 response in the Swagger UI?
screen shot 2018-12-17 at 11 47 11 am

PR #112, opened 2018-12-17
BCDA-645: Create /_health endpoint

embh:
### Fixes [BCDA-645](https://jira.cms.gov/browse/BCDA-645)
We need a way to check the health of the API.

### Proposed changes:
Add a health check endpoint that returns a response based on database connectivity. Note that this ticket originally included Blue Button connectivity, but that will be checked by the worker application instead.

### Change Details
* Creates unauthenticated endpoint, `/_health`
  * Checks database with GORM`Ping()`
  * Returns `200 OK` if database is accessible and `502 Bad Gateway` if not
  * Includes tests in api_test.go

### Security Implications
None

### Acceptance Validation
screen shot 2018-12-17 at 12 00 15 pm

### Feedback Requested
Any

### DIFF
--- a/Dockerfiles/Dockerfile.integration_test
+++ b/Dockerfiles/Dockerfile.integration_test
@@ -25,4 +25,4 @@ WORKDIR /go/src/github.com/CMSgov/bcda-app/test/integration_test
 
 # Use simple tester command once the fork gets merged to main repo
 # CMD ["tester", "-var", "PROTOCOL=http", "-var", "HOST=api:3000", "version.toml", "swagger.toml", "metadata.toml", "jobs.toml", "eob.toml"]
-CMD ["go", "run", "/go/src/github.com/whytheplatypus/tester/main.go", "-var", "PROTOCOL=http", "-var", "HOST=api:3000", "version.toml", "swagger.toml", "metadata.toml", "jobs.toml", "eob.toml", "patient.toml"]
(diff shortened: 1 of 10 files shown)


msnook:
Since the BB Backend check is being moved to the worker health check, should we update this ticket to ensure we actually implement that check: https://jira.cms.gov/browse/BCDA-661

msnook:
Do we need to do anything special (comments, etc) to ensure this endpoint shows up appropriately in Swagger?

msnook:
Should we add a simple `health` integration test?  
https://github.com/CMSgov/bcda-app/tree/master/test/integration_test

rnagle:
> Since the BB Backend check is being moved to the worker health check, should we update this ticket to ensure we actually implement that check: https://jira.cms.gov/browse/BCDA-661

I'll update BCDA-661 with notes on the worker health check implementation option we discussed last week.

> Do we need to do anything special (comments, etc) to ensure this endpoint shows up appropriately in Swagger?

I don't think we need things like `/_version` and `/_health` included in our Swagger docs. IMO they're unrelated to the product and more for internal use by us.

> Should we add a simple `health` integration test?
> https://github.com/CMSgov/bcda-app/tree/master/test/integration_test

I am +1 for this idea.

embh:
Tester and Postman integration tests added!

> I don't think we need things like /_version and /_health included in our Swagger docs. IMO they're unrelated to the product and more for internal use by us.

That's what I was thinking, and I've been wondering if the data path belongs there as an endpoint-- it's a file, not really an API resource or action. If we remove `/_version` and `/data`, we could set the Swagger UI base path as `/api/v1` and wouldn't need it prefixed on the metadata, export, and status endpoints. Maybe a topic for another conversation.

PR #114, opened 2018-12-17
BCDA-628: Move models from auth to bcda

DeirdreHolub:
### Completes [BCDA-628](https://jira.cms.gov/browse/BCDA-628)

Our ACO and User models are in the auth package, but the API needs them to exist independent of the auth implementation. Since we are preparing to more rigorously separate auth from the api, we need to move these models to the bcda/models package. The token model remains in auth.

### Change Details
- ACO model and tests moved to bcda/models
- User model and tests moved to bcda/models
- Extracted shared code for reading PKI keys into secutils package
- Added some missing tests

### Security Implications
None

### Acceptance Validation
All tests pass.

### Feedback Requested

Did I miss anything?

I had to change the docker-compose.test.yml file to correct an env path var that had (apparently) not been used by tests before.

### DIFF
--- a/bcda/encryption/encryption_test.go
+++ b/bcda/encryption/encryption_test.go
@@ -7,7 +7,6 @@ import (
 	"crypto/rsa"
 	"crypto/sha256"
 	"errors"
-	"github.com/CMSgov/bcda-app/bcda/auth"
 	"github.com/CMSgov/bcda-app/bcda/database"
 	"github.com/CMSgov/bcda-app/bcda/models"
 	"github.com/CMSgov/bcda-app/bcda/testUtils"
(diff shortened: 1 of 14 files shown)


bcda/models/models.go
```diff
@@ -1,19 +1,28 @@
...
 )
 
 func InitializeGormModels() *gorm.DB {
+	fmt.Print("Intitialize bcda models")
 	db := database.GetGORMDbConnection()
 	defer db.Close()
 
 	// Migrate the schema
 	// Add your new models here
+	// This should probably not be called in production
```
knollfear: this is fine to run in production.  it only adds columns and doesn't delete any data.

bcda/encryption/encryption_test.go
```diff
@@ -38,14 +37,14 @@ func (s *EncryptionTestSuite) TestEncryptBytes() {
 	// Make a random String for encrypting
 	testBytes := []byte(uuid.NewRandom().String())
 	// Encrypt the sting and get the key back
-	encryptedBytes, encryptedKey, err := EncryptBytes(auth.GetATOPublicKey(), testBytes, "TEST")
```
knollfear: You can remove lines 32/3 above.  The old structure was finding these keys from 2 different levels of code and now that value is what is in the YML file and should not need to be overridden here.

os.Setenv("ATO_PUBLIC_KEY_FILE", "../../shared_files/ATO_public.pem")
os.Setenv("ATO_PRIVATE_KEY_FILE", "../../shared_files/ATO_private.pem")

DeirdreHolub: This PR is about moving the models, not fixing configuration related to a different PR and code change. For traceability, I think it's important to keep PRs focused on the work of the ticket, and the changes you describe here should be in a separate ticket.

PR #115, opened 2018-12-18
BCDA-629: Add an auth provider interface and an alpha plugin

embh:
### Fixes [BCDA-629](https://jira.cms.gov/browse/BCDA-629)
We'd like to have an interface to be implemented by auth plugins.

### Proposed changes:
Create an auth provider interface and alpha plugin.

### Change Details
* Adds auth/provider.go, an interface of provider operations
* Adds auth/alpha.go, an implementation of the interface (returns only "Not yet implemented" errors)
  * auth/alpha_test.go is a suite of simple tests for the plugin

### Security Implications
None

### Acceptance Validation
All tests pass.
screen shot 2018-12-18 at 12 02 53 pm

### Feedback Requested
I'm a bit out of the loop on auth; happy to make any adjustments you need for this to work with your implementation.
Any opinions on using pointer vs value receivers here? My inclination is pointer, but value could be preferable if we're not modifying the plugin in any of the methods. https://golang.org/doc/faq#methods_on_values_or_pointers

### DIFF
--- a/bcda/auth/plugin/alpha.go
+++ b/bcda/auth/plugin/alpha.go
@@ -1,0 +1,43 @@
+package auth
+
+import (
+	"errors"
+)
+
+type AlphaAuthPlugin struct{}
+
+func (p *AlphaAuthPlugin) RegisterClient(params []byte) error {
+	return errors.New("Not yet implemented")
+}
+
+func (p *AlphaAuthPlugin) UpdateClient(params []byte) error {
+	return errors.New("Not yet implemented")
+}
+
+func (p *AlphaAuthPlugin) DeleteClient(params []byte) error {
+	return errors.New("Not yet implemented")
+}
+
+func (p *AlphaAuthPlugin) GenerateClientCredentials(params []byte) ([]byte, error) {
+	return nil, errors.New("Not yet implemented")
+}
+
+func (p *AlphaAuthPlugin) RevokeClientCredentials(params []byte) error {
+	return errors.New("Not yet implemented")
+}
+
+func (p *AlphaAuthPlugin) RequestAccessToken(params []byte) ([]byte, error) {
+	return nil, errors.New("Not yet implemented")
+}
+
+func (p *AlphaAuthPlugin) RevokeAccessToken(params []byte) error {
+	return errors.New("Not yet implemented")
+}
+
+func (p *AlphaAuthPlugin) ValidateAccessToken(params []byte) ([]byte, error) {
+	return nil, errors.New("Not yet implemented")
+}
+
+func (p *AlphaAuthPlugin) DecodeAccessToken(params []byte) ([]byte, error) {
+	return nil, errors.New("Not yet implemented")
+}
--- a/bcda/auth/plugin/alpha.go
+++ b/bcda/auth/plugin/alpha.go
@@ -2,16 +2,18 @@ package auth
 
 import (
 	"errors"
+
+	jwt "github.com/dgrijalva/jwt-go"
 )
 
 type AlphaAuthPlugin struct{}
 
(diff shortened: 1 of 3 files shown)


bcda/auth/plugin/alpha.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (p *AlphaAuthPlugin) RevokeAccessToken(params []byte) error {
+	return errors.New("Not yet implemented")
+}
+
+func (p *AlphaAuthPlugin) ValidateAccessToken(params []byte) ([]byte, error) {
+	return nil, errors.New("Not yet implemented")
+}
+
+func (p *AlphaAuthPlugin) DecodeAccessToken(params []byte) ([]byte, error) {
```
DeirdreHolub: DecodeAccessToken(token string) (Token, error)

A Token is a very well known thing (RFCs and all that), and we have a Token type. We should be explicit here in this return value. (This method will be used internally by revoke and validate)

bcda/auth/plugin/alpha.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (p *AlphaAuthPlugin) RequestAccessToken(params []byte) ([]byte, error) {
+	return nil, errors.New("Not yet implemented")
+}
+
+func (p *AlphaAuthPlugin) RevokeAccessToken(params []byte) error {
+	return errors.New("Not yet implemented")
+}
+
+func (p *AlphaAuthPlugin) ValidateAccessToken(params []byte) ([]byte, error) {
```
DeirdreHolub: ValidateAccessToken(token string) error 

I can't think of anything we would return other than the error.

bcda/auth/plugin/alpha.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (p *AlphaAuthPlugin) RevokeClientCredentials(params []byte) error {
+	return errors.New("Not yet implemented")
+}
+
+func (p *AlphaAuthPlugin) RequestAccessToken(params []byte) ([]byte, error) {
+	return nil, errors.New("Not yet implemented")
+}
+
+func (p *AlphaAuthPlugin) RevokeAccessToken(params []byte) error {
```
DeirdreHolub: RevokeAccessToken(token string) error 

Tokens are passed around the net as base64 encoded strings. Internally, these methods will decode the string, validate the token, and then do the thing. Where the type is as well known as this, I think it's worth being explicit.

bcda/auth/plugin/alpha.go
```diff
@@ -0,0 +1,43 @@
...
+
+func (p *AlphaAuthPlugin) GenerateClientCredentials(params []byte) ([]byte, error) {
+	return nil, errors.New("Not yet implemented")
+}
+
+func (p *AlphaAuthPlugin) RevokeClientCredentials(params []byte) error {
+	return errors.New("Not yet implemented")
+}
+
+func (p *AlphaAuthPlugin) RequestAccessToken(params []byte) ([]byte, error) {
```
DeirdreHolub: As discussed below, I think the return value is a specific type, Token. The params are correct as []byte, because different providers of tokens might require different parameters to construct tokens. But a Token is a Token.

bcda/auth/plugin/alpha.go
```diff
@@ -0,0 +1,43 @@
+package auth
+
+import (
+	"errors"
+)
+
+type AlphaAuthPlugin struct{}
+
+func (p *AlphaAuthPlugin) RegisterClient(params []byte) error {
```
DeirdreHolub: RegisterClient(params []byte) ([]byte, error)

Need a return value; it will contain (among other things) the id by which the AuthProvider knows this client. Otherwise, we don't know what id to tell it to perform all the client-related methods on.

PR #118, opened 2018-12-19
Knollfear/bcda 664 Log bad Token ID's

knollfear:
### Fixes [BCDA-664](https://jira.cms.gov/browse/BCDA-664)

When 'bad' tokens are presented we do not log anything about the actual token used.  Logging the token ID might be helpful in some cases when trying to track down a problem.  

### Proposed changes:

Track the token ID for rejected requests under the following cases:
    Revoked or blacklisted token
   Invalid token (expired most likely)
    Mismatch between ACO and Token
    
Other cases might result in exposure of PII because a user made a request with a "token" value that happened to be PII.
### Security Implications

The changes made do log something about a token, but the information logged should be free from PII.

### Acceptance Validation

Make requests with invalid tokens and verify that appropriate log messages are generated

### Feedback Requested

Any other appropriate logging or if there are concerns about what is being logged.

### DIFF
--- a/bcda/auth/middleware.go
+++ b/bcda/auth/middleware.go
@@ -64,7 +64,12 @@ func RequireTokenAuth(next http.Handler) http.Handler {
 				ctx := context.WithValue(r.Context(), "token", token)
 				next.ServeHTTP(w, r.WithContext(ctx))
 				return
+			} else {
+				log.Error("Blacklisted Token with ID: %v was rejected", token.Claims.(jwt.MapClaims)["id"])
 			}
+
+		} else {
+			log.Error("Invalid Token with ID: %v was rejected", token.Claims.(jwt.MapClaims)["id"])
 		}
 
 		oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, "", responseutils.TokenErr)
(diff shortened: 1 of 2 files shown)


bcda/auth/middleware.go
```diff
@@ -64,7 +64,12 @@ func RequireTokenAuth(next http.Handler) http.Handler {
 				ctx := context.WithValue(r.Context(), "token", token)
 				next.ServeHTTP(w, r.WithContext(ctx))
 				return
+			} else {
+				log.Error("Blacklisted Token with ID: %v was rejected", token.Claims.(jwt.MapClaims)["id"])
 			}
+
+		} else {
+			log.Error("Invalid Token with ID: %v was rejected", token.Claims.(jwt.MapClaims)["id"])
```
dhgreene: I know enough to be concerned that this is where PII could get logged, but not enough to be certain the preceding logic is enough to have already filtered it out.  It should be a well-formed token, but still . . .  What's the worst case we could see get logged here?

knollfear: Right now only tokens signed with our key pair will work. So this should only ever be the UUID of the token thast we put in it when we created it.  I am ONLY logging when we find the expected claims to prevent logging when someone presents "bearer <SOCIAL_SECURITY_NUMBER>" as their token.

PR #121, opened 2018-12-20
Dj/bcda 632

DeirdreHolub:
### Completes [BCDA-632](https://jira.cms.gov/browse/BCDA-632) *and* [BCDA-633](https://jira.cms.gov/browse/BCDA-633)

In order to complete these before leaving on Xmas vacation, I needed to combine in one PR. The commits are discrete enough that if you really want to, you can separate them. I'll leave that up to y'all.

### Proposed changes:

This PR implements the RegisterClient() and RequestAccessToken() methods for the alpha auth plugin.

### Change Details

See the description [here](https://confluence.cms.gov/pages/viewpage.action?pageId=128963277)

### Security Implications
None. The methods are currently not accessible.

### Acceptance Validation
Tests included! (but no batteries)


bcda/models/models.go
```diff
@@ -96,7 +98,7 @@ type User struct {
 	Name  string    `json:"name"`                                   // name
 	Email string    `json:"email"`                                  // email
 	Aco   ACO       `gorm:"foreignkey:AcoID;association_foreignkey:UUID"`
-	AcoID uuid.UUID `gorm:"type:char(36)" json:"aco_id"` // aco_id
+	AcoID uuid.UUID `gorm:"type:char(36)" json:"aco_id"` 			// aco_id
```
embh: Can we get the ID from the ACO model instead of having an additional field?

bcda/auth/plugin/alpha.go
```diff
@@ -1,15 +1,60 @@
...
+
+	acoUUID, ok := paramsMap["clientID"].(string)
+	if !ok {
+		return empty, errors.New("missing or otherwise invalid clientID")
+	}
+
+	// We'll check carefully in this method, because we're returning something to be used as an id
+	// Normally, a plugin would treat this value as a black box external key, but this implementation is
+	// intimate with the API. So, we're going to protect against accidental bad things
+	if acoUUID == "" || len(acoUUID) > 36 {
```
embh: Do we need this when we have the regex matcher right after it?

bcda/auth/plugin/alpha.go
```diff
@@ -20,18 +65,86 @@ func (p *AlphaAuthPlugin) DeleteClient(params []byte) error {
...
+		Value:  tokenString,
+		Active: true,
+		Token:  jwtToken,
+	}
+
+	if err = db.Create(&token).Error; err != nil {
+		return jwtToken, err
+	}
+
+	return jwtToken, err // really want to return auth.Token here, but first let's get this all working
```
embh: Should we have the return type be `auth.Token` in the interface?

bcda/models/models.go
DeirdreHolub: I think you're asking about the semantics of GORM. How to describe associations with GORM: http://doc.gorm.io/associations.html

We are using the 'belongs to' approach. We could possibly have used the 'has one' approach described later in the page, but I don't know whether or how the resulting tables and constraints would differ.

bcda/auth/plugin/alpha.go
DeirdreHolub: Conventional wisdom is that regex is expensive, so always make it the last check, after you've ruled out other failures. I also like the documentation aspect of length checking the string, particularly when length is a requirement of the format.

bcda/auth/plugin/alpha.go
DeirdreHolub: yes. I want to do that in the next few changes, where we actually start working with the tokens.

PR #122, opened 2018-12-21
BCDA-681 Feature: Create-token command now persist to the db

tbellj:
Fixes [BCDA-681](https://jira.cms.gov/browse/BCDA-681)

Problem: The `create-token` command does not result in any information being persisted to the DB.

Proposed changes:
bcda/main.go - added code to persist token to the db
bcda/main_test.go - added test for revoking a token created from the "create-token" command

Security Implications:
None

Acceptance Validation:
Call the create-token command and be sure that information is persisted to the DB. Revoke-token should also successfully execute.

Feedback Requested:
Any

### DIFF
--- a/bcda/main.go
+++ b/bcda/main.go
@@ -386,14 +386,24 @@ func createAccessToken(acoID, userID string) (string, error) {
 		return "", errors.New(strings.Join(errMsgs, "\n"))
 	}
 
+	db := database.GetGORMDbConnection()
+	defer db.Close()
+	var aco models.ACO
+	var user models.User
+
+	if db.First(&aco, "UUID = ?", acoID).RecordNotFound() {
+		return "", fmt.Errorf("unable to locate ACO with id of %s", acoID)
+	}
+	if db.First(&user, "UUID = ?", userID).RecordNotFound() {
+		return "", fmt.Errorf("unable to locate User with id of %s", userID)
+	}
+
 	authBackend := auth.InitAuthBackend()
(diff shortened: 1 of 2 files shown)


bcda/main.go
```diff
@@ -386,14 +386,24 @@ func createAccessToken(acoID, userID string) (string, error) {
...
+	defer db.Close()
+	var aco models.ACO
+	var user models.User
+
+	if db.First(&aco, "UUID = ?", acoID).RecordNotFound() {
+		return "", fmt.Errorf("unable to locate ACO with id of %s", acoID)
+	}
+	if db.First(&user, "UUID = ?", userID).RecordNotFound() {
+		return "", fmt.Errorf("unable to locate User with id of %s", userID)
+	}
```
embh: Should we also check that the user is part of the ACO?

knollfear: agreed, but let's take this a step further and remove the need for an ACO entirely.  Users are scoped to a single ACO, just figure out the ACO from the user.  The command will then just take a user ID as the only input.  Making this change will just require deleting some existing code.

bcda/main.go
```diff
@@ -361,18 +356,10 @@ func createUser(acoID, name, email string) (string, error) {
 	return user.UUID.String(), nil
 }
 
-func createAccessToken(acoID, userID string) (string, error) {
+func createAccessToken(userID string) (string, error) {
 	errMsgs := []string{}
-	var acoUUID, userUUID uuid.UUID
+	var userUUID uuid.UUID
```
knollfear: This seems unused

PR #123, opened 2018-12-27
knollfear/bcda-600 Postman in CI/CD

knollfear:
### Fixes [BCDA-600](https://jira.cms.gov/browse/BCDA-600)

We have a collection of postman tests that can be executed on a command line with Newman.  We should set those up to run in jenkins as part of the release process.  This is the dockerfiles and environments for that to happen

### Proposed changes:

Adding docker file for newman tests
adding make file command for running test suite locally
renaming postman collections to just be more CL friendly

### Security Implications

This does generate a token in dev, but it is not persisted

### Acceptance Validation

Run the tests locally with "make env=dev newman" 
See the step in jenkins when build/package runs

image

### Feedback Requested

input on the design would be welcome, devops is not my forte


Makefile
```diff
@@ -29,6 +29,13 @@ test:
 	docker-compose up -d db queue
 	docker-compose -f docker-compose.test.yml up --build --force-recreate --exit-code-from unit_test unit_test
 
+newman:
+	# This target should be executed by passing in an argument for the environment (dev/test/sbx)
+	# and if needed a token.
+	# For example: make newman env=test token=<MY_TOKEN>
+	docker build . -f Dockerfiles/Dockerfile.newman_test -t newman_test
```
msnook: Since this is intended to execute against a running application, you'll need to bring up the app similar to the other test targets.  Additionally, the test container should really be tied into the `docker-compose.test.yml` file so all the networking just "works".

test/api.postman_environment.json
```diff
@@ -0,0 +1,27 @@
+{
+	"id": "5cca63fb-26dd-4a87-8463-a55e7f56551f",
```
msnook: Are these config files needed?  Can the config params just be passed in at runtime instead of requiring a file for each env? 
To make life easy in local testing env, could default everything to `http` and `api:3000`.

Makefile
```diff
@@ -29,6 +29,18 @@ test:
...
 	docker-compose -f docker-compose.test.yml up --build --force-recreate --exit-code-from unit_test unit_test
 
+postman:
+	# This target should be executed by passing in an argument for the environment (dev/test/sbx)
+	# and if needed a token.
+	# For example: make postman env=test token=<MY_TOKEN>
+	docker-compose -f docker-compose.test.yml run postman_test test/$(env).postman_environment.json --global-var "token=$(token)"
+
+
+postman-local:
```
msnook: Do we need two postman targets?  Might get confusing as to which one to use in which situation. Can we get away with just one?

Makefile
```diff
@@ -29,6 +29,18 @@ test:
...
+	docker-compose up -d
+	sleep 30
+	docker-compose -f docker-compose.test.yml run postman_test test/local.postman_environment.json
```
msnook: To ensure we’re not using cached images or containers, can we build and force recreate (similar to the other make targets)?

msnook: We are going to need to capture the exit code from the container similar to the other make targets. Otherwise the exit code won’t propagate to Travis properly.

Makefile
knollfear: One spins up the local environment and the other one doesn't.  I don't see any reason to spin that up if we are just testing against an external source.  I will add a comment so that it is more clear.

Makefile
knollfear: sure.  I will make this change shortly.

knollfear: It looks like run actually just returns the exit code by default with no additional specifications required.

Dockerfiles/Dockerfile.postman_test
```diff
@@ -0,0 +1,5 @@
+FROM postman/newman
+WORKDIR /go/src/github.com/CMSgov/bcda-app
+COPY . .
+RUN newman -v
```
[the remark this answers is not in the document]
knollfear: I can remove it.  I had some weird problems getting it to work and I suspected a bug in the latest version.

Makefile
knollfear: confirmed on this build  https://travis-ci.org/CMSgov/bcda-app/builds/473104369?utm_source=github_status&utm_medium=notification

Makefile
```diff
@@ -29,6 +29,22 @@ test:
 	docker-compose up -d db queue
 	docker-compose -f docker-compose.test.yml up --build --force-recreate --exit-code-from unit_test unit_test
 
+postman:
```
msnook: This will need a `docker-compose up -d` added to it, to ensure the app is running before running tests.

Makefile
```diff
@@ -29,6 +29,22 @@ test:
...
+	# This target should be executed by passing in an argument for the environment (dev/test/sbx)
+	# and if needed a token.
+	# For example: make postman env=test token=<MY_TOKEN>
+	docker-compose -f docker-compose.test.yml build --no-cache postman_test
+	docker-compose -f docker-compose.test.yml run --rm postman_test test/$(env).postman_environment.json --global-var "token=$(token)"
+
+
+postman-local:
```
msnook: Why do we need `postman-local` target?  Can't we just use the `postman` target for both situations?
For example, when running in scenario where token is not needed: `make postman env=api`
And when running in scenario where token is needed: `make postman env=test token=<MY_TOKEN>`

Makefile
knollfear: No, we don't.  If we are running postman against dev having it running locally is completely unnecessary and will have no positive effect.  If it isn't running on dev then local docker commands won't solve that.

Makefile
knollfear: added in an ifeq statement to make this more responsive and only require a single command

.travis.yml
```diff
@@ -47,3 +47,4 @@ script:
   - make test
   - make smoke-test
   - make integration-test
+  - make postman-local
```
msnook: With the new changes, this should just be `postman`, right?

Makefile
```diff
@@ -60,4 +72,4 @@ debug-worker:
 	@-bash -c "trap 'docker-compose stop' EXIT; \
 		docker-compose -f docker-compose.yml -f docker-compose.debug.yml run --no-deps -T --rm -v $(shell pwd):/go/src/github.com/CMSgov/bcda-app worker dlv debug"
 
-.PHONY: docker-build docker-bootstrap load-fixtures test debug-api debug-worker api-shell worker-shell package release smoke-test integration-test
+.PHONY: docker-build docker-bootstrap load-fixtures test debug-api debug-worker api-shell worker-shell package release smoke-test integration-test postman postman-local
```
msnook: Can remove `postman-local`.

PR #124, opened 2018-12-31
Knollfear/bcda 657 Variable SIzed ACOs for alpha tokens

knollfear:
### Fixes [BCDA-657](https://jira.cms.gov/browse/BCDA-657)

Currently all ACO's in the alpha program get a test ACO with 50 beneficiaries.  To further test our system we need to be able to create tokens for larger sized ACO's
### Proposed changes:

All specification of "dev", "small", "medium" , or "Large" sized ACO when creating a token.  Use a CLI variable similar to the TTL variable.

### Change Details

altered beneficiary query assignment to use different ACO's as the baseline.  
added size argument to Alpha token creation command
added some tests

### Security Implications

no security implications.  THis just changes the number of synthetic beneficiaries we are already using

### Acceptance Validation

run tests
make an alpha ACO to verify that it has the right number of beneficiaries

### Feedback Requested

any


DeirdreHolub:
I've commented several times in meetings that creating an ACO with a big population is problematic because the first task a user tackles is just to get stuff working (can I connect properly and download the data? can I parse it and do something useful with it?), and for that you want a small set of data so it doesn't take forever. After you do that, then you move on to 'can I do this at volume'. I think this should be a task to 'add X more beneficiaries to this ACO'.

knollfear:
> I've commented several times in meetings that creating an ACO with a big population is problematic because the first task a user tackles is just to get stuff working (can I connect properly and download the data? can I parse it and do something useful with it?), and for that you want a small set of data so it doesn't take forever. After you do that, then you move on to 'can I do this at volume'. I think this should be a task to 'add X more beneficiaries to this ACO'.

I don't necessarily disagree with you.  However, that isn't what the ticket said to do.  If you want to re litigate the ticket feel free.  I did this the way I did because that was how the ticket was speced.

DeirdreHolub:
>> I don't necessarily disagree with you. However, that isn't what the ticket said to do. If you want to re litigate the ticket feel free. I did this the way I did because that was how the ticket was speced.
>
Agreed. That's why this is a comment, not part of the review.

PR #126, opened 2019-01-03
bcda-630: add decodeAccessToken and unit test for alpha plugin

alexhova:
### Fixes [BCDA-630](https://jira.cms.gov/browse/BCDA-630)
adds a working `decodeAccessToken()` func and an associated unit test for the alpha plugin

### Proposed changes:
- make `decodeAccessToken()` work
- write a simple unit test to test it

### Acceptance Validation
test passes!

### DIFF
--- a/bcda/auth/plugin/alpha.go
+++ b/bcda/auth/plugin/alpha.go
@@ -11,11 +11,17 @@ import (
 	"github.com/CMSgov/bcda-app/bcda/database"
 	"github.com/CMSgov/bcda-app/bcda/models"
 
-	"github.com/dgrijalva/jwt-go"
+	jwt "github.com/dgrijalva/jwt-go"
 	"github.com/pborman/uuid"
 )
 
 type AlphaAuthPlugin struct{}
+
+type CustomClaims struct {
+	Aco string `json:"aco"`
+	ID  string `json:"id"`
+	jwt.StandardClaims
+}
 
 // This implementation expects one value in params, an id the API knows this client by in string form
 // It returns a single string as well, being the clientID this implementation knows this client by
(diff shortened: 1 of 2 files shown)


bcda/auth/plugin/alpha.go
```diff
@@ -154,13 +156,24 @@ func (p *AlphaAuthPlugin) ValidateAccessToken(token string) error {
 }
 
 func (p *AlphaAuthPlugin) DecodeAccessToken(token string) (jwt.Token, error) {
-	return jwt.Token{}, errors.New("Not yet implemented")
+	t, err := jwt.Parse(token, func(token *jwt.Token) (interface{}, error) {
+		if _, ok := token.Method.(*jwt.SigningMethodRSA); !ok {
+			return nil, fmt.Errorf("Unexpected signing method: %v", token.Header["alg"])
```
alexhova: is there a better way to handle/log this error?

DeirdreHolub:
If you search for GetJWTClaims, you'll find at least most (all?) of the places where a token's claims are examined. There are several. I was hoping to clean up the handling of claims, probably using the ParseWithClaims method. An example here: https://godoc.org/github.com/dgrijalva/jwt-go#pkg-examples. The last line in the example is catching my particular attention.

bcda/auth/plugin/alpha_test.go
```diff
@@ -95,9 +96,10 @@ func (s *AlphaAuthPluginTestSuite) TestValidateAccessToken() {
 }
 
 func (s *AlphaAuthPluginTestSuite) TestDecodeAccessToken() {
-	t, err := s.p.DecodeAccessToken("")
+	ts, _ := auth.InitAuthBackend().GenerateTokenString("504", "911")
```
DeirdreHolub: You should use UUIDs as params here, since that's what we use in the tokens we generate. You can use uuid.NewRandom().String(), or you can consult the test for GenerateTokenString() in auth/backend_test.go for fixture data UUIDs. This will important with respect to parsing claims and validation.

bcda/auth/plugin/alpha.go
embh: Probably down to personal preference, but I think this error handling makes sense, with it bubbling up and being returned from `DecodeAccessToken()` later. The calling function can decide how to handle it and whether it should be logged.

PR #128, opened 2019-01-09
BCDA-637: Implement RevokeClientCredentials() on alpha plugin

embh:
### Fixes [BCDA-637](https://jira.cms.gov/browse/BCDA-637)
`RevokeClientCredentials()` is not yet implemented in our alpha plugin.

### Proposed changes:
Implement `RevokeClientCredentials()` in alpha.go.

### Change Details
* Looks up ACO by client ID, users by ACO, and tokens by users. Calls `RevokeToken()` on any tokens found.
* Test added in alpha_test.go that will need to be fully implemented with `RevokeToken()`.

### Security Implications
No changes.

### Acceptance Validation
Input is the same. Currently returns an error due to `RevokeToken()` not being implemented yet.

### Feedback Requested
Do we want to return errors in the case of the ACO, users, or tokens not being found? Or should it return nil because if they're not found, there's nothing to revoke?

### DIFF
--- a/bcda/auth/plugin/alpha.go
+++ b/bcda/auth/plugin/alpha.go
@@ -118,7 +118,7 @@ func (p *AlphaAuthPlugin) RevokeClientCredentials(params []byte) error {
 	}
 	log.Info(fmt.Sprintf("%d token(s) revoked", revokedCount))
 	if len(errs) > 0 {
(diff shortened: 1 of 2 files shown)


DeirdreHolub:
needs BCDA-635

PR #129, opened 2019-01-10
BCDA-636: implement GenerateClientCredentials on Alpha plugin

DeirdreHolub:
### Fixes [BCDA-636](https://jira.cms.gov/browse/BCDA-636)

GenerateClientCredentials() needs to be implemented on the alpha plugin. The intent of this method is to generate new or to replace existing credentials with new ones.

### Proposed changes:

Verifies that ACO has a ClientID (i.e., RegisterClient() has been called for the ACO previously).
Will use RevokeClientCredentials() to revoke any existing credentials (needs merge of BCDA-637)
Uses CreateAccessToken() to generate credentials.
Returns a token string.
Adds tests. 

### Security Implications
No changes.

### Acceptance Validation
All tests pass.

### Feedback Requested
Any objections to revoking existing credentials as part of this call?

### DIFF
--- a/bcda/auth/plugin/alpha.go
+++ b/bcda/auth/plugin/alpha.go
@@ -83,12 +83,10 @@ func (p *AlphaAuthPlugin) GenerateClientCredentials(params []byte) ([]byte, erro
 		return empty, fmt.Errorf("ACO %s does not have a registered client", clientID)
 	}
 
-	/* uncomment when PR 128 implementing BCDA-637 is merged
 	err = p.RevokeClientCredentials([]byte(fmt.Sprintf(`{"clientID":"%s"}`, clientID)))
 	if err != nil {
 		return empty, fmt.Errorf("unable to revoke existing credentials for ACO %s because %s", clientID, err)
 	}
-	*/
 
 	jwtToken, err := p.RequestAccessToken([]byte(params))
 	if err != nil {
--- a/bcda/auth/plugin/alpha.go
+++ b/bcda/auth/plugin/alpha.go
@@ -29,26 +29,24 @@ type CustomClaims struct {
 // It returns a single string as well, being the clientID this implementation knows this client by
 // NB: Other implementations will probably expect more input, and will certainly return more data
 func (p *AlphaAuthPlugin) RegisterClient(params []byte) ([]byte, error) {
-	var empty []byte
-
 	acoUUID, err := GetParamString(params, "clientID")
 	if err != nil {
-		return empty, err
+		return nil, err
 	}
 
 	// We'll check carefully in this method, because we're returning something to be used as an id
 	// Normally, a plugin would treat this value as a black box external key, but this implementation is
 	// intimate with the API. So, we're going to protect against accidental bad things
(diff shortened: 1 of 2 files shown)


bcda/auth/plugin/alpha.go
```diff
@@ -68,7 +68,36 @@ func (p *AlphaAuthPlugin) DeleteClient(params []byte) error {
 
 // can treat as a no-op or call RequestAccessToken
 func (p *AlphaAuthPlugin) GenerateClientCredentials(params []byte) ([]byte, error) {
-	return nil, errors.New("not yet implemented")
+	var empty []byte
```
embh: It might be a little more readable to call this `responseBody` or something along those lines, then setting it to equal ``fmt.Sprintf(`{"tokenString":"%s"}`, tokenString)`` when successful.

bcda/auth/plugin/alpha.go
```diff
@@ -68,7 +68,36 @@ func (p *AlphaAuthPlugin) DeleteClient(params []byte) error {
...
+	clientID, err := GetParamString(params, "clientID")
+	if err != nil {
+		return nil, err
```
embh: Why nil rather than an empty byte slice?

bcda/auth/plugin/alpha_test.go
```diff
@@ -33,6 +34,22 @@ func (s *AlphaAuthPluginTestSuite) SetupTest() {
...
+var connections = make(map[string]*gorm.DB)
+
+func (s *AlphaAuthPluginTestSuite) BeforeTest(suiteName, testName string) {
+	connections[testName] = database.GetGORMDbConnection()
+}
+
+func (s *AlphaAuthPluginTestSuite) AfterTest(suiteName, testName string) {
+	c, ok := connections[testName]
+	if !ok {
+		s.FailNow("WTF? no db connection for %s", testName)
```
embh: 😂

bcda/auth/plugin/alpha_test.go
```diff
@@ -33,6 +34,22 @@ func (s *AlphaAuthPluginTestSuite) SetupTest() {
 	s.p = new(AlphaAuthPlugin)
 }
 
+var connections = make(map[string]*gorm.DB)
+
+func (s *AlphaAuthPluginTestSuite) BeforeTest(suiteName, testName string) {
+	connections[testName] = database.GetGORMDbConnection()
```
embh: This is a really neat solution!

bcda/auth/plugin/alpha.go
DeirdreHolub: I thought of that as well, but that seems to run against the go gestalt of nil-ish response. We may also change the signature of the methods, so let's defer this conversation until then.

bcda/auth/plugin/alpha.go
DeirdreHolub: Handling was inconsistent throughout the file. It turns out that in go, an empty slice is nil until it is assigned a value, making this a completely pointless variable. It's also not idiomatic go, so I've  replaced these all with nil.

PR #131, opened 2019-01-10
bcda-631: Implement ValidateAccessToken() for alpha plugin

alexhova:
### Fixes [BCDA-631](https://jira.cms.gov/browse/BCDA-631)
implements ValidateAccessToken() in the alpha plugin

### Proposed changes:
- made ValidateAccessToken() a real function
- wrote a unit test that tests a valid token and an invalid one

### Acceptance Validation
tests pass and code continues to work!


embh:
This looks good to me, but is a little different from what the ticket describes (the moving code part). I'd like to get confirmation from @DeirdreHolub that there isn't anything else that needs to change before approving.

bcda/auth/plugin/alpha_test.go
```diff
@@ -145,8 +145,13 @@ func (s *AlphaAuthPluginTestSuite) TestRevokeAccessToken() {
...
 
 func (s *AlphaAuthPluginTestSuite) TestValidateAccessToken() {
-	err := s.p.ValidateAccessToken("")
-	assert.Equal(s.T(), "not yet implemented", err.Error())
+	userID := uuid.NewRandom().String()
+	acoID := uuid.NewRandom().String()
+	ts, _ := auth.InitAuthBackend().GenerateTokenString(userID, acoID)
+	err := s.p.ValidateAccessToken(ts)
+	assert.Nil(s.T(), err)
+	err = s.p.ValidateAccessToken("this.token.fails")
```
DeirdreHolub: This test should construct a structurally valid token that is invalid because 

- it is missing required claims
- the signature is invalid (because it was signed using some other key
- it was signed using the wrong algorithm

bcda/auth/plugin/alpha.go
```diff
@@ -193,7 +192,11 @@ func (p *AlphaAuthPlugin) RevokeAccessToken(token string) error {
 }
 
 func (p *AlphaAuthPlugin) ValidateAccessToken(token string) error {
-	return errors.New("not yet implemented")
+	t, err := p.DecodeAccessToken(token)
+	if err != nil {
+		return err
+	}
+	return t.Claims.(*CustomClaims).Valid()
```
DeirdreHolub: See auth/middleware.go for additional validations that should be incorporated here. Specifically:
* signed with the right algorithm in ParseToken()
* blacklisted check in RequireTokenAuth()
* aco string matches pattern and is in id in RequireTokenACOMatch() (this one should always be checked, not conditionally checked as it is now)

bcda/auth/plugin/alpha_test.go
DeirdreHolub: Instead of calling it CustomClaims, can we rename it to AllClaims, or something along those lines? It does contain all of them, not just the custom claims ;)

knollfear: The second 2 can be done externally on the JWT.io site and work off of a static token string.

bcda/auth/plugin/alpha.go
```diff
@@ -248,18 +247,37 @@ func revokeAccessTokenByID(tokenID uuid.UUID) error {
...
+	if err != nil {
+		return err
+	}
+
+	b := isBlacklisted(t)
+	if b {
+		return fmt.Errorf("blacklisted Token with ID: %v was rejected", t.Claims.(*AllClaims).ID)
+	}
+
+	return t.Claims.(*AllClaims).Valid()
```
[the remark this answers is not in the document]
alexhova: `t.Valid` returns a bool and is populated when you parse a token.  `Claims.Valid()` runs all of the standard verifications and returns an error.  It's a little nicer to have the error over just `false`

bcda/auth/plugin/alpha.go
```diff
@@ -248,18 +247,37 @@ func revokeAccessTokenByID(tokenID uuid.UUID) error {
...
+}
+
+func isBlacklisted(token jwt.Token) bool {
+	claims := token.Claims.(*AllClaims)
+
+	db := database.GetGORMDbConnection()
+	defer db.Close()
+
+	return !db.Find(token, "UUID = ? AND active = ?", claims.ID, false).RecordNotFound()
```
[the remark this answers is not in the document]
alexhova: yyr

bcda/auth/plugin/alpha_test.go
alexhova: that's what i did!  you can't even sign it if you don't have the required claims though it seems.

bcda/auth/plugin/alpha.go
alexhova: the right algorithm is technically checked in `DecodeAccessToken()` which is the first line of this validation func

bcda/auth/plugin/alpha_test.go
alexhova: I thought `CustomClaims` was a good name.  I know it's not _just_ custom claims.  but it is our custom set of claims including some standard and some custom claims.  I changed it to `AllClaims` anyway but feel free to change again in a later PR if you think it could be clearer

bcda/auth/plugin/alpha.go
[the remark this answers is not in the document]
alexhova: i agree with you on the variable naming but it is [go convention](https://github.com/golang/go/wiki/CodeReviewComments#variable-names)
since go is statically typed, we can tell easily the type of any variable

bcda/auth/plugin/alpha.go
```diff
@@ -248,18 +247,44 @@ func revokeAccessTokenByID(tokenID uuid.UUID) error {
...
-func (p *AlphaAuthPlugin) ValidateAccessToken(token string) error {
-	return errors.New("not yet implemented")
+func (p *AlphaAuthPlugin) ValidateAccessToken(tokenString string) error {
+	t, err := p.DecodeAccessToken(tokenString)
+	if err != nil {
+		return err
+	}
+
+	c := t.Claims.(*AllClaims)
+
```
knollfear: I think you should be checking the validity of the claim here  "t.Claims.(*AllClaims).Valid()" is checkabel based on internal locally stored info.  The other 2 checks rely on a DB call which is fundamentally more expensive than a local check.  If we can exit before doing them we should.

alexhova: ✔️

bcda/auth/plugin/alpha.go
```diff
@@ -248,18 +247,65 @@ func revokeAccessTokenByID(tokenID uuid.UUID) error {
...
+	return nil
+}
+
+func isBlacklisted(token jwt.Token) bool {
+	c := token.Claims.(*AllClaims)
+
+	db := database.GetGORMDbConnection()
+	defer db.Close()
+
+	return !db.Find(&token, "UUID = ? AND active = ?", c.ID, false).RecordNotFound()
```
DeirdreHolub: This function will return false when the token id is not found in the db. That's not a Good Thing.

DeirdreHolub: Then someone would be allowed to use the system if they submit a token with a non-existent database id. That's not good in any circumstance. @embh I think you're being too literal here ;) An alternative is to put the check on token id existence in separate code before the blacklist check. That approach would fit the semantics better.

embh: I would definitely err on the side of keeping this literal so it's clear what the value returned from this function is telling you. Checking that the ID is in the database separately makes sense (assuming that's something we need to check).

PR #133, opened 2019-01-14
BCDA-635: Implement RevokeAccessToken()

embh:
### Fixes [BCDA-635](https://jira.cms.gov/browse/BCDA-635)
The `RevokeAccessToken()` function in our alpha plugin is not yet implemented.

### Proposed changes:
Implement `RevokeAccessToken()` and use it in the CLI.

### Change Details
* Implements `RevokeAccessToken()` and tests
* Uses `RevokeAccessToken()` for the `revoke-token` CLI command
* Removes old token revocation code and tests
* Adds `revokeAccessTokenByID()` to the plugin, called by `RevokeAccessToken()` and `RevokeClientCredentials()`

### Security Implications
No changes.

### Acceptance Validation
Tokens can still be successfully revoked from the CLI.

### Feedback Requested
Are we okay with the approach of the having a private function to revoke a token that accepts the ID, or do we want to change the parameter(s) for the public function? Any other thoughts?

### DIFF
--- a/bcda/auth/plugin/alpha.go
+++ b/bcda/auth/plugin/alpha.go
@@ -10,6 +10,7 @@ import (
 	"github.com/CMSgov/bcda-app/bcda/auth"
 	"github.com/CMSgov/bcda-app/bcda/database"
 	"github.com/CMSgov/bcda-app/bcda/models"
+	"github.com/jinzhu/gorm"
 
 	"github.com/dgrijalva/jwt-go"
 	"github.com/pborman/uuid"
@@ -110,7 +111,7 @@ func (p *AlphaAuthPlugin) RevokeClientCredentials(params []byte) error {
 	var errs []string
 	revokedCount := 0
 	for _, t := range tokens {
-		err := p.RevokeAccessToken(fmt.Sprintf(`{"token":"%s"}`, t.TokenString))
+		err := revokeAccessTokenByID(t.UUID)
 		if err != nil {
 			log.Error(err)
 			errs = append(errs, err.Error())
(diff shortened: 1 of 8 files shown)


bcda/auth/plugin/alpha.go
```diff
@@ -188,8 +189,31 @@ func (p *AlphaAuthPlugin) RequestAccessToken(params []byte) (jwt.Token, error) {
 }
 
 // lookup token and set active flag to false
-func (p *AlphaAuthPlugin) RevokeAccessToken(token string) error {
-	return errors.New("not yet implemented")
+func (p *AlphaAuthPlugin) RevokeAccessToken(tokenString string) error {
+
+	backend := auth.InitAuthBackend()
+	claims := backend.GetJWTClaims(tokenString)
```
alexhova: would we prefer to use `DecodeAccessToken()` and the `CustomClaims` struct in this plugin now?

bcda/auth/plugin/alpha_test.go
```diff
@@ -140,8 +135,33 @@ func (s *AlphaAuthPluginTestSuite) TestRequestAccessToken() {
...
+	_, tokenString, _ := s.AuthBackend.CreateToken(user)
+
+	assert := assert.New(s.T())
+
+	err := s.p.RevokeAccessToken(userID)
+	assert.NotNil(err)
+
+	err = s.p.RevokeAccessToken(tokenString)
+	assert.Nil(err)
+	jwtToken, err := s.AuthBackend.GetJWToken(tokenString)
```
alexhova: and similarly using the plugin method to decode here

bcda/auth/plugin/alpha.go
```diff
@@ -217,9 +218,30 @@ func (p *AlphaAuthPlugin) RequestAccessToken(params []byte) (jwt.Token, error) {
...
+
+	if c, ok := t.Claims.(*CustomClaims); ok {
+		return revokeAccessTokenByID(uuid.Parse(c.ID))
+	}
+
+	return errors.New("could not read token claims")
+}
+
+func revokeAccessTokenByID(tokenID uuid.UUID) error {
+	db := database.GetGORMDbConnection()
```
DeirdreHolub: missing defer db.close()

PR #136, opened 2019-01-16
BCDA-661: Implement worker health check

embh:
### Fixes [BCDA-661](https://jira.cms.gov/browse/BCDA-661)
We would like to be able to monitor the health of worker processes.

### Proposed changes:
Write worker health status to a log file.

### Change Details
* Worker application will check connections to the database and Blue Button at a set interval and log to a file
* Uses environment variables:
  * `WORKER_HEALTH_LOG` for the location of the log file
  * `WORKER_HEALTH_INT_SEC` for the interval at which workers should write to the log, in seconds. If this is not set, workers will not log health status, so we can disable it without redeploying.
* Log entries are JSON-formatted: `{"bb":"(ok|error)","db":"(ok|error)","id":"(uuid)","level":"info","msg":"","time":"yyyy-MM-ddTHH:mm:ssZ","type":"health"}`
* This does not include EC2-related monitoring and recovery, per team discussion yesterday.

### Security Implications
None

### Acceptance Validation
Confirmed that invalid database or Blue Button URLs lead to "error" values in the log entries.

screen shot 2019-01-16 at 9 10 30 am
screen shot 2019-01-16 at 9 17 14 am
screen shot 2019-01-16 at 9 10 21 am

### Feedback Requested
Any.


bcda/health/health.go
```diff
@@ -0,0 +1,28 @@
+package health
+
+import (
+	"database/sql"
+	"os"
+
+	"github.com/CMSgov/bcda-app/bcda/client"
+)
+
```
msnook: Just a thought, and out of scope for this ticket, but should we add a check to ensure we can access the Swagger docs?  Since they are hosted via a different path, maybe we can include it in our health check for BCDA API?  Thoughts?  Maybe we can create a separate ticket for this to be done at a later time.

msnook:
Do we need follow-up tickets for Splunk (ingest the file if we aren't already), and Ops (somehow detect if we are unhealthy and attempt to self-heal via ASG)?

bcdaworker/health.go
```diff
@@ -0,0 +1,56 @@
...
+type HealthLogger struct {
+	Logger *logrus.Logger
+}
+
+func (l *HealthLogger) Log() {
+	entry := &HealthLogEntry{Logger: logrus.NewEntry(l.Logger)}
+
+	logFields := logrus.Fields{}
+	logFields["type"] = "health"
+	logFields["id"] = uuid.NewRandom()
```
dhgreene: Interesting idea.  Is a UUID for every message a pattern we're following?

embh: It was requested in @rnagle's comment on the ticket. Now I'm wondering if we should be including some kind of identifier for the worker. 🤔

bcda/health/health.go
tbellj: we kinda already hit the swagger docs in new relic, but that is currently only polling every hour. If we need to do it more often than that then I agree with Matt.

msnook: I was thinking that adding it to the health check would ensure that we fail the deployment if swagger is unreachable at the time we deploy... that way we can roll back to the last stable deployment automatically rather than deploying and letting new relic alerts tell us that swagger is unreachable (after we've already deployed)

PR #145, opened 2019-01-21
BCDA-396 Feature: Add Coverage endpoint for bulk export of patient data

tbellj:
Fixes [BCDA-396](https://jira.cms.gov/browse/BCDA-396)

Problem:
Added new endpoint.

Change Details:
bcda/api.go - added "coverage" option for a bulk request.
bcda/api_test.go - added test case for new endpoint
bcda/client/bluebutton.go - added new function for APIClient interface
bcda/router.go - added new coverage endpoint
bcdaworker/main.go - added new case for coverage function
docker-compose.yml - added ENABLE_COVERAGE_EXPORT config var
test/postman_test/BCDA_Tests.postman_collection.json- added postman tests
test/postman_test/BCDA_Tests_Sequential.postman_collection.json - added postman tests
test/smoke_test/smoke_test.sh - added smoke tests

 Security Implications
None, but env flags should be used to control enabling/disabling of this endpoint

Acceptance Validation:
Yes

Feedback Requested
Any and all

### DIFF
--- a/bcda/api.go
+++ b/bcda/api.go
@@ -98,8 +98,12 @@ func bulkPatientRequest(w http.ResponseWriter, r *http.Request) {
 	bulkRequest("Patient", w, r)
 }
 
+func bulkCoverageRequest(w http.ResponseWriter, r *http.Request) {
+	bulkRequest("Coverage", w, r)
+}
+
 func bulkRequest(t string, w http.ResponseWriter, r *http.Request) {
-	if t != "ExplanationOfBenefit" && t != "Patient" {
+	if t != "ExplanationOfBenefit" && t != "Patient" && t != "Coverage" {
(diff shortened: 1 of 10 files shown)


msnook:
Good stuff.

From a documentation perspective, I suppose all the Swagger and other doc materials are going to be handled in BCDA-777?

From an ops perspective, how will the config variables and smoke test be updated?  More specifically, do we have a ticket to do the ops piece of this, or should it be done as part of BCDA-396?  At minimum, we need the following to be done:
1) Add `Coverage` endpoint to be tested as part of the smoke tests from Jenkins in AWS (`dev` only for now)
2) Add `ENABLE_COVERAGE_EXPORT` config variable in `dev`: https://github.com/CMSgov/bcda-ops/tree/master/config/api/dev

PR #147, opened 2019-01-23
Replace [most] uses of backend.CreateToken

DeirdreHolub:
### Fixes [BCDA-761](https://jira.cms.gov/browse/BCDA-761)
Replaces uses of auth/backend.CreateToken function with the auth/Provider.RequestAccessToken function.

### Proposed changes:
CLI and /api/v1/token endpoint now use auth provider method.
middleware uses of backend will be addressed by BCDA-764.

### Change Details
CLI and token end point.

### Security Implications
No new security concerns

### Acceptance Validation
All tests continue to pass.

### Reviewer Notes
* added a fixture user that is assigned to 'ACO Dev'
* reworked some tests because RequestAccessToken() requires that the User be associated with the ACO (CreateToken did not).

### DIFF
--- a/bcda/auth/plugin/alpha_test.go
+++ b/bcda/auth/plugin/alpha_test.go
@@ -189,20 +189,20 @@ func (s *AlphaAuthPluginTestSuite) TestRequestAccessToken() {
 func (s *AlphaAuthPluginTestSuite) TestRevokeAccessToken() {
 	db := connections["TestRevokeAccessToken"]
 
-	const UserID, AcoID = "EFE6E69A-CD6B-4335-A2F2-4DBEDCCD3E73", "DBBD1CE1-AE24-435C-807D-ED45953077D3"
+	const userID, acoID = "EFE6E69A-CD6B-4335-A2F2-4DBEDCCD3E73", "DBBD1CE1-AE24-435C-807D-ED45953077D3"
 	assert := assert.New(s.T())
 
 	// Good Revoke test
-	jwtToken, err := s.p.RequestAccessToken([]byte(fmt.Sprintf(`{"clientID": "%s", "ttl": 720}`, AcoID)))
+	jwtToken, err := s.p.RequestAccessToken([]byte(fmt.Sprintf(`{"clientID": "%s", "ttl": 720}`, acoID)))
 	if err != nil {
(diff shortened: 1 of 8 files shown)


bcda/auth/plugin/alpha_test.go
```diff
@@ -151,14 +151,20 @@ func (s *AlphaAuthPluginTestSuite) TestRevokeClientCredentials() {
 	}
 	db.Save(&user)
 
-	token, _, _ := s.AuthBackend.CreateToken(user)
+	params := fmt.Sprintf(`{"clientID":"%s", "ttl":720}`, user.AcoID.String())
+	_, err := s.p.GenerateClientCredentials([]byte(params))
```
embh: Want to make @dhgreene aware of this, since it will have to get changed with https://github.com/CMSgov/bcda-app/pull/140 if this PR is merged first.

bcda/auth/plugin/alpha_test.go
```diff
@@ -183,22 +189,28 @@ func (s *AlphaAuthPluginTestSuite) TestRequestAccessToken() {
 func (s *AlphaAuthPluginTestSuite) TestRevokeAccessToken() {
 	db := connections["TestRevokeAccessToken"]
 
-	userID, acoID := "EFE6E69A-CD6B-4335-A2F2-4DBEDCCD3E73", "DBBD1CE1-AE24-435C-807D-ED45953077D3"
-	var user models.User
-	db.Find(&user, "UUID = ? AND aco_id = ?", userID, acoID)
-	// Good Revoke test
-	_, tokenString, _ := s.AuthBackend.CreateToken(user)
-
+	const UserID, AcoID = "EFE6E69A-CD6B-4335-A2F2-4DBEDCCD3E73", "DBBD1CE1-AE24-435C-807D-ED45953077D3"
```
embh: Capitalizing these const names exports them, right? Do we need that?

DeirdreHolub: I don't think variables inside a method are exported. These are constant values (can't be changed or assigned), and there is a convention for indicating that in naming. In C and Java, the convention is all caps, like USER_ID and ACO_ID. In go, the convention is initial cap only. Weird and confusing to me, but it is what it is.

PR #148, opened 2019-01-23
bcda-762 make sure all database connections are closed

alexhova:
### Fixes [BCDA-762](https://jira.cms.gov/browse/BCDA-762)
we don't always close db connections when we open them right now.  this closes them all and logs an error if we are unable to close the connection.

### Proposed changes:
add `utils.go` to the database package with a function that closes a db connection and logs an error if unable to do so.

### DIFF
--- a/bcda/main.go
+++ b/bcda/main.go
@@ -9,7 +9,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/dgrijalva/jwt-go"
+	jwt "github.com/dgrijalva/jwt-go"
 
 	"github.com/CMSgov/bcda-app/bcda/auth"
 	"github.com/CMSgov/bcda-app/bcda/auth/plugin"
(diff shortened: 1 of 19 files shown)


bcda/main.go
```diff
@@ -610,7 +613,9 @@ func cleanupArchive(hrThreshold int) error {
 }
 
 func createAlphaEntities(acoSize string) (aco models.ACO, err error) {
-	tx := database.GetGORMDbConnection().Begin()
+	db := database.GetGORMDbConnection()
+	defer database.Close(db)
```
DeirdreHolub: This function already has a defer on the transaction. We can't close the connection before that transaction handling executes.

knollfear: Defers execute in a last in first out order so as I don't think this is an issue.  The other defer will execute first and then the db connection will close.

alexhova: we should be good to go https://tour.golang.org/flowcontrol/13

PR #150, opened 2019-01-25
knollfear/bcda-709 Add expires header to completed jobs

knollfear:
### Fixes [BCDA-709](https://jira.cms.gov/browse/BCDA-709)

Completed job status results should include an expires header to indicate when the file will no longer be available.
### Proposed changes:

added expires header
added necessary unit tests

### Change Details

added header
added additional check to unit test

### Security Implications

no PII is impacted

### Acceptance Validation

Run tests
Run job manually and inspect headers.

### DIFF
--- a/bcda/api.go
+++ b/bcda/api.go
@@ -54,8 +54,6 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-const JobTimeout = time.Hour * 24
-
 /*
   	swagger:route GET /api/v1/ExplanationOfBenefit/$export bulkData bulkEOBRequest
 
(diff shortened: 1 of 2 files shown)


bcda/api.go
```diff
@@ -54,6 +54,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const JobTimeout = time.Hour * 24
```
embh: Could we use the `ARCHIVE_THRESHOLD_HR` for this if it's available (default to 24 hours if not) to match the `archive-job-files` CLI action?

dhgreene:
Looks good, and everything works locally.  Given that the job expiration time is not just mentioned here, is there a better place for a constant defining that as 1 day?

bcda/api.go
knollfear: I added a GetJobTimeout() method that is controlled by that environment variable.  I thought the value was passed into the job as a CLI.  This is a much better solution.  I also updated 710 with the same logic.

PR #151, opened 2019-01-28
knollfear/bcda-710 Adding 410 (gone) response code to old jobs

knollfear:
### Fixes [BCDA-710](https://jira.cms.gov/browse/BCDA-710)

Expired and Archived jobs don't currently return any reasonable response.  They should return an error and an http status code of 410 (Gone)
### Proposed changes:

Add additional cases tot eh switch statement on the job status endpoint
Added a catch in the completed case for jobs that were over 24 hours old.

### Security Implications

no PII should be affected

### Acceptance Validation

Unit tests should run
If you want to manually expire/archive a job in your local DB and make a request that would be jazzy too.

### Feedback Requested

Any, maybe testing strategy given that this is dependent on other jobs running and time passing

### DIFF
--- a/bcda/api.go
+++ b/bcda/api.go
@@ -54,6 +54,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const JobTimeout = time.Hour * 24
+
 /*
   	swagger:route GET /api/v1/ExplanationOfBenefit/$export bulkData bulkEOBRequest
 
@@ -232,6 +234,7 @@ func bulkRequest(t string, w http.ResponseWriter, r *http.Request) {
 		200:bulkResponseBody
 		400:ErrorModel
         404:ErrorModel
+        410:ErrorModel
 		500:FHIRResponse
 */
 func jobStatus(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


bcda/api.go
```diff
@@ -275,6 +276,12 @@ func jobStatus(w http.ResponseWriter, r *http.Request) {
 	case "Failed":
 		responseutils.WriteError(&fhirmodels.OperationOutcome{}, w, http.StatusInternalServerError)
 	case "Completed":
+		// If the job should be expired, but the cleanup job hasn;t run for some reason, still respond with 410
```
tbellj: hasn't

bcda/api.go
```diff
@@ -331,6 +338,11 @@ func jobStatus(w http.ResponseWriter, r *http.Request) {
 		}
 
 		w.WriteHeader(http.StatusOK)
+	case "Archived":
+		fallthrough
+	case "Expired":
+		w.Header().Set("Expires", job.CreatedAt.Add(GetJobTimeout()).String())
+		responseutils.WriteError(&fhirmodels.OperationOutcome{}, w, http.StatusGone)
```
embh: The `Expires` header and 410 HTTP status should convey the message that the job expired, but I'm wondering if we should still use `CreateOpOutcome()` here to put some details in the OperationOutcome. Maybe a separate ticket?

bcda/api.go
```diff
@@ -275,6 +276,13 @@ func jobStatus(w http.ResponseWriter, r *http.Request) {
 	case "Failed":
 		responseutils.WriteError(&fhirmodels.OperationOutcome{}, w, http.StatusInternalServerError)
 	case "Completed":
+		// If the job should be expired, but the cleanup job hasn't run for some reason, still respond with 410
```
tbellj: okay this makes sense but if the cleanup hasn't run then we have bigger issues. So im thinking, "is it correct logic to put this here?" But from our users perspective they don't and shouldn't care about our internal jobs so this makes sense in these cases. 1) the job is completed but we're notifying them that it's going to be expiring. 2) The cleanup job didn't run for some reason but they're still correctly notified that it's expired with the 410 status. Plus I think the plan is to have some kind of cloud watch monitoring to make sure those files were successfully archived so hopefully we won't run into that latter scenario :)

knollfear: I hear you and I agree, but by the same token, this code can only help us conform the software to our expectations of how it works so I think it should stay.  It is like the elevator with multiple levels of fail safes.

PR #154, opened 2019-01-30
bcda-764: retrofit router.go to use new auth provider

alexhova:
### Fixes [BCDA-764](https://jira.cms.gov/browse/BCDA-764)
middleware.go and therefore router.go uses the old backend implementation of auth.  it should use the new provider implementation.

### Proposed changes:
- move the alpha plugin into the auth package
- move GetAuthProvider() to auth package
- simplify auth middleware to use provider and router to use new middleware
- remove AllClaims type
- fix tests

### Security Implications
this introduces some changes to how requests are authorized and validated.  i don't believe this introduced any security concerns but it does touch a vital area security-wise

### Acceptance Validation
tests passed locally.  let's see what travis says

### DIFF
--- a/bcda/auth/provider.go
+++ b/bcda/auth/provider.go
@@ -8,17 +8,31 @@ import (
 
 // Provider is an interface for operations performed through an authentication provider.
 type Provider interface {
+	// Ask the auth Provider to register a software client for the ACO identified by localID.
 	RegisterClient(params []byte) ([]byte, error)
+
+	// Update data associated with the registered software client identified by clientID
 	UpdateClient(params []byte) ([]byte, error)
+
+	// Delete the registered software client identified by clientID, revoking an active tokens
 	DeleteClient(params []byte) error
 
+	// Generate new or replace existing Credentials for the given clientID
 	GenerateClientCredentials(params []byte) ([]byte, error)
+
+	// Revoke any existing Credentials for the given clientID
 	RevokeClientCredentials(params []byte) error
 
+	// Request an access token with a specific time-to-live for the given clientID
 	RequestAccessToken(params []byte) (Token, error)
+
+	// Revoke a specific access token identified in a base64 encoded token string
 	RevokeAccessToken(tokenString string) error
 
+	// Assert that a base64 encoded token string is valid for accessing the BCDA API
 	ValidateJWT(tokenString string) error
+
+	// Decode a base64 encoded token string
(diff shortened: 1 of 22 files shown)


bcda/auth/provider.go
```diff
@@ -17,3 +21,13 @@ type Provider interface {
 	ValidateJWT(tokenString string) error
 	DecodeJWT(tokenString string) (jwt.Token, error)
 }
+
+func GetAuthProvider() Provider {
```
embh: With this being in `auth` now, could we call it `GetProvider()`?

bcda/router.go
```diff
@@ -49,7 +48,7 @@ func NewDataRouter() http.Handler {
 	r := chi.NewRouter()
 	m := monitoring.GetMonitor()
 	r.Use(ConnectionClose)
-	r.With(auth.ParseToken, logging.NewStructuredLogger(), auth.RequireTokenAuth,
```
embh: We need the token in the context before logging. Without it, `sub`, `aco`, and `token_id` are missing from the request log entries (vs. master, for instance)

alexhova: nice catch!  I can't figure out a good way to test it though...

bcda/auth/provider.go
alexhova: yep!

bcda/router.go
DeirdreHolub: https://github.com/sirupsen/logrus#testing

alexhova: emily created a [ticket](https://jira.cms.gov/browse/BCDA-847)!

alexhova: and i can confirm that locally the token context is being logged:
```
api_1            | {"aco":"0c527d2e-2e8a-4808-b11d-0fa06baf8254","http_method":"GET","http_proto":"HTTP/1.1","http_scheme":"http","level":"info","msg":"request started","remote_addr":"172.21.0.1:38406","sub":"6baf8254-2e8a-4808-b11d-0fa00c527d2e","time":"2019-02-01T20:41:13Z","token_id":"611be673-834a-4069-b4fe-3643a3900508","ts":"Fri, 01 Feb 2019 20:41:13 UTC","uri":"http://localhost:3000/data/100/0c527d2e-2e8a-4808-b11d-0fa06baf8254.ndjson","user_agent":"PostmanRuntime/7.6.0"}
```

PR #164, opened 2019-02-06
knollfear/bcda-740B revisions with Jenn regarding the static site.

knollfear:
### Fixes [BCDA-740](https://jira.cms.gov/browse/BCDA-740)

Jenn had some feedback on the design that I am patching up.

### Proposed changes:

changed levels of some content
changed up menu
changed font
bolded CTA header
changed gradient size for big monitors
made links open externally if they are not on the cms domain
other things??

### Change Details

markdown changes
CSS changes
header html changes

### Security Implications

no PII is affected

### Acceptance Validation

Jenn reviewed the static site.  It should still look good locally

### Feedback Requested

Any

### DIFF
--- a/bcda-site-static/index.md
+++ b/bcda-site-static/index.md
@@ -8,6 +8,8 @@ gradient: "blueberry-lime-background"
 subnav-link-gradient: "blueberry-lime-link"
 sections:
   - Overview
+  - What We've Learned
+  - How Can I Get Access
   - About the Data
 ctas:
   -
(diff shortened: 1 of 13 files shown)


bcda/_site/404.html
```diff
@@ -8,20 +8,20 @@
 	<!-- Metadata for Twitter -->
 	<meta name="twitter:title" content="Your awesome title" />
 	<meta name="twitter:description" content="Write an awesome description for your new site here. You can edit this line in _config.yml. It will appear in your document head meta (for Google search resu..." />
-	<meta name="twitter:image" content="/assets/img/default-share-image.jpg" />
+	<meta name="twitter:image" content="http://localhost:4000/assets/img/default-share-image.jpg" />
```
embh: I think it's correct to switch these from relative to absolute URLs, but shouldn't they use the domain of the site?

knollfear: _site is supposed to be ignored.  It gets built during the deploy process.  I was running it locally (not even in our app)  and it put in the full path.  When I rebuilt it with the docker container it reverted to a relative path which I think is actually correct.

rnagle: FWIW, I don't think anyone will ever see this `404` page unless we configure the go `FileServer` to respond with it in the event of a `404`.

bcda-site-static/index.md
```diff
@@ -19,53 +21,56 @@ ctas:
...
+{:#what-we-ve-learned}
 
-    Developers, analysts, and administrators at ACOs have been instrumental in shaping CMS’ approach to this API. With their feedback, the team is implementing the following elements:
+   Developers, analysts, and administrators at ACOs have been instrumental in shaping CMS’ approach to this API. With their feedback, the team is implementing the following elements:
 
-    * Providing clear, human-readable narrative documentation to aid all users’ use of the API and the data that is shared
-    * Using resilient NDJSON ([New Line Delimited JSON](http://ndjson.org)) rather than fixed-width files in response to requests for delimited information
-    * Providing bulk beneficiary claims data through an automated retrieval process that requires minimal hands-on intervention to receive
-    * Formatting data in accordance with robust Fast Health Interoperability Resource ([FHIR](https://www.hl7.org/fhir/overview.html)) specifications
+   * Providing clear, human-readable narrative documentation to aid all users’ use of the API and the data that is shared
```
rnagle: Nit-picking: the bullet points in this section do not end with any punctuation.

rnagle:
One other thing I noticed when testing locally, some of the font files requested in CSS result in `404` errors:

image

Looking in `bcda/_site/assets/` directory, they are definitely missing. Is that something we can/should fix?

bcda/_site/404.html
embh: Good point, NBD as it's not being used at the moment. But I think in general these Twitter and Open Graph URLs are expected to be absolute. They are in the examples here: https://developer.twitter.com/en/docs/tweets/optimize-with-cards/guides/getting-started

I believe Twitter has a tool for checking how the meta tags will be read, so maybe we want to do that when the site is public.

bcda-site-static/index.md
knollfear: this is per the source doc from Elissa.  I am happy to change, but need approval from her.

knollfear:
> One other thing I noticed when testing locally, some of the font files requested in CSS result in `404` errors:
> 
> image
> 
> Looking in `bcda/_site/assets/` directory, they are definitely missing. Is that something we can/should fix?

I commented them out in the source CSS file.  It should stop happening.

PR #166, opened 2019-02-06
knollfear/bcda-868 Adding in Open Graph and Favicon images.

knollfear:
### Fixes [BCDA-868](https://jira.cms.gov/browse/BCDA-868)

To look the best our static site should have OG and twitter card images
### Proposed changes:

Made a share image
added CMS favicon png

### Change Details

Made a share image
added CMS favicon png
### Security Implications

no PII is affected
### Acceptance Validation

Open the site locally and verify you get a favicon
Look at the head of the HTML and ensure it is correctly formed.
When it is live, test it in the facebook open graph debugger.

### Feedback Requested

Any


knollfear:
> Is `layout.image` defined somewhere? We should set that value or replace the default for `twitter:image` and `og:image` so they use absolute URLs.

layout.ANYTHING can be set in the front matter of the markdown and will be available to the page when it gets built.  We can set that value on any markdown page by adding this to the top of the markdown file:

---
layout: home
title:  "Beneficiary Claims Data API"
date:   2017-10-30 09:21:12 -0500
image: some/directory/image.jpg
...
---

Setting the url value would require that jekyll be aware of the environment it will be deployed to which is kind of a heavy devops lift right now because we will need to get it passed forward all the way from jenkins.  I can spend the time to do it, but I can confirm that relative url's are fine for facebook.  If we had an S3 bucket I could use to host these files instead of serving them locally I think I could get that turned around relatively quickly.

embh:
Facebook does say to use absolute URLs: https://developers.facebook.com/docs/sharing/webmasters/faq/#faq_702972769845238 Maybe they handle relative image URLs also, and I believe they may also default to an image pulled from the page content, but I think it's safer to follow the guideline.

Time to add an environment value to Jekyll is certainly a consideration, so we may need someone else to weigh in on effort vs. priority. Maybe @rnagle or @msnook ?

rnagle:
@knollfear For now, we can use absolute URLs for the sandbox environment in the meta tags. This way we can avoid the matter of passing the env name to Jekyll at build time. It's only important that the images resolve correctly in sandbox anyway.

Later, we can implement a per-env build strategy so that the URLs are correct for the respective environments. Blue Button has an implementation we can borrow from (i.e., separate config files for each env, like [_config-prod.yml](https://github.com/CMSgov/bluebutton-site-static/blob/master/_config-prod.yml) and an env selector in the [Jenkinsfile](https://github.com/CMSgov/bluebutton-site-static/blob/master/ops/Jenkinsfile)).

knollfear:
> @knollfear For now, we can use absolute URLs for the sandbox environment in the meta tags. This way we can avoid the matter of passing the env name to Jekyll at build time. It's only important that the images resolve correctly in sandbox anyway.
> 
> Later, we can implement a per-env build strategy so that the URLs are correct for the respective environments. Blue Button has an implementation we can borrow from (i.e., separate config files for each env, like [_config-prod.yml](https://github.com/CMSgov/bluebutton-site-static/blob/master/_config-prod.yml) and an env selector in the [Jenkinsfile](https://github.com/CMSgov/bluebutton-site-static/blob/master/ops/Jenkinsfile)).

I reviewed their jenkins file and we can do it like they did, I am just nervous about committing to it in time and having it work correctly.  I have hardcoded in a url value to the config.yml file which when used to build the site locally hardcodes the absolute values correctly.

bcda-site-static/_config.yml
```diff
@@ -19,7 +19,7 @@ email: bc-api@googlegroups.com
 description: >- # this means to ignore newlines until "baseurl:"
   Shared Savings Program ACOs can use this FHIR-based API to retrieve bulk claims data related to their assigned beneficiaries. Under construction; feedback invited.
 baseurl: "" # the subpath of your site, e.g. /blog
-url: "" # the base hostname & protocol for your site, e.g. http://example.com
+url: "http://sandbox.bcda.cms.gov" # This is a kludge for HIMMS, we can fix it later as needed. the base hostname & protocol for your site, e.g. http://example.com
```
embh: 👍 Showing up as expected in the `meta` tags now. Should this be https though?

rnagle: Oh yeah, +1 for using `https` here.

PR #168, opened 2019-02-07
BCDA-794: fix sonarqube issues

alexhova:
### Fixes [BCDA-794](https://jira.cms.gov/browse/BCDA-794)

Sonarqube is showing some issues that are causing quality gate failures

### Proposed changes:

- fix the issues that are causing quality gate failures


bcda/servicemux/servicemux_test.go
```diff
@@ -18,7 +18,7 @@ type ServiceMuxTestSuite struct {
 }
 
 func (s *ServiceMuxTestSuite) TestNew() {
-	sm := New("127.0.0.1:0")
+	sm := New(os.Getenv("SM_TEST_ADDR"))
```
DeirdreHolub: I get that the quality check in Sonar is no hard-coded IPs, but later on in the same test there's an assertion that sm == "127.0.0.1". It seems the test[s] expect this to be the loopback address, in which case why not just create a const LoopbackAddress (or something like that) and use it? This way, we can minimize the change and not have to change docker configs here and in ops.

bcda/servicemux/servicemux_test.go
```diff
@@ -82,7 +82,7 @@ var testHandler http.HandlerFunc = func(w http.ResponseWriter, r *http.Request)
 	}
 }
 
-func (s *ServiceMuxTestSuite) TestServe_serveHTTPS() {
+func (s *ServiceMuxTestSuite) TestServeServeHTTPS() {
```
DeirdreHolub: Can we change the 'ServeServe' part of these names to just 'Serve'. When they were separated with _ , it was following a given/when/should pattern in test naming. Without the _, it's hard to read.

bcda/servicemux/servicemux_test.go
alexhova: yeah, i thought about that.  using a const still counts as a hard-coded IP address though.  I could use a config file for just this test, even though that seems a little silly.  i was trying to find somewhere that the loopback address was defined in http or a similar package but no dice so far.

DeirdreHolub: Often there is a way to express an exception to a rule by putting a comment in the code. Do we know what inspection tool Sonar is using, and can we find it's exception notation and use it here?

bcda/servicemux/servicemux_test.go
alexhova: done!

bcda/servicemux/servicemux_test.go
alexhova: There doesn't seem to be a way to ignore individual lines with comments in sonarqube.  only full files and directories

alexhova: i ended up going with a json config file to hold the ip address

DeirdreHolub:
sonarqube and golang: https://docs.sonarqube.org/display/PLUG/Importing+GoVet%2C+GoLint+and+GoMetaLinter+Issues+Reports
I'm guessing if you follow the directions in the link below, you'll be able to use a constant:
https://github.com/alecthomas/gometalinter#comment-directives

alexhova:
This led to the creation of two tickets: https://jira.cms.gov/browse/BCDA-902 and https://jira.cms.gov/browse/BCDA-903

PR #171, opened 2019-02-08
BCDA-812: implement registerClient() method for Okta auth plugin

DeirdreHolub:
### Implements [BCDA-812](https://jira.cms.gov/browse/BCDA-812)
This is the first method of the auth Provider interface to be given a working implementation in the Okta Plugin. All other tickets are dependent on it, as one must have credentials in order to acquire a token.

### Proposed changes:
1. Changed method signature to use specific typed parameters. Originally we thought the arguments might vary with plugin, so we used json for parameters. However, the parameters won't vary.
2. Use value semantics for the receiver in plugin implementations. The types are based on the empty struct so there is nothing to modify, hence no need for pointer receivers.
3. CreateAlphaToken() handles either plugin.
4. Simple support for changing the plugin at runtime added to Provider. This is done for testing support. See BCDA-882 for more details.
5. Because the okta client (BCDA-808) is not ready yet, and because testing will probably benefit from a mock, adds a simple mocktaclient.

### Dependencies
BCDA-809, for the real Okta client
BCDA-668, to add the BCDA_AUTH_PROVIDER env var 

### Security Implications
Register client returns credentials. The secret key is not saved to the database, cached in application memory or logged. This is similar to alpha handling of access tokens.

### Acceptance Validation
Existing tests continue to work.
Okta plugin and unit tests continue to work.
If the BCDA_AUTH_PROVIDER env variable is set to 'okta' (any casing), the okta plugin will be used for all unit tests that call GetProvider() (main, api, middleware, router), smoke test, and postman tests. Most will fail at this time, but without panics and with error messages.
Return from create-alpha-token CLI command should be usable by existing ansible playbook.

### Feedback Requested
Any

### DIFF
--- a/bcda/auth/alpha.go
+++ b/bcda/auth/alpha.go
@@ -7,7 +7,7 @@ import (
 	"regexp"
 	"time"
 
-	jwt "github.com/dgrijalva/jwt-go"
+	"github.com/dgrijalva/jwt-go"
 	"github.com/jinzhu/gorm"
 	"github.com/pborman/uuid"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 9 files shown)


bcda/auth/alpha_test.go
```diff
@@ -53,37 +50,25 @@ func (s *AlphaAuthPluginTestSuite) AfterTest(suiteName, testName string) {
...
 
-	c, err = s.p.RegisterClient([]byte(`{"clientID":""}`))
+	c, err = s.p.RegisterClient("")
 	assert.NotNil(s.T(), err)
-	assert.Nil(s.T(), c)
+	assert.Empty(s.T(), c.ClientID)
 	assert.Contains(s.T(), err.Error(), "provide a non-empty string")
 
-	c, err = s.p.RegisterClient([]byte(`{"clientID": "correct length, but not a valid UUID"}`))
+	c, err = s.p.RegisterClient("correct length, but not a valid UUID")
```
embh: Whoa. It's perfect. 😮

bcda/main.go
```diff
@@ -417,34 +417,41 @@ func createAlphaToken(ttl int, acoSize string) (s string, err error) {
 		return
 	}
 
-	authProvider := auth.GetProvider()
+	// authProvider := auth.GetProvider()
```
embh: Can we drop this line altogether or will we be using it later?

bcda/main.go
```diff
@@ -56,6 +57,7 @@ func init() {
 		log.Info("Failed to open error log file; using default stderr")
 	}
 	monitoring.GetMonitor()
+	log.Info(fmt.Sprintf(`Auth is made possible by %T`, reflect.TypeOf(auth.GetProvider())))
```
embh: In the logs when I run this locally, I get: `{"level":"info","msg":"Auth is made possible by *reflect.rtype","time":"2019-02-12T13:46:21Z"}`

bcda/auth/provider.go
```diff
@@ -2,14 +2,58 @@ package auth
...
+		switch (strings.ToLower(name)) {
+		case Okta:
+			providerName = name
+		case Alpha:
+			providerName = name
+		default:
+			log.Infof(`Unknown providerName %s; using %s`, name, providerName)
+		}
+	}
+	log.Infof(`Auth is made possible by %s`, providerName)
```
embh: This one prints the correct value, though I notice we're not using the structured logging here. Is there a ticket to add that?

embh:
We already talked about this, but for the record: moktaclient.go has only a couple of functions right now. Is the plan for it to evolve into a mock that implements the provider interface?

bcda/auth/provider.go
DeirdreHolub: BCDA-866

DeirdreHolub:
> We already talked about this, but for the record: moktaclient.go has only a couple of functions right now. Is the plan for it to evolve into a mock that implements the provider interface?

It will mock what the okta client does for us, so that we can use it as a test mock. It will match the methods of the okta client, not the provider.

bcda/auth/provider.go
DeirdreHolub: and I added the formatting statement to init()

bcda/auth/provider.go
```diff
@@ -2,14 +2,59 @@ package auth
...
+)
+
+var providerName = Alpha
+
+func init () {
+	log.SetFormatter(&log.JSONFormatter{})
+	SetProvider(strings.ToLower(os.Getenv(`BCDA_AUTH_PROVIDER`)))
+}
+
+func SetProvider(name string) {
```
alexhova: should this return the provider as well for convenience?

bcda/auth/alpha.go
```diff
@@ -21,44 +21,40 @@ type AlphaAuthPlugin struct{}
 // This implementation expects one value in params, an id the API knows this client by in string form
 // It returns a single string as well, being the clientID this implementation knows this client by
 // NB: Other implementations will probably expect more input, and will certainly return more data
-func (p *AlphaAuthPlugin) RegisterClient(params []byte) ([]byte, error) {
-	acoUUID, err := GetParamString(params, "clientID")
-	if err != nil {
-		return nil, err
-	}
+func (p AlphaAuthPlugin) RegisterClient(localID string) (Credentials, error) {
```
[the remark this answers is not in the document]
embh: On the topic of value vs. pointer receivers: https://github.com/golang/go/wiki/CodeReviewComments#receiver-type

bcda/auth/alpha.go
```diff
@@ -7,7 +7,7 @@ import (
 	"regexp"
 	"time"
 
-	jwt "github.com/dgrijalva/jwt-go"
```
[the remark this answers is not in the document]
DeirdreHolub: Go does not require that package file names and package names match. I don't know why your IDE sets up an alias when the package is actually named jwt. In contrast, my IDE constantly removes the unnecessary alias.

DeirdreHolub:
> I have a couple of general questions about the code here. These changes look good but it doesn't seem to fully implement a RegisterClient method with Okta. Are we blocked by not having the OktaClient yet?

Yes, we are blocked in that we aren't actually using Okta yet. That's why there is a mock client. Since I know the Okta API call, I wanted to get everything else ready in advance.

bcda/auth/alpha.go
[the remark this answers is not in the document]
DeirdreHolub: > what does the name `localID` mean? isn't it more of a `clientID`?

For the API, we have an id for the ACO. Okta will have an id for the ACO's system key. These are not the same id. RegisterClient is the process of creating an Okta client (with an Okta client id). localID is a piece of data that we send to Okta that will help correlate the two systems.

bcda/auth/provider.go
DeirdreHolub: I think having the set of a get/set pair return a value would be unexpected. I was trying to keep the mechanics similar to setting an env var to select the auth provider, since that's the standard use case. In other words, you set the type name, and then that's what you get from that point forward.

PR #174, opened 2019-02-13
BCDA-623 Feature: Standardize handling of app dirs

tbellj:
Replace 623 with the JIRA ticket number: -->
Fixes [BCDA-623](https://jira.cms.gov/browse/BCDA-623)

Problem:
Application directories were previously created lazily and now they're created on app startup.

Change Details:
Dockerfiles/Dockerfile.unit_test - added archive dir env variable
bcda/main.go - moved directory creation for the payload and archive directories to the initial startup.
bcda/main_test.go - modified test to handle new changes
bcdaworker/main.go - added the staging directory creation to the initial startup 

Security Implications:
None

Feedback Requested:
Any and all


bcda/main.go
```diff
@@ -59,6 +60,20 @@ func init() {
 	log.Info(fmt.Sprintf(`Auth is made possible by %T`, auth.GetProvider()))
 }
 
+func createAPIDirs() {
+	payload := os.Getenv("FHIR_PAYLOAD_DIR")
+	err := os.MkdirAll(payload, os.ModePerm)
```
embh: `os.ModePerm` = 777, right? Do we want these locked down a bit more, maybe 644?

tbellj: changed to 0750 and it works but willing to try something else if people have ideas.

tbellj: changed to 0744 after discussing with @knollfear. Good stuff!

PR #180, opened 2019-02-18
Dj/bcda 814

DeirdreHolub:
### Fixes [BCDA-814](https://jira.cms.gov/browse/BCDA-814)
We need to decode and verify Okta tokens. Unlike alpha, where we manage the server certificates, we have to obtain and use our Okta authorization server's public key[s].

### Proposed changes:
* gets and refreshes an in memory stash of Okta authorization server public keys
* establishes an OktaBackend interface to support testing using either a mock double or the actual oktabackend
* expands mokta, the mock double, to have methods for keys and tokens
* implements the missing oktabackend method for RegisterClient(). 
* makes sure code that calls okta only does so when environment variables and/or build tags are set correctly.

### Change Details
What ^^^ said.

### Security Implications
Does not change existing profile. No sensitive data is stored in code or other files.

### Acceptance Validation
There are four scenarios of testing supported. 

First is the default unit test mode, in which the unit test suite of the okta.go, the plugin itself, are run. These tests use the mokta backend in order to isolate testing our plugin behavior from okta. These tests are always run as part of the suite, and hence by `make test`.

Second is working on oktabackend mode. When you are working on adding a behavior to the oktabackend, a separate and standalone unit test for the oktabackend.go is in oktabackend_test.go. It is intended only to be run by hand and to be helpful in local development of code connecting to Okta itself. Environment variables listed below are required, and you must  set them yourself. To run the tests by hand:

`cd bcda/auth/client`
`go test -tags=okta -v`

Third, you can run the provider suite of unit tests where the provider used is Okta by setting the BCDA_AUTH_PROVIDER=okta. This is the only testing scenario where you need to set the provider. Until the entire provider interface is implemented, running the suite this way will generate errors for code that uses unimplemented methods.

Fourth, to do manual tests against the API running with the Okta plugin in docker, you have to modify the docker-compose.yml file to set all the environment variables.

export BCDA_AUTH_PROVIDER=okta .       (only needed for testing/running okta plugin as a provider)
export OKTA_CLIENT_TOKEN=<token>
export OKTA_CLIENT_ORGURL=https://cms-sandbox.oktapreview.com
export OKTA_OAUTH_SERVER_ID=<server id>
export OKTA_OAUTH_SERVER_METADATA=

### Feedback Requested
Any.


bcda/auth/oktabackend.go
```diff
@@ -0,0 +1,305 @@
...
+// Gets the current set of public server signing keys. This code treats all failures as fatal,
+// since failures are likely to result in an unrecoverable state for the application.
+func getPublicKeys() map[string]rsa.PublicKey {
+	var (
+		body []byte
+		err  error
+		keys map[string]rsa.PublicKey
+	)
+
+	body, err = get(os.Getenv("OKTA_OAUTH_SERVER_METADATA"))
```
embh: Since we have multiple URLs being created, can we generate this one based on the base URL and admin ID env vars?

bcda/auth/oktabackend.go
```diff
@@ -0,0 +1,305 @@
...
+	var client = &http.Client{Timeout: time.Second * 10,}
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", "", err
+	}
+
+	if resp.StatusCode != 201 {
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			log.Printf("failed to create client %s because %v", localID, err)
```
embh: I think it's only error messages that shouldn't be complete sentences (capitalized + with punctuation). Not necessary for log messages.

bcda/auth/oktabackend.go
```diff
@@ -0,0 +1,321 @@
...
+	return ob
+}
+
+func (ob *OB) PublicKeyFor(id string) (rsa.PublicKey, bool) {
+	key, ok := publicKeys[id]
+	log.Printf("invalid key id %s presented", id)
+	return key, ok
+}
+
+func (ob *OB) AddClientApplication(localID string) (string, string, error) {
```
alexhova: What do you think about moving this func to `oktaclient.go`?  That way, we could consolidate the external communication with okta in one place: the logging of okta requests/responses, the handling of errors, the building of requests and headers, etc.

bcda/auth/oktabackend.go
```diff
@@ -0,0 +1,321 @@
...
+}
+
+type Cli struct {
+	Include []string `json:"include"`
+}
+
+// Update the Auth Server's access policy to include our new client application. Otherwise, that application
+// will not be able to use the server. To do this, we first get the current list of clients, add the new
+// server to the inclusion list, and put it back to the server
+func addClientToPolicy(clientID string) error {
```
alexhova: think about moving this to `oktaclient.go` as well

bcda/auth/oktabackend.go
```diff
@@ -0,0 +1,321 @@
...
+	Use       string `json:"use"`
+	E         string `json:"e"`
+	N         string `json:"n"`
+}
+
+type KeyList struct {
+	Keys []*RsaJWK `json:"keys"`
+}
+
+func get(urlString string) ([]byte, error) {
```
alexhova: also think about moving this function and the places it is called to the `oktaclient.go`.

bcda/auth/oktabackend.go
alexhova: I'm thinking single responsibility principle here

bcda/auth/client/oktaclient.go
```diff
@@ -31,94 +42,213 @@ func init() {
...
+// separate from init for testing
+func config() error {
+	// required env vars
+	oktaBaseUrl = os.Getenv("OKTA_CLIENT_ORGURL")
+	oktaServerID = os.Getenv("OKTA_OAUTH_SERVER_ID")
 	oktaToken := os.Getenv("OKTA_CLIENT_TOKEN")
-	if oktaToken == "" {
-		logger.Fatal("No Okta token found; please set OKTA_CLIENT_TOKEN")
+
+	// report missing env vars
```
embh: Could we tidy up comments a bit? What's happening here, for example, is evident from the code below.

bcda/auth/provider.go
```diff
@@ -28,18 +30,18 @@ func SetProvider(name string) {
 		case Alpha:
 			providerName = name
 		default:
-			log.Infof(`Unknown providerName %s; using %s`, name, providerName)
+			log.Infof(`unknown providerName %s; using %s`, name, providerName)
```
embh: It looks like this and some other log messages were switched to not be capitalized, but I think that is a pattern we want to maintain (not to be confused with error text where this format is the standard practice)

Gopkg.toml
```diff
@@ -37,10 +37,6 @@
   branch = "master"
   name = "github.com/eug48/fhir"
 
-[[constraint]]
-  branch = "master"
-  name = "github.com/okta/okta-sdk-golang"
```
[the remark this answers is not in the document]
DeirdreHolub: I considered that. However, that track lead down a rabbit hole. Is this a general okta client, or is it our purpose-built okta client? should each okta call be in its in own method, and then there are additional methods that orchestrate among them? should orchestration happen in the plugin? How do the requestIDs track? Where are requestIDs initiated? I think we would better served to implement more of the smaller, single call methods, and then look back at this one.

bcda/auth/client/oktaclient.go
```diff
@@ -31,94 +42,213 @@ func init() {
...
-	filter := query.NewQueryParams(query.WithQ(email))
-	users, resp, err := client.User.ListUsers(filter)
+type Cli struct {
+	Include []string `json:"include"`
+}
+
+// Update the Auth Server's access policy to include our new client application. Otherwise, that application
+// will not be able to use the server. To do this, we first get the current list of clients, add the new
+// server to the inclusion list, and put it back to the server
+func addClientToPolicy(clientID string, requestID uuid.UUID) error {
```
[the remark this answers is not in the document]
DeirdreHolub: I considered that. However, that track lead down a rabbit hole. Is this a general okta client, or is it our purpose-built okta client? should each okta call be in its in own method, and then there are additional methods that orchestrate among them? should orchestration happen in the plugin? How do the requestIDs track? Where are requestIDs initiated? I think we would better served to implement more of the smaller, single call methods, and then look back at this one.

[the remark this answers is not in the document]
alexhova: i was picturing that oktaclient.go would have methods that only make one call each (with logging and everything).  then, an okta.go method may call more than one oktaclient.go method

bcda/auth/client/oktaclient.go
```diff
@@ -31,94 +42,213 @@ func init() {
...
 
-	// The email in OKTA_EMAIL should represent a test user present in the Okta sandbox environment
-	oktaTestUser, success := os.LookupEnv("OKTA_EMAIL")
-	if !success {
-		logRequest(reqId).Warn("Unable to perform Okta HealthCheck.  Please set OKTA_EMAIL to match a test user account.")
-		return errors.New("cannot perform Okta health check without OKTA_EMAIL configured ")
+// Returns an OktaClient. An OktaClient is always created, whether or not it is currently able to converse with Okta.
+func NewOktaClient() *OktaClient {
+	var err error
+	once.Do(func() {
```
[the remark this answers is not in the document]
DeirdreHolub: We keep a list of our okta authorization server's public keys so that we can decode / validate tokens. We only need to get the keys the first time an OktaClient value is created.

bcda/auth/client/oktaclient.go
```diff
@@ -31,94 +42,213 @@ func init() {
...
 	}
 
-	if len(users) >= 1 {
-		logResponse(resp.StatusCode, reqId).Info("Okta ping request successful")
-		return nil
+	req.Header.Add("Content-Type", "application/json")
+	req.Header.Add("Accept", "application/json")
+	req.Header.Add("Authorization", oktaAuthString)
+
+	logRequest(requestID)
```
dhgreene: Wouldn't we want to include some information that shows what method we're in?  And if not PII, something about what we're trying to create?

bcda/auth/client/oktaclient.go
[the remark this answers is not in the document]
DeirdreHolub: That would break the interface that supports Mokta.

bcda/auth/client/oktaclient.go
dhgreene: And isn't it a good practice to log that we're trying a request before our first possible return value?

docker-compose.test.yml
```diff
@@ -12,8 +12,12 @@ services:
       - ATO_PRIVATE_KEY_FILE=../../shared_files/ATO_private.pem
       - OKTA_CLIENT_ORGURL=https://cms-sandbox.oktapreview.com
       - OKTA_EMAIL=shawn@bcda.aco-group.us
-# Add the Okta token below and uncomment
+# Add the Okta values below and uncomment
+#     - BCDA_AUTH_PROVIDER=okta
 #     - OKTA_CLIENT_TOKEN=<token>
+#     - OKTA_OAUTH_SERVER_METADATA=
```
embh: It looks like we can remove `OKTA_OAUTH_SERVER_METADATA` here and in docker-compose.yml, right? Since the URL is being created using the base URL + server ID.

bcda/auth/client/oktaclient.go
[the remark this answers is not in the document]
alexhova: i really need to wrap my head around that.  essentially, there should be a mokta method with stubbed out responses for each oktaclient method.  is that right?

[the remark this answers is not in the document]
DeirdreHolub: Begin from the OktaBackend interface in okta.go. This is the interface that must be honored by values passed in to NewOktaAuthPlugin(backend OktaBackend). Thus, Mokta and OktaClient implement that interface. As more plugin methods get implemented, this interface will acquire additional methods that must be implemented in Mokta and OktaClient.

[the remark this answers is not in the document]
alexhova: thanks for the explanation.  i'll update my pr after this is merged

bcda/auth/client/oktaclient.go
DeirdreHolub: The request is actually sent on the next line of code.

bcda/auth/client/oktaclient.go
```diff
@@ -31,94 +42,210 @@ func init() {
...
+		logger.Errorf("No public keys available for server because %s", err)
+		// our practice is to not stop the app, even when it's in a state where it can do nothing but emit errors
+		// methods called on this ob value will result in errors until the publicKeys map is successfully updated
 	}
+	return &OktaClient{}
+}
+
+func (oc *OktaClient) PublicKeyFor(id string) (rsa.PublicKey, bool) {
+	key, ok := publicKeys[id]
+	logger.Warnf("Invalid key id %s presented", id)
```
dhgreene: Won't this always warn whether the key is valid or not?

bcda/auth/client/oktaclient.go
```diff
@@ -31,94 +42,210 @@ func init() {
...
 	}
+
+	req.Header.Add("Content-Type", "application/json")
+	req.Header.Add("Accept", "application/json")
+	req.Header.Add("Authorization", oktaAuthString)
+
+	var client = &http.Client{Timeout: time.Second * 10,}
+	resp, err := client.Do(req)
+	if err != nil {
+		logError(err, requestID)
```
dhgreene: Mea culpa: I named the log helper methods in a way that promotes misunderstanding.  This line (and all the other logging lines in this function) will give you a *logrus.Entry object.  Unless you assign it (`l := logError(err, requestID)`) or chain it into a method (`logError(err, requestID).Error("oh no!")`) it's not going to do anything.

bcda/auth/client/oktaclient.go
dhgreene: OK, I can get on board with a practice of just logging the Okta call, not the beginning of the method.  So I'm still concerned:

- As mentioned below (for line 186 as of this revision), this won't do anything unless it's `logRequest(requestID).Info("adding Okta client")`
- While we're at it, since I don't think the client ID is PII, why don't we log it too?  `logRequest(requestID).Info("adding Okta client: %v", localID)`

bcda/auth/client/oktajwk.go
```diff
@@ -0,0 +1,155 @@
...
+	u, err := url.Parse(urlString)
+	if err != nil {
+		logger.Errorf("invalid url %s because %v", urlString, err)
+		return nil, err
+	}
+
+	var client = &http.Client{Timeout: time.Second * 10,}
+	res, err := client.Get(u.String())
+	if err != nil {
+		logger.Errorf("net failure to %s because %v", u.String(), err)
```
dhgreene: Having chosen to use structured logging, we should be putting our common arguments in separate fields.  And we should standardize the field names.  Out of the scope of this PR would be a logging library to help with that.  But in the meantime, try:

`logger.WithFields(logrus.Fields{"error": err, "url": u.String()}).Error("net failure in get")`

NB: don't use more than 2 `.WithFields()` calls on a single logrus call.

bcda/auth/client/oktaclient.go
DeirdreHolub: How about if we rename these methods "loggerWithError()" and so forth? That seems like a clearer expression of purpose.

dhgreene: That name works for me.

alexhova: `errorLogger()`? 🤷‍♂️

PR #181, opened 2019-02-20
knollfear/bcda-820 upgrading to GO 1.11.5

knollfear:
### Fixes [BCDA-820](https://jira.cms.gov/browse/BCDA-820)

Upgrade to go 1.11.5 to keep consistent and up to date on the go version we are using
### Proposed changes:

adjusted docker files and made 2 minor code changes to be go 1.11 compliant.

### Security Implications

PII is not affected by this change
### Acceptance Validation

tests should run and deployment should work

### Feedback Requested

any


knollfear:
> Looks good with all the tests passing and everything seems to work fine. Changelogs don't show anything that would kill us. I just wonder if we should change this as well: https://github.com/CMSgov/bcda-app/blob/master/.travis.yml#L14
> Shouldn't really matter since we still bring up the containers with 1.11 to run the tests but maybe just in case or just for completeness?

I added it.  I updated it the last time, but I couldn't remember the file name to update it.

PR #183, opened 2019-02-20
BCDA-817 : implement GenerateClientCredentials() for okta plugin

alexhova:
### Fixes [BCDA-817](https://jira.cms.gov/browse/BCDA-817)
We need to implement the `GenerateClientCredentials()` func to satisfy our auth provider interface with the okta plugin

### Proposed changes:
- Create func in `oktaclient.go` to generate a new client secret for a client in okta
- Change `GenerateClientCredentials()` to accept a `clientID` and return the `Credentials` type
- Implement `GenerateClientCredentials()` in `okta.go` so that it is functional

### Acceptance Validation
- `go test -tags=okta` passes from the client directory with the proper okta-related env vars set for oktaclient_test.go
- `make test` passes using mokta as the oktabackend


bcda/auth/alpha.go
```diff
@@ -54,32 +54,28 @@ func (p AlphaAuthPlugin) DeleteClient(params []byte) error {
...
+		return Credentials{}, fmt.Errorf("ACO %s does not have a registered client", clientID)
 	}
 
 	err = p.RevokeClientCredentials([]byte(fmt.Sprintf(`{"clientID":"%s"}`, clientID)))
 	if err != nil {
-		return nil, fmt.Errorf("unable to revoke existing credentials for ACO %s because %s", clientID, err)
+		return Credentials{}, fmt.Errorf("unable to revoke existing credentials for ACO %s because %s", clientID, err)
 	}
 
+	params := []byte(fmt.Sprintf(`{"clientID":"%s", "ttl":%d}`, clientID, 1))
```
alexhova: I put a magic number here...

DeirdreHolub: can't do this; it will create an access token lasting 1 hour in alpha.

bcda/auth/okta_test.go
```diff
@@ -44,9 +43,14 @@ func (s *OktaAuthPluginTestSuite) TestDeleteClient() {
 }
 
 func (s *OktaAuthPluginTestSuite) TestGenerateClientCredentials() {
-	r, err := s.o.GenerateClientCredentials([]byte("{}"))
-	assert.Nil(s.T(), r)
-	assert.Equal(s.T(), "not yet implemented", err.Error())
+	validClientID := "0oaj4590j9B5uh8rC0h7"
```
dhgreene: Do we want this hard-coded twice (also in `oktaclient_test.go`)?

alexhova: only other idea i had was an env var.  do we want to go that route?

dhgreene: It _feels_ more ephemeral than something we want to put in code, and it will be a different value for production.  If env vars are all we've got at the moment, maybe that's better than a surprise later on.

alexhova: well, it'll pass mokta with any string of length 20.  it's for testing against the real okta in oktaclient.go that we'll need a real clientID for

dhgreene: OK, for the sake of oktaclient.go I'm guessing the test suite won't ever run against the production environment, so that issue is less concerning.  We'll try not to delete the referenced client on pain of a new PR.  😬

alexhova: once we settle on an established "testing client" in okta, i think it will make sense to keep the id as an env var

DeirdreHolub: The future solution is to create the data we need in Okta for a test and then clean it up. For example, this test could use RegisterClient() to create a client, then GenerateClientCredentials() to generate it new credentials, and then it can prove that the new credentials are, in fact, different from the original. Before we do that, though, we need a delete client function. I opened BCDA-951 for improving testing. Add comments there.

bcda/auth/client/oktaclient.go
```diff
@@ -258,6 +254,49 @@ func refreshKeys() {
 	}
 }
 
+func (oc *OktaClient) GenerateNewClientSecret(clientID string) (string, error) {
+	url := os.Getenv("OKTA_CLIENT_ORGURL") + "/oauth2/v1/clients/" + clientID + "/lifecycle/newSecret"
```
DeirdreHolub: should use the package global `oktaBaseUrl` because it's set once in config.

bcda/auth/client/oktaclient.go
```diff
@@ -258,6 +254,49 @@ func refreshKeys() {
...
+	return cs, nil
+}
+
+func client() *http.Client {
+	return &http.Client{Timeout: time.Second * 10}
+}
+func addRequestHeaders(req *http.Request) {
+	req.Header.Set("Accept", "application/json")
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Authorization", "SSWS "+os.Getenv("OKTA_CLIENT_TOKEN"))
```
DeirdreHolub: This one should use the package global `oktaAuthString` because it's set once in config.

bcda/auth/client/oktaclient.go
```diff
@@ -229,9 +227,7 @@ func addClientToPolicy(clientID string, requestID uuid.UUID) error {
 	req.Header.Add("Authorization", oktaAuthString)
```
DeirdreHolub: This line and the two above can be replaced by your new addRequestHeaders function. There's also two other places where the same replacement can be made. Let's not leave our future selves checking to see why it was done one way in some methods and not in others!

PR #184, opened 2019-02-22
BCDA-886: Optimize Docker and Makefile configurations

msnook:
### Fixes [BCDA-886](https://jira.cms.gov/browse/BCDA-886)

The team has begun to identify slowness in the build/test process on local machines, primarily due to excessively large Docker images, the built-in (within Makefile) necessity for images to be re-built, and not leveraging caches properly.  

### Proposed changes:

This PR contains the app-related changes.
The ops-related changes can be found in [this PR](https://github.com/CMSgov/bcda-ops/pull/135).

### Change Details

Changes to reduce image size:

- Using Alpine where possible.  One disadvantage of this is that those images based on Alpine do not have `bash`; instead `sh` can be used when `exec`ing into a container.
- Only copying necessary elements from repo to image instead of entire repo
- Leveraging a `.dockerignore` file to exclude items from being copied into new images
- Removed `unit-test`, `smoke-test`, and `performance-test` from the compose stack; replaced with one single shared `tests` service.

Changes to speed up execution time:

- Removed environment variables from Dockerfiles and into the compose stack (no need to rebuild the image after an environment variable change).
- Removed all `sleep` from the test targets in the Makefile.  It is the responsibility of the user to ensure the stack is up and running prior to pulling the trigger on test execution.
- Separated `lint` and `unit-test` to allow the user to have more control over running tests in isolation.
- No need to build a new image each time tests are run.
- Forced the removal of the test container following each test execution.

Other details:

- Updated the `README`.  Note that from a user perspective, a little bit has changed.  Namely, the prerequisite for execution of tests is environment configuration.  To keep the execution time of tests low, users are now responsible for ensuring that the images have been built, compose stack is up and running, and database has been seeded properly.  
- Additionally, image creation via `make docker-bootstrap` should only be required if the Dockerfiles change (which is a rare occurrence, I hope). 

### Security Implications
None.

### Acceptance Validation

- Reduced the number of services by two (added `tests` service; removed `unit-test`, `smoke-test`, and `performance-test`)
- Reduced the size of the images by >50% (confirmed by executing `docker image ls | grep bcda`)
- Reduced the time it takes to execute tests (confirmed by executing `/usr/bin/time make test`)
- [Build/Package](https://bcda-ci.adhocteam.us/job/BCDA%20-%20Build%20and%20Package/422/parameters/) passed with this feature branch
- [Deployment](https://bcda-ci.adhocteam.us/job/BCDA%20-%20Deploy/347/parameters/) passed with this feature branch
- [Smoke test](https://bcda-ci.adhocteam.us/job/BCDA%20-%20Test%20-%20Integration%20Smoke/311/parameters/) passed with this feature branch
- [Performance test ](https://bcda-ci.adhocteam.us/job/BCDA%20-%20Test%20-%20Performance/9/parameters/)passed with this feature branch

### Feedback Requested
Please review.


README.md
```diff
@@ -27,13 +27,43 @@ docker-compose up
...
 
-Run tests and produce test metrics:
+Run tests and produce test metrics.  
+The items identified above in the `Build/Start` section are prerequisites to running tests.  
+In order to keep the test feedback loop optimized, the following items must be handled by the caller (and are not handled by the test targets):
+- Ensuring the compose stack is up and running
+- Ensuring the database has been seeded
+- Managing images/containers (if Dockerfile changes have occurred, an image rebuild is required and won't occur as part of the test targets)
+
+1. Run golang linter and gosec:
```
knollfear: Technically this one test does not require the docker-compose up to be run.

PR #187, opened 2019-02-26
Knollfear/bcda 922 Prep for splitting job into chunks

knollfear:
### Fixes [BCDA-922](https://jira.cms.gov/browse/BCDA-922)

This ticket is one of 3 steps to get us to the point where we can parallelize job processing in smaller chunks. 
File names from jobs are all statically set as <acoID>.ndjson they should be dynamically/randomly generated
Job status should be pulling the list of files associated with the job from the jobKey set associated with the job
Unencrypted files should get job keys with a static encryption key

### Proposed changes:

Removed tokenACOmatch middleware
added tokenJobMatch middleware
removed token check from job status view
added encrypted key to output file item list

### Security Implications

TokenJobMatch middleware is a security function.  It should be reviewed carefully to ensure that it behaves correctly.

### Acceptance Validation

Tests should pass
This change should be mostly a no op change except for changing the file name of the job result.

### Feedback Requested

ANy.  Mostly about TokenJobMatch middleware


embh:
Could you update the metadata endpoint to reflect this change? It has `/data/[jobID]/[acoID].ndjson` for the data file path. It will need to change in the sample /metadata response in API.md also.

knollfear:
> Could you update the metadata endpoint to reflect this change? It has `/data/[jobID]/[acoID].ndjson` for the data file path. It will need to change in the sample /metadata response in API.md also.

Fixed in latest commit.

bcda/api.go
```diff
@@ -242,26 +241,8 @@ func jobStatus(w http.ResponseWriter, r *http.Request) {
 	db := database.GetGORMDbConnection()
 	defer database.Close(db)
 
-	i, err := strconv.Atoi(jobID)
-	if err != nil {
-		log.Print(err)
-		oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, "", responseutils.Processing)
-		responseutils.WriteError(oo, w, http.StatusBadRequest)
-		return
```
embh: I think we should keep this check on the job ID. We know `/jobs/abc` is a bad request, so we shouldn't query the database with that value.

bcda/router.go
```diff
@@ -41,8 +41,8 @@ func NewDataRouter() http.Handler {
 	r := chi.NewRouter()
 	m := monitoring.GetMonitor()
 	r.Use(auth.ParseToken, logging.NewStructuredLogger(), HSTSHeader, ConnectionClose)
-	r.With(auth.RequireTokenAuth, auth.RequireTokenACOMatch).
-		Get(m.WrapHandler("/data/{jobID}/{acoID}.ndjson", serveData))
+	r.With(auth.RequireTokenAuth, auth.RequireTokenJobMatch).
+		Get(m.WrapHandler("/data/{jobId}/{fileName}", serveData))
```
embh: If the aim is to be consistent with the way we write "jobId"/"jobID", the latter is preferred in Go. Can we change it for the `/jobs/` path on line 29 and keep it `jobID` here?

bcda/api.go
```diff
@@ -373,9 +356,9 @@ func jobStatus(w http.ResponseWriter, r *http.Request) {
 */
 func serveData(w http.ResponseWriter, r *http.Request) {
 	dataDir := os.Getenv("FHIR_PAYLOAD_DIR")
-	acoID := chi.URLParam(r, "acoID")
-	jobID := chi.URLParam(r, "jobID")
-	http.ServeFile(w, r, fmt.Sprintf("%s/%s/%s.ndjson", dataDir, jobID, acoID))
+	fileName := chi.URLParam(r, "fileName")
+	jobID := chi.URLParam(r, "jobId")
```
embh: Same as comment in router.go-- let's keep `jobID` to be consistent.

bcda/api.go
embh: Nvm, I see this moved to middleware.

bcda/auth/middleware.go
```diff
@@ -63,25 +65,41 @@ func RequireTokenAuth(next http.Handler) http.Handler {
...
 		token := r.Context().Value("token").(*jwt.Token)
+
+		db := database.GetGORMDbConnection()
+		defer database.Close(db)
+
+		i, err := strconv.Atoi(jobID)
+		if err != nil {
+			log.Error(err)
+			oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, "", responseutils.Processing)
+			responseutils.WriteError(oo, w, http.StatusInternalServerError)
```
embh: 400 or 404 would be more appropriate when the job ID in the URL is invalid.

bcda/auth/middleware.go
```diff
@@ -63,25 +65,41 @@ func RequireTokenAuth(next http.Handler) http.Handler {
...
 		}
 
-		re := regexp.MustCompile("/([a-fA-F0-9]{8}(?:-[a-fA-F0-9]{4}){3}-[a-fA-F0-9]{12})(?:-error)?.ndjson")
-		urlUUID := re.FindStringSubmatch(r.URL.String())[1]
-		if uuid.Equal(uuid.Parse(claims["aco"].(string)), uuid.Parse(string(urlUUID))) {
-			next.ServeHTTP(w, r)
-		} else {
-			log.Error(fmt.Errorf("token with ID: %v does not match ACO in request url", claims["id"]))
-			respond(w, http.StatusUnauthorized)
+		acoId := claims["aco"].(string)
```
embh:
```suggestion
		acoID := claims["aco"].(string)
```

bcda/auth/middleware.go
```diff
@@ -63,25 +65,41 @@ func RequireTokenAuth(next http.Handler) http.Handler {
...
+
+		var job models.Job
+		err = db.Find(&job, "id = ? and aco_id = ?", i, acoId).Error
```
embh:
```suggestion
		err = db.Find(&job, "id = ? and aco_id = ?", i, acoID).Error
```

bcda/router.go
```diff
@@ -26,7 +26,7 @@ func NewAPIRouter() http.Handler {
 		if os.Getenv("ENABLE_COVERAGE_EXPORT") == "true" {
 			r.With(auth.RequireTokenAuth, ValidateBulkRequestHeaders).Get(m.WrapHandler("/Coverage/$export", bulkCoverageRequest))
 		}
-		r.With(auth.RequireTokenAuth).Get(m.WrapHandler("/jobs/{jobId}", jobStatus))
+		r.With(auth.RequireTokenAuth, auth.RequireTokenJobMatch).Get(m.WrapHandler("/jobs/{jobId}", jobStatus))
```
embh: Awesome. Great that we're enforcing this on the job status now.

bcda/router.go
knollfear: Not easily.  This syntax puts it in as a URLParam so I would need somethign wonky like:
jobID := chi.URLParam(r, "jobId")
if jobID == nil{
jobID := chi.URLParam(r, "jobID")
}
which seems worse.  If we are going to do anything I would like to refactor it all to jobID.  I started that process but it felt like a rabbit hole and abandoned it.  If you support it I am happy to take the time to do it.

bcda/api.go
knollfear: Same comment as above.

bcda/auth/middleware.go
knollfear: I stuck with 500 because there was a test that was expecting it.  I am happy to change the test though.

bcda/router.go
rnagle: It'd be great to make this consistent/use go's convention, but wow are there are a lot of places in the code and tests that this touches. Might be better to backlog a ticket for this, as long as there's no imminent harm in leaving as-is.

bcda/auth/middleware.go
rnagle: +1 for a 404 here.

bcda/router.go
knollfear: This has been refactored

bcda/auth/middleware.go
knollfear: all returns for this function are now 404

PR #188, opened 2019-02-26
BCDA-950: Read CCLF8 and CCLF9

embh:
### Fixes [BCDA-950](https://jira.cms.gov/browse/BCDA-950)
For attribution, we need an ETL process for CCLF8 and CCLF9 files. This ticket is the first step (extract) of that process.

### Proposed changes:
* Add two new CLI commands, `import-cclf8` and `import-cclf9` to extract data from CCLF8/9 filenames and content

### Change Details
This parses the filenames and content for ACO IDs, HICNs, MBIs, and other values and writes them to stdout. Nothing is saved to the database yet. Command usage is:
`bcda import-cclf8 --file path/to/file`
It's possible that we may need more or less of the information later. This is just getting us started on the process.

### Security Implications
The data provided for testing is synthetic and contains no PII.

### Acceptance Validation
screen shot 2019-02-26 at 5 31 36 pm
screen shot 2019-02-26 at 5 32 19 pm

### Feedback Requested
Any


bcda/cli.go
```diff
@@ -0,0 +1,135 @@
+package main
```
knollfear: I think this should be cclf.go maybe?

embh: I was thinking we could eventually move the other CLI functions-- basically everything after `setUpApp()`-- from main.go to cli.go. Or do you think it makes more sense to have them split into separate files based on what they do?

knollfear: That is fine with me; I just wasn't sure.

PR #189, opened 2019-02-27
BCDA-813 Feature: Add Okta RequestAccessToken()

dhgreene:
### Fixes [BCDA-813](https://jira.cms.gov/browse/BCDA-813)
Implements RequestAccessToken() for the Okta client

Having taken over this ticket after @embh already implemented the core feature, all credit for good code goes to @embh and I'll take ownership of any mistakes.

### Testing:
- [ ] From project base dir: `make unit-test`
- [ ] From `bcda/auth`: `DATABASE_URL=postgresql://postgres:toor@localhost:5432/bcda?sslmode=disable go test -run 'Okta*' -v`
- [ ] From `bcda/auth/client` and with env vars (detailed below) set: `go test -v -tags=okta`

## Environmental variables
`OKTA_CLIENT_ORGURL`
`OKTA_OAUTH_SERVER_ID`
`OKTA_CLIENT_TOKEN`
`OKTA_CLIENT_ID`
`OKTA_CLIENT_SECRET`

### Proposed changes:

- Change method signature from `(params []byte)` to `(creds Credentials, ttl int)`

### Security Implications
No additional security implications

### Feedback Requested
- Is the returned token sufficient for our needs?
- All other feedback


bcda/auth/client/oktaclient.go
```diff
@@ -184,7 +247,9 @@ func addClientToPolicy(clientID string, requestID uuid.UUID) error {
 
 	// not calling logRequest() because this is a step of AddClientApplication
 
-	resp, err := client().Do(req)
+	var client = &http.Client{Timeout: time.Second * 10}
+	resp, err := client.Do(req)
```
alexhova: let's use this func i made just to reuse code and do `client().Do(req)`
https://github.com/CMSgov/bcda-app/blob/7ee08580546daba2c233a0c95048764dc2ec2cf4/bcda/auth/client/oktaclient.go#L285-L287

bcda/auth/client/oktaclient.go
```diff
@@ -149,6 +164,54 @@ func (oc *OktaClient) AddClientApplication(localID string) (string, string, erro
...
+func (oc OktaClient) RequestAccessToken(creds Credentials) (OktaToken, error) {
+	requestID := uuid.NewRandom()
+
+	tokenURL := fmt.Sprintf("%s/oauth2/%s/v1/token", oktaBaseUrl, oktaServerID)
+	req, err := http.NewRequest("POST", tokenURL, nil)
+	if err != nil {
+		return OktaToken{}, err
+	}
+
+	req.Header.Add("Accept", "application/json")
```
alexhova: lets reuse the addRequestHeaders() func i made here
https://github.com/CMSgov/bcda-app/blob/7ee08580546daba2c233a0c95048764dc2ec2cf4/bcda/auth/client/oktaclient.go#L288-L292

bcda/auth/client/oktaclient.go
```diff
@@ -149,6 +164,54 @@ func (oc *OktaClient) AddClientApplication(localID string) (string, string, erro
 	return clientID, clientSecret, nil
 }
 
+func (oc OktaClient) RequestAccessToken(creds Credentials) (OktaToken, error) {
```
alexhova: other funcs use the pointer to OktaClient as the receiver.  should we do that here as well?

bcda/auth/client/oktaclient.go
alexhova: as well as everywhere else we can

bcda/auth/client/oktaclient_test.go
```diff
@@ -74,6 +74,22 @@ func (s *OTestSuite) TestAddClientApplication() {
...
+	assert.IsType(s.T(), OktaToken{}, t)
+	assert.Nil(s.T(), err)
+
+	t, err = s.oc.RequestAccessToken(Credentials{ClientID: "", ClientSecret: ""})
+	assert.IsType(s.T(), OktaToken{}, t)
+	assert.NotNil(s.T(), err)
+}
+
 func (s *OTestSuite) TestGenerateNewClientSecret() {
 	validClientID := "0oaj4590j9B5uh8rC0h7"
```
alexhova: we can use your os.Getenv("OKTA_CLIENT_ID") here if you want to make this change while you're at it

bcda/auth/client/oktaclient.go
```diff
@@ -25,6 +26,18 @@ var oktaServerID string
...
 var once sync.Once
 
+type OktaToken struct {
+	AccessToken string `json:"access_token"`
+	TokenType   string `json:"token_type"`
+	ExpiresIn   int    `json:"expires_in"`
+	Scope       string `json:"scope"`
+}
+
+type Credentials struct {
```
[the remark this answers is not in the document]
dhgreene: I ran into an import loop, so ended up duplicating them.  I'd be happy to hear there's a good place to put mutual objects.  The client is particularly blind to objects defined elsewhere.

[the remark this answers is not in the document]
alexhova: ahh ok that makes sense. well, it's harmless. we should come back to the organization of code at some point though

bcda/auth/client/oktaclient.go
dhgreene: The Content-Type is actually different in this code.  I'm not sure that happens often enough to add complexity to your addRequestHeaders() func.

bcda/auth/client/oktaclient.go
dhgreene: Great!

dhgreene:
> should we make any changes to docker-compose.yml or docker-compose.test.yml for the env vars you added (client id and secret)?

No, I wouldn't think so (yet).  They're only applicable to tests that must be run manually.

bcda/auth/client/oktaclient_test.go
dhgreene: But then it would change the secret I carefully put in my environment.  😄

alexhova: this is just the clientID.  not the secret

bcda/auth/client/oktaclient.go
```diff
@@ -149,6 +163,54 @@ func (oc *OktaClient) AddClientApplication(localID string) (string, string, erro
...
+	params.Set("client_id", creds.ClientID)
+	params.Set("client_secret", creds.ClientSecret)
+	params.Set("grant_type", "client_credentials")
+	params.Set("scope", "bcda_api")
+	req.URL.RawQuery = params.Encode()
+
+	httpClient := &http.Client{}
+
+	logRequest(requestID).Print("requesting Okta access token")
+	resp, err := httpClient.Do(req)
```
alexhova: one more place you can use `client()`.  that'll also ensure it has a timeout

DeirdreHolub: +1

bcda/auth/client/oktaclient_test.go
dhgreene: But the following code changes the secret associated with it.  We'd have to have a separate client ID for this test.  I'll back down from my previous position.

PR #192, opened 2019-02-28
BCDA-957: Add CMS-format ID to ACOs

embh:
### Fixes [BCDA-957](https://jira.cms.gov/browse/BCDA-957)
For attribution using CCLF8 and CCLF9, beneficiaries are associated with ACOs by their CMS IDs. We'll need to store this ACO ID value (AXXXX). (Beneficiary IDs handled by separate ticket.)

### Proposed changes:
Add CMS ID to ACO model while retaining our internal ACO UUID.

### Change Details
* Adds string/char(5) `CMSID` to `ACO` model/table
* Updates `create-aco` CLI command to accept optional CMS ID
* Moves `createACO()` from main.go to cli.go (recently added file for CLI functions)

### Security Implications
None

### Acceptance Validation
ACO can be created from CLI with or without CMS ID (if valid format).
screen shot 2019-02-28 at 2 08 54 pm

### Feedback Requested
Any

### DIFF
--- a/bcda/cli.go
+++ b/bcda/cli.go
@@ -19,15 +19,17 @@ func createACO(name, cmsID string) (string, error) {
 		return "", errors.New("ACO name (--name) must be provided")
 	}
 
+	var cmsIDPt *string
 	// ACO ID is not required, but must match `AXXXX` if provided
 	if cmsID != "" {
 		acoIDFmt := regexp.MustCompile(`^A\d{4}$`)
 		if !acoIDFmt.MatchString(cmsID) {
 			return "", errors.New("ACO CMS ID (--cms-id) is invalid")
 		}
+		cmsIDPt = &cmsID
 	}
 
-	acoUUID, err := models.CreateACO(name, cmsID)
+	acoUUID, err := models.CreateACO(name, cmsIDPt)
 	if err != nil {
 		return "", err
 	}
--- a/bcda/cli.go
+++ b/bcda/cli.go
@@ -20,7 +20,6 @@ func createACO(name, cmsID string) (string, error) {
 	}
 
 	var cmsIDPt *string
-	// ACO ID is not required, but must match `AXXXX` if provided
 	if cmsID != "" {
 		acoIDFmt := regexp.MustCompile(`^A\d{4}$`)
 		if !acoIDFmt.MatchString(cmsID) {
--- a/bcda/cli.go
+++ b/bcda/cli.go
@@ -10,7 +10,30 @@ import (
 	"regexp"
 	"strconv"
 	"time"
+
+	"github.com/CMSgov/bcda-app/bcda/models"
 )
+
+func createACO(name, cmsID string) (string, error) {
+	if name == "" {
+		return "", errors.New("ACO name (--name) must be provided")
+	}
+
+	// ACO ID is not required, but must match `AXXXX` if provided
+	if cmsID != "" {
+		acoIDFmt := regexp.MustCompile(`^A\d{4}$`)
+		if !acoIDFmt.MatchString(cmsID) {
+			return "", errors.New("ACO CMS ID (--cms-id) is invalid")
+		}
+	}
+
+	acoUUID, err := models.CreateACO(name, cmsID)
+	if err != nil {
+		return "", err
+	}
+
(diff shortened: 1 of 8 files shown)


bcda/models/models.go
```diff
@@ -52,8 +53,9 @@ type JobKey struct {
 type ACO struct {
 	gorm.Model
 	UUID     uuid.UUID `gorm:"primary_key; type:char(36)" json:"uuid"` // uuid
-	Name     string    `json:"name"`                                   // name
-	ClientID string    `json:"client_id"`                              // software client id
+	CMSID    string    `gorm:"type:char(5)" json:"cms_id"`
```
knollfear: can/should we enforce uniqueness here?  What would happen if we had 2 acos with ID of A1234?

bcda/cli.go
```diff
@@ -10,8 +10,33 @@ import (
...
+	"github.com/CMSgov/bcda-app/bcda/models"
 )
 
+func createACO(name, cmsID string) (string, error) {
+	if name == "" {
+		return "", errors.New("ACO name (--name) must be provided")
+	}
+
+	var cmsIDPt *string
+	// ACO ID is not required, but must match `AXXXX` if provided
```
alexhova: i think this comment is explained in the clear, readable code below and can be removed.  you know i have strong opinions about inline comments and find them distracting so if you want to keep it, that's cool too!

bcda/cli.go
```diff
@@ -10,8 +10,33 @@ import (
...
+	if cmsID != "" {
+		acoIDFmt := regexp.MustCompile(`^A\d{4}$`)
+		if !acoIDFmt.MatchString(cmsID) {
+			return "", errors.New("ACO CMS ID (--cms-id) is invalid")
```
alexhova: woah is this how you do string interpolation in go?  that's sweet 👍

bcda/models/models.go
```diff
@@ -52,8 +53,9 @@ type JobKey struct {
 type ACO struct {
 	gorm.Model
 	UUID     uuid.UUID `gorm:"primary_key; type:char(36)" json:"uuid"` // uuid
-	Name     string    `json:"name"`                                   // name
-	ClientID string    `json:"client_id"`                              // software client id
+	CMSID    *string   `gorm:"type:char(5); unique" json:"cms_id"`
+	Name     string    `json:"name"`      // name
```
alexhova: do these comments do anything in documentation or otherwise? for consistency, maybe we should remove them all or add a similar one to `CMSID`

bcda/models/models.go
```diff
@@ -82,11 +84,11 @@ func GetATOPrivateKey() *rsa.PrivateKey {
 	return secutils.OpenPrivateKeyFile(atoPrivateKeyFile)
 }
 
-func CreateACO(name string) (uuid.UUID, error) {
+func CreateACO(name string, cmsID *string) (uuid.UUID, error) {
```
alexhova: why a pointer here?  it works fine but just wondering for my own knowledge

db/api.sql
```diff
@@ -1,5 +1,6 @@
 create table acos (
   uuid uuid not null primary key,
+  cms_id char(5) unique null,
```
alexhova: why is cms_id nullable?  for backwards compatibility?

bcda/cli.go
embh: 👍 Hooray! I didn't know if it would be helpful to explain the business rule or not. Happy to remove it.

bcda/models/models.go
embh: I was wondering that about the `json` ones. I don't know where we'd be using those because they're not included in our Swagger models, and we wouldn't be unmarshalling ACO JSON objects. I'll try removing.

db/api.sql
embh: Yep, exactly. Without this we'd need to add lots of IDs matching the CMS format, and there aren't many available for testing purposes.

bcda/models/models.go
embh: To allow it to be null for adding to the db. We could also assume empty string means nil and not use a pointer, but this seemed a little simpler.

PR #193, opened 2019-02-28
BCDA-956 Feature: add Okta token generation to cli "create-token" command

dhgreene:
### Fixes [BCDA-956](https://jira.cms.gov/browse/BCDA-956)

For ops purposes, we need to be able to get an Okta token.  This PR extends the existing CLI command `create-token` to work when using an OktaAuthPlugin auth provider.

### Proposed changes:

- Change the `--user-id` argument to `--id` to account for both user-id's and client-id's
- Add a `--secret` argument, which is ignored when using the AlphaAuthPlugin auth provider.
- Too many `switch` statements in the code and tests!

### Security Implications
This is strictly a dev-only tool.  There should be no change to the security posture with this PR.

### Acceptance Validation
screen shot 2019-02-28 at 2 44 02 pm

### Testing
Get a token by running:
`docker exec -it bcda-app_api_1 sh -c 'BCDA_AUTH_PROVIDER=okta OKTA_OAUTH_SERVER_ID=<okta_server_id> OKTA_CLIENT_TOKEN=token_not_used_but_required tmp/bcda create-token --id <a_valid_client_id> --secret <this_client_secret>'`

and (using an alpha token included in the test suite)
`docker exec -it bcda-app_api_1 sh -c 'tmp/bcda create-token --id 82503A18-BF3B-436D-BA7B-BAE09B7FFD2F'`

### Feedback Requested
- Does it work with both auth providers?
- Were the switch statements or other code smells acceptable compromises?
- Is there adequate test coverage for this stage of development?

### DIFF
--- a/bcda/auth/alpha.go
+++ b/bcda/auth/alpha.go
@@ -140,39 +140,52 @@ func (p AlphaAuthPlugin) RevokeClientCredentials(params []byte) error {
 	return nil
 }
 
-// generate a token for the id (which user? just have a single "user" (alpha2, alpha3, ...) per test cycle?)
-// params are currently acoId and ttl; not going to introduce user until we have clear use cases
+// generate a token for the ACO, either for a specified UserID or (if not provided) any user in the ACO
 func (p AlphaAuthPlugin) RequestAccessToken(creds Credentials, ttl int) (Token, error) {
-	db := database.GetGORMDbConnection()
-	defer database.Close(db)
-
+	var userUUID, acoUUID uuid.UUID
+	var user models.User
+	var err error
 	token := Token{}
 
-	acoUUID := creds.ClientID
-	if acoUUID == "" {
-		return token, errors.New("no ACO ID provided")
-	}
-
-	aco, err := getACOFromDB(acoUUID)
-	if err != nil {
-		return token, err
-	}
-
-	// I arbitrarily decided to use the first user. An alternative would be to make a specific user
-	// that represents the client. I have no strong opinion here other than not creating stuff in the db
-	// unless we're willing to live with it forever.
-	var user models.User
-	if err = db.First(&user, "aco_id = ?", aco.UUID).Error; err != nil {
-		return token, errors.New("no user found for " + aco.UUID.String())
+	if creds.UserID == "" && creds.ClientID == "" {
+		return token, fmt.Errorf("must provide either UserID or ClientID")
 	}
 
 	if ttl < 0 {
 		return token, fmt.Errorf("invalid TTL: %d", ttl)
 	}
 
+	db := database.GetGORMDbConnection()
+	defer database.Close(db)
+
+	// Get both acoUUID and userUUID based on creds.UserID if provided
+	if creds.UserID != "" {
+		if db.First(&user, "UUID = ?", creds.UserID).RecordNotFound() {
+			return token, fmt.Errorf("unable to locate User with id of %s", creds.UserID)
+		}
+
+		userUUID = user.UUID
+		acoUUID = user.ACOID
+	} else {
+		aco, err := getACOFromDB(creds.ClientID)
+		if err != nil {
+			return token, err
+		}
+
+		// I arbitrarily decided to use the first user. An alternative would be to make a specific user
+		// that represents the client. I have no strong opinion here other than not creating stuff in the db
+		// unless we're willing to live with it forever.
+		if err = db.First(&user, "aco_id = ?", aco.UUID).Error; err != nil {
+			return token, errors.New("no user found for " + aco.UUID.String())
+		}
+
+		userUUID = user.UUID
+		acoUUID = aco.UUID
+	}
+
 	token.UUID = uuid.NewRandom()
-	token.UserID = user.UUID
-	token.ACOID = aco.UUID
+	token.UserID = userUUID
+	token.ACOID = acoUUID
 	token.IssuedAt = time.Now().Unix()
 	token.ExpiresOn = time.Now().Add(time.Hour * time.Duration(ttl)).Unix()
 	token.Active = true
--- a/bcda/auth/alpha.go
+++ b/bcda/auth/alpha.go
@@ -158,8 +158,12 @@ func (p AlphaAuthPlugin) RequestAccessToken(creds Credentials, ttl int) (Token, 
 	db := database.GetGORMDbConnection()
 	defer database.Close(db)
 
-	// Get both acoUUID and userUUID based on creds.UserID if provided
 	if creds.UserID != "" {
+		userUUID = uuid.Parse(creds.UserID)
+		if userUUID == nil {
+			return token, fmt.Errorf("user ID must be a UUID")
+		}
+
 		if db.First(&user, "UUID = ?", creds.UserID).RecordNotFound() {
 			return token, fmt.Errorf("unable to locate User with id of %s", creds.UserID)
 		}
@@ -172,9 +176,6 @@ func (p AlphaAuthPlugin) RequestAccessToken(creds Credentials, ttl int) (Token, 
 			return token, err
 		}
 
-		// I arbitrarily decided to use the first user. An alternative would be to make a specific user
-		// that represents the client. I have no strong opinion here other than not creating stuff in the db
-		// unless we're willing to live with it forever.
 		if err = db.First(&user, "aco_id = ?", aco.UUID).Error; err != nil {
 			return token, errors.New("no user found for " + aco.UUID.String())
 		}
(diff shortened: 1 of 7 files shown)


bcda/main_test.go
```diff
@@ -243,51 +243,75 @@ func (s *MainTestSuite) TestCreateUser() {
...
-	args = []string{"bcda", "create-token", "--abcd", "efg"}
-	err = s.testApp.Run(args)
+	args := []string{"bcda", "create-token", "--abcd", "efg"}
+	err := s.testApp.Run(args)
 	assert.Equal("flag provided but not defined: -abcd", err.Error())
 	assert.Contains(buf.String(), "Incorrect Usage: flag provided but not defined")
+	buf.Reset()
+
+	switch auth.GetProvider().(type) {
+	case auth.OktaAuthPlugin:
```
msnook: No need for a `case` statement here.  All the scenarios being tested in this switch are negative cases.  In order to increase test coverage, we should set the env var in the test case, execute tests, then set the env var back.  That way we get more test coverage.  There are lots of examples in our existing test suite where we override env vars and then set them back.

bcda/main_test.go
```diff
@@ -640,52 +648,58 @@ func (s *MainTestSuite) TestCleanArchive() {
...
-	assert.Equal(0, buf.Len())
-	buf.Reset()
-
-	// Positive case - revoke a token passing in a valid token string (alpha)
-	args = []string{"bcda", "revoke-token", "--access-token", alphaTokenString}
-	err = s.testApp.Run(args)
-	assert.Nil(err)
-	assert.Equal("Access token has been deactivated\n", buf.String())
-	buf.Reset()
+	switch auth.GetProvider().(type) {
```
knollfear: originalAuthProvider := auth.GetProviderName()
        defer auth.SetProvider(originalAuthProvider)
	auth.SetProvider("alpha")
is better than this switch imho

bcda/auth/alpha.go
```diff
@@ -140,39 +140,52 @@ func (p AlphaAuthPlugin) RevokeClientCredentials(params []byte) error {
...
-	var user models.User
-	if err = db.First(&user, "aco_id = ?", aco.UUID).Error; err != nil {
-		return token, errors.New("no user found for " + aco.UUID.String())
-	}
+	db := database.GetGORMDbConnection()
+	defer database.Close(db)
 
-	if ttl < 0 {
-		return token, fmt.Errorf("invalid TTL: %d", ttl)
+	// Get both acoUUID and userUUID based on creds.UserID if provided
```
alexhova: Is this comment necessary? I think this is explained by the well-written code below it already

bcda/auth/alpha.go
```diff
@@ -140,39 +140,52 @@ func (p AlphaAuthPlugin) RevokeClientCredentials(params []byte) error {
...
+		acoUUID = user.ACOID
+	} else {
+		aco, err := getACOFromDB(creds.ClientID)
+		if err != nil {
+			return token, err
+		}
+
+		// I arbitrarily decided to use the first user. An alternative would be to make a specific user
+		// that represents the client. I have no strong opinion here other than not creating stuff in the db
+		// unless we're willing to live with it forever.
```
alexhova: I think these comments can be removed as well.  it's obvious by the use of `db.First(&user)` that we are arbitrarily selecting the first user

bcda/auth/okta.go
```diff
@@ -76,15 +76,21 @@ func (o OktaAuthPlugin) RevokeClientCredentials(params []byte) error {
 }
 
 func (o OktaAuthPlugin) RequestAccessToken(creds Credentials, ttl int) (Token, error) {
-	if creds.ClientID == "" {
+	clientID := creds.ClientID
+	// Also accept clientID via creds.UserID to match alpha auth implementation
```
alexhova: conversely, i think this comment makes sense to have.  it's obvious _how_ the code works but we don't know _why_ we're doing it without the comment. 👍

bcda/auth/alpha.go
```diff
@@ -140,39 +140,52 @@ func (p AlphaAuthPlugin) RevokeClientCredentials(params []byte) error {
...
-	}
+	db := database.GetGORMDbConnection()
+	defer database.Close(db)
 
-	if ttl < 0 {
-		return token, fmt.Errorf("invalid TTL: %d", ttl)
+	// Get both acoUUID and userUUID based on creds.UserID if provided
+	if creds.UserID != "" {
+		if db.First(&user, "UUID = ?", creds.UserID).RecordNotFound() {
+			return token, fmt.Errorf("unable to locate User with id of %s", creds.UserID)
```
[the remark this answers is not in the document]
alexhova: you can't return nil as type `Token` from this func though since the func signature declares a `Token` as the first output.  so it's kind of interesting that you can return nil instead of type `Error`, which is definitely that idiomatic Go we love.  that is allowed because the zero value of the `error` type is actually `nil`.  `Token`'s nil value is `Token{}` for contrast.  kinda cool!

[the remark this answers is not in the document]
alexhova: and looks like the zero value of structs is just the struct with zero values for each field: https://play.golang.org/p/QDohNnjEpp0

bcda/main_test.go
dhgreene: Thanks!

bcda/auth/alpha.go
dhgreene: OK.

bcda/auth/alpha.go
dhgreene: These comments are preserved from the previous version, but I agree they don't express our current concerns.

bcda/auth/okta.go
dhgreene: 👍

bcda/main.go
```diff
@@ -390,32 +393,18 @@ func createUser(acoID, name, email string) (string, error) {
 	return user.UUID.String(), nil
 }
 
-func createAccessToken(userID string) (string, error) {
+func createAccessToken(userID string, secret string) (string, error) {
```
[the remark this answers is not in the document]
dhgreene: 👍

bcda/main.go
```diff
@@ -390,32 +393,18 @@ func createUser(acoID, name, email string) (string, error) {
...
-	var userUUID uuid.UUID
 
 	if userID == "" {
-		errMsgs = append(errMsgs, "User ID (--user-id) must be provided")
-	} else {
-		userUUID = uuid.Parse(userID)
-		if userUUID == nil {
-			errMsgs = append(errMsgs, "User ID must be a UUID")
-		}
+		errMsgs = append(errMsgs, "ID (--id) must be provided")
```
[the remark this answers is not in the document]
dhgreene: Yes, with only one thing we're checking here, that cleans up a lot of code.

bcda/main_test.go
```diff
@@ -243,51 +243,59 @@ func (s *MainTestSuite) TestCreateUser() {
...
 
-	// Bad User UUID
-	args = []string{"bcda", "create-token", "--user-id", BADUUID}
+	// Alpha auth section
+	originalAuthProvider := auth.GetProviderName()
+	auth.SetProvider("alpha")
+	// Test alpha auth bad ID
+	args = []string{"bcda", "create-token", "--id", badUUID}
 	err = s.testApp.Run(args)
-	assert.Equal("User ID must be a UUID", err.Error())
```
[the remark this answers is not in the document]
dhgreene: We're no longer verifying that it's a UUID in `main.go`, but I've taken what I hope is the spirit of your comment and moved that check to `alpha.go`. In that location, the message won't say anything about the CLI syntax.  I've brought back a modified assertion.

bcda/main_test.go
```diff
@@ -243,51 +243,59 @@ func (s *MainTestSuite) TestCreateUser() {
...
+	assert.Nil(err)
+	assert.NotNil(buf)
+	accessTokenString := strings.TrimSpace(buf.String())
+	assert.Nil(err)
+	assert.NotEmpty(accessTokenString)
+	buf.Reset()
+
+	// Okta auth section
+	auth.SetProvider("okta")
+
```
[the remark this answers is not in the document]
dhgreene: Removed

bcda/main_test.go
dhgreene: 💯

bcda/main_test.go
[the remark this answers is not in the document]
dhgreene: @knollfear let me know if this change meets your concern.

dhgreene:
@alexhova and @msnook I believe I've implemented your suggestions.

PR #194, opened 2019-02-28
Modify Okta create-alpha-token output to include client name

DeirdreHolub:
### Fixes [BCDA-968](https://jira.cms.gov/browse/BCDA-968)
We need the client name in the output.

### Proposed changes:
For Okta, replaced first line of output with client name

### Security Implications
No change in security posture.

### Acceptance Validation
All unit and integration tests run.

To hand test:

`make docker-bootstrap`

edit docker-compose.yml and set these env variables:
BCDA_AUTH_PROVIDER=okta
OKTA_OAUTH_SERVER_ID=_server id_
OKTA_CLIENT_TOKEN=_token_

`docker-compose up`
`docker exec -it bcda-app_api_1 sh -c 'tmp/bcda create-alpha-token --size Dev'`

### Feedback Requested
Any.

### DIFF
--- a/bcda/auth/client/oktaclient.go
+++ b/bcda/auth/client/oktaclient.go
@@ -36,6 +36,7 @@ type OktaToken struct {
 type Credentials struct {
 	ClientID     string
 	ClientSecret string
+	ClientName   string
 }
 
 func init() {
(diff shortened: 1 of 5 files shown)


bcda/auth/client/oktaclient.go
```diff
@@ -112,23 +112,24 @@ func (oc *OktaClient) PublicKeyFor(id string) (rsa.PublicKey, bool) {
 	return key, ok
 }
 
-func (oc *OktaClient) AddClientApplication(localID string) (string, string, error) {
+func (oc *OktaClient) AddClientApplication(localID string) (string, string, string, error) {
```
dhgreene: We're starting to have a confusing number of string return values.  Would it be a good idea to name them (here, in`okta.go` and in `mokta.go`)?

`func (oc *OktaClient) AddClientApplication(localID string) (clientID string, clientSecret string, clientName string, err error)`

bcda/main.go
```diff
@@ -483,8 +483,7 @@ func createAlphaToken(ttl int, acoSize string) (s string, err error) {
 		result = fmt.Sprintf("%s\n%s\n%s", expiresOn, tokenId, token.TokenString)
 
 	case auth.OktaAuthPlugin:
-		expiresOn := time.Now().AddDate(1, 0, 0).Format(time.RFC850)
-		result = fmt.Sprintf("%s\n%s\n%s", expiresOn, creds.ClientID, creds.ClientSecret)
+		result = fmt.Sprintf("%s\n%s\n%s", creds.ClientName, creds.ClientID, creds.ClientSecret)
```
dhgreene: Is it time yet to change the name of `createAlphaToken`?

DeirdreHolub: I think we want to leave that to when we actually the alpha impl to return credentials.

bcda/auth/client/oktaclient.go
dhgreene: Now that you've gone to the trouble of naming these, I notice you've added `ClientName` to Credentials{}, which makes it a good candidate for a return type.

DeirdreHolub: I started down that path (which is why I put it in there), but that doing so ends in a method that does nothing but copy one struct into another, which did not feel idiomatic and leads down the path of refactoring how credentials are handled through the entire auth package, not just the oktaclient. I don't want to go there for this PR.

PR #196, opened 2019-03-06
Knollfear/bcda 923 multi thread jobs

knollfear:
### Fixes [BCDA-923](https://jira.cms.gov/browse/BCDA-923)

Implement multi threading to allow multiple threads to pull data from BB simultaneously
### Proposed changes:

When the job is requested it now enques many jobs instead of one.  This happens in the Request/Response cycle but happens so fast it seems fine even for 10,000 jobs created.

### Security Implications

PII should not be affected

### Acceptance Validation

Unit tests.  you may also want to manually override BCDA_FHIR_MAX_RECORDS and submit some jobs that way.  1 is a confirmed valid value.
also, submit a job for a large/extra large ACO. to get a feel for how it will work.  (reset the job size before you do this)

### Feedback Requested

any


bcda/api.go
```diff
@@ -153,6 +175,11 @@ func bulkRequest(t string, w http.ResponseWriter, r *http.Request) {
 	}
 	defer rows.Close()
 	var id string
+	var jobCount = 0
+	maxBeneficiaries, err := strconv.Atoi(os.Getenv("BCDA_FHIR_MAX_RECORDS"))
```
msnook: You can use `getEnvInt` to automatically set the default if the env var doesn't exist.  
Here is an example: https://github.com/CMSgov/bcda-app/blob/master/bcda/api.go#L534

bcda/api.go
```diff
@@ -53,6 +54,9 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+//Setting a default here.  It may need to change.  It also shouldn't really be used much.
+const BCDA_FHIR_MAX_RECORDS = 10000
```
msnook: Highly recommend naming this something different.  It is a bit confusing that it is named exactly the same as the env var.

bcda/api.go
```diff
@@ -162,49 +189,75 @@ func bulkRequest(t string, w http.ResponseWriter, r *http.Request) {
...
+				oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, "", responseutils.Processing)
+				responseutils.WriteError(oo, w, http.StatusInternalServerError)
+				return
+			}
+			jobCount++
+			beneficiaryIds = []string{}
+		}
 	}
+	// Get the last chunk of beneficiaries queued up if there are any.
+	if len(beneficiaryIds) > 0 {
```
msnook: Looks like a lot of DRY in this conditional block.  How can we alleviate the repeated code?

bcda/models/models.go
```diff
@@ -33,14 +34,43 @@ func InitializeGormModels() *gorm.DB {
 type Job struct {
 	gorm.Model
 	ACO        ACO       `gorm:"foreignkey:ACOID;association_foreignkey:UUID"` // aco
-	ACOID      uuid.UUID `gorm:"primary_key; type:char(36)" json:"aco_id"`
```
msnook: Was this intentional?

unit_test.sh
```diff
@@ -23,7 +23,7 @@ usql $TEST_DB_URL -f db/test_synthetic_beneficiaries.sql
 usql $TEST_DB_URL -f db/worker.sql
 
 echo "Running unit tests and placing results/coverage in test_results/${timestamp} on host..."
-DATABASE_URL=$TEST_DB_URL QUEUE_DATABASE_URL=$TEST_DB_URL gotestsum --junitfile test_results/${timestamp}/junit.xml -- -race ./... -coverprofile test_results/${timestamp}/testcoverage.out 2>&1 | tee test_results/${timestamp}/testresults.out
+DATABASE_URL=$TEST_DB_URL QUEUE_DATABASE_URL=$TEST_DB_URL gotestsum --format standard-verbose --junitfile test_results/${timestamp}/junit.xml -- -race ./... -coverprofile test_results/${timestamp}/testcoverage.out 2>&1 | tee test_results/${timestamp}/testresults.out
```
msnook: Wasn't there some reason for using the default format because it didn't clutter the Jenkins/Travis output?

docker-compose.yml
```diff
@@ -73,6 +74,7 @@ services:
       - BB_CLIENT_KEY_FILE=../shared_files/bb-dev-test-key.pem 
       - BB_SERVER_LOCATION=https://fhir.backend.bluebutton.hhsdevcloud.us
       - BB_TIMEOUT_MS=500
+      - BCDA_FHIR_MAX_RECORDS=10000
```
msnook: I didn't see this env var being used by `bcdaworker`.  Is this needed?

docker-compose.test.yml
```diff
@@ -21,6 +21,7 @@ services:
       - FHIR_ARCHIVE_DIR=../bcdaworker/archive
       - OKTA_CLIENT_ORGURL=https://cms-sandbox.oktapreview.com
       - OKTA_EMAIL=shawn@bcda.aco-group.us
+      - BCDA_FHIR_MAX_RECORDS=100
```
msnook: I didn't see this env var being used by the tests. Is this needed?

unit_test.sh
knollfear: this was a mistake to commit.  I will fix it.

bcda/models/models.go
knollfear: It was.  This was never intended to be a primary key for Job.

msnook: 👍

bcda/api.go
knollfear: I named it that way on purpose so it would be clear that it represented the same thing.  I can change it to BCDA_FHIR_MAX_RECORDS_DEFAULT

bcda/api.go
knollfear: great.  I had no idea this existed.

bcda/api.go
knollfear: I tried to figure that out and didn't have any good ideas.  It is doing literally 3 things:
Marshall the job args
Que the job
increment the job count
The rest of it is necessary handling for the errors.  I can do something like args, err := dojobMarshall(int(newJob.ID), acoId, userId, beneficiaryIds, t, encrypt) but that hardly seems worth it and a function with that many args makes me sad.

docker-compose.yml
knollfear: I wasn't sure.  I would rather add it now than be super confused about what is happening later when it isn't there.

docker-compose.test.yml
knollfear: I wasn't sure. I would rather add it now than be super confused about what is happening later when it isn't there.

bcda/api.go
msnook: What if the first conditional is 
```
if len(beneficiaryIds) >= maxBeneficiaries || len(beneficiaryIds) > 0
   DO STUFF
```

and the second conditional is just 
```
if len(beneficiaryIds) > 0
   DO MORE STUFF
```

docker-compose.yml
msnook: I think the answer is... if you use it in the `bcdaworker` code, then it should be here.  Otherwise, it shouldn't.  This is important because this is where we usually grab the env vars required for S3/AWS.

bcda/api.go
knollfear: That isn't it.  I am giving you $123 in 20's.  After I give you 6 twenties I still owe you $3.  That code is to make sure I get you the other $3

knollfear: I think I fixed this.  I just needed a count to know how far along I was in the process

knollfear:
> Looking good. Just a couple requested changes. Also, since this is kind of modifying the way we handle jobs, it might make sense to try to test this a little more thorough via integration tests. Any ideas on how we might be able to slightly modify our smoke test to ensure that we have a regression test in place for this work? Maybe we can trigger two or more requests simultaneously? I wonder if we could take advantage of `&&` and string a few tests together...
> Here: https://github.com/CMSgov/bcda-app/blob/master/test/smoke_test/smoke_test.sh#L7-L12
> And here: https://github.com/CMSgov/bcda-ops/blob/master/jenkins_files/Jenkinsfile.smoke_test#L97-L102
> 
> It would be sweet if we could execute all those tests in parallel.
> 
> Thinking more on this... would this approach even hit your code since it is separate requests? 🤔

I think some of that is out of bounds for this ticket.  however, I did make changes to the smoke test runner to parse all parts of the output section to make sure that we are checking all files created by the job, not just the first one.

knollfear:
submitting new PR for code that got refactored

PR #197, opened 2019-03-07
BCDA-723: Update DB to be used with BB DPR

msnook:
### Fixes [BCDA-723](https://jira.cms.gov/browse/BCDA-723)

BB DPR only contains negative `patient_id` values.

### Proposed changes:

Create a helper script to migrate a database to be used with BB DPR.

### Change Details

- Helper script duplicates the data in `beneficiaries` table, adding all `patient_id` with negative values.

### Acceptance Validation

Smoke test is passing in `dev`

### Feedback Requested
Please review.


msnook:
I am going to not deliver this to baseline to avoid confusion, since I believe it will only be a one-time use.

PR #198, opened 2019-03-07
WIP: Require that the OKTA secret values be passed in on command line.

msnook:
### Fun with env vars

To reduce the risk of committing/pushing actual secret values in the Docker configuration files, we should leverage environment values that are passed to Docker on command line.

### Proposed changes:

The Okta environment should be set in the container by passing an environment value on the command line (or by exporting them in your shell).

### Change Details

Updated the `docker-compose.yml` and `docker-compose.test.yml` to set the Okta values using environment values.

### Security Implications
None.

### Acceptance Validation

Very minimal testing performed.  Using this PR as a starting point to get the ball rolling.  

Here is sample usage, setting just one value.  Obviously, this illustrates how we'd use it to set all the values dynamically with no risk of committing/pushing anything to the repository.

Command:
```BCDA_AUTH_PROVIDER=okta docker-compose up```

Output:
screen shot 2019-03-07 at 9 54 00 am

### Feedback Requested
Is this an acceptable approach?  Any gotchas?  We'll probably need to update the `Makefile` in some areas to make sure all targets can accept the environment values too.  
Please consider all scenarios you use regularly.

### CODE REVIEW

File: docker-compose.test.yml [hunk 1]
```diff
@@ -21,11 +21,10 @@ services:
...
       - OKTA_CLIENT_ORGURL=https://cms-sandbox.oktapreview.com
       - OKTA_EMAIL=shawn@bcda.aco-group.us
-# Add the Okta values below and uncomment
-#     - BCDA_AUTH_PROVIDER=okta
-#     - OKTA_CLIENT_TOKEN=<token>
-#     - OKTA_OAUTH_SERVER_METADATA=
-#     - OKTA_OAUTH_SERVER_ID
+      - OKTA_CLIENT_TOKEN=${OKTA_CLIENT_TOKEN}
+      - BCDA_AUTH_PROVIDER=${BCDA_AUTH_PROVIDER}
+      - OKTA_OAUTH_SERVER_METADATA=${OKTA_OAUTH_SERVER_METADATA}
```
DeirdreHolub: We no longer need this variable.

msnook: @DeirdreHolub are you referring to `OKTA_OAUTH_SERVER_METADATA`?

msnook: @DeirdreHolub Done.  Removed.

msnook:
> Remove deprecated env var OKTA_OAUTH_SERVER_METADATA

Done!

PR #200, opened 2019-03-08
BCDA-1030: implement strategy for handling secrets safely

alexhova:
### Fixes [BCDA-1030](https://jira.cms.gov/browse/BCDA-1030)
This provides instructions and a template file for handling secrets safely to reduce the chance of incidents occurring and to be more convenient for local development

### Proposed changes:
- add `.env.sh` to `.gitignore` so it is never committed
- add `.env.sh` file to the bcda keybase team
- update `README.md` to provide instructions for using this strategy

### Feedback Requested
- is the README clear and the strategy safe?


msnook:
I think this will cause the same problem.  If someone edits `env.sh` with their env vars, they might forget that it was edited and then commit/push and we will have another incident.  I think we need to avoid having any versioned file that sets an env var.  Perfectly fine to set it in `.bash_profile`, or `.env` or on command line.

dhgreene:
I really appreciate the thought put in on this.

>some other options are: (1) fire up the compose stack specifying the Okta vars on the command line, (2) fire up the compose stack after the Okta vars have been exported in the shell, or (3) using a separate docker-compose file (not versioned) which overrides the Okta vars.

I don't see all of these examples yet, just variations on (2).

>I think it's important to show many safe options. I think it would also be useful also to include our command line runs

Should this be (4) (passing temporary env vars to non-docker tests)?

alexhova:
@dhgreene Thanks.  I just thought it would be best to have _one_ solution that we all use so we can be on the same page and iterate further if necessary.  This works for docker (since we get env vars in docker compose files) and it works for local development and testing.  As long as we're hitting all of our bases where secrets are used.  I thought it best to keep this solution simple so it is adopted and used.  We can always refine later.  Do you still think it would be better to include more options?  This question is for @DeirdreHolub and @msnook as well

dhgreene:
@alexhova Documenting one strategy for clarity is fine, and your latest commit helps with that.

Perhaps some of the other strategies belong more appropriately to the `Test` section, as a later improvement.

PR #201, opened 2019-03-12
BCDA-963: Beneficiary:ACO relationship to many:many

embh:
### Fixes [BCDA-963](https://jira.cms.gov/browse/BCDA-963)
BCDA does not currently allow for the fact that beneficiaries may belong to multiple ACOs.

### Proposed changes:
Change the ACO:beneficiary relationship from one:many to many:many.

### Change Details
* Removes ACO ID from beneficiary model and table
* Adds `acos_beneficiaries` join table
* Updates synthetic beneficiaries SQL scripts and `AssignAlphaBeneficiaries()` to use new structure

### Security Implications
None

### Acceptance Validation
Tests continue to pass.

### Feedback Requested
Any


bcda/main_test.go
```diff
@@ -44,6 +44,7 @@ func (s *MainTestSuite) SetupSuite() {
 
 func (s *MainTestSuite) SetupTest() {
 	s.testApp = setUpApp()
+	models.InitializeGormModels()
```
knollfear: should this be autoMigrate()?  also, if we are doing this in the test setup should we remove it from the tests that are calling it now?  I see it in a few tests below like TestArchiveExpiringWithThreshold

PR #202, opened 2019-03-13
knollfear/BCDA-1068 adding decryption util files from gists to our repo

knollfear:
### Fixes [BCDA-1068](https://jira.cms.gov/browse/BCDA-1068)

Decryption utils are hosted in gists on people's personal github pages.  We should gasther those resources into our repo so that it instills more confidence in our users that the code is legit and the whole team can work on them as needed.
### Proposed changes:

copied gist files into decryption_utils folder

### Security Implications

no PII is affected.  These are code samples that will not be executed by our application

### Acceptance Validation

Just verify the necessary files are there and with the right names.


dhgreene:
Looks good except the CI

PR #203, opened 2019-03-14
BCDA-1031 Simplify token handling

DeirdreHolub:
### Fixes [BCDA-1031](https://jira.cms.gov/browse/BCDA-1031)

The goal of this PR is to normalize and simplify the handling of JWT tokens.

This PR is predicated on the following assumptions:

* We are changing the alpha auth implementation to use client credentials instead of a token for authentication
* We will no longer need to store a token for authentication purposes
* The auth.Token model should therefore be considered deprecated
* The ACO ID and User ID are not present in the Okta token; they must be looked up
* The API should be insulated from the inner workings of auth

### Proposed changes:

Puts the decoded token and identity values into the request context. Decoded values come from tokens verified to be signed by our server and to have not expired. Additional authorization occurs in RequireTokenAuth(). Only auth code should look at the token claims; API code should rely on the values in AuthData struct that holds the identity values. We do this to insulate API code from the differences among Provider tokens.

### Change Details

* Updated the jwt-go library to its most current maintained version. This gives us an expired token error, instead of a message string that must be parsed to confirm expiration. It is also 2 years newer than our current version.
* Introduced AuthData, which contains the identity values extracted by auth. AuthData is always put into the request context if a token is present in the request. 
  If the router passes the request on to RequireTokenAuth(), additional inspections are done to authorize access.
* Propagated these changes across auth and the api
* Standardized handling of tokens in the auth code with CommonClaims
* Replaced some uses of old auth.Backend methods with use of auth.Provider() methods
* Removed much direct use of auth.Backend outside of alpha.go

### Security Implications
No changes to the security posture. 

### Acceptance Validation

All tests pass.

### Feedback Requested

Any.


karlmdavis:
Just wanted to say: this is a beautifully detailed and useful PR description!

(I'd also encourage you to be this verbose in your commit messages, so that all of this wonderful context is available to future users of `git blame`.)

bcda/auth/middleware_test.go
```diff
@@ -47,17 +48,15 @@ func (s *MiddlewareTestSuite) CreateRouter() http.Handler {
...
 func (s *MiddlewareTestSuite) SetupTest() {
 	s.SetupAuthBackend()
-	validClaims := jwt.MapClaims{
-		"sub": "82503A18-BF3B-436D-BA7B-BAE09B7FFD2F",
-		"aco": "DBBD1CE1-AE24-435C-807D-ED45953077D3",
-		"id":  "d63205a8-d923-456b-a01b-0992fcb40968",
-		"iat": time.Now().Unix(),
-		"exp": time.Now().Add(time.Duration(999999999)).Unix(),
+	userID := "82503A18-BF3B-436D-BA7B-BAE09B7FFD2F"
+	acoID := "DBBD1CE1-AE24-435C-807D-ED45953077D3"
```
dhgreene: I see that BCDA-1040 will remove the userID from the alpha tokens.  Not here, but should we also align alpha and Okta as to the identifier being returned for the ACO?

bcda/auth/middleware.go
```diff
@@ -34,11 +39,43 @@ func ParseToken(next http.Handler) http.Handler {
...
+				if err != nil {
+					log.Errorf("no aco for clientID %s because %v", claims.ClientID, err)
+					next.ServeHTTP(w, r)
+					return
+				}
+
+				db := database.GetGORMDbConnection()
+				defer database.Close(db)
+
+				var user models.User
```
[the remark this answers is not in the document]
DeirdreHolub: yep yep. I was being very disciplined about not fixing all the things ;)

bcda/auth/middleware_test.go
DeirdreHolub: If you mean for constructing the token, I think the answer there will be that the plugins need to have a testToken() method (not exported) that will generate test tokens. That's one of the reasons I focused so much on getting a standard pattern going for creating them. If we do that, the same tests will work for either plugin.

PR #204, opened 2019-03-14
BCDA-973 Feature: Add /auth/token endpoint

dhgreene:
### Fixes [BCDA-973](https://jira.cms.gov/browse/BCDA-973)
We need an API endpoint for ACO's to get temporary access tokens.  They will present a client id and secret, and our auth provider(s) will verify and give a token if authorized.

### Proposed changes:
- A new function `GetAccessToken()` is used instead of `RequestAccessToken()` because `RequestAccessToken()` in alpha does not require a secret, and does require a TTL
- A number of testing updates solved issues that cropped up during development (especially trying to test with Okta as auth provider)
- Add endpoint to Swagger

### Security Implications
Since this will be open to all visitors in the public sandbox, and that token is the gateway to all data, any path that leads to releasing a token needs to be carefully reviewed.

### Acceptance Validation
Screen Shot 2019-03-14 at 11 35 27 AM

### Feedback Requested
- Are we using the right return values and parameter names?
- Is there more test coverage that we could/should have?
- Are the testing changes appropriate?

### DIFF
--- a/bcda/api.go
+++ b/bcda/api.go
@@ -17,16 +17,14 @@
      - application/json
 
      Security:
-	 - api_key:
-	 - basic_auth:
+     - api_key:
+
 
      SecurityDefinitions:
      api_key:
           type: apiKey
           name: Authorization
-		  in: header
-	 basic_auth:
-		  type: basic
+          in: header
  swagger:meta
 */
 package main
--- a/bcda/api.go
+++ b/bcda/api.go
@@ -17,13 +17,16 @@
      - application/json
 
      Security:
-     - api_key:
+	 - api_key:
+	 - basic_auth:
 
      SecurityDefinitions:
      api_key:
           type: apiKey
           name: Authorization
-          in: header
+		  in: header
+	 basic_auth:
+		  type: basic
  swagger:meta
 */
 package main
@@ -366,13 +369,12 @@ func serveData(w http.ResponseWriter, r *http.Request) {
 
 	Get access token
 
-	Verifies parameters client_id and secret, and returns a JWT token that can be presented to the other API endpoints.
-
-	Consumes:
-	- application/x-www-form-urlencoded
+	Verifies Basic authentication credentials, and returns a JWT token that can be presented to the other API endpoints.
 
 	Produces:
 	- application/json
+
+	Security: basic_auth
 
 	Schemes: https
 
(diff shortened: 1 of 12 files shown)


bcda/api_test.go
```diff
@@ -701,13 +701,14 @@ func (s *APITestSuite) TestAuthToken() {
...
 	assert.Equal(s.T(), http.StatusBadRequest, s.rr.Code)
-	assert.Empty(s.T(), s.rr.Body)
 
-	req.Header.Add("client_id", "not_a_client_id")
-	req.Header.Add("secret", "not_a_secret")
+	// // The following change in error status can be tested with alpha auth (presumably along with success paths)
+	// // once GetAccessToken() is implemented
+	//
+	// req.Header.Add("client_id", "not_a_client_id")
+	// req.Header.Add("secret", "not_a_secret")
```
DeirdreHolub: FYI, this is supposed to be a basic authentication header, which is one header value, not two. See https://en.wikipedia.org/wiki/Basic_access_authentication.

bcda/api.go
```diff
@@ -361,6 +361,50 @@ func serveData(w http.ResponseWriter, r *http.Request) {
...
+	Verifies parameters client_id and secret, and returns a JWT token that can be presented to the other API endpoints.
+
+	Consumes:
+	- application/x-www-form-urlencoded
+
+	Produces:
+	- application/json
+
+	Schemes: https
+
```
DeirdreHolub: Please make swagger docs reflect basic auth header required for this endpoint. https://swagger.io/docs/specification/2-0/authentication/basic-authentication/

bcda/api.go
```diff
@@ -361,6 +361,50 @@ func serveData(w http.ResponseWriter, r *http.Request) {
...
+	Responses:
+		200: tokenResponse
+		400: missingCredentials
+		401: invalidCredentials
+		500: serverError
+*/
+func getAuthToken(w http.ResponseWriter, r *http.Request) {
+	clientId := r.Header.Get("client_id")
+	secret := r.Header.Get("secret")
```
DeirdreHolub: This endpoint is supposed to be using basic authentication.

dhgreene: Thanks for pointing that out.  Code updated to basic auth.

bcda/api.go
dhgreene: @DeirdreHolub I've edited swagger docs to change to basic auth, but `swagger generate spec -o swaggerui/swagger.json -m` fails at the moment.  Passing code over to you.

bcda/main_test.go
```diff
@@ -328,6 +327,29 @@ func (s *MainTestSuite) TestCreateAlphaTokenCLI() {
 	l2 := strings.Split(anotherTokenInfo, "\n")
 	assert.NotEqual(l1[0], l2[0], "alpha expiration dates should be different (%s == %s)", l1[0], l2[0])
 	assert.NotEqual(l1[1], l2[1], "alpha token uuids should be different (%s == %s)", l1[1], l2[1])
+
+	// The following Okta tests should pass if you have OKTA_CLIENT_TOKEN set
```
alexhova: this can be uncommented if we're going to use it to test.  otherwise, let's delete so it doesn't confuse

DeirdreHolub:
@knoll got swagger working with two forms of auth.

PR #205, opened 2019-03-14
BCDA-963: (Part 2) Migration and synthetic data fixes

embh:
### Fixes [BCDA-963](https://jira.cms.gov/browse/BCDA-963)
Addendum to https://github.com/CMSgov/bcda-app/pull/201

Note that the migration script is intended to run **before** the GORM auto migration.

### Change Details
* Includes migration script to:
  * Create new `acos_beneficiaries` table
  * Copy associations from `beneficiaries` to `acos_beneficiaries`
  * Rename or remove old columns and constraints from `beneficiaries`
* Specifies columns in synthetic beneficiaries insert scripts
* Fixes Beneficiary preload on ACOBeneficiary `Find()` in api.go


db/migrations/20190314-fix-01.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE acos_beneficiaries (
+  aco_id uuid NOT NULL,
+  beneficiary_id int NOT NULL
+);
+
```
DeirdreHolub: Don't you want some constraints on this table? i.e., I'm assuming you don't want to allow duplicate (aco_id, beneficiary_id) tuples? and what about foreign keys constraints? maybe that's elsewhere?

knollfear: I think it is here.  I remembered seeing it in the initial code review
https://github.com/CMSgov/bcda-app/blob/cd413cc3abbdfc167c74e3cc47ba52db045d6493/bcda/models/models.go#L33

embh: Yep, what Knoll said. It's relying on GORM to add those constraints.

PR #206, opened 2019-03-18
BCDA-1017 use clientID and clientSecret for alpha plugin

alexhova:
### Fixes [BCDA-1017](https://jira.cms.gov/browse/BCDA-1017)
To make the alpha auth plugin more like okta, we should register a client with a clientID and clientSecret like we do in okta

### Proposed changes:
- add AlphaSecret to aco model
- hash alpha secret when registering and save to db
- return client ID and secret in Credentials from register client
- generate random 20 digit client secret when registering for alpha

### Feedback Requested
Does this meet the minimum requirements for opensbx?


bcda/auth/alpha.go
```diff
@@ -1,47 +1,55 @@
...
-	// use as our clientId for all the methods below. We could come up with yet another numbering scheme, or generate
-	// more UUIDs, but I can't see a benefit in that. Plus, we will know just looking at the DB that any aco
-	// whose client_id matches their UUID was created by this plugin.
-	return Credentials{ClientID: localID}, nil
+	hash := Hash{}
+	hashedSecret := hash.Generate(s)
+
+	db := database.GetGORMDbConnection()
+	defer database.Close(db)
+	aco.AlphaSecret = hashedSecret
```
DeirdreHolub: need to set ClientID here also, since it's now formally being used that way.

bcda/models/models.go
```diff
@@ -62,6 +62,7 @@ type ACO struct {
 	CMSID            *string   `gorm:"type:char(5)" json:"cms_id"`
 	Name             string    `json:"name"`
 	ClientID         string    `json:"client_id"`
+	AlphaSecret      string
```
DeirdreHolub: add a json spec for consistency with other fields: `json:"alpha_secret"`

bcda/auth/alpha_test.go
```diff
@@ -53,18 +53,14 @@ func (s *AlphaAuthPluginTestSuite) TestRegisterClient() {
 	c, err := s.p.RegisterClient(auth.KnownFixtureACO)
 	assert.Nil(s.T(), err)
 	assert.NotNil(s.T(), c)
+	assert.NotNil(s.T(), c.ClientSecret)
 	assert.Equal(s.T(), auth.KnownFixtureACO, c.ClientID)
 
 	c, err = s.p.RegisterClient("")
 	assert.NotNil(s.T(), err)
 	assert.Empty(s.T(), c.ClientID)
 	assert.Contains(s.T(), err.Error(), "provide a non-empty string")
```
DeirdreHolub: We should add a test for making sure we can successfully compare the hashed secret in the db to the one received in the result.

bcda/auth/alpha.go
```diff
@@ -1,47 +1,55 @@
...
 
-	if matched, err := regexp.MatchString("^[a-fA-F0-9]{8}(?:-[a-fA-F0-9]{4}){3}-[a-fA-F0-9]{12}$", localID); !matched || err != nil {
-		return Credentials{}, errors.New("expected a valid UUID string")
+	aco, err := getACOFromDB(localID)
+	if err != nil {
+		return Credentials{}, err
 	}
 
-	if _, err := getACOFromDB(localID); err != nil {
+	s, err := generateClientSecret()
```
DeirdreHolub: can't generate a secret if there's already one in the db

bcda/auth/alpha.go
alexhova: ✔️

bcda/models/models.go
alexhova: ✔️

bcda/auth/alpha.go
alexhova: oh wait.  it already is the clientID right?

PR #210, opened 2019-03-18
Dj/bcda 1025

DeirdreHolub:
### Fixes [BCDA-1025](https://jira.cms.gov/browse/BCDA-1025)
For tokens obtained through via the AccessToken() method, the following changes apply:
* JWT_EXPIRATION_DELTA now indicates in minutes, not hours, the ttl of a tokens issued.
* It is no longer possible to override the ttl of tokens at runtime. The ttl on tokens is set at server startup. It defaults to 60 minutes, but can be overridden by setting the env JWT_EXPIRATION_DELTA.
* access tokens are not saved in the database

Note that tokens previously obtained through the CLI should be unaffected by these changes.

Further note that these tokens are not obtainable until https://github.com/CMSgov/bcda-app/pull/204 is merged.

### Security Implications
Other than their shorter TTL, these tokens are identical to tokens previously issued.

### Acceptance Validation
All existing tests pass, and all tests added pass. All integration tests pass.

### Feedback Requested
Any.


karlmdavis:
> access tokens are not saved in the database

Just wondering how this interacts/plays with auditing requirements? Is any sort of identifying record/hash/ID saved or logged when an access token is issued? Should one be? (Genuinely don't know -- just curious.)

DeirdreHolub:
I amended only the message on the last commit because I thought it was not clear.

DeirdreHolub:
> > access tokens are not saved in the database
> 
> Just wondering how this interacts/plays with auditing requirements? Is any sort of identifying record/hash/ID saved or logged when an access token is issued? Should one be? (Genuinely don't know -- just curious.)

We plan on logging them.

bcda/api_test.go
```diff
@@ -135,7 +135,8 @@ func (s *APITestSuite) TestBulkEOBRequestUserDoesNotExist() {
 	acoID := "dbbd1ce1-ae24-435c-807d-ed45953077d3"
 	userID := "82503a18-bf3b-436d-ba7b-bae09b7ffdff"
 	tokenID := "665341c9-7d0c-4844-b66f-5910d9d0822f"
-	ad := auth.AuthData{ACOID: acoID, UserID: userID, TokenID: tokenID}
+	ad := auth.AuthData{ACOID: acoID, UserID: userID, TokenID: tokenID,}
```
alexhova: why this change?  idt its necessary unless it spans multiple lines

bcda/api_test.go
```diff
@@ -135,7 +135,8 @@ func (s *APITestSuite) TestBulkEOBRequestUserDoesNotExist() {
...
+
```
alexhova: extra blank line here

bcda/api_test.go
```diff
@@ -795,5 +796,5 @@ func TestAPITestSuite(t *testing.T) {
 }
 
 func makeContextValues(acoID string, userID string) (data auth.AuthData) {
-	return auth.AuthData{ACOID: acoID, UserID: userID, TokenID: uuid.NewRandom().String()}
+	return auth.AuthData{ACOID: acoID, UserID: userID, TokenID: uuid.NewRandom().String(),}
```
alexhova: same as above.  not sure its necessary

bcda/auth/alpha.go
```diff
@@ -35,14 +35,16 @@ func (p AlphaAuthPlugin) RegisterClient(localID string) (Credentials, error) {
...
-	hashedSecret := hash.Generate(s)
+	hashedSecret := NewHash(s)
 
 	db := database.GetGORMDbConnection()
 	defer database.Close(db)
 	aco.ClientID = localID
-	aco.AlphaSecret = hashedSecret
-	db.Save(&aco)
+	aco.AlphaSecret = hashedSecret.String()
+	err = db.Save(&aco).Error
```
alexhova: thank you for this!

bcda/auth/alpha.go
```diff
@@ -148,7 +150,27 @@ func (p AlphaAuthPlugin) RevokeClientCredentials(clientID string) error {
 	return nil
 }
 
-// generate a token for the ACO, either for a specified UserID or (if not provided) any user in the ACO
+// AcessToken manufactures an access token for the given credentials
```
alexhova: just to be complete, theres a minor typo here Acess to Access

bcda/auth/alpha.go
```diff
@@ -148,7 +150,27 @@ func (p AlphaAuthPlugin) RevokeClientCredentials(clientID string) error {
...
+func (p AlphaAuthPlugin) AccessToken(credentials Credentials) (string, error) {
+	if credentials.ClientSecret == "" || credentials.ClientID == "" {
+		return "", fmt.Errorf("missing or incomplete credentials")
+	}
+	aco, err := getACOByClientID(credentials.ClientID)
+	if err != nil {
+		return "", fmt.Errorf("invalid credentials; %s",err)
+	}
+	// when we have ClientSecret in ACO, adjust following line
```
alexhova: we do as of a few minutes ago!

bcda/auth/alpha.go
```diff
@@ -179,7 +201,8 @@ func (p AlphaAuthPlugin) RequestAccessToken(creds Credentials, ttl int) (Token,
 		userUUID = user.UUID
 		acoUUID = user.ACOID
 	} else {
-		aco, err := getACOFromDB(creds.ClientID)
+		var aco models.ACO
```
alexhova: why declare it this way?  just curious.  we don't seem to use the empty aco afterwards

bcda/auth/alpha_test.go
```diff
@@ -2,10 +2,11 @@ package auth_test
 
 import (
 	"fmt"
+	"regexp"
 	"testing"
 	"time"
 
-	jwt "github.com/dgrijalva/jwt-go"
+	"github.com/dgrijalva/jwt-go"
```
alexhova: hah thanks.  wont be doing that anymore :]

bcda/auth/alpha.go
DeirdreHolub: The aco is used on the next line? It's catching the return value from the function, which is not the typical pattern of how we get objects out of the db. I can't use := on the next line because err is actually declared at the top of the function. I decided against moving the decl of aco up because it's only used inside this else block. I suppose another option would have been to just shadow the err in the outer block, but shadowed vars == bad juju.

bcda/api_test.go
DeirdreHolub: You're right.

bcda/auth/alpha.go
```diff
@@ -148,12 +150,27 @@ func (p AlphaAuthPlugin) RevokeClientCredentials(clientID string) error {
...
 
-// Verify credentials (client ID and secret) and return access token
-func (p AlphaAuthPlugin) GetAccessToken(creds Credentials) (Token, error) {
-	return Token{}, fmt.Errorf("GetAccessToken not implemented in Alpha")
+// MakeAccessToken manufactures an access token for the given credentials
+func (p AlphaAuthPlugin) MakeAccessToken(credentials Credentials) (string, error) {
+	if credentials.ClientSecret == "" || credentials.ClientID == "" {
+		return "", fmt.Errorf("missing or incomplete credentials")
+	}
+	aco, err := getACOByClientID(credentials.ClientID)
```
alexhova: does this func exist?  or am i missing something?

alexhova: oh, it's defined in okta.go ...  how does this work?

DeirdreHolub: Yes. It's in auth/middleware.go.

alexhova: oh ok so its defined right here in okta.go :https://github.com/CMSgov/bcda-app/blob/d76da634cc9c77ae1386057b0690c7f2e773d7f2/bcda/auth/okta.go#L176

it doesn't need to be exported because we're using it in the same package.  should we put it in a more "shared" place since it is used by alpha _and_ okta?

PR #211, opened 2019-03-19
BCDA-1022 createAlphaToken generates a client ID and secret

alexhova:
### Fixes [BCDA-1022](https://jira.cms.gov/browse/BCDA-1022)

to make alpha more like okta, createAlphaToken needs to generate a client ID and secret to be used to request and access token

### Proposed changes:
- add ClientName to credentials for output of createAlphaToken to be like okta
- registerclient instead of creating a token in in the createAlphaToken cli command
- adjust tests to reflect new createAlphaToken functionality


bcda/main.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/CMSgov/bcda-app/bcda/monitoring"
 	"github.com/CMSgov/bcda-app/bcda/servicemux"
 
-	"github.com/bgentry/que-go"
+	que "github.com/bgentry/que-go"
```
DeirdreHolub: You need to fix your IDE so that it stops putting in these duplicate aliases. What's happening here is that the package is name que, but the file is not. That's perfectly legal in go.

alexhova: hmm yeah, i guess the point is is that when you use it, you call it "que".  so you can tell how it will be used from the import statement.  i'll find where the setting is and switch it off

alexhova: using goimports on save instead of goreturns solved it!

bcda/main_test.go
```diff
@@ -264,7 +245,7 @@ func (s *MainTestSuite) TestCreateAlphaTokenCLI() {
 	var outputPattern *regexp.Regexp
 	switch auth.GetProvider().(type) {
 	case auth.AlphaAuthPlugin:
```
DeirdreHolub: There shouldn't be any difference in the output patterns between the two plugins now. Please remove the switch.

alexhova: the clientID is not a uuid though i believe

alexhova: in okta that is

DeirdreHolub: Both plugins should be return client name, client id, and client secret now. I don't think we should be validating the client id's structure here; that kind of specific testing belongs in the plugin unit tests. We should validate here only that we got outputs for all three fields, and there is nothing unacceptable about those outputs.

DeirdreHolub:
> hold this guy until we have a convo with matt ... it will break the existing behavior

We had this convo, and we're all prepared.

PR #214, opened 2019-03-20
BCDA-534: Redact Bearer Token from URL Logs

embh:
### Fixes [BCDA-534](https://jira.cms.gov/browse/BCDA-534)
It's possible for Bearer tokens to end up in our Splunk logs if provided in the query string (malformed request).

### Proposed changes:
Replace `[string]` with `<redacted>` when `Bearer%20[string]` is found in a request URL before writing it to the log.

### Security Implications
This will prevent a security incident if a user places a Bearer token in the query string.

### Acceptance Validation
Includes tests and can be seen in log entries.
Screen Shot 2019-03-20 at 5 20 24 PM

### Feedback Requested
Any

### DIFF
--- a/bcda/logging/middleware.go
+++ b/bcda/logging/middleware.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"regexp"
+	"strings"
 	"time"
 
 	"github.com/CMSgov/bcda-app/bcda/auth"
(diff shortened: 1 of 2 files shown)


bcda/logging/middleware_test.go
```diff
@@ -188,6 +188,20 @@ func (s *LoggingMiddlewareTestSuite) TestPanic() {
 	os.Setenv("BCDA_REQUEST_LOG", reqLogPathOrig)
 }
 
+func (s *LoggingMiddlewareTestSuite) TestRedact() {
```
tbellj: nice test work

bcda/logging/middleware.go
```diff
@@ -90,3 +92,12 @@ func (l *StructuredLoggerEntry) Panic(v interface{}, stack []byte) {
 		"panic": fmt.Sprintf("%+v", v),
 	})
 }
+
+func Redact(uri string) string {
+	re := regexp.MustCompile(`Bearer%20([^&]+)(?:&|$)`)
+	submatches := re.FindAllStringSubmatch(uri, -1)
```
tbellj: is there a chance for this to return nil?

embh: There shouldn't be. If there are no matches, `FindAllStringSubmatch()` returns an empty slice, causing `Redact()` to return the original string. The third part of `TestRedact()` has an example of that, using a URL that doesn't have `Bearer` in it.

PR #215, opened 2019-03-22
BCDA-336: Check Blue Button backend certificate

embh:
### Fixes [BCDA-336](https://jira.cms.gov/browse/BCDA-336)
The Blue Button dev environment certificate can change. Currently we are checking the certificate (or not) based on URL, but using an environment variable instead would allow us to configure it by environment, regardless of BB URL.

### Proposed changes:
Add `BB_CHECK_CERT` environment variable that, when `false`, indicates that the Blue Button backend certificate should not be checked.

### Change Details
* Creates `BB_CHECK_CERT` that is used in bluebutton.go. If not explicitly set to false, the certificate will be checked.
* Adds tests for missing or invalid keypair and CA cert files

### Security Implications
By default, the Blue Button backend certificate will be checked. `BB_CHECK_CERT` needs to be set to `false` to skip adding the certificate to the Blue Button client TLS config.

### Acceptance Validation
Tests pass.

### Feedback Requested
Any.

### DIFF
--- a/bcda/client/bluebutton.go
+++ b/bcda/client/bluebutton.go
@@ -3,13 +3,15 @@ package client
 import (
 	"crypto/tls"
 	"crypto/x509"
-	"errors"
 	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
 	"strconv"
+	"strings"
 	"time"
+
+	"github.com/pkg/errors"
 
 	"github.com/CMSgov/bcda-app/bcda/monitoring"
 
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -44,6 +44,7 @@ services:
       - BB_CLIENT_CERT_FILE=../shared_files/bb-dev-test-cert.pem
       - BB_CLIENT_KEY_FILE=../shared_files/bb-dev-test-key.pem
       - BB_SERVER_LOCATION=https://fhir.backend.bluebutton.hhsdevcloud.us
+      - BB_CHECK_CERT=false
```
tbellj: is there a corresponding ops ticket for this environment variable?

tbellj: is BB_CHECK_CERT needed for both worker and api or just the worker container? Note: I know nothing about docker

bcda/client/bluebutton.go
```diff
@@ -52,23 +54,26 @@ func NewBlueButtonClient() (*BlueButtonClient, error) {
 	keyFile := os.Getenv("BB_CLIENT_KEY_FILE")
 	cert, err := tls.LoadX509KeyPair(certFile, keyFile)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "could not load Blue Button keypair")
```
tbellj: didn't know the wrap function existed. Nice!

docker-compose.yml
embh: Shouldn't need to set it for our AWS environments because we do want to check the BB certificate from those (which is the default)

embh: You're right! Should be worker only.

tbellj: makes sense

PR #216, opened 2019-03-25
Knollfear/bcda 991

knollfear:
### Fixes [BCDA-991](https://jira.cms.gov/browse/BCDA-991)

The snippets being used for decryption need to be tested as part of our normal test suite to ensure that they are compatible with our current code base on an ongoing basis.
### Proposed changes:

Add execution of decryption utils to our test suite
add necessary requirements to our docker image to all for the tests to run
### Security Implications

PII is not affected.  This container is not deployed anywhere.

### Acceptance Validation

I was able to verify that all 3 utils ran successfully locally.


test/smoke_test/bcda_client.go
```diff
@@ -203,22 +225,96 @@ func main() {
...
+							"--project", "../../decryption_utils/C#",
+							"decrypt.cs",
+							"--file", filename,
+							"--pk", privateKeyFile,
+							"--key", encryptedKey,
+						}
+
+						fmt.Println("Running the C# decryptor externally...")
+
+						// C# puts a bunch of nonsense in the file the first time it runs so this run is just ignored.
```
dhgreene: Does the extra content come 1) the first time it's run after installing, 2) the first time in a smoke test, or 3) the first of every pair of times it's running now?  I'd guess it's 1), and that running it once with an empty argument list after invoking the installation script would be sufficient.

knollfear: It is 1.  I was trying to run it earlier in the process, but I got some non zero return codes and it aborted.  It def adds a few seconds to the run, but it shouldn't really be a problem to do this twice.

dhgreene:
I also get 17 new files in the `decryption_utils/C#/` path when running the smoke test.  We should either add this folder to `.gitignore` or check in the files.

knollfear:
> I also get 17 new files in the `decryption_utils/C#/` path when running the smoke test. We should either add this folder to `.gitignore` or check in the files.

updated gitignore and pushed.  I think it worked, would appreciate a test to confirm

dhgreene:
> > I also get 17 new files in the `decryption_utils/C#/` path when running the smoke test. We should either add this folder to `.gitignore` or check in the files.
> 
> updated gitignore and pushed. I think it worked, would appreciate a test to confirm

👍 That takes care of it.

test/smoke_test/smoke_test.sh
```diff
@@ -3,10 +3,15 @@
...
-go run bcda_client.go decrypt_util.go -host=api:3000 -endpoint=Coverage -encrypt=false
\ No newline at end of file
+echo $PATH
+echo "Running Coverage Encrypted"
+go run bcda_client.go -host=api:3000 -endpoint=Coverage
+echo "Running EOB Encrypted"
+go run bcda_client.go -host=api:3000 -endpoint=ExplanationOfBenefit
+echo "Running Patient Encrypted"
+go run bcda_client.go -host=api:3000 -endpoint=Patient
+echo "Running Coverage Encrypted"
```
msnook: Why run Coverage encrypted twice?

test/smoke_test/smoke_test.sh
```diff
@@ -3,10 +3,15 @@
...
+go run bcda_client.go -host=api:3000 -endpoint=Coverage
+go run bcda_client.go -host=api:3000 -endpoint=ExplanationOfBenefit -encrypt=false
```
msnook: Any interest in echoing the unencrypted tests, similar to what you did above?

test/smoke_test/bcda_client.go
```diff
@@ -203,22 +225,96 @@ func main() {
...
+						}
+						fmt.Println("Running the Python decryptor externally...")
+
+						// #nosec
+						if cmdOut, err = exec.Command(cmdName, cmdArgs...).Output(); err != nil {
+							fmt.Fprintln(os.Stderr, "There was an error running the Python decryption util: ", err)
+							os.Exit(1)
+						}
+
+						output = string(cmdOut)
```
msnook: Can we write it to file instead?  I think that will be the standard use-case by clients.

test/smoke_test/bcda_client.go
```diff
@@ -145,6 +146,27 @@ func isValidNDJSONFile(filename string) bool {
 	return isValid
 }
 
+func isValidNDJSONText(data string) bool {
```
msnook: is this function necessary?  can't we just reuse `isValidNDJSONFile`?

test/smoke_test/smoke_test.sh
knollfear: This was a debug thing.  I removed it.

test/smoke_test/smoke_test.sh
knollfear: added.

knollfear:
> The removal of the `decrypt_util` file is going to break the smoke test in Jenkins. I think we're going to need an accompanying PR for bcda-ops.
> Take a look here: https://github.com/CMSgov/bcda-ops/blob/master/jenkins_files/Jenkinsfile.smoke_test#L99-L100

I will fix this and submit a PR shortly.

test/smoke_test/bcda_client.go
knollfear: no, because the output isn't a file.

test/smoke_test/bcda_client.go
knollfear: except that it isn't.  The enryption examples all pipe the output to a file, none of the utils create a file.  If the utils start making files we can adjust this to use files.

msnook: Fair enough.  I see this now in the examples:
```
> decrypted_output.ndjson
```

It is, however, a change to how the existing smoke test executes.  I have found it valuable to have the decrypted file be spit out as part of the test execution so that debugging can occur on that decrypted file if a test fails format validation (or if we just want to look at the content).  If we write to string in memory, validate it, and then move on - then we have no output from the test to look at if we need to debug.
Would it make sense to pipe it to a file within the test client so that we have the results of the decryption in case we need to debug it?  
`decrypted_output_go.ndjson`
`decrypted_output_python.ndjson`
`decrypted_output_c.ndjson`

knollfear: In testing I was able to do fmt.Println(output) when I needed to see the files.  I don't really know how we would even start to get at the files in Travis or jenkins if we needed to see them.  Putting the output in the logs is actually the easiest to troubleshoot.

msnook: But we currently aren't doing `fmt.Println(output)` (and it would really clutter the logs if we did).  So if something fails in the Jenkins pipeline, we have no way of debugging.  
However, if a file is created, we can go onto the EC2 instance and look at it.  (This technique was used heavily when troubleshooting BB DPR integration).  In the legacy decryption mechanism, we are controlling where it is written here: https://github.com/CMSgov/bcda-app/blob/master/test/smoke_test/decrypt_util.go#L60

knollfear: I added in code to print the first bad line it encounters.  This will give us a starting point for debugging.

PR #218, opened 2019-03-27
BCDA-1032 implement delete client func for alpha

alexhova:
### Fixes [BCDA-1032](https://jira.cms.gov/browse/BCDA-1032)
We need a way to delete a client in the alpha auth plugin so that Eric and Jason can sleep at night.

### Proposed changes:
- implement  the DeleteClient() func for alpha

### Change Details
Since registering a client in alpha is really just generating a clientID and an alphaSecret on the ACO model, to delete the client, we will simply set these values on the db record for the ACO back to an empty string

### Feedback Requested
Anything you can think of!

### DIFF
--- a/bcda/auth/models.go
+++ b/bcda/auth/models.go
@@ -1,6 +1,8 @@
 package auth
 
 import (
+	"errors"
+
 	"github.com/jinzhu/gorm"
 	_ "github.com/jinzhu/gorm/dialects/postgres"
 	"github.com/pborman/uuid"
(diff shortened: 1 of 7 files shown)


bcda/auth/models.go
```diff
@@ -49,3 +51,17 @@ func (t *Token) AfterFind() error {
 	}
 	return err
 }
+
+func GetACOByClientID(clientID string) (models.ACO, error) {
```
dhgreene: I like that this is moved to a plugin-neutral location.

alexhova: thanks, i do too.  i figured since it is a db select, the models would be a good place for it

PR #219, opened 2019-03-28
BCDA-923: Chunk Files and Process them Concurrently

knollfear:
### Fixes [BCDA-923](https://jira.cms.gov/browse/BCDA-923)

We should be processing files from BB in multiple jobs at the same time.  This solves a ton of problems with large files, encryption and speed.
### Proposed changes:

refactor how jobs get enqueued to allow for variable sized jobs controlled by an environment variable.

### Security Implications

PII should not be affected

### Acceptance Validation

Tests complete
File downloads still work
on dev/test/sandbox it still needs to work as I refactored a query that previously took too long to complete

### Feedback Requested

any


bcda/api.go
```diff
@@ -149,67 +148,43 @@ func bulkRequest(t string, w http.ResponseWriter, r *http.Request) {
...
-		}
-		beneficiaryIDs = append(beneficiaryIDs, beneficiary.BlueButtonID)
-	}
-
 	// TODO: this checks for ?encrypt=false appended to the bulk data request URL
 	// By default, our encryption process is enabled but for now we are giving users the ability to turn
 	// it off
 	// Eventually, we will remove the ability for users to turn it off and it will remain on always
-	var encrypt = true
+	var encrypt bool = true
```
embh: `bool` is unnecessary because it's implied by the value. I think it used to be there, but VSCode complained so I removed it.

bcda/models/models.go
```diff
@@ -40,14 +45,93 @@ func InitializeGormModels() *gorm.DB {
...
+	if job.Status == "Completed" {
+		return true, nil
+	}
+	db := database.GetGORMDbConnection()
+	defer database.Close(db)
+
+	var completedJobs int64
+	db.Model(&JobKey{}).Where("job_id = ?", job.ID).Count(&completedJobs)
+
+	if int(completedJobs) >= job.JobCount {
```
embh: Is it possible for there to be more completed jobs than the count?

bcda/models/models.go
```diff
@@ -40,14 +45,93 @@ func InitializeGormModels() *gorm.DB {
...
+
+		staging := fmt.Sprintf("%s/%d", os.Getenv("FHIR_STAGING_DIR"), job.ID)
+		err := os.Remove(staging)
```
embh: `CheckCompleted` sounds like it's looking to see if the job is done, so it's not clear that there would be actions taken here like removing the staging directory. Could we rename to make that apparent, or make it a separate function?

bcda/models/models.go
```diff
@@ -67,6 +151,22 @@ type ACO struct {
 	ACOBeneficiaries []*ACOBeneficiary
 }
 
+func (aco *ACO) GetBeneficiaryIDs() (beneficiaryIDs []string, err error) {
+	db := database.GetGORMDbConnection()
+	defer database.Close(db)
+
+	beneficiaryJoin := "inner join beneficiaries on acos_beneficiaries.beneficiary_id = beneficiaries.id"
```
embh: 🎉 So much better than doing a zillion queries!

bcda/models/models.go
```diff
@@ -67,6 +151,22 @@ type ACO struct {
...
+	if err = db.Table("acos_beneficiaries").Joins(beneficiaryJoin).Where("acos_beneficiaries.aco_id = ?", aco.UUID).Pluck("beneficiaries.blue_button_id", &beneficiaryIDs).Error; err != nil {
+		log.Errorf("Error retrieving ACO-beneficiaries for ACO ID %s: %s", aco.UUID.String(), err.Error())
+		return nil, err
+	} else if len(beneficiaryIDs) == 0 {
+		log.Errorf("Retrieved 0 ACO-beneficiaries for ACO ID %s", aco.UUID.String())
+		return nil, fmt.Errorf("Retrieved 0 ACO-beneficiaries for ACO ID %s", aco.UUID.String())
```
embh: For the error being returned, not capitalizing is the standard practice https://github.com/golang/go/wiki/CodeReviewComments#error-strings (makes sense to capitalize in the previous line for the log message though)

bcda/testConstants/constants.go
```diff
@@ -0,0 +1,6 @@
+package testConstants
+
+const DEVACOUUID = "0c527d2e-2e8a-4808-b11d-0fa06baf8254"
+const SMALLACOUUID = "3461C774-B48F-11E8-96F8-529269fb1459"
+const MEDIUMACOUUID = "C74C008D-42F8-4ED9-BF88-CEE659C7F692"
+const LARGEACOUUID = "8D80925A-027E-43DD-8AED-9A501CC4CD91"
```
msnook: Very necessary.  Thank you for doing that.

bcda/models/models.go
knollfear: This should still be idempotent.  The staging dir it is deleting was housing all of the unencrypted BB files for just this job.  They should have all been moved at this point and this is just housekeeping.  It HAS to happen so I just wanted to make sure it was done in the easiest way to not mess up.

bcda/models/models.go
knollfear: I don't think so, but I get nervous about 0 vs 1 indexed things and I didn't want to get into a weird permanent pending status if they did have more.

bcda/models/models.go
knollfear: will fix in incoming commit

bcda/testConstants/constants.go
knollfear: I just couldn't keep track of it all in my head.  We probably need a ticket to actually start using all of these.

bcda/api.go
knollfear: will remove in incoming commit.

bcda/models/models.go
```diff
@@ -40,14 +45,93 @@ func InitializeGormModels() *gorm.DB {
...
+	}
+
+	return false, nil
+}
+
+func (job *Job) GetEnqueJobs(encrypt bool, t string) (enqueJobs []*que.Job, err error) {
+	var jobIDs []string
+
+	var rowCount = 0
+	maxBeneficiaries := utils.GetEnvInt("BCDA_FHIR_MAX_RECORDS", BCDA_FHIR_MAX_RECORDS_DEFAULT)
```
karlmdavis: Am I following correctly, that you assume (& enforce) that each ACO will have <= 10K beneficiaries?

If so, why?

knollfear: I do not.  I enforce that I will que up a new internal job when I get over 10K beneficiaries.  If an ACO has 55,000 benes I will que up 5 jobs with 10K benes and one more with 5,000 benes.  This is our attempt to parallelize things while not taking Blue Button down.  The que has a max simultaneous workers which will be limited by how many connections BB can handle simultaneously.

karlmdavis: Cool!

Some questions:

1. Does each job just run a single request to the backend at a time, or does it split out the requests over goroutines or some such?
2. What's the current default for max concurrent workers?
3. Are workers 1:1 with EC2 instances, or something else?

knollfear: 1. This seemed like an easier task than doing go routines because it achieves the same end result with a lot less work by leveraging an existing worker queue.  each Queue job will run through the bene list sequentially until it completes.  When all the Queue jobs are done the Job is done and ready to be downloaded by the end user.
2. I think it is 2, but the expectation is that it will be much higher than that in production.  there is a var called "WORKER_POOL_SIZE" that controls it
3. There should be many workers per EC2 instance.

bcda/main.go
```diff
@@ -445,15 +435,9 @@ func createAlphaToken(ttl int, acoSize string) (string, error) {
 	return msg, nil
 }
 
+// This has 100+ references to refactor
```
embh: Thought I commented on this in my earlier review, but must not have saved it. I see only ~10 calls for `getEnvInt()`. Moving it altogether would be preferable to keeping this old one around as a wrapper.

bcda/models/models.go
karlmdavis: What job primitive are workers run in then? Separate process per worker? Separate thread? Separate goroutine?

bcda/models/models.go
embh: Understood, and idempotence isn't in question. Functions shouldn't have side effects. Could the staging directory deletion be a separate function, called when this one returns `true`? That would also make it possible to test the behavior of job completion and file cleanup separately.

bcda/main.go
knollfear: ok.  I will toss in a ticket for it in this sprint if that is ok.  I think my search included unit test results which was obviously not extra place I need to change this.

bcda/models/models.go
knollfear: Returning True just means that the job is completed, it could have been completed previously or it could be marked as completed just now.  

This is just a task that 100% has to happen when the job is finally marked as complete.  If I split it up I am concerned that the next time someone uses it they won't know about this requirement and will choose not to do it.  By keeping it here they can't make that mistake.  The dir it is deleting is scoped only to this job and won't affect any other jobs.

The way I see it if I close my front door I need to lock it.  It isn't two things, it is one thing close the door and lock it.  A function called CloseDoor() ends in a closed and locked door and I can't end up leaving the door unlocked.

I discussed with Ryan in my 1 on 1 today and he seemed fine with it.

bcda/models/models.go
knollfear: The workers are running using QueueGO which is a Go port of Ruby Queue.  I don't have a great understanding of the internals, I just know it queues jobs, limits how many jobs can run simultaneously and won't fall apart if the system goes down.  


    How many EC2 instances do I have to run workers on? 
       Not sure.  I think this isn't a static value and is subject to auto scaling.  I think each EC2 instance has a fixed number of workers.
    How many user jobs/requests will I allow to be worked at once?
    a Job is what our end user experiences, a queueJob is what I am adding.  Each worker thread will run one queueJob at a time.  This could be one Job or a different job for each worker.  The queue works a queueJob until it is done and then grabs another one.  The size of the worker pool limits how many queueJobs will be worked simultaneously which limits the load on BB.
    
If payloads are split into multiple files/chunks, what is the max size of those chunks?

one queueJob makes one file.  By limiting the number of benes the file size is inherently limited and we don't have any concurrency issues with file writes.  It would be theoretically possible for the 10,000 benes with the most claims to all end up in the same file chunk and the file would be rather large, but I feel that at 10,000 the law of averages has to kick in and file sizes will be within a small range of each other.  To tune it we can alter the number of benes per file if the file sizes are too big or too small.
    How many requests to the backend Data Server will those EC2 instances collectively be running concurrently?  This is the number of worker threads.  We can increase or decrease this number easily as needs demand.

If you want to have a call or a meeting on this I am happy to organize one.

knollfear: I think ultimately the management is actually pretty simple and needs to satisfy only 2 constraints; how many benes can I put in a file before it starts to become "too big" (limits of encryption and file download times are limiting factors) and how many simultaneous calls to BB can I make before I break it.  

The max bene parameter will control how big the files get because barring some unexpected statistical bias 10,000 benes in a file be approximately 10,000 x the average file size per bene.  The goal is not to even come in the neighborhood of the limits of encryption file size but to have manageable sized files for our users to get them in a reasonable amount of time and not have to ever restart a 1 hour download from scratch.  10,000 is definitely just a guess, but it puts an upper limit on the number of files at around 20 per ACO (biggest ACO is +-200,000 benes afaik).  If these files are all pushing 64 GB we have a much different problem than any of us realize.  

The max requests to BB is determined by the size of the worker pool.  If we have 100 workers that is 100 max connections to BB; if we have 10,000 workers that is 10,000 connections.  If we can have 10,000 connections at a time, but only have 100 queuejobs cooking at a time we are not running in an optimized way, but this is going to be an order of magnitude faster than what we have now and has the added benefit of not writing a ton of complicated threading code ourselves.

karlmdavis: Couple follow-up questions:

1. What controls how many worker EC2 instances are running?
2. Is `WORKER_POOL_SIZE` applied per worker EC2 instance, or globally?

I'm wondering if autoscaling is going to come in Kool-Aid Man style and suddenly double or triple the `WORKER_POOL_SIZE`, sending way more connections to the backend than expected.

msnook: Hi Karl,
1) For the worker EC2 instances, we currently have it configured in [this manner](https://github.com/CMSgov/bcda-ops/blob/master/terraform/opensbx/main.tf#L646-L648) (2 desired, 2 min, 2 max).  We scale up and down based on various auto-scaling policies defined [here](https://github.com/CMSgov/bcda-ops/blob/master/terraform/modules/asg/main.tf#L120).  Obviously these are tweakable as we learn more about handling of load.
2) `WORKER_POOL_SIZE` is applied globally across each EC2 instance that spins up. The default is 2 (for now), but obviously tweakable.

bcda/main.go
knollfear: This is in as BCDA-1128

bcda/models/models.go
knollfear: I looked in my copy of "Clean Code" because I think we are both directly arguing for points he makes in that book and this is a real judgement call.  Chapter 3 (Functions) says functions should "Do One Thing" and "Have No Side Effects".

> So, another way to know that a function is doing more than 'one thing' is if you can extract another function from it with a name that is not merely a restatement of its implementation.

  I would state that the abstraction of checkComplete()  is this:
`Check if a job is in a completed status and if it was not previously in a completed status clean up the job's staging directory.`
Any further refactoring would, in my opinion, unnecessarily complicate things and force the caller of the function to take important steps that might be missed

>Side Effects are lies.  Your function promises to do one thing, but it also does other hidden things.  Sometimes it will make unexpected changes to the variables of its own class.  Sometimes it will make them to parameters passed into the function or to system globals.  In either case they are devious and damaging mistruths that often result in strange temporal couplings and order dependencies

I feel that the side effects of this function are sufficiently tied to the job in question that they do not have any global effects and will not cause any temporal couplings or dependencies.  The creation of a job key only happens after the file in question has been encrypted and moved out of the staging directory.  As such, when this function decides to delete the staging directory it should be deleting an empty directory and if it isn't, we have solved a problem (storing unencrypted BB data) that we didn't know we had in the first place.

bcda/models/models.go
knollfear: I think it is 4 max, but with the worker pool limited across instance I don't think it will matter too much because spinning up another instance won't increase the number of running tasks.   I think that brings up a different point of what auto scaling will accomplish and what problems it will solve, but that is a discussion for another time.

knollfear: I made a tech spec back in December for this and updated it recently with a new diagram if you want to take a look at that it might answer some questions.

https://confluence.cms.gov/display/BCDA/Tech+Spec%3A+Worker

knollfear:
> This looks good. A lot going on - so I would encourage that you deploy your feature branch to `dev` and make sure it can pass all the tests prior to merging to `master`.

I deployed to dev and successfully ran job 6683 with the expected results.

bcda/models/models.go
karlmdavis: > WORKER_POOL_SIZE is applied globally across each EC2 instance that spins up. The default is 2 (for now), but obviously tweakable.

If it's constrained globally, that covers my main concern. Out of curiosity, what's enforcing/synchronizing that constraint globally?

msnook: @karlmdavis To clarify, the global definition of `WORKER_POOL_SIZE` is the value that each EC2 instance will inherit.  So, if our `WORKER_POOL_SIZE` is set to 2, then each EC2 instance that spins up will have a worker pool size of 2.

karlmdavis: > if our WORKER_POOL_SIZE is set to 2, then each EC2 instance that spins up will have a worker pool size of 2

Okay, then we're back to me being a bit concerned. Some background:

* In order to produce data in a timely fashion, BCDA will likely want to have hundreds of concurrent requests going. Certainly 150, and I would not be at all surprised to see that eventually scale up to 1000.
* Each of those requests is super lightweight: if we can support serving it with just a couple of servers (and we can) then you definitely shouldn't need more hardware than that as a client.
* During autoscaling events (e.g. deploys, health check failures, scaling up if you ever add a policy for that, etc.) your number of workers could easily go to 150% of baseline, and perhaps sometimes 200%.

That strikes me as a risky design. Given that it's your design, it presumably doesn't seem like a bad idea to you, though. Can you help me understand why you don't find that a bit concerning?

msnook: Hi @karlmdavis, kindly requesting that we move the design discussion to this Tech Spec:
https://confluence.cms.gov/display/BCDA/Tech+Spec%3A+Worker

I have populated the comments section of the doc above with some information that might be helpful.

bcda/models/models.go
embh: My other suggestion was to rename the function. It sounds like this logic is all closely tied together, so can we call it something that will more clearly imply what it does?

knollfear: renamed to CheckJobCompletedAndCleanup

PR #221, opened 2019-03-29
removing getEnvInt and putting it in utils

knollfear:
### Fixes [BCDA-1128](https://jira.cms.gov/browse/BCDA-1128)

getEnvInt was in main and would be useful to the rest of the system.  I moved it into utils so it could be used in other parts of the software.
### Proposed changes:
Remove getEnvInt from main and move it into utils

### Change Details

exactly what was described
### Security Implications

none

### Acceptance Validation

tests pass.  This should be a no op change.

### Feedback Requested

any


bcda/testUtils/utils.go
```diff
@@ -140,3 +141,14 @@ func RandomBase64(n int) string {
 	}
 	return base64.StdEncoding.EncodeToString(b)
 }
+
+func GetEnvInt(varName string, defaultVal int) int {
```
embh: Is this meant to be here _and_ in utils/common.go?

knollfear: no. I moved it to testutils at first by mistake.  It should just be in utils now.

embh:
Is this ready to be reviewed? The description looks like it might be in progress still...?

bcda/testUtils/utils.go
knollfear: > Is this ready to be reviewed? The description looks like it might be in progress still...?

This sb gtg now.

PR #222, opened 2019-04-01
BCDA 1036 Maintenance: Remove unused alpha auth functions and tests

dhgreene:
### Addresses [BCDA-1036](https://jira.cms.gov/browse/BCDA-1036)
At this stage of refactoring alpha auth, the following functions are unneeded and/or implemented in a way that we do not intend to use them:

* `GenerateClientCredentials()`
* `RevokeClientCredentials()`
* `RevokeAccessToken()`

### Proposed changes:
* Remove these functions and tests that reference them.
* Add a comment that addresses a security concern.

### Security Implications
No change is expected to the security of the app.

### Feedback Requested
Are new tests required to replace any of the deleted tests?


bcda/auth/alpha_test.go
```diff
@@ -101,78 +99,6 @@ func (s *AlphaAuthPluginTestSuite) TestDeleteClient() {
...
-
-	// quick and dirty register client
-	aco.ClientID = aco.UUID.String()
-	err = connections["TestGenerateClientCredentials"].Save(&aco).Error
-	assert.Nil(s.T(), err, "wtf? %v", err)
-	email := fmt.Sprintf("%s@genclientcredstest.com", testUtils.RandomHexID())
-	user, err := models.CreateUser("Fake User", email, aco.UUID)
-	assert.Nil(s.T(), err, "wtf? %v", err)
-
-	r, err = s.p.GenerateClientCredentials(clientID, 0)
```
knollfear: As long as we are going to leave in the method signatures I think we should be running the tests on them.  I think this line can stay and just check that an error comes back.

bcda/auth/alpha_test.go
```diff
@@ -101,78 +99,6 @@ func (s *AlphaAuthPluginTestSuite) TestDeleteClient() {
...
-
-	clientID := user.ACOID.String()
-	_, err := s.p.GenerateClientCredentials(clientID, 0)
-	if err != nil {
-		assert.FailNow(s.T(), fmt.Sprintf(`can't create client credentials for %s because %s`, user.ACOID.String(), err))
-	}
-
-	assert := assert.New(s.T())
-
-	err = s.p.RevokeClientCredentials(aco.ClientID)
```
knollfear: same idea here as above.  Just run this and verify that the error comes back.  That way if the code gets reverted by mistake we might catch it in this test.

bcda/main_test.go
```diff
@@ -552,33 +552,15 @@ func (s *MainTestSuite) TestRevokeToken() {
...
-	assert.Nil(err)
-
 	buf := new(bytes.Buffer)
 	s.testApp.Writer = buf
 
 	// Negative case - attempt to revoke a token passing in a blank token string
 	args := []string{"bcda", "revoke-token", "--access-token", ""}
-	err = s.testApp.Run(args)
+	err := s.testApp.Run(args)
 	assert.Equal("Access token (--access-token) must be provided", err.Error())
```
knollfear: Should this CLI just return an error directly or possibly be removed?  I am not sure why we would run just this test and not just erase the whole CLI.  If we do leave it, I think we should ensure that it returns the expected errors.

Name:     "revoke-token",
			Category: "Authentication tools",
			Usage:    "Revoke an access token",

dhgreene: The CLI should continue to function with Okta as the provider, but `main_test.go` isn't the right place to require Okta-specific environmental variables to be set.  Also, later on we should be re-implementing at least this feature for Alpha auth.

knollfear: That makes sense.  But this test sets the provider to alpha, what is the expected behavior if someone runs this CLI with the alpha auth provider?  Can we make sure we get the expected result, even if it is an error?

PR #223, opened 2019-04-03
BCDA-921 Feature: Ingest of cclf9 data into db

tbellj:
Fixes [BCDA-921](https://jira.cms.gov/browse/BCDA-921)

Problem:
Previously had no way of storing data from a file ingest. This moves us in the right direction for receiving cclf files.

Proposed changes:
bcda/cli.go - saves ingest data to the database
bcda/models/models.go - added the new cclf9 data model

Security Implications
This code will be ingesting PII/PHI into our DB. Removal of these source files will be critical and implemented in ticket: BCDA-1159

Acceptance Validation
Yes, tested and got the right results

image

Feedback Requested
Any and all


bcda/models/models.go
```diff
@@ -80,10 +81,22 @@ type ACOBeneficiary struct {
...
+	PrevNum      string `json:"previous_number"`
+	PrevsEfctDt  string `json:"effective_date"`
+	PrevsObsltDt string `json:"obsolete_date"`
+}
+
 func (*ACOBeneficiary) TableName() string {
 	return "acos_beneficiaries"
 }
 
+func (*CCLF9) TableName() string {
```
knollfear: Is this function used?  Should we be doing this on other models?

bcda/cli.go
```diff
@@ -113,14 +117,19 @@ func importCCLF9(filePath string) error {
 	for sc.Scan() {
 		b := sc.Bytes()
 		if len(bytes.TrimSpace(b)) > 0 {
-			fmt.Printf("\nXREF: %s\n", b[0:1])
-			fmt.Printf("Current identifier: %s\n", b[currIDStart:currIDEnd])
-			fmt.Printf("Previous identifier: %s\n", b[prevIDStart:prevIDEnd])
-			fmt.Printf("Previous identifier effective date: %s\n", b[prevIDEffDateStart:prevIDEffDateEnd])
-			fmt.Printf("Previous identifier obsolete date: %s\n", b[prevIDObsDateStart:prevIDObsDateEnd])
+
+			cclf9 := models.CCLF9{
```
msnook: Do we need the xref so that we know if the number is an MBI or HICN?

bcda/models/models.go
```diff
@@ -80,10 +81,22 @@ type ACOBeneficiary struct {
 	// Join model needed for additional fields later, e.g., AttributionDate
 }
 
+type CCLF9 struct {
```
msnook: Do we need the xref so that we know if the number is an MBI or HICN?

bcda/models/models.go
```diff
@@ -80,10 +81,22 @@ type ACOBeneficiary struct {
...
+	gorm.Model
+	CurrentNum   string `json:"current_number"`
```
msnook: The CCLF8 model includes some metadata... I wonder if we need to be storing that here too?
1) ACO ID - do we need to tie this data back to an ACO?
2) File date
3) File performance year
4) Imported date/created at

bcda/cli.go
```diff
@@ -113,14 +117,19 @@ func importCCLF9(filePath string) error {
...
+				CurrentNum:   string(b[currIDStart:currIDEnd]),
+				PrevNum:      string(b[prevIDStart:prevIDEnd]),
+				PrevsEfctDt:  string(b[prevIDEffDateStart:prevIDEffDateEnd]),
+				PrevsObsltDt: string(b[prevIDObsDateStart:prevIDObsDateEnd]),
+			}
+			err = db.Create(&cclf9).Error
+			if err != nil {
+				return err
+			}
```
msnook: We'll need to clean the files up, but not sure if we want to handle that here (function side effect) or handle it via a separate call.  I wonder if @embh has thought about this for CCLF9?

bcda/models/models.go
tbellj: not sure about other models but gorm kept screwing up the table name by putting in an underscore

bcda/models/models.go
tbellj: yessss

bcda/models/models.go
tbellj: not so sure about the other stuff, but we get the created at and deleted at for free in gorm.

bcda/cli.go
tbellj: Yeah a general cleanup should happen for all the cclf files, I say a separate function to handle   the cleanup of everything.

bcda/models/models.go
embh: Without it I'm pretty sure it tries to pluralize automatically, so it wasn't getting this one right

bcda/cli.go
msnook: OK, I think we'll need a separate ticket for this.  Can you create one?

tbellj: yessir

bcda/models/models.go
knollfear: that is a DB level created at though.  The CCLF9 has its own date that I think people would be interested in knowing.

tbellj: are u referring to the previous effective and obsolete dates that come in that file? Or the actual delivery date of that file?

knollfear: I think the date embedded in the name of the file.

bcda/cli.go
```diff
@@ -82,6 +83,9 @@ func importCCLF8(filePath string) error {
 }
 
 func importCCLF9(filePath string) error {
```
msnook: Wondering if this will work for us: 
https://github.com/CMSgov/bcda-app/blob/54a1de48cb7263e31d5b1e570019a3a5fdcea2e6/bcda/main.go#L304-L331

Basically, this requires that the caller (for sake of conversation, lets say its Jenkins on a cron job) knows the name of the file to be ingested.  I'm not sure that is going to be feasible, mainly because we don't want Jenkins (or any caller really) to have access to the directory where these files will live - so how will they know the file name?  
I wonder if we need to consider a single CLI command called something like `importCCLF` which accepts a directory path as an argument.  Internally, that function will iterate through all the files in the dir (could be 0 files), determine if each file is cclf0/8/9, and then call the internal functions we have written to process 0, 8, and 9.  Basically, processing all files it finds instead of just a single file at a time.

Curious to hear thoughts from @embh and @knollfear on this.

embh: I like the idea of passing a directory path to a general CCLF import command. The imports of all three files are related, with CCLF0 for validating the rows in the others, and the application could clean up all the files together (or at least indicate that they're ready for cleanup, if we decide to make that a separate job). Making a new command to check a directory and determine which CCLF # each file is sounds like it might be worthy of a new ticket.

tbellj: good idea. If it's implemented in this ticket or a new one, doesn't matter to me :)

msnook: I created this ticket https://jira.cms.gov/browse/BCDA-1184

bcda/cli_test.go
```diff
@@ -150,6 +153,15 @@ func (s *CLITestSuite) TestImportCCLF9() {
 	args = []string{"bcda", "import-cclf9", "--file", "../shared_files/cclf/T.A0001.ACO.ZC9Y18.D181120.T1000010"}
```
msnook: Can we have a test case for each file type?  Or does this test handle all scenarios?  
From the spec:
```The filename convention for the Medicare Shared Savings Program in Table 9 is P.A****.ACO.ZC9Y**.Dyymmdd.Thhmmsst.
The filename convention for the NGACO Model in Table 9 is P.V***.ACO.ZC9.Dyymmdd.Thhmmsst.
The filename convention for the CEC Model in Table 9 is P.CEC.ZC9.Dyymmdd.Thhmmsst.
The filename convention for the VT APM Model in Table 9 is P.F***.ACO.ZC9.Dyymmdd.Thhmmsst```

File: bcda/cli.go [hunk 8]
```diff
@@ -113,14 +117,20 @@ func importCCLF9(filePath string) error {
 	for sc.Scan() {
 		b := sc.Bytes()
 		if len(bytes.TrimSpace(b)) > 0 {
-			fmt.Printf("\nXREF: %s\n", b[0:1])
-			fmt.Printf("Current identifier: %s\n", b[currIDStart:currIDEnd])
-			fmt.Printf("Previous identifier: %s\n", b[prevIDStart:prevIDEnd])
-			fmt.Printf("Previous identifier effective date: %s\n", b[prevIDEffDateStart:prevIDEffDateEnd])
-			fmt.Printf("Previous identifier obsolete date: %s\n", b[prevIDObsDateStart:prevIDObsDateEnd])
+
+			cclf9 := models.CCLF9{
```
msnook: Do the IDs need to be associated with an ACO?  I think we can get the ACO ID from the filename.  Is there value in storing it?

bcda/cli_test.go
knollfear: This is a good flag.  However, I think the place for this test is in getCCLFFileMetaData below.  That has a regex that validates the file name.

embh: I thought BCDA was working with SSP ACOs only, at least at first, so we shouldn't need to test filename formats for the other models right now. @epfrankle , can you confirm?

knollfear: I also think more restrictive is better here.  If we did for some reason get a CCLFX file for a non SSP ACO we know we shouldn't have it and we should not import it.

epfrankle: Confirmed that we are working with SSP ACOs to begin, and also want to leave ourselves available to work with other ACOs in the future.

msnook:
@tbellj Can you update the `Security Implications` section to clearly indicate that this processing will be ingesting PII/PHI into the DB.

bcda/cli.go
tbellj: I can store it from the filename no problem. Have to think on what this means...but will add it if this is what we want.

msnook: Trying to think about what this means.  Will we be able to do what we need to do if the ACO ID is _not_ associated with the ID references which have changed?

msnook: ^ I think this is the only open action.  Do we need ACO ID in the DB?  It can't hurt, right?  @embh @knollfear Thoughts?

embh: I can't think of how we'd use the ACO ID, because the beneficiaries' identifiers should be considered changed across the board, not just in the context of that ACO. Since we would be able to get it later from the filename, not sure it's worth having separated out. But it can't hurt. We will be saving ACO ID in its own column for CCLF8 files, if that makes any difference.

tbellj: yeah to what @embh said, the aco id would go "wherever we decided for the overall ingest" but we could separate it out if that's what we want to do.

bcda/cli.go
```diff
@@ -108,15 +111,20 @@ func importCCLF0(fileMetadata cclfFileMetadata) (map[string]cclfFileValidator, e
...
 }
 
 func importCCLF8(fileMetadata cclfFileMetadata) error {
 	fmt.Printf("Importing CCLF8 file %s...\n", fileMetadata.name)
 	log.Infof("Importing CCLF8 file %s...", fileMetadata.name)
 
-	if fileMetadata.filePath == "" {
-		return errors.New("file path must be provided")
+	if (cclfFileMetadata{}) == fileMetadata {
+		err := errors.New("File CCLF8 not found")
```
embh: Golang recommends not capitalizing error strings https://github.com/golang/go/wiki/CodeReviewComments#error-strings

bcda/cli.go
```diff
@@ -200,13 +231,23 @@ func importCCLF9(fileMetadata cclfFileMetadata) error {
 	for sc.Scan() {
 		b := sc.Bytes()
 		if len(bytes.TrimSpace(b)) > 0 {
-			fmt.Printf("\nXREF: %s\n", b[0:1])
-			fmt.Printf("Current identifier: %s\n", b[currIDStart:currIDEnd])
-			fmt.Printf("Previous identifier: %s\n", b[prevIDStart:prevIDEnd])
-			fmt.Printf("Previous identifier effective date: %s\n", b[prevIDEffDateStart:prevIDEffDateEnd])
-			fmt.Printf("Previous identifier obsolete date: %s\n", b[prevIDObsDateStart:prevIDObsDateEnd])
+
+			cclf9 := models.CCLF9{
```
embh: I wonder if there's something we can call this model that's more descriptive than `CCLF9`. That sounds like it refers to the file instead of its contents. The only idea I have is `CCLFBeneficiaryXRef` but there is likely a better name than that. 🤔

bcda/cli.go
```diff
@@ -200,13 +231,23 @@ func importCCLF9(fileMetadata cclfFileMetadata) error {
...
+				XrefIndicator: string(b[0:1]),
+				CurrentNum:    string(b[currIDStart:currIDEnd]),
+				PrevNum:       string(b[prevIDStart:prevIDEnd]),
+				PrevsEfctDt:   string(b[prevIDEffDateStart:prevIDEffDateEnd]),
+				PrevsObsltDt:  string(b[prevIDObsDateStart:prevIDObsDateEnd]),
+			}
+			err = db.Create(&cclf9).Error
+			if err != nil {
+				return errors.Wrap(err, "could not create CCLF9 record")
```
embh: Same here with naming, it sounds like the file couldn't be stored rather than a row from it.

bcda/cli_test.go
```diff
@@ -223,26 +215,66 @@ func (s *CLITestSuite) TestImportCCLF8_InvalidMetadata() {
...
 	assert := assert.New(s.T())
+	db := database.GetGORMDbConnection()
+	defer database.Close(db)
 
-	filePath := "../shared_files/cclf/T.A0001.ACO.ZC9Y18.D181120.T1000010"
-	metadata, err := getCCLFFileMetadata(filePath)
-	metadata.filePath = filePath
-	assert.Nil(err)
-	err = importCCLF9(metadata)
+	db.Unscoped().Delete(&models.CCLF9{})
```
embh: Maybe also clear out all the `CCLFFile` records so there aren't any hanging out in the table from other tests that could pollute the results of this one.

bcda/cli_test.go
```diff
@@ -223,26 +215,66 @@ func (s *CLITestSuite) TestImportCCLF8_InvalidMetadata() {
...
+		filePath:  "../shared_files/cclf/T.A0001.ACO.ZC9Y18.D181120.T1000010",
+		perfYear:  18,
+		name:      "T.A0001.ACO.ZC9Y18.D181120.T1000010",
+	}
+
+	err := importCCLF9(cclf9metadata)
 	assert.Nil(err)
 
+	file := models.CCLFFile{}
+	db.First(&file, "aco_cms_id = ?", acoID)
```
embh: I know my CCLF8 tests searched by ACO ID too, but do you think it would be better to query on filename? Or maybe it doesn't matter because we're trying to confirm that this one file was imported, and that should be the only matching row in the table.

bcda/cli.go
tbellj: oooo good catch!

bcda/cli_test.go
tbellj: yep :)

bcda/cli_test.go
tbellj: changed both of ours to search by filename. I think this was a good find also

bcda/cli.go
tbellj: right now I'm a keep it as CCLF9 for the name. This pr has been sitting here for a hot second, so I'd like to think if this was a real concern it would have popped up before now.

bcda/cli.go
embh: To follow up on this: to be able to use the filename to get the ACO ID, we need the CCLF9 xref records to be associated with a file, so we need the ID from the file in this model.

bcda/cli.go
```diff
@@ -277,7 +318,7 @@ func importCCLFDirectory(filePath string) (success, failure, skipped int, err er
 
 		cclfvalidator, err := importCCLF0(cclf0)
 		if err != nil {
-			log.Errorf("Failed to import CCLF0 file: %v, Skipping CCLF8 file: %v and CCLF9 file: %v ", cclf0, cclf8, cclf9)
+			log.Errorf("failed to import CCLF0 file: %v, Skipping CCLF8 file: %v and CCLF9 file: %v ", cclf0, cclf8, cclf9)
```
embh: Not important to change this now, but wanted to highlight that these `log.Error` calls are different from the rule of lowercase messages for errors. It's writing directly to the error level log, so this message wouldn't get wrapped as it bubbles up. It can be a complete sentence (or whatever we want 😄 )

PR #224, opened 2019-04-03
BCDA-779 SAMHSA filtering

knollfear:
### Fixes [BCDA-779](https://jira.cms.gov/browse/BCDA-779)

BB has agreed to filter the SAMHSA information from the EOB endpoint if we include the get param "excludeSAMHSA=true" in our request for the data
### Proposed changes:

Add "excludeSAMHSA=true" to the params of the EOB requests from blue button

### Change Details

added function for getDefaultParams.  It is just one for now, but there could be more
added extra param for EOB endpoint
improved tests a bit to be able to confirm that we are in fact making the correct request.

### Security Implications

The affected PII is the data that comes from BB.  This does not affect our security posture.

### Acceptance Validation

I downloaded a file with and without the samhsa filtering (comment out line 108 in bluebutton.go and reboot containers) after decrypting the files I did a diff on the command line to determine if any records had been removed.  6 records were missing which I confirmed with Karl were the correct records to remove.

image
unfiltered file is on the right side.
### Feedback Requested

any


msnook:
Nice.  The only thing that worries me is - what if BB regresses and now all of a sudden we start providing unfiltered data?  Probably out of scope of this ticket, but do we need some redundancy in place to ensure we don't accidentally send unfiltered data in the event that BB has problems?

knollfear:
> Nice. The only thing that worries me is - what if BB regresses and now all of a sudden we start providing unfiltered data? Probably out of scope of this ticket, but do we need some redundancy in place to ensure we don't accidentally send unfiltered data in the event that BB has problems?

Karl has you covered.  If you make a request for whaterver.bb.com?not_a_real_option=7 the server responds with a 400 error.  This means that if BB stops accepting the param or otherwise changes it we will just get nothing which is the desired outcome in those instances.

msnook:
> > Nice. The only thing that worries me is - what if BB regresses and now all of a sudden we start providing unfiltered data? Probably out of scope of this ticket, but do we need some redundancy in place to ensure we don't accidentally send unfiltered data in the event that BB has problems?
> 
> Karl has you covered. If you make a request for whaterver.bb.com?not_a_real_option=7 the server responds with a 400 error. This means that if BB stops accepting the param or otherwise changes it we will just get nothing which is the desired outcome in those instances.

I was more thinking along the lines of the logic behind the filtering regressing at BB, not necessarily the actual param acceptance/rejection.  If they do some sort of optimization on the filtering and it somehow breaks and actually sends us unfiltered data, then we are on the hook I believe.

knollfear:
> > > Nice. The only thing that worries me is - what if BB regresses and now all of a sudden we start providing unfiltered data? Probably out of scope of this ticket, but do we need some redundancy in place to ensure we don't accidentally send unfiltered data in the event that BB has problems?
> > 
> > 
> > Karl has you covered. If you make a request for whaterver.bb.com?not_a_real_option=7 the server responds with a 400 error. This means that if BB stops accepting the param or otherwise changes it we will just get nothing which is the desired outcome in those instances.
> 
> I was more thinking along the lines of the logic behind the filtering regressing at BB, not necessarily the actual param acceptance/rejection. If they do some sort of optimization on the filtering and it somehow breaks and actually sends us unfiltered data, then we are on the hook I believe.

I am not sure the solution there.  If we are "verifying" the filtering we might as well just do the filtering ourselves.  Do we want to go back and redo the risk assessment of having BB do this for us?  I assumed we just accepted that risk when we offloaded this work.

tbellj:
I agree with the double checking. If we implement some verification then we might as well just do it on our end. Not sure what all that entails and would like to avoid that if possible but it would be nice to know that BB logic didn't regress.

msnook:
Out-of-scope ticket created to discuss and possibly implement validation: https://jira.cms.gov/browse/BCDA-1168

bcda/client/bluebutton_test.go
```diff
@@ -20,8 +20,9 @@ type BBTestSuite struct {
 
 func (s *BBTestSuite) SetupTest() {
 	s.ts = httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.Header().Set("Content-Type", "application/fhir+json")
-		fmt.Fprint(w, `{ "test": "ok" }`)
+		w.Header().Set("Content-Type", r.URL.Query().Get("_format"))
+		response := fmt.Sprintf("{ \"test\": \"ok\"; \"url\": %v}", r.URL.String())
```
tbellj: ohhhh so does the printed out url contain the added "excludeSAMHSA=true" now?

knollfear: sort of.  The BB client makes a request to this TLSServer instead of actual blue button.  All I did was rewrite the response to include the url that the client made the request to.  This lets me determine exactly what params were added to the request.

PR #226, opened 2019-04-03
BCDA-1044 remove user from alpha token and client registration

alexhova:
### Fixes [BCDA-1044](https://jira.cms.gov/browse/BCDA-1044)

User is no longer needed to create tokens now that we use a clientID and clientSecret instead

### Proposed changes:
- only allow clientID and clientSecret for alpha token creation
- remove userID from token model


bcda/main.go
```diff
@@ -374,11 +375,11 @@ func createUser(acoID, name, email string) (string, error) {
 }
 
 func createAccessToken(ID string, secret string) (string, error) {
-	if ID == "" {
-		return "", errors.New("ID (--id) must be provided")
+	if uuid.Parse(ID) == nil {
```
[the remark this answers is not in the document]
alexhova: good call.  i'm thinking we just check for nil then and let each plugin validate its own way

bcda/api.go
```diff
@@ -149,7 +149,10 @@ func bulkRequest(t string, w http.ResponseWriter, r *http.Request) {
 	}
 
 	acoID := ad.ACOID
-	userID := ad.UserID
+	user := models.User{}
+	// Arbitrarily use the first user in order to satisfy foreign key constraint "jobs_user_id_fkey" until user is removed from jobs table
+	db.First(&user)
```
dhgreene: It's somewhat surprising that a possible mismatch between the job's acoID and user's acoID doesn't cause a problem, but I can confirm it doesn't, and this is ok.  One more step along the refactoring road!

db/api.sql
```diff
@@ -27,7 +27,6 @@ create table jobs (
 
 create table tokens (
   uuid uuid not null primary key,
-  user_id uuid not null references users,
```
dhgreene: What's our migration strategy for the AWS persistent databases?

bcda/api.go
alexhova: yep, i thought the same thing!

PR #228, opened 2019-04-04
BCDA-1108 Feature: use /auth/token instead of /api/v1/token in tests

dhgreene:
### Fixes [BCDA-1108](https://jira.cms.gov/browse/BCDA-1108)
Now that the alpha auth scheme has a public-facing endpoint for getting tokens (`/auth/token`), we can remove our temporary private token endpoint (`/api/v1/token`).  

The complicating factor is that with `/auth/token` authentication is required.

### Proposed changes:
I explored two strategies for having a `client_id` and `secret available for tests:
1. Put them in database fixtures
2. Generate them as needed

The team was more in favor of 2., though lessons learned from 1. are sparking some discussion.

### Change Details
- Generate `client_id` and `secret` when a token will be needed in tests
  - For Postman and smoke tests this means configuring the Makefile to run the CLI and pass credentials to the executables
- Remove `/api/v1/token` and related tests
- Fix a couple of bugs along the way
- Add test coverage for bugs, as well as filling out tests for `/auth/token`

### Security Implications
Ephemeral secrets are shown in the result of `make` commands.  This should not be a security issue.

### Feedback Requested
- Especially in the Makefile, is there a (different) preferred way to generate, parse, and pass credentials?
- Is more test coverage needed?
- Is there an environment in which we need to run unit, smoke, or Postman tests in which this strategy will not work?


Makefile
```diff
@@ -17,7 +19,9 @@ lint:
 	docker-compose -f docker-compose.test.yml run --rm tests gosec ./...
 
 smoke-test:
-	docker-compose -f docker-compose.test.yml run --rm -w /go/src/github.com/CMSgov/bcda-app/test/smoke_test tests sh smoke_test.sh
+	docker exec -it bcda-app_api_1 sh -c 'tmp/bcda create-alpha-token --size dev' > /tmp/creds
```
msnook: I think we should try to avoid saving the output of `create-alpha-token` to a temp file.  Instead, maybe we can leverage some nifty commands like `cut` to save the output of the commands into vars?

bcda/router.go
```diff
@@ -28,9 +28,6 @@ func NewAPIRouter() http.Handler {
 		}
 		r.With(auth.RequireTokenAuth, auth.RequireTokenJobMatch).Get(m.WrapHandler("/jobs/{jobID}", jobStatus))
 		r.Get(m.WrapHandler("/metadata", metadata))
-		if os.Getenv("DEBUG") == "true" {
```
msnook: Bye Bye!!!

tbellj:
so answering @snook, I'm not familiar with the auth code at all so just reading through this now. Are clientId and clientSecret now required for receiving a token? The performance test just uses the stored alpha tokens for TEST in Jenkins. The code itself in performance.go never calls the /api/v1/token endpoint just uses the tokens as parameters.

msnook:
> so answering @snook, I'm not familiar with the auth code at all so just reading through this now. Are clientId and clientSecret now required for receiving a token? The performance test just uses the stored alpha tokens for TEST in Jenkins. The code itself in performance.go never calls the /api/v1/token endpoint just uses the tokens as parameters.

Ahh, then we might be good there.  And, thinking about it, we might be good with our standard smoke test in Jenkins too since we simply use the saved secret that is stored in Jenkins (and don't attempt to get a new token each time we test).  So I think to be safe, if we just deploy this feature branch to `dev` to make sure it functions properly, then we should be good to go.

dhgreene:
@msnook @tbellj Thanks for the verification and suggestions.  The code has now been tested in `dev` with credentials stored in Jenkins, and a [PR](https://github.com/CMSgov/bcda-ops/pull/171) is open in the ops repo to support that change.

msnook:
@dhgreene Looks like the branch became out of date.  Could you update the branch with `master` prior to final review?

msnook:
@alexhova @DeirdreHolub I have reviewed this from a test/ops perspective and this looks great.  David and I were able to execute the feature branch in the Jenkins pipeline.  I think it would be a good idea to get a review from an auth perspective as well.

bcda/main.go
```diff
@@ -422,10 +422,11 @@ func createAlphaToken(ttl int, acoSize string) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("could not register client for %s (%s) because %s", aco.UUID.String(), aco.Name, err.Error())
 	}
-	aco.ClientID = creds.ClientID
+
 	db := database.GetGORMDbConnection()
 	defer database.Close(db)
-	err = db.Save(&aco).Error
+	// Only update ClientID; the AlphaSecret may already be in the database, and we'd overwrite it with an empty string
```
DeirdreHolub: This comment at first confused me, so I had to go read the code. The source of confusion is that the API and Alpha Auth are using the same table in the same database. Alpha auth needs to remember both the client_id and alpha_secret when it registers a client, but the API only remembers the client_id when it receives and forwards the credentials to output. This CLI command "belongs" to the API, so it only touches client_id. Finally, because this same command is used whether it's being called with Alpha Auth or Okta Auth, it has to save the client_id.

dhgreene: Comment updated to something hopefully more useful.

bcda/api_test.go
```diff
@@ -665,20 +671,55 @@ func (s *APITestSuite) TestGetToken() {
 
 func (s *APITestSuite) TestAuthToken() {
 	s.SetupAuthBackend()
-	req := httptest.NewRequest("POST", "/auth/token", nil)
 
+	// Missing authorization header
+	req := httptest.NewRequest("POST", "/auth/token", nil)
 	handler := http.HandlerFunc(getAuthToken)
+	fmt.Println("No authorization header")
```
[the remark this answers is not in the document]
dhgreene: No problem.  Println()'s removed and test split.

PR #232, opened 2019-04-10
BCDA-1184 Make CLI to import directory of CCLF files

knollfear:
### Fixes [BCDA-1184](https://jira.cms.gov/browse/BCDA-1184)

In order to effectively schedule CCLF file imports we need to be able to scan in a directory and have it import all of the files in that directory.
### Proposed changes:

Removed CCLF9 and CCLF8 import tasks
added import-cclf-directory command to import a whole directory of files
changed method signature for CCLF 8/9 to accept a cclffilemetadata object instead of a string.

### Change Details

changed method signature for CCLF 8/9 to accept a cclffilemetadata object instead of a string.

### Security Implications

no PII should be impacted.  this leverages existing code that has PII, but this function doesn't affect that

### Acceptance Validation

Make sure tests are the right tests
I was able to "import" the existing sample files

### Feedback Requested

Are the tests the right tests?
anything else is welcome.


bcda/cli.go
```diff
@@ -157,3 +148,61 @@ func getCCLFFileMetadata(filePath string) (cclfFileMetadata, error) {
...
+
+	err = filepath.Walk(filePath, sortCCLFFiles(&cclf0, &cclf8, &cclf9))
+	if err != nil {
+		return 0, 0, err
+	}
+	for _, file := range cclf8 {
+		//todo match cclf0 and validate
+		err = importCCLF8(file)
+		if err != nil {
+			failure += 1
```
msnook: We should probably log the failure with accompanying filename.

bcda/cli.go
```diff
@@ -157,3 +148,61 @@ func getCCLFFileMetadata(filePath string) (cclfFileMetadata, error) {
...
+			continue
+		}
+		log.Info(fmt.Sprintf("Successfully imported CCLF8 file: %v", file))
+		success += 1
+	}
+	for _, file := range cclf9 {
+		//todo match cclf0 and validate
+		err = importCCLF9(file)
+		if err != nil {
+			failure += 1
```
msnook: Same comment as above... failure logging.

bcda/cli.go
```diff
@@ -157,3 +148,61 @@ func getCCLFFileMetadata(filePath string) (cclfFileMetadata, error) {
...
+}
+
+func sortCCLFFiles(cclf0, cclf8, cclf9 *[]cclfFileMetadata) filepath.WalkFunc {
+	return func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		// Directories are not CCLF files
+		if info.IsDir() {
+			return nil
```
msnook: Should we log if we encounter this?  Maybe we received the files in a format we weren't expecting (i.e., wrapped in a dir).

msnook: Also, do we have a test case for this?

bcda/cli_test.go
```diff
@@ -176,4 +169,102 @@ func (s *CLITestSuite) TestGetCCLFFileMetadata() {
...
+	assert.Contains(buf.String(), "Failed to import 0 files.")
+	buf.Reset()
+
+	// dir has 3 files, but 2 will be ignored because of bad file names.
+	args = []string{"bcda", "import-cclf-directory", "--directory", "../shared_files/cclf_BadFileNames/"}
+	err = s.testApp.Run(args)
+	assert.Nil(err)
+	assert.Contains(buf.String(), "Completed CCLF import.")
+	assert.Contains(buf.String(), "Successfully imported 1 files.")
+	assert.Contains(buf.String(), "Failed to import 0 files.")
```
msnook: Wondering if we need to be informed that we failed to import 2 files with bad filenames.  Looking at the end results, it looks like we had no problems (i.e., no failures to import files).  But in reality, we actually had issues with 2 files.

bcda/cli.go
```diff
@@ -157,3 +148,61 @@ func getCCLFFileMetadata(filePath string) (cclfFileMetadata, error) {
...
+		return 0, 0, err
+	}
+	for _, file := range cclf8 {
+		//todo match cclf0 and validate
+		err = importCCLF8(file)
+		if err != nil {
+			failure += 1
+			continue
+		}
+		log.Info(fmt.Sprintf("Successfully imported CCLF8 file: %v", file))
```
msnook: I think it would be great to have a unit test case which validates this log statement - ensuring we are truly outputting the filename.  With that test in place, it will ensure we don't regress from a logging transparency perspective.

bcda/cli.go
```diff
@@ -157,3 +148,61 @@ func getCCLFFileMetadata(filePath string) (cclfFileMetadata, error) {
...
+		success += 1
+	}
+	for _, file := range cclf9 {
+		//todo match cclf0 and validate
+		err = importCCLF9(file)
+		if err != nil {
+			failure += 1
+			continue
+		}
+		log.Info(fmt.Sprintf("Successfully imported CCLF9 file: %v", file))
```
msnook: Similar comment to above:
I think it would be great to have a unit test case which validates this log statement - ensuring we are truly outputting the filename.  With that test in place, it will ensure we don't regress from a logging transparency perspective.

bcda/cli_test.go
```diff
@@ -107,49 +107,42 @@ func (s *CLITestSuite) TestCreateACO() {
...
-	args = []string{"bcda", "import-cclf9", "--file", "../shared_files/cclf/T.A0001.ACO.ZC9Y18.D181120.T1000010"}
-	err = s.testApp.Run(args)
-	assert.Nil(err)
+	metadata.cclfNum = 9
+	metadata.filePath = ""
+	err = importCCLF9(metadata)
+	assert.NotNil(err)
 }
 
 func (s *CLITestSuite) TestGetCCLFFileMetadata() {
```
msnook: Absolutely love all of these test cases !!!

bcda/cli.go
knollfear: I made folders with CCLF like file names.  It looks like git might have just ignored them.  I will move some things around to get it to be committed.

bcda/cli.go
knollfear: I couldn't get the import as it stands to fail.  I imported bad files constantly, but they just roll through.  I suggested a solution in slack, but it seemed kind of out of scope since the imports were already resolved by other pending commits.

shared_files/cclf_BadFileNames/T.A0001.ACO.ZC8Y18.D181120.T1000009
```diff
@@ -0,0 +1,6 @@
+1A69B98CD30203031401M 25013010891940-10-271107710NA                              FnameOne                      A              LnameOne                                032016-08-012017-11-01Address L1b1                                 Address L2b1                                 Address L3b1                            Address L4b1                            Address L5b1                            Address L6b1                            CITY1                                                                                               MD211171111
```
tbellj: forgive me if this is a silly question, but did we generate the bad file names or did they give us a sample of bad files?

bcda/cli.go
knollfear: I think this is a concern for importCCLF8, not this code.

bcda/cli_test.go
knollfear: I added in a skipped count to the return

shared_files/cclf_BadFileNames/T.A0001.ACO.ZC8Y18.D181120.T1000009
knollfear: We made them ourselves.

msnook:
@knollfear Technically, this is the entrypoint/driver of the application for ingesting PII/PHI into the system.  I think we should probably update the `Security Implications` section of the PR to reflect that, for full transparency.

PR #233, opened 2019-04-11
BCDA-1143 Maintenance: Separate Auth API Methods from BCDA API

dhgreene:
### Fixes [BCDA-1143](https://jira.cms.gov/browse/BCDA-1143)
The goal is to separate the auth routing code (as much as possible), and create stubs for the next few auth endpoints.

### Proposed changes:
Reduce knowledge of auth internals by moving auth-related routing and route testing to the `/bdca/auth/` folder.

### Change Details
Added (by taking away code from `/bcda/*.go`:
- `/bcda/auth/api.go`
- `/bcda/auth/api_test.go`
- `/bcda/auth/router.go`
- `/bcda/auth/router_test.go`

Created stubs for four `/auth/group` verbs.  
Note that:
- No Swagger documentation was added, since these will be private and unavailable to the audience using our Swagger docs.
- An `auth_test` package had to be created for `/bcda/auth/api_test.go` to prevent an import loop with testUtils.

### Security Implications
The new `/auth/group` endpoints are only available when the environment contains `DEBUG=true`.  We'll likely have a more sophisticated way of making them available to ACO-MS later.

### Feedback Requested
- Are these the correct new locations for this code?
- Has too much/too little been moved?


bcda/router.go
```diff
@@ -29,13 +29,16 @@ func NewAPIRouter() http.Handler {
...
 	})
-	r.Post(m.WrapHandler("/auth/token", getAuthToken))
 	r.Get(m.WrapHandler("/_version", getVersion))
 	r.Get(m.WrapHandler("/_health", healthCheck))
 	r.Get(m.WrapHandler("/_auth", getAuthInfo))
 	return r
 }
 
+func NewAuthRouter() http.Handler {
+	return auth.NewAuthRouter(auth.ParseToken, logging.NewStructuredLogger(), HSTSHeader, ConnectionClose)
```
dhgreene: > ParseToken is for external consumption by users of auth, because the token claims it expects are for the API. Token claims for its auth admin endpoints like group would be different. ParseToken should not be middleware for this call.

@DeirdreHolub Is this just a semantic issue, since the parameters are named "middleware" in `auth.NewAuthRouter`?  The API is still in full control over whether `ParseToken` is called.  It's used in exactly this way 3 lines up in a call to `chi.With()` (which also names these parameters as middleware).  The one change is that it's now being passed to `auth.NewAuthRouter`.

From a risk perspective I'm less worried about whether we're now hiding `ParseToken` inside auth, and more worried whether there are other customizations to an `http.Handler` that might need to be made in other contexts or by other clients.

dhgreene: ParseToken removed.  Thanks for clarifying!

PR #235, opened 2019-04-12
BCDA-1134 Documentation: improve swagger workflow

dhgreene:
### Fixes [BCDA-1134](https://jira.cms.gov/browse/BCDA-1134)
We have multiple authentication methods on Swagger now:
1) `/auth/token` accepts client credentials with Basic authentication
2) The resulting token is used in the bulkData endpoints via bearer token (what Swagger calls "API authentication")

This leads to some potential confusion as users navigate the Swagger docs.

### Proposed changes:
- Add appropriate text to differentiate authentication methods when the user clicks the "authentication" box near the top of the page
- Change the order of authentication methods in the "authentication" dialog to reflect order of usage
- Change the order of the API groups so that it's intuitive to get a token from `/auth/token` before using the authenticated bulkData endpoints.
- Remove the lock icon from the metadata endpoints, so as to make it clear they do not require authentication.

### Change Details
- API group ordering is made possibly by defining the tags in the schema, which is not directly supported by `go-swagger`.  However, `go-swagger` allows for direct creation of a YAML file: in this case, `/bcda/swaggerui/tags.yml`.
- Not changed: the ticket requested that the words "username" and "password" be changed to "client id" and "secret" for Basic authentication.  This might be possible by editing the content of the Swagger page after it's generated (e.g. with Javascript), but does not seem to be possible during Swagger page generation.

### Security Implications
These changes affect documentation only.

### Acceptance Validation
Screen Shot 2019-04-12 at 12 00 01 PM

Screen Shot 2019-04-12 at 11 59 33 AM

### Testing
Review the screenshots above or:
1. Rebuild the `static_site` container
1. Run `docker-compose up static_site`

### Feedback Requested
- Swagger wording, ordering, and grouping suggestions are welcome
- In particular, the text in the `Name` field of the bearer token authentication might be a bit much.  What do you think?
- The Basic authentication prompts are for `username` and `password`, but we are looking for users to provide `client ID` and `secret`.  Is this a problem worth spending significant effort changing?


bcda/swaggerui/tags.yml
```diff
@@ -0,0 +1,4 @@
+tags:
```
knollfear: Nice.  This is great.

bcda/api.go
```diff
@@ -16,18 +16,13 @@
...
 
-     Security:
-     - api_key:
-
      SecurityDefinitions:
-     api_key:
+     bearer_token:
           type: apiKey
-          name: Authorization
+          name: For bulkData endpoints. 1) Put your credentials in Basic Authentication, 2) get your token from /auth/token, 3) put your token here, after the word "Bearer"
```
knollfear: I think new lines would be jazzy if that is possible.
I also propose this:
For BulkData Endpoints require a Bearer Token (Do we have a userguide explanation? can we link it?).  
1) Put your credentials in Basic Authentication (is an anchor link possible?)
2) Request a bearer token from /auth/token
3) Put "Bearer {TOKEN}" in this field (no quotes) using the bearer token retrieved in step 2

dhgreene: @knollfear 
- Best I can tell, newlines are not an option given that it's HTML as produced by YAML as processed through GoDoc.  We're stretching the `Name` field a bit far as it is!
- HTML tags also seem to be escaped, so no links.

I've adapted your suggestions to the text at the top of the page as well as the "Authorize" dialog:

Screen Shot 2019-04-15 at 8 01 46 PM

Screen Shot 2019-04-15 at 8 02 26 PM

karlmdavis: > * Best I can tell, newlines are not an option given that it's HTML as produced by YAML as processed through GoDoc.  We're stretching the `Name` field a bit far as it is!

Just 'cause I was curious, I went and looked and found this: <https://golang.org/pkg/go/doc/#ToHTML>. Which is dumb; not allowing for MarkDown or some more advanced embedded formatting than what they've got seems short-sighted. Oh well.

dhgreene: @karlmdavis Thanks for that.  Yes, it all works as your reference says for the top of the page.  But in the middle of the `bearer_token` `Name:` field links aren't handled.  We'll keep that in mind for future enhancements.

PR #236, opened 2019-04-15
BCDA-1159 add CLI to delete files in a dir

knollfear:
### Fixes [BCDA-1159](https://jira.cms.gov/browse/BCDA-1159)

In order to ensure that CCLF files are removed after import we need a CLI to delete all of the files in a specific Directory.  To prevent accidental deletion of any random dir the CLI has an ENV var as the argument which should prevent removal of an arbitrary directory by mistake.
### Proposed changes:

Added a CLI

### Change Details

Added a CLI and associated tests

### Security Implications

This code deletes file in order to increase our security posture by ensuring that files do not exist longer than necessary

### Acceptance Validation

Run the CLI (Set an ENV Var first)

### Feedback Requested

Any

### DIFF
--- a/bcda/cli.go
+++ b/bcda/cli.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/CMSgov/bcda-app/bcda/models"
+	log "github.com/sirupsen/logrus"
 )
 
 func createACO(name, cmsID string) (string, error) {
@@ -156,4 +157,30 @@ func getCCLFFileMetadata(filePath string) (cclfFileMetadata, error) {
 	metadata.timestamp = t
 
 	return metadata, nil
+}
+
+func deleteDirectory(dirToDelete string) (filesDeleted int, err error) {
+	log.Info(fmt.Sprintf("preparing to delete directory '%v'", dirToDelete))
+	f, err := os.Open(dirToDelete)
+	if err != nil {
+		return 0, err
+	}
+	files, err := f.Readdir(-1)
+	if err != nil {
+		return 0, err
+	}
+	err = f.Close()
+	if err != nil {
+		return 0, err
+	}
+
+	for _, file := range files {
+		log.Info(fmt.Sprintf("deleting %v", file.Name()))
+		err = os.Remove(filepath.Join(dirToDelete, file.Name()))
+		if err != nil {
+			return 0, err
+		}
+	}
+
+	return len(files), nil
 }
--- a/bcda/main.go
+++ b/bcda/main.go
@@ -70,7 +70,7 @@ func setUpApp() *cli.App {
 	app.Name = Name
 	app.Usage = Usage
 	app.Version = version
-	var acoName, acoCMSID, acoID, userName, userEmail, tokenID, tokenSecret, accessToken, ttl, threshold, acoSize, filePath string
+	var acoName, acoCMSID, acoID, userName, userEmail, tokenID, tokenSecret, accessToken, ttl, threshold, acoSize, filePath, envVar string
 	app.Commands = []cli.Command{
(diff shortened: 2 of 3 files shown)


bcda/cli.go
```diff
@@ -211,3 +211,29 @@ func sortCCLFFiles(cclf0, cclf8, cclf9 *[]cclfFileMetadata, skipped *int) filepa
 		return nil
 	}
 }
+
+func deleteDirectory(dirToDelete string) (filesDeleted int, err error) {
+	log.Info(fmt.Sprintf("preparing to delete directory '%v'", dirToDelete))
+	f, err := os.Open(dirToDelete)
```
alexhova: the linter doesn't like this.  where does the dirToDelete var come from?

knollfear: I will nosec it.  It comes from the name of an environment variable that gets passed in as the argument to the command.

alexhova: ok cool.  i guess the other option would be to pull from the environment in this func.  but i haven't looked at it enough to know if that makes sense for the use case.  i'll review again when after whatever change you make!

bcda/main.go
```diff
@@ -326,6 +326,29 @@ func setUpApp() *cli.App {
...
+			Usage:    "Delete the files in a directory specified by the passed Environment Variable",
+			Flags: []cli.Flag{
+				cli.StringFlag{
+					Name:        "envvar",
+					Usage:       "Name of Env Var that determines the dir to delete",
+					Destination: &envVar,
+				},
+			},
+			Action: func(c *cli.Context) error {
+				dirToDelete := os.Getenv(envVar)
```
msnook: We could pass in any env var here.  Should we make sure its truly a filepath?

bcda/cli.go
```diff
@@ -211,3 +211,29 @@ func sortCCLFFiles(cclf0, cclf8, cclf9 *[]cclfFileMetadata, skipped *int) filepa
 		return nil
 	}
 }
+
+func deleteDirectory(dirToDelete string) (filesDeleted int, err error) {
+	log.Info(fmt.Sprintf("preparing to delete directory '%v'", dirToDelete))
```
msnook: Technically, we aren't deleting the directory, right?  We are just deleting the files in the directory.

bcda/main.go
knollfear: added a check to ensure it is a dir before moving on.

bcda/cli.go
knollfear: Renamed function to be more clear.

bcda/cli.go
knollfear: I just added a comment that tells the linter I am aware of this flag and have decided it is ok.

bcda/main.go
msnook: I guess nothing is stopping me from defining my own env var and executing this like:
```
ROOT_DIR=/ bcda delete-dir-contents
```

So now I'm second guessing the approach of requiring that an env var be passed in.  And I think if we keep it open to _delete any dir_ we could run into accidental issues (not protecting ourselves from accidental file system corruption).  

What is the downside of making this CLI very specific like `delete-cclf-files`, and deleting only the files in that directory (which is specified via env var)?

knollfear: I took our conversation to heart and refactored to accept a directory instead of an env var.

bcda/cli.go
embh: `filepath.Clean(dirToDelete)` should prevent the linter warning on it without needing `#nosec`

knollfear: Added.  Thanks. I was not aware of that option.

PR #238, opened 2019-04-16
BCDA-1185 Feature: Added CCLF0 validation

tbellj:
Fixes [BCDA-1185](https://jira.cms.gov/browse/BCDA-1185)

Problem:
We did not have a way to validate cclf8 and cclf9 files before this code.

Proposed changes:
A synthetic cclf0 file is now being used to validate our synthetic attribution files.

Change Details:
bcda/cli.go - added cclf0 validation logic
bcda/cli_test.go - added test cases for cclf0 import
shared_files/cclf/ - added synthetic cclf0 file (matches our synthetic 8 & 9)
shared_files/cclf0/ - added synthetic cclf0 file (doesn't match)
hared_files/cclf_BadFileNames/ - added synthetic cclf0 file (matches our synthetic 8 & 9)

Security Implications:
It's being used with files that have PII/PHI, but this particular change is just to validate those files.

Acceptance Validation
Yes, see screen shot below.
Screen Shot 2019-04-15 at 8 11 44 PM

Screen Shot 2019-04-15 at 8 19 21 PM

Feedback Requested
Any and all.


bcda/cli.go
```diff
@@ -67,18 +120,28 @@ func importCCLF8(fileMetadata cclfFileMetadata) error {
...
+
 	for sc.Scan() {
 		b := sc.Bytes()
-		if len(bytes.TrimSpace(b)) > 0 {
+		bytelength := len(bytes.TrimSpace(b))
+		if bytelength > 0 && bytelength <= validator.maxRecordLength {
+			count++
+			// currently only errors if there are more records than we expect.
+			if count > validator.totalRecordCount {
+				return errors.New("maximum record count reached: did not match cclf0 validation")
```
msnook: Can we log as much as possible here?  Filename being validated?  Expected count vs actual count?  Anything else?

bcda/cli.go
```diff
@@ -67,18 +120,28 @@ func importCCLF8(fileMetadata cclfFileMetadata) error {
...
+			}
 			fmt.Printf("\nMBI: %s\n", b[mbiStart:mbiEnd])
 			fmt.Printf("HICN: %s\n", b[hicnStart:hicnEnd])
+		} else {
+			return errors.New("incorrect record length: did not match cclf0 validation")
```
msnook: Same as above.  Can we log as much as possible here? Filename being validated? Expected count vs actual count? Anything else?

bcda/cli.go
```diff
@@ -101,17 +164,27 @@ func importCCLF9(fileMetadata cclfFileMetadata) error {
...
+			if count > validator.totalRecordCount {
+				return errors.New("maximum record count reached: did not match cclf0 validation")
+			}
 			fmt.Printf("\nXREF: %s\n", b[0:1])
 			fmt.Printf("Current identifier: %s\n", b[currIDStart:currIDEnd])
 			fmt.Printf("Previous identifier: %s\n", b[prevIDStart:prevIDEnd])
 			fmt.Printf("Previous identifier effective date: %s\n", b[prevIDEffDateStart:prevIDEffDateEnd])
 			fmt.Printf("Previous identifier obsolete date: %s\n", b[prevIDObsDateStart:prevIDObsDateEnd])
+		} else {
+			return errors.New("incorrect record length: did not match cclf0 validation")
```
msnook: Same logging comments as above.  We should try to log as much as possible on failure.

bcda/cli.go
```diff
@@ -156,26 +229,41 @@ func importCCLFDirectory(filePath string) (success, failure, skipped int, err er
...
-	}
-	for _, file := range cclf9 {
-		//todo match cclf0 and validate
-		err = importCCLF9(file)
-		if err != nil {
-			failure += 1
-			continue
+
+		for _, file = range cclf8 {
+			if file.acoID == aco {
```
msnook: If we don't get a match here, then that means that a CCLF8 (or CCLF9) file exists without an accompanying CCLF0 file, correct?  This means that the 8 and/or 9 file will not be imported.  Can we log this scenario?  It would be very hard to troubleshoot that scenario without logging.
To do that, I think we'd have to somehow keep track of files that we imported and compare against total files in the directory :(

bcda/cli.go
tbellj: filename validation already happens earlier on. I can add extra logging for the expected vs. the actual though.

bcda/cli.go
tbellj: same response above

bcda/cli.go
tbellj: the logging happens here: importCCLFDirectory as the error will bubble up.

bcda/cli.go
```diff
@@ -67,18 +120,28 @@ func importCCLF8(fileMetadata cclfFileMetadata) error {
 	)
 
 	sc := bufio.NewScanner(file)
+	count := 0
+	validator := cclfFileValidator["CCLF8"]
```
msnook: I think it is possible for cclfFileValidator to be nil/empty (this might happen if CCLF0 file contents doesn't contain an entry for CCLF8 or CCLF9, or if the file is empty).  We should probably ensure that that the validator object contains the data we want.

bcda/cli.go
msnook: If we fail validation here, I don't see anywhere indicating exactly which file failed.  We'll need transparency into knowing exactly the file that failed, so we can potentially request a re-send.

bcda/cli.go
msnook: Doesn't look like that function reports the exact file name that failed though.

bcda/cli.go
```diff
@@ -156,26 +229,41 @@ func importCCLFDirectory(filePath string) (success, failure, skipped int, err er
...
+					continue
+				}
+				log.Info(fmt.Sprintf("Successfully imported CCLF9 file: %v", file))
+				success++
+			}
 		}
-		log.Info(fmt.Sprintf("Successfully imported CCLF9 file: %v", file))
-		success += 1
 	}
+
```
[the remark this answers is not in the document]
knollfear: Does it matter if a CCLF0 file fails to find the associated CCLF8/9?  Maybe we should change sortCCLFFiles(cclf0, cclf8, cclf9 *[]cclfFileMetadata) to sortCCLFFiles(*map[string][]cclfFileMetadata) to ease up on the sorting load?  I think we could get that done in a single pass instead of up to 600 passes as it is now.

bcda/cli.go
tbellj: This won't even execute if the there's an error in reading the CCLF0 file. The ingestion of cclf0 happens first and if there's an error we log it and return. Because the ingestion of 8 and 9 has to be validated first, it doesn't make sense to continue if we have an empty cclf0 file or missing data points. But this is part of a larger discussion we had about what to do if cclf8 was weird and cclf9 was okay and vice versa.

bcda/cli.go
```diff
@@ -101,17 +164,27 @@ func importCCLF9(fileMetadata cclfFileMetadata) error {
...
+	count := 0
+	validator := cclfFileValidator["CCLF9"]
+
 	for sc.Scan() {
 		b := sc.Bytes()
-		if len(bytes.TrimSpace(b)) > 0 {
+		bytelength := len(bytes.TrimSpace(b))
+		if bytelength > 0 && bytelength <= validator.maxRecordLength {
+			count++
+			// currently only errors if there are more records than we expect.
```
msnook: I think this processing could potentially lead to a file that is only partially ingested.  In other words, we validate in realtime (while ingesting) instead of validating the file up-front (followed by ingesting).  I think we should do the latter, because we don't want to consider ingesting any parts of a file that isn't valid.

bcda/cli.go
```diff
@@ -45,7 +46,59 @@ type cclfFileMetadata struct {
...
-func importCCLF8(fileMetadata cclfFileMetadata) error {
+type cclfFileValidator struct {
+	totalRecordCount int
+	maxRecordLength  int
+}
+
+func importCCLF0(fileMetadata cclfFileMetadata) (map[string]cclfFileValidator, error) {
+	var validator map[string]cclfFileValidator
+	if fileMetadata.filePath == "" {
+		return nil, errors.New("file path (--file) must be provided")
```
embh: The `--file` reference here is from when CCLF8/9 were imported directly with CLI commands, instead of the first step of looking at the whole directory that we have now. So that flag can be removed from the message

bcda/cli.go
```diff
@@ -45,7 +46,59 @@ type cclfFileMetadata struct {
...
+	if fileMetadata.cclfNum != 0 {
+		return nil, errors.New("invalid CCLF0 filename")
+	}
+
+	file, err := os.Open(filepath.Clean(fileMetadata.filePath))
+	if err != nil {
+		return nil, err
+	}
+
+	fmt.Printf("File contains %s data for ACO %s at %s.\n", fileMetadata.env, fileMetadata.acoID, fileMetadata.timestamp)
```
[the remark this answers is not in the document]
embh: When this is running from the CLI, it could be helpful to have some messaging so you can see where it is in the process. Maybe print something (not necessarily this message - another holdover from the CCLF8/9 import PoC) to stdout _and_ log it?

bcda/cli.go
tbellj: gotcha! it's okay to log the filenames right?

bcda/cli.go
tbellj: yep this was clarified in slack. On it

bcda/cli.go
tbellj: thinking on this, there's several little things that can happen here.

bcda/cli.go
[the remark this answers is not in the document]
tbellj: right I was just following the other logic of cclf8 and 9. Do we want to log and write to stdout?

bcda/cli.go
```diff
@@ -156,26 +229,41 @@ func importCCLFDirectory(filePath string) (success, failure, skipped int, err er
...
+		for _, file = range cclf8 {
+			if file.acoID == aco {
+				err = importCCLF8(file, cclfValidator)
+				if err != nil {
+					log.Error(err)
+					failure++
+					continue
+				}
+				log.Info(fmt.Sprintf("Successfully imported CCLF8 file: %v", file))
+				success++
```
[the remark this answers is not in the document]
tbellj: not seeing the value of a break here. you might have to explain that in person to me, too much to type :), also was thinking that the continue's that were already in place made sense as a particular file failed to import.

bcda/cli.go
```diff
@@ -156,26 +229,41 @@ func importCCLFDirectory(filePath string) (success, failure, skipped int, err er
...
+		for _, file = range cclf9 {
+			if file.acoID == aco {
+				err = importCCLF9(file, cclfValidator)
+				if err != nil {
+					log.Error(err)
+					failure++
+					continue
+				}
+				log.Info(fmt.Sprintf("Successfully imported CCLF9 file: %v", file))
+				success++
```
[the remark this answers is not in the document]
tbellj: same response above

bcda/cli.go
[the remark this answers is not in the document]
tbellj: to your first question, we have to answer that as a team. And I think the way you wrote sortCCLFFiles(cclf0, cclf8, cclf9 *[]cclfFileMetadata) is fine for now, at least I didn't see a problem with it as I was coding.

bcda/cli.go
tbellj: yep was thinking transactional at the time, but we discussed this in slack so will change.

bcda/cli.go
tbellj: oooh good catch!

bcda/cli.go
embh: Seconding @msnook's comment. If the CCLF0 file is empty,`sc.Scan()` won't even be loop-able, right? So `importCCLF0()` could return an unpopulated `validator` map and no error.

tbellj: ahh yeah I see what you'll are saying, I was thinking more of a ingest error not an empty map. good stuff

bcda/cli.go
msnook: In an external Slack discussion with Eric:
```
The CCLF file names are okay for logging, since the chances of incorporating PII / PHI is low since it's internal CMS naming conventions
```

bcda/cli.go
embh: This might be moving into the realm of requirements rather than implementation, but do we know for sure how the files will arrive? We know they will be zipped, but could CCLF0, CCLF8, and CCLF9 be together in one archive? Or is each CCLF file zipped individually? If they're grouped into archives, that could make things easier for us, to confirm that there is a matching set of 0, 8, and 9 together.

bcda/cli.go
```diff
@@ -45,21 +46,79 @@ type cclfFileMetadata struct {
...
-func importCCLF8(fileMetadata cclfFileMetadata) error {
-	if fileMetadata.filePath == "" {
-		return errors.New("file path (--file) must be provided")
+type cclfFileValidator struct {
+	totalRecordCount int
+	maxRecordLength  int
+}
+
+func importCCLF0(fileMetadata cclfFileMetadata) (map[string]cclfFileValidator, error) {
+	if (cclfFileMetadata{}) == fileMetadata {
```
embh: 🆒 Didn't know about this way of checking if the struct is empty!

bcda/cli.go
```diff
@@ -45,21 +46,79 @@ type cclfFileMetadata struct {
...
+		return nil, err
+	}
+
+	file, err := os.Open(filepath.Clean(fileMetadata.filePath))
+	if err != nil {
+		log.Error(err)
+		return nil, err
+	}
+
+	log.Infof("File contains %s data for ACO %s at %s.\n", fileMetadata.env, fileMetadata.acoID, fileMetadata.timestamp)
```
embh: The details about environment, ACO, and timestamp were mostly for verification when this was a PoC, so we probably don't need all that now. With `importCCLFDirectory()` logging when the import is done or failed, maybe we could just say here that the CCLF0 file is being read, with the filename.

bcda/cli.go
```diff
@@ -45,21 +46,79 @@ type cclfFileMetadata struct {
...
+				if err != nil {
+					log.Error(err)
+					return nil, err
+				}
+				validator[filetype] = cclfFileValidator{totalRecordCount: count, maxRecordLength: length}
+			}
+		}
+	}
+
+	if len(validator) < 2 {
```
embh: Should we be more specific in checking that there are CCLF8 and CCLF9 entries in the map? Seems possible (though very unlikely) that CCLF0 could have two CCLF8 rows but no CCLF9 rows, for example, and we wouldn't realize we hadn't gotten CCLF9 info.

bcda/cli.go
```diff
@@ -45,21 +46,79 @@ type cclfFileMetadata struct {
...
+		log.Error(err)
+		return nil, err
 	}
+	return validator, nil
+}
 
+func importCCLF8(fileMetadata cclfFileMetadata) error {
 	if fileMetadata.cclfNum != 8 {
-		return errors.New("invalid CCLF8 filename")
+		err := errors.New("invalid CCLF8 file number")
```
embh: This is probably going to get overwritten with my PR (by a very similar message). Brace yourself. 😄

bcda/cli.go
```diff
@@ -150,31 +207,65 @@ func getCCLFFileMetadata(filePath string) (cclfFileMetadata, error) {
...
+				cclf8 = cclf
+			} else if cclf.cclfNum == 9 {
+				cclf9 = cclf
+			}
+		}
+
+		cclfvalidator, err := importCCLF0(cclf0)
 		if err != nil {
-			failure += 1
+			log.Errorf("Failed to import CCLF0 file: %v, Skipping CCLF8 file:%v and CCLF9 file:%v ", cclf0, cclf8, cclf9)
```
embh:
```suggestion
			log.Errorf("Failed to import CCLF0 file: %v, Skipping CCLF8 file: %v and CCLF9 file: %v ", cclf0, cclf8, cclf9)
```

bcda/cli.go
```diff
@@ -150,31 +207,65 @@ func getCCLFFileMetadata(filePath string) (cclfFileMetadata, error) {
...
-			failure += 1
-			continue
+			log.Errorf("Failed to validate CCLF8 file: %v", cclf8)
+			failure++
+		} else {
+			if err = importCCLF8(cclf8); err != nil {
+				log.Errorf("Failed to import CCLF8 file: %v ", cclf8)
+				failure++
+			} else {
+				log.Info(fmt.Sprintf("Successfully imported CCLF8 file: %v", cclf8))
```
embh:
```suggestion
				log.Infof("Successfully imported CCLF8 file: %v", cclf8)
```

bcda/cli.go
```diff
@@ -150,31 +207,65 @@ func getCCLFFileMetadata(filePath string) (cclfFileMetadata, error) {
...
+		err = validate(cclf9, cclfvalidator)
+		if err != nil {
+			log.Errorf("Failed to validate CCLF9 file: %v", cclf9)
+			failure++
+		} else {
+			if err = importCCLF9(cclf9); err != nil {
+				log.Errorf("Failed to import CCLF9 file: %v ", cclf9)
+				failure++
+			} else {
+				log.Info(fmt.Sprintf("Successfully imported CCLF9 file: %v", cclf9))
```
embh:
```suggestion
				log.Infof("Successfully imported CCLF9 file: %v", cclf9)
```

bcda/cli.go
```diff
@@ -201,17 +293,64 @@ func sortCCLFFiles(cclf0, cclf8, cclf9 *[]cclfFileMetadata, skipped *int) filepa
...
 		}
-		if metadata.cclfNum == 0 {
-			*cclf0 = append(*cclf0, metadata)
-		} else if metadata.cclfNum == 8 {
-			*cclf8 = append(*cclf8, metadata)
-		} else if metadata.cclfNum == 9 {
-			*cclf9 = append(*cclf9, metadata)
+
+		// if we get multiple sets of files relating to the same aco for attribution purposes, concat the year
+		key := fmt.Sprintf(metadata.acoID+"_%d", metadata.timestamp.Year())
```
embh: I'm not clear on what this is handling-- is it for the claims run-out period?

bcda/cli.go
```diff
@@ -201,17 +293,64 @@ func sortCCLFFiles(cclf0, cclf8, cclf9 *[]cclfFileMetadata, skipped *int) filepa
...
+
+	sc := bufio.NewScanner(file)
+	count := 0
+	var key string
+
+	if fileMetadata.cclfNum == 8 {
+		key = "CCLF8"
+	} else if fileMetadata.cclfNum == 9 {
+		key = "CCLF9"
+	}
```
embh: Should we return an error if it's not 8 or 9?

bcda/cli.go
tbellj: @knollfear mentioned that we could potentially have multiple sets of files relating to attribution. Based on this being true, a regular ole' (key: acoID) here wouldn't work but can change if this isn't a real case.

bcda/cli.go
tbellj: we could add an error here, but my original thinking was that since the validate function only ever gets called with cclf8/cclf9 metadata then we would be okay.

bcda/cli.go
embh: Ah, got it. The run-out period is when that could occur. In that case, we'd want to use the performance year from the filename, indicated by the `Y**` part. Two files from the same month would both have the current year in the timestamp, vs. the performance year being different (e.g., one `Y18` and one `Y19`). 

The CCLF8 branch parses performance year in `getCCLFFileMetadata()`, so maybe it makes sense to copy that function over to this branch?

bcda/cli.go
tbellj: yeah that's a good check, I'll add that 👍

bcda/cli_test.go
```diff
@@ -106,6 +106,86 @@ func (s *CLITestSuite) TestCreateACO() {
 	buf.Reset()
 }
 
+func (s *CLITestSuite) TestImportCCLF0() {
+	assert := assert.New(s.T())
+
+	cclf0filePath := "../shared_files/cclf/T.A0001.ACO.ZC0Y18.D181120.T1000011"
+	cclf0metadata, err := getCCLFFileMetadata(cclf0filePath)
```
embh: I'd recommend creating the `cclfFileMetadata` struct directly here rather than using `getCCLFFileMetadata()` to generate it. If something breaks in that function, it should break only the tests written specifically for it.

bcda/cli_test.go
```diff
@@ -106,6 +106,86 @@ func (s *CLITestSuite) TestCreateACO() {
...
+	cclf8metadata.filePath = cclf8filePath
+	assert.Nil(err)
+
+	cclf9filePath := "../shared_files/cclf/T.A0001.ACO.ZC9Y18.D181120.T1000010"
+	cclf9metadata, err := getCCLFFileMetadata(cclf9filePath)
+	cclf9metadata.filePath = cclf9filePath
+	assert.Nil(err)
+
+	// positive
+	cclfvalidator, err := importCCLF0(cclf0metadata)
```
embh: Same as comment above here-- it will test `validate()` more specifically if we create the exact `cclfFileMetadata` structs and validator maps that we want the test to use.

bcda/cli_test.go
tbellj: yessss 👍

bcda/cli_test.go
```diff
@@ -172,88 +230,103 @@ func (s *CLITestSuite) TestGetCCLFFileMetadata() {
 	assert.Equal(expTime, metadata.timestamp)
 	assert.Nil(err)
 
+	expTime, _ = time.Parse(time.RFC3339, "2019-01-19T20:13:01Z")
+	metadata, err = getCCLFFileMetadata("/path/T.A0002.ACO.ZC0Y18.D190119.T2013010")
+	assert.Equal("test", metadata.env)
+	assert.Equal("A0002", metadata.acoID)
+	assert.Equal(0, metadata.cclfNum)
+	assert.Equal(expTime, metadata.timestamp)
```
embh: Can we also check that `perfYear` gets set to `18`?

PR #239, opened 2019-04-16
BCDA 1164: Remove API dependency on AuthTestSuite

DeirdreHolub:
### Fixes [BCDA-1164](https://jira.cms.gov/browse/BCDA-1164)

Prior to this PR, the API used auth's base test suite as its base test suite, in order to set up the auth system for API testing. This PR provides an independent test suite for the API and a mechanism for it to set up auth for unit test purposes.

### Change Details

- auth uses a single PK pair to use as the server's signing keys when setting up the backend for testing. Previously it created different pk pairs for every test which it ignored unless ResetBackend() was called.
- the alpha backend continues to test that it handles invalid signing keys correctly
- a PK pair suitable for testing is located in shared_files
- the api now uses its own test suites
- testUtils/utils.go provides helper methods for setting and restoring the environment variables for signing keys. This approach allows us not to worry about order of initialization and order of execution issues when running tests.

### Not Changed

- I considered but rejected adding a separate env variable for test keys because that would have required environment aware runtime code.
- The visibility of values in the auth package that no longer have to be exported for the API. I want this PR to focus on just breaking the test suite dependency.

### Security Implications
- This does not change the existing security profile The "test" keys are still RSA keys using RSA-512 for token hashing.
- These keys are not used in upper environments.
- All non-test code remains unchanged.

### Acceptance Validation
All unit, smoke and integration tests run.

### Feedback Requested
Any. Please note the following:

- Tests will fail with a cryptic panic if you do not have the test keys in shared files. If you experience this, rebuild your docker containers
- Similarly, postman tests may fail to get a token if you have not rebuilt your docker containers.


shared_files/README.md
```diff
@@ -0,0 +1,20 @@
+# Shared Files
+
+What's in here and why?
+
+* cclf directory
+  * sample cclf files related to figuring out attribution of beneficiary data
+* synthetic_benficiary_data
```
dhgreene:
```suggestion
* synthetic_beneficiary_data
```

shared_files/README.md
```diff
@@ -0,0 +1,20 @@
...
+  * see README.me in this directory
```
dhgreene:
```suggestion
  * see README.md in this directory
```

PR #240, opened 2019-04-16
BCDA-964/BCDA-958: Save CCLF8 data into database

embh:
### Fixes [BCDA-964](https://jira.cms.gov/browse/BCDA-964), [BCDA-958](https://jira.cms.gov/browse/BCDA-958)
The CLI has a proof of concept for reading CCLF8 files, but does not yet store the file metadata or contents.

### Proposed changes:
Create data model for CCLF8 file metadata and beneficiaries, and save to database as files are read.

### Change Details
* Creates two new data models and tables:
  * `CCLF8File` model - `cclf8_files` table
  * `CCLF8Beneficiary` - `cclf8_beneficiaries` table
* Saves file metadata, one row per file, into `cclf8_files`
* Saves beneficiary HICN and MBI, per row in file, into `cclf8_beneficiaries`

Note that the new tables currently stand alone, and our next steps for attribution will determine how they are used. We will likely need foreign key associations:
* `cclf8_files.aco_cms_id`-`acos.cms_id`
* `cclf8_beneficiaries.beneficiary_id`-`beneficiaries.id`

### Security Implications
This change allows the storage of beneficiary identifiers, HICN and MBI, found in CCLF8 files. ACO IDs and other metadata about the files are also stored.

### Acceptance Validation
`TestImportCCLF8()` in cli_test.go confirms that data can be read and imported from a sample CCLF8 file. The sample file used contains synthetic data.

### Feedback Requested
Any

### DIFF
--- a/bcda/cli.go
+++ b/bcda/cli.go
@@ -3,7 +3,6 @@ package main
 import (
 	"bufio"
 	"bytes"
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -11,8 +10,10 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 
+	"github.com/CMSgov/bcda-app/bcda/database"
 	"github.com/CMSgov/bcda-app/bcda/models"
 )
 
@@ -39,12 +40,13 @@ func createACO(name, cmsID string) (string, error) {
 }
 
 type cclfFileMetadata struct {
+	name      string
 	env       string
 	acoID     string
 	cclfNum   int
+	perfYear  int
 	timestamp time.Time
 	filePath  string
-	perfYear  int
 }
 
 type cclfFileValidator struct {
@@ -110,8 +112,15 @@ func importCCLF0(fileMetadata cclfFileMetadata) (map[string]cclfFileValidator, e
 }
 
 func importCCLF8(fileMetadata cclfFileMetadata) error {
+	fmt.Printf("Importing CCLF8 file %s...\n", fileMetadata.name)
+	log.Infof("Importing CCLF8 file %s...", fileMetadata.name)
+
+	if fileMetadata.filePath == "" {
+		return errors.New("file path must be provided")
+	}
+
 	if fileMetadata.cclfNum != 8 {
-		err := errors.New("invalid CCLF8 file number")
+		err := fmt.Errorf("expected CCLF file number 8, but was %d", fileMetadata.cclfNum)
 		log.Error(err)
 		return err
 	}
@@ -122,7 +131,21 @@ func importCCLF8(fileMetadata cclfFileMetadata) error {
 		return err
 	}
 
-	log.Infof("File contains %s data for ACO %s at %s.\n", fileMetadata.env, fileMetadata.acoID, fileMetadata.timestamp)
+	cclf8File := models.CCLFFile{
+		CCLFNum:         8,
+		Name:            fileMetadata.name,
+		ACOCMSID:        fileMetadata.acoID,
+		Timestamp:       fileMetadata.timestamp,
+		PerformanceYear: fileMetadata.perfYear,
+	}
+
+	db := database.GetGORMDbConnection()
+	defer database.Close(db)
+
+	err = db.Create(&cclf8File).Error
+	if err != nil {
+		return errors.Wrap(err, "could not create CCLF8 file record")
+	}
 
 	const (
 		mbiStart, mbiEnd   = 0, 11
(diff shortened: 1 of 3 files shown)


bcda/models/models.go
```diff
@@ -265,6 +270,32 @@ func AssignAlphaBeneficiaries(db *gorm.DB, aco ACO, acoSize string) error {
 	return db.Exec(s).Error
 }
 
+type CCLF8File struct {
```
knollfear: Is there any reason to make this specifically CCLF8 files?  If we used the CCLFMetaData struct defined in CLI.go (with minor changes) we could store CCLF8 and 9 files in here with minimal effort.

bcda/models/models.go
```diff
@@ -265,6 +270,32 @@ func AssignAlphaBeneficiaries(db *gorm.DB, aco ACO, acoSize string) error {
...
+
+func (*CCLF8File) TableName() string {
+	return "cclf8_files"
+}
+
+// "The MBI has 11 characters, like the Health Insurance Claim Number (HICN), which can have up to 11."
+// https://www.cms.gov/Medicare/New-Medicare-Card/Understanding-the-MBI-with-Format.pdf
+type CCLF8Beneficiary struct {
+	gorm.Model
+	FileID        uint
```
knollfear: I think this is a "belongs to" relationship and should have this format:
File CCLF8File (or CCLFMetadata)
FileID uint

bcda/models/models.go
```diff
@@ -265,6 +270,32 @@ func AssignAlphaBeneficiaries(db *gorm.DB, aco ACO, acoSize string) error {
 	return db.Exec(s).Error
 }
 
+type CCLF8File struct {
+	gorm.Model
+	Name            string
```
knollfear:
```suggestion
	Name            string `gorm:"not null;unique"`
```
This should always exist and will be unique.  If it isn't we have a duplicate import and should throw an error probably.

bcda/models/models.go
embh: Good idea. Since this applies to CCLF9 also, @tbellj, any thoughts? Do you want to do that in your PR first?

bcda/models/models.go
embh: The beneficiaries do belong to a file, but I don't know that it's necessary to make a relationship to that type here just yet, before we know how we're using these models. Would having the file as a field of `CCLF8Beneficiary` impact the database tables or constraints? There is a foreign key for `cclf8_beneficiaries.file_id` to `cclf8_files.id`.

bcda/cli.go
```diff
@@ -117,34 +148,41 @@ func importCCLF9(fileMetadata cclfFileMetadata) error {
...
-	t, err := time.Parse("D060102.T150405", filenameDate)
-	if err != nil || t.IsZero() {
-		return metadata, fmt.Errorf("failed to parse date '%s' from filename", filenameDate)
+	cclfNum, err := strconv.Atoi(filenameMatches[3])
+	if err != nil {
+		return metadata, err
 	}
 
-	cclfNum, err := strconv.Atoi(filenameMatches[3])
+	perfYear, err := strconv.Atoi(filenameMatches[4])
```
tbellj: does this return just the year or the entire date string D190417?

embh: It grabs the two digits from the `Y**` part of the filename, so for 2019 files it would get just `19`.

bcda/models/models.go
knollfear: I think we need both.  I haven't done it in practice, but I think it would let us do this

db.Model(&user).Related(&profile)
from http://doc.gorm.io/associations.html
or this:
// Query - Find out all related associations
db.Model(&user).Association("Languages").Find(&languages)
from http://doc.gorm.io/associations.html#association-mode

Maybe we can just add it in when the need arises though.

bcda/models/models.go
tbellj: hrmmm now thinking after seeing these new changes that it'll make more sense if the cclf0 validation pr got merged first. This is a lot more changes since the last time I looked at it :)

embh: Kept `cclfMetadata` as a separate struct, but made a shared `CCLFFile` model for 8 and 9.

bcda/cli_test.go
```diff
@@ -167,17 +167,67 @@ func (s *CLITestSuite) TestValidate() {
...
+	assert := assert.New(s.T())
 
 	metadata.cclfNum = 9
+	metadata.filePath = "../shared_files/cclf/T.A0001.ACO.ZC8Y18.D181120.T1000009"
+	err := importCCLF8(metadata)
+	assert.NotNil(err)
+	assert.EqualError(err, "expected CCLF file number 8, but was 9")
+
+	metadata.cclfNum = 8
+	metadata.filePath = ""
```
tbellj: metadata.filePath="", is this a real use case?

embh: Shouldn't happen, but we have error handling on it, so wanted to demonstrate that it's functioning.

tbellj: yep makes sense. I ended up removing the error handling for that from my code and replaced it with the empty struct check instead.

PR #241, opened 2019-04-18
BCDA-1148: add models for group management features

alexhova:
### Fixes [BCDA-1148](https://jira.cms.gov/browse/BCDA-1148)
We need a way to store the data about groups in the db and a way to model the data in the app.

### Proposed changes:
- create models for groups to be stored in the db
- create models to unmarshal group json object
- add test for example usage

### Security Implications
There shouldn't be any security impact on this one.  Groups don't contain any sensitive data.

### Feedback Requested
Check out the test I added.  Is this the best way to use these json object?  What is the best way to get the raw json into a usable struct?


alexhova:
> This looks like it works, and is optimized for unstructured data. To the degree that we are dealing with known types of data and their relationships, though, I'd rather take full advantage of our RDMS and use traditional tables/columns be able to have constraints (rather than storing JSON in a single column).
> 
> * There should be a limited number of scopes and resources, for instance; should we put business logic at the database level enforcing that?
> * When we're testing for authorization, it would be nice to not have to jump through hoops to test for a user's group membership in potentially multiple groups.
> 
> @alexhova How much different would it be to change these types to `db.Model`s?
> @DeirdreHolub How likely is it that the structure of what we're storing will change frequently?
> 
> Looking to spark conversation, not discredit the good work here.

the original idea was to keep it as a json object so we could iterate on the structure if it changes.  and when we're more certain on the structure to create those relations.

however, now I'm thinking after writing the test here that this is a bit complicated to handle in the app.  although we could write helper functions that are used to access and update these fields within the objects, it will still need to be changed when we eventually switch to utilizing the multiple.  i don't think there is any advantage to having this in a nosql-style storage and may actually cause some data redundancy.  if we update a `system` for instance, do we have to update it in each record?  how will we even know where those are?
@DeirdreHolub thoughts?

DeirdreHolub:
I expect changes to this data structure. We have no feedback from potential users of this API yet, and that feedback may prompt change. We also have open questions in the tech spec that may prompt change. It is possible, for example, that this json structure will turn out to be instructions we use to drive external object creation. I have a bias against creating tables (types) and relations until I have some level of confidence in their stability, and I haven't gotten there yet.

To Alex's question  

> if we update a system for instance, do we have to update it in each record? how will we even know where those are?

The API's PUT method is specifically stated to support only replacing the entire json structure because of the uncertainties described above.

We need to discuss internally what increases confidence that we have the right data structure. Among the evidence that I look for is something else needing the data. That encryption keys for ACOs is coming our way will probably lead to a system table, for example.

Bottom line, I think these are great comments and welcome thinking-out-loud. I am inclined to go with what Alex has for now, understanding that it won't remain this way for very long, but allows us to progress with getting the API methods in place. There's a built in assumption here that the API methods won't be changing even if we change the json structure.

bcda/models/models_test.go
```diff
@@ -277,3 +278,28 @@ func (s *ModelsTestSuite) TestGetBeneficiaryIDs() {
 	assert.Equal(100, len(beneficiaryIDs))
 
 }
+
+func (s *ModelsTestSuite) TestGroupStructs() {
```
DeirdreHolub: The essential things the testing needs to do: 
(a) unmarshal a json string into the structs (as it will do in an api handler)
(b) save the json to the db
(c) get the json out of the db into structs
(d) make sure the input of a matches the output of (c) (marshal the output from (c) and compare strings with (a)'s input?) 
I see (b) and (c), but not (a) and (d)

bcda/models/models.go
```diff
@@ -270,6 +272,32 @@ func AssignAlphaBeneficiaries(db *gorm.DB, aco ACO, acoSize string) error {
 	return db.Exec(s).Error
 }
 
+type Group struct {
+	ID   string         `gorm:"primary_key" json:"id"`
+	Data postgres.Jsonb `json:"data"`
```
DeirdreHolub: I believe it is an option to use a string here, since we are validating the structure of the json before it goes in and when it comes out simply by marshaling it in/out of structs and are not planning on using postgres's jsonb query capabilities.

DeirdreHolub:
> What is the best way to get the raw json into a usable struct?

see bcda/auth/mokta.go for an example of an annotated struct that marshals / unmarshals from raw string json.

PR #242, opened 2019-04-19
BCDA-638 Documentation: add description of credentials

dhgreene:
### Fixes [BCDA-638](https://jira.cms.gov/browse/BCDA-638)
Up to this point we've been distributing long-lived access tokens.  Our auth providers are now ready to use credentials to create short-lived access tokens via `/auth/token`.  The goal of this PR is to document that change in the user guide, encryption walkthrough, and anywhere else needed.

### Proposed changes:
* Publish client ID and client secret instead of a token
* Document new auth steps, including screenshots
* Change terminology (`s/token/credentials/`) where needed

### Change Details
* Inserting screenshots in the encryption walkthrough required re-numbering.  To avoid potential cache issues, the screenshot naming scheme was also changed.
* A typo in the `pip install` instruction was corrected
* Some additional examples were added to the troubleshooting section

### Security Implications
There is a minor change in security: we're publishing credentials for `opensbx` instead of a token.  While no additional access is granted, the credentials allow for a new action (token creation).  This is a planned change, and brings us up to the point we had originally targeted for the release of `opensbx`.

### Acceptance Validation
Screen Shot 2019-04-19 at 2 49 56 PM
Screen Shot 2019-04-19 at 2 50 29 PM
Screen Shot 2019-04-19 at 2 50 41 PM

### Feedback Requested
* Typos, inaccuracies
* Visual distractions (for instance, line wrapping on the client secret and Basic authentication header would be ideal)
* Anything else!
* Note that @epfrankle will be reviewing the edits next sprint

PR #244, opened 2019-04-23
BCDA-1248 Documentation: Semantic links

dhgreene:
### Fixes [BCDA-1248](https://jira.cms.gov/browse/BCDA-1248)
This is a follow-on PR from #242 (BCDA-648), which updated the static site to document using client credentials with the `/auth/token` endpoint.  This PR includes edits suggested by our research team to clarify and increase accessibility.  

### Proposed changes:
- Use full URL's in text (avoid e.g. `/auth/token` in favor of `https://sandbox.bcda.cms.gov/auth/token`)
- Link to meaningful descriptions, not "find stuff **`here`**"

### Security Implications
No additional credentials or PII are exposed, and no additional functionality is documented.

### Acceptance Validation
Full link for tokens
Semantic link for getting token
Use your credentials link

### Feedback Requested
- Do the screenshots above accurately capture the intended changes?
- Are there additional changes that should be included?


epfrankle:
Question on the "getting a token" link - where does that link go now? Is it to obtain credentials? Should that be a first step before "click the authorize button"?

dhgreene:
> Question on the "getting a token" link - where does that link go now? Is it to obtain credentials? 

It goes to the section of the user guide that publishes the credentials.

> Should that be a first step before "click the authorize button"?

The very first step in the "Gathering the tools" section has the same link, asking them to have the credentials ready for a later step.  Either @knollfear or I felt it was important to link again here, at the point the credentials are actually used.  I'm happy to update it again to reduce awkwardness.

knollfear:
> > Question on the "getting a token" link - where does that link go now? Is it to obtain credentials?
> 
> It goes to the section of the user guide that publishes the credentials.
> 
> > Should that be a first step before "click the authorize button"?
> 
> The very first step in the "Gathering the tools" section has the same link, asking them to have the credentials ready for a later step. Either @knollfear or I felt it was important to link again here, at the point the credentials are actually used. @epfrankle I'm happy to update it again to reduce awkwardness.

I know I have been adamantly opposed to opening internal (same domain) links in a new window, however, in this case we want people to pop out, grab something and come back.  I think that might justify opening it in a new window.

dhgreene:
> . . . in this case we want people to pop out, grab something and come back. I think that might justify opening it in a new window.

@knollfear OK, anything that makes it better.  Done.

epfrankle:
For the "use your credentials" link, another thought: what if we linked "use your credentials from the previous step" and finished the sentence with "to get an access token in this step" - unlinked. So:

[Use your credentials from the previous step](https://sandbox.bcda.cms.gov/user_guide.html#authentication-and-authorization) to get an access token in this step. 

Thoughts?

dhgreene:
> [Use your credentials from the previous step](https://sandbox.bcda.cms.gov/user_guide.html#authentication-and-authorization) to get an access token in this step.

@epfrankle I like it better than I expected.  😸 That's now in the PR, and the screenshot above is updated.

PR #246, opened 2019-04-25
BCDA-1246 add public encryption key model and db table

alexhova:
### Fixes [BCDA-1246](https://jira.cms.gov/browse/BCDA-1246)
we need a model and a db table to store public encryption keys for systems integrated with us.  this PR adds that model

### Proposed changes:
- add the EncryptionKey gorm model
- add it to the automigrate so it is created as a table in the db

### Security Implications
no security concerns here.  we're just adding a db table.  it is not being used yet


dhgreene:
I think we should also add tests for any new tables.

bcda/models/models.go
```diff
@@ -340,3 +349,9 @@ type jobEnqueueArgs struct {
 	// TODO: remove `Encrypt` when file encryption disable functionality is ready to be deprecated
 	Encrypt bool
 }
+
+type EncryptionKey struct {
+	gorm.Model
+	Body     string `json:"body"`
+	SystemID uint   `json:"system_id"`
```
DeirdreHolub: Need to add an FK reference to System here. It would be another field in the struct (above the SystemID), like so:
	System         models.System 
 `gorm:"foreignkey:SystemID;association_foreignkey:ID"` 

See how auth.Token references ACO as a foreign key; also Job and JobKey

bcda/models/models.go
```diff
@@ -304,10 +310,13 @@ type Resource struct {
...
-	SoftwareID string `json:"software_id"`
-	ClientName string `json:"client_name"`
-	ClientURI  string `json:"client_uri"`
+	gorm.Model
+	GroupID        string          `json:"group_id"`
+	ClientID       string          `json:"client_id"`
+	SoftwareID     string          `json:"software_id"`
+	ClientName     string          `json:"client_name"`
+	ClientURI      string          `json:"client_uri"`
+	EncryptionKeys []EncryptionKey `gorm:"foreignkey:SystemID" json:"encryption_keys"`
```
DeirdreHolub: should be []*EncryptionKey, with no annotation. See auth.Token model Token/ACO and bcda.models Job/JobKey

bcda/models/models_test.go
```diff
@@ -288,19 +288,28 @@ func (s *ModelsTestSuite) TestGroupStructs() {
 		Resource{ID: "xxx", Name: "BCDA API", Scopes: []string{"bcda-api"}},
 		Resource{ID: "eft", Name: "EFT CCLF", Scopes: []string{"eft-app:download", "eft-data:read"}},
 	}
-	systems := []System{
-		System{ClientID: "4tuhiOIFIwriIOH3zn", SoftwareID: "4NRB1-0XZABZI9E6-5SM3R", ClientName: "ACO System A", ClientURI: "https://www.acocorpsite.com"},
-	}
-	groupData := GroupData{Name: name, Users: users, Scopes: scopes, Resources: resources, Systems: systems}
+	groupData := GroupData{Name: name, Users: users, Scopes: scopes, Resources: resources}
```
DeirdreHolub: Where is the systems object of the group in the json? Even though systems have their own separate table, they will still be in the json request that the API has to deal with, need to be decoded properly, and most likely be stored in the json column. In other words, we will want to keep the last group definition we got from the user via the API. 

I think this test would be easier to follow if you embedded the json out of the tech spec as a string, converted it to bytes, unmarsheled it into a struct and carried on from there. That is what will happen in the API code.

alexhova: `Group` is what is being stored in the database.  so `Systems` are an array in the `Group` json

alexhova: it's there if you print a `Group`

alexhova: >I think this test would be easier to follow if you embedded the json out of the tech spec as a string, converted it to bytes, unmarsheled it into a struct and carried on from there. That is what will happen in the API code.

That's more complex now because we decided to go with the jsonb field.  so it is actually different than the spec already.

alexhova: but i will try it with the `data` field we added

PR #247, opened 2019-04-26
BCDA-1225 & BCDA-1200 Feature: Cleanup process for cclf files

tbellj:
Fixes [BCDA-1225](https://jira.cms.gov/browse/BCDA-1225)
Fixes [BCDA-1200](https://jira.cms.gov/browse/BCDA-1200)

Problem:
Before we had no way of cleaning/deleting files after ingestion, potentially sitting there to retry forever. Successfully imported files were also lingering around.

Proposed changes:
Successfully imported files now move to the pending deletion dir via an env var.
Files that fail to import are now moved to the pending deletion dir only if they are beyond our set threshold (within 24 hours) via an env var. 

Change Details:
.gitignore - added ignore for pending deletion dir
bcda/cli.go - added cleanup function
bcda/cli_test.go - test cases for the cleanup function

Security Implications:
**Yes** this code deals with PII and PHI removal. It is critical that this works so we can maintain our good security posture going forward.

Acceptance Validation:
Yes
Screenshot?? Files are moving around not sure what the screenshot would look like

Feedback Requested
Any and all


.gitignore
```diff
@@ -15,4 +15,5 @@ bcda-site-static/.jekyll-metadata
 bcdaworker/tmpdata/*
 .env.sh
 decryption_utils/C#/bin
-decryption_utils/C#/obj
\ No newline at end of file
+decryption_utils/C#/obj
+bcda/pending_delete_dir/*
```
msnook: Should this be added to dockerignore too?

bcda/cli.go
```diff
@@ -40,30 +40,34 @@ func createACO(name, cmsID string) (string, error) {
...
 
 type cclfFileValidator struct {
 	totalRecordCount int
 	maxRecordLength  int
 }
 
-func importCCLF0(fileMetadata cclfFileMetadata) (map[string]cclfFileValidator, error) {
-	fmt.Printf("Importing CCLF0 file %s...\n", fileMetadata.name)
-	log.Infof("Importing CCLF0 file %s...", fileMetadata.name)
+const deleteThresholdHr = 8
```
msnook: Can you add a comment indicating exactly what this represents?  Also, since it's a constant, it should be all caps I believe.

bcda/cli.go
```diff
@@ -466,3 +482,29 @@ func deleteDirectoryContents(dirToDelete string) (filesDeleted int, err error) {
...
+
+func cleanupCCLF(cclfmap map[string][]*cclfFileMetadata) {
+	for _, cclflist := range cclfmap {
+		for _, cclf := range cclflist {
+			newpath := fmt.Sprintf("%s/%s", os.Getenv("PENDING_DELETION_DIR"), cclf.name)
+			if !cclf.imported {
+				// check the timestamp on the failed files
+				elapsed := time.Since(cclf.deliveryDate).Hours()
+				if int(elapsed) > (24 - deleteThresholdHr) {
+					err := os.Rename(cclf.filePath, newpath)
```
msnook: I think we should log this.  This is an indication that the file is doomed (we retried it for a 24 hour period, and still it won't ingest - so we need to specify in logs that we are deleting a file that was never ingested).

bcda/cli.go
embh: Looks like constants follow the usual naming format in Go, not all caps like in some languages. https://stackoverflow.com/questions/22688906/go-naming-conventions-for-const

bcda/cli.go
```diff
@@ -324,6 +327,7 @@ func importCCLFDirectory(filePath string) (success, failure, skipped int, err er
 			continue
 		} else {
 			log.Infof("Successfully imported CCLF0 file: %v", cclf0)
+			cclf0.imported = true
```
msnook: What happens if the CCLF0 file imported successfully, but the 8 and 9 were not yet present on our system (they hadn't been sent yet)?  If that happens, then because CCLF0 has `imported=true`, it will be cleaned up.  The 8 and 9 will never be retried properly on the next ingestion, because the CCLF0 was already removed.  I think they should be marked as a group, instead of individually.

bcda/cli.go
```diff
@@ -349,10 +354,13 @@ func importCCLFDirectory(filePath string) (success, failure, skipped int, err er
 				failure++
 			} else {
 				log.Infof(fmt.Sprintf("Successfully imported CCLF9 file: %v", cclf9))
+				cclf9.imported = true
 				success++
 			}
 		}
 	}
```
knollfear: in reference to the CCLF0 comment above, we could mark CCLF0 imported = true here if CCLF8 AND CCLF9 are imported

cclf0.imported = cclf8.imported && cclf9.imported

bcda/cli.go
```diff
@@ -373,26 +381,34 @@ func sortCCLFFiles(cclfmap *map[string][]cclfFileMetadata, skipped *int) filepat
 		}
 		metadata, err := getCCLFFileMetadata(info.Name())
 		metadata.filePath = path
+		metadata.deliveryDate = info.ModTime()
```
knollfear: I think we need to be checking this in the existing tests.  Maybe even touch a file and see if it changes, open it and see that it doesn't?

bcda/cli.go
```diff
@@ -466,3 +482,29 @@ func deleteDirectoryContents(dirToDelete string) (filesDeleted int, err error) {
...
 }
+
+func cleanupCCLF(cclfmap map[string][]*cclfFileMetadata) {
+	for _, cclflist := range cclfmap {
+		for _, cclf := range cclflist {
+			newpath := fmt.Sprintf("%s/%s", os.Getenv("PENDING_DELETION_DIR"), cclf.name)
+			if !cclf.imported {
+				// check the timestamp on the failed files
+				elapsed := time.Since(cclf.deliveryDate).Hours()
+				if int(elapsed) > (24 - deleteThresholdHr) {
```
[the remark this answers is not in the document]
msnook: Yeah, I think this is an artifact of me and Terrance possibly over-engineering this in our initial discussion.  I agree with Knoll that it should be simplified.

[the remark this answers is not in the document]
tbellj: yep we definitely over-engineered this lol. I can't even think of why we decided this

bcda/cli.go
tbellj: exactly

bcda/cli.go
tbellj: good catch!

bcda/cli.go
```diff
@@ -117,16 +121,16 @@ func importCCLF0(fileMetadata cclfFileMetadata) (map[string]cclfFileValidator, e
...
-
-	if (cclfFileMetadata{}) == fileMetadata {
+func importCCLF8(fileMetadata *cclfFileMetadata) error {
+	if fileMetadata == nil {
 		err := errors.New("file CCLF8 not found")
 		log.Error(err)
 		return err
 	}
 
+	fmt.Printf("Importing CCLF8 file %s...\n", fileMetadata.name)
```
embh: Could we have the stdout and log messages either before or after all the error handling, instead of in between?

bcda/cli_test.go
```diff
@@ -521,3 +588,36 @@ func (s *CLITestSuite) makeDirToDelete(filePath string) {
...
+		log.Fatal(err)
+	}
+	cclfDeletion := os.Getenv("PENDING_DELETION_DIR")
+	err = os.MkdirAll(cclfDeletion, 0744)
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+
+func (s *CLITestSuite) resetFiles(resetPath string) {
```
embh: This is great. Hooray for cleaning up after our tests!

bcda/cli_test.go
```diff
@@ -447,7 +453,68 @@ func (s *CLITestSuite) TestImportCCLFDirectory() {
...
+		timestamp: fileTime,
+		filePath:  "../shared_files/cclf/T.A0001.ACO.ZC9Y18.D181120.T1000010",
+		imported:  true,
+	}
+	cclfmap["A0001_18"] = []*cclfFileMetadata{cclf0metadata, cclf8metadata, cclf9metadata}
+	cleanupCCLF(cclfmap)
+
+	files, err := ioutil.ReadDir(os.Getenv("PENDING_DELETION_DIR"))
+	if err != nil {
+		log.Fatal(err)
```
embh: Does using `log.Fatal()` here kill the whole run of all unit tests if there's an error? Might want to try https://godoc.org/github.com/stretchr/testify/require#FailNow

bcda/cli.go
tbellj: so the idea is to have it before the error handling, I remember we talked about that specifically in the code before. But we can't log fileMetadata.name if fileMetadata is null hence why it's after the null check.

bcda/cli_test.go
tbellj: Is this something we're going to implement in all of our testing code? Shooting for consistency here.

bcda/cli_test.go
```diff
@@ -576,6 +599,67 @@ func (s *CLITestSuite) TestImportCCLFDirectory() {
...
+		timestamp: fileTime,
+		filePath:  "../shared_files/cclf/T.A0001.ACO.ZC9Y18.D181120.T1000010",
+		imported:  true,
+	}
+	cclfmap["A0001_18"] = []*cclfFileMetadata{cclf0metadata, cclf8metadata, cclf9metadata}
+	cleanupCCLF(cclfmap)
+
+	files, err := ioutil.ReadDir(os.Getenv("PENDING_DELETION_DIR"))
+	if err != nil {
+		s.FailNow("failed to get the PENDING_DELETION_DIR env variable", err)
```
knollfear: Is this really true?  What if I set the env var to this\directory\isnt\real?

PR #248, opened 2019-04-26
BCDA-1223: Ingest Zipped CCLF Files

embh:
### Fixes [BCDA-1223](https://jira.cms.gov/browse/BCDA-1223)
The BCDA CLI CCLF file ingestion process currently expects text files, but ACO-OS will provide zip files.

### Proposed changes:
Read CCLF files as zipped archives.

### Change Details
* Updated CCLF0, CCLF8, and CCLF9 import logic to unzip files before reading enclosed text files
* Converted files used for testing to zip files containing the old text files
* Added `importCCLF()` function to reduce duplicated code in `importCCLF8()` and `importCCLF9()`
* Added tests for the possibility of archive contents being split into multiple files (e.g., CCLF8 archive containing two files of CCLF8 data)

### Security Implications
This adds a file unzip step to the process by which BCDA ingests CCLF files. The test files have been zipped, but contain the same synthetic data.

### Acceptance Validation
Tests pass, including new tests on archives where synthetic data has been split into two text files.

### Feedback Requested
Any


bcda/cli.go
```diff
@@ -118,94 +131,87 @@ func importCCLF0(fileMetadata cclfFileMetadata) (map[string]cclfFileValidator, e
...
 
-	if fileMetadata.cclfNum != 9 {
-		err := fmt.Errorf("expected CCLF file number 9, but was %d", fileMetadata.cclfNum)
+	r, err := zip.OpenReader(filepath.Clean(fileMetadata.filePath))
+	if err != nil {
+		err := errors.Wrapf(err, "could not read CCLF%d archive %s", fileMetadata.cclfNum, fileMetadata.name)
 		log.Error(err)
 		return err
 	}
+	defer r.Close()
```
knollfear: should this be above the error handling?  If we get an error r won't be closed

bcda/cli.go
```diff
@@ -215,38 +221,34 @@ func importCCLF9(fileMetadata cclfFileMetadata) error {
...
-				PrevNum:       string(b[prevIDStart:prevIDEnd]),
-				PrevsEfctDt:   string(b[prevIDEffDateStart:prevIDEffDateEnd]),
-				PrevsObsltDt:  string(b[prevIDObsDateStart:prevIDObsDateEnd]),
-			}
-			err = db.Create(&cclf9).Error
-			if err != nil {
-				return errors.Wrap(err, "could not create CCLF9 cross reference record")
+	for i, f := range r.File {
+		log.Infof("Reading file #%d from archive %s", i, fileMetadata.name)
+		rc, err := f.Open()
```
knollfear: defer f.close()?

bcda/cli.go
```diff
@@ -418,24 +411,43 @@ func validate(fileMetadata cclfFileMetadata, cclfFileValidator map[string]cclfFi
...
-			count++
+	count := 0
+	validator := cclfFileValidator[key]
 
-			// currently only errors if there are more records than we expect.
-			if count > validator.totalRecordCount {
-				err := fmt.Errorf("maximum record count reached for file %s, Expected record count: %d, Actual record count: %d ", key, validator.totalRecordCount, count)
+	for i, f := range r.File {
+		log.Infof("Reading file #%d from archive %s", i, fileMetadata.name)
+		rc, err := f.Open()
```
knollfear: defer f.close()

bcda/cli.go
```diff
@@ -77,26 +81,35 @@ func importCCLF0(fileMetadata cclfFileMetadata) (map[string]cclfFileValidator, e
...
-					log.Error(err)
-					return nil, err
-				}
-				length, err := strconv.Atoi(string(bytes.TrimSpace(b[recordLengthStart:recordLengthEnd])))
-				if err != nil {
-					log.Error(err)
-					return nil, err
+	for i, f := range r.File {
+		log.Infof("Reading file #%d from archive %s", i, fileMetadata.name)
+		rc, err := f.Open()
```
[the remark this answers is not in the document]
embh: Good catch! It'll be `rc.Close()` because `Open()` returns a `ReadCloser` (https://golang.org/pkg/io/#ReadCloser) for the file.

bcda/cli.go
embh: I think the Go convention is to return nil for the thing being created if there's an error, so there wouldn't be a reader we could close. Looks to be how they handle it here: https://golang.org/src/archive/zip/reader.go?s=911:960#L40

bcda/cli.go
```diff
@@ -77,26 +81,35 @@ func importCCLF0(fileMetadata cclfFileMetadata) (map[string]cclfFileValidator, e
...
+			err = errors.Wrapf(err, "could not read file %s in CCLF0 archive %s", f.Name, fileMetadata.name)
+			log.Error(err)
+			return nil, err
+		}
+		sc := bufio.NewScanner(rc)
+		for sc.Scan() {
+			b := sc.Bytes()
+			if len(bytes.TrimSpace(b)) > 0 {
+				filetype := string(bytes.TrimSpace(b[fileNumStart:fileNumEnd]))
+				if filetype == "CCLF8" || filetype == "CCLF9" {
```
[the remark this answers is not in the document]
embh: That's a good point. Didn't we have a ticket floating around about validation we wanted to add in the future? Would be great to get it documented there or in a new ticket.

PR #251, opened 2019-05-01
BCDA-1130 Auth: increase security of stored hashes

dhgreene:
### Fixes [BCDA-1130](https://jira.cms.gov/browse/BCDA-1130)
We are following the recommendations in [NIST 800-63b](https://pages.nist.gov/800-63-3/sp800-63b.html#sec5) to decrease the likelihood of a successful attack by implementing the pbkdf2 key-derivation protocol when storing our secrets. This adds a random "salt" to the secret and drastically increases the time to brute-force an attack on the stored secrets by iterating the hash more than 10,000 times.

### Proposed changes:
- Import `golang.org/x/crypto/pbkdf2`
- Use the new library when creating/storing hashes
- Tune the number of iterations as high as possible
- Document and test to ensure we stay safe and compatible

### Data Migration Strategy
No changes are made to the database in this code.  However, it does break any existing client secrets. 
 At the moment client id's & secrets should only exist in the `dev` and `test` environments.  Except for a set of credentials used by Jenkins, they should be ephemeral.  The migration strategy is therefore:
- Delete any existing client ID's and secrets from the `acos` table
- Deploy this code
- Set up new credentials for Jenkins to use in smoke and postman tests in `dev` and `test`

### Security Implications
See SIA for details.  Note that we are comfortable with the idea that alterations to this code (including number of iterations) will force credential rotation in production.

### Testing
- Make sure to run `dep ensure` before running tests
- If you wish to see the actual hash execution time on your computer
   - Set the format for `gotestsum` in `/unit_test.sh`: Line 27: . . . `gotestsum --format standard-verbose` . . .
   - Search the output of `make unit-test` for `Hash creation`

### Feedback Requested
- Are there additional areas of code which could use comments?
- Do any existing comments need to be clarified?
- Are there additional unit, integration, smoke, or other tests that we should create?
- Should we adjust the initial number of hash iterations?  The security recommendation is to have them as high as we can tolerate.  This code aims to add 1s to all calls to `/auth/token`--can we tolerate more?


bcda/auth/backend.go
```diff
@@ -23,17 +33,38 @@ func init() {
 // Hash is a cryptographically hashed string
 type Hash string
 
-// NewHash creates a hashed string from a source string, return it as a new Hash value.  While this is not
-// intended for production password hashing, as a matter of principle we intend to add a salt and implement
-// bcrypt or another hash appropriate for storing secrets in ticket BCDA-1130.
-func NewHash(source string) Hash {
-	sum := sha256.Sum256([]byte(source))
-	return Hash(fmt.Sprintf("%x", sum))
+// NewHash creates a hashed string from a source string, return it as a new Hash value.
```
knollfear: // NewHash creates a Hash value from a source string
// The HashValue consists of the salt and hash separated by a colon  ( : )
// If the source of randomness fails it returns an error.

bcda/auth/backend.go
```diff
@@ -23,17 +33,38 @@ func init() {
...
+
+	h := pbkdf2.Key([]byte(source), salt, hashIter, hashKeyLen, sha512.New)
+	return Hash(fmt.Sprintf("%s:%s", base64.StdEncoding.EncodeToString(salt), base64.StdEncoding.EncodeToString(h))), nil
 }
 
 // IsHashOf accepts an unhashed string, which it first hashes and then compares to itself
 func (h Hash) IsHashOf(source string) bool {
-	return h == NewHash(source)
+	// Avoid comparing with an empty source so that a hash of an empty string is never successful
+	if source == "" {
```
knollfear: I get that it is invalid and unhelpful to hash "", but what would be the harm in just hashing ""?

bcda/auth/backend_test.go
```diff
@@ -123,13 +125,52 @@ func (s *BackendTestSuite) TestInitAuthBackend() {
...
+	hash := auth.Hash("RFbxyt3Fa/zQTZ324QbgTRv/yj5cXdaq5CdVDBvliN8=:8c7EH5a/I76UFgwO0F8GRTD1pfqDvrbrwN5qbuvKUZO90BuBm4WQFQynhnKE1gtvmF5xJlyoEPeqnK44TfSMuQ==")
+	assert.True(s.T(), hash.IsHashOf(uuidString), "Possible change in hashing parameters or algorithm.  Known input/output does not match.  Merging this code will result in invalidating credentials.")
+}
+
+func (s *BackendTestSuite) TestHashEmpty() {
+	hash, err := auth.NewHash("")
+	assert.Nil(s.T(), err)
+	assert.False(s.T(), hash.IsHashOf(""))
+}
+
```
knollfear: Can we get a test for an invalid source hash?  For example "RFbxyt3Fa/zQTZ324QbgTRv/yj5cXdaq5CdVDBvliN8=:8c7EH5a/I76UFgwO0F8GRTD1pfqDvrbrwN5qbuvKUZO90BuB:m4WQFQynhnKE1gtvmF5xJlyoEPeqnK44TfSMuQ==" will fail because it has 3 sections instead fo 2 when split on the colon.  sb an easy way to bump up that coverage number.

bcda/auth/backend.go
dhgreene: Because it's being used to verify credentials, this is the last line of defense in case some weird error mode allowed a hashed "" value to be saved as the client secret.

bcda/auth/backend.go
```diff
@@ -23,17 +33,40 @@ func init() {
...
-// intended for production password hashing, as a matter of principle we intend to add a salt and implement
-// bcrypt or another hash appropriate for storing secrets in ticket BCDA-1130.
-func NewHash(source string) Hash {
-	sum := sha256.Sum256([]byte(source))
-	return Hash(fmt.Sprintf("%x", sum))
+// NewHash creates a Hash value from a source string
+// The HashValue consists of the salt and hash separated by a colon ( : )
+// If the source of randomness fails it returns an error.
+func NewHash(source string) (Hash, error) {
+	salt := make([]byte, saltSize)
```
knollfear: If it is invalid to hash "", and it won't match on the hash compare, why hash "" at all?  Can this just short circuit and return "invalid source" error or similar?

dhgreene: That's reasonable in most cases.  However, because this is a core function to our authentication, what I'm trying to cover is a future case if misunderstanding, or the case of unexpected failure in our the libraries, or some other failure case that leads to inadvertently passing "" to `pbkdf2.Key()`.  So by definition we can't short-circuit it completely at the input stage.  Where we call the `NewHash()` at the moment, we're providing a random secret anyway.  But I'll add a check here too.  😃

bcda/auth/backend.go
```diff
@@ -1,17 +1,28 @@
...
+	"strings"
 
 	"github.com/dgrijalva/jwt-go"
 	log "github.com/sirupsen/logrus"
+	"golang.org/x/crypto/pbkdf2"
 
 	"github.com/CMSgov/bcda-app/bcda/utils"
 )
 
+// The time for hash comparison should be about 1s.  Increase hashIter if this is significantly faster in production.
```
DeirdreHolub: How are we going to know anything about the performance characteristics of the hash at runtime? I didn't see any logging of runtime performance measurements, such that we could set up a splunk alert if the measurement was outside of an acceptable range.

dhgreene: Good thought.  The speed check has been moved from a unit test to a `warning` level log which we can alert on in production environments.

bcda/auth/backend.go
```diff
@@ -48,7 +50,15 @@ func NewHash(source string) (Hash, error) {
 		return Hash(""), err
 	}
 
+	start := time.Now()
 	h := pbkdf2.Key([]byte(source), salt, hashIter, hashKeyLen, sha512.New)
+	hashCreationTime := time.Since(start)
+
+	if hashCreationTime < hashMinTime {
```
DeirdreHolub: What if it's taking too long? I think we should have an upper bound here as well. Maybe 3 sec?

bcda/auth/backend_test.go
```diff
@@ -140,28 +137,9 @@ func (s *BackendTestSuite) TestHashUnique() {
 	assert.NotEqual(s.T(), hash1.String(), hash2.String())
 }
 
-func (s *BackendTestSuite) TestHashTime() {
```
DeirdreHolub: We should still be testing to make sure we get the log alerts if we exceed either bounds.

dhgreene: Yes, now that both bounds create logs, we can add Splunk alerts for both cases.

PR #252, opened 2019-05-01
BCDA-1232 Implement mechanism to get BBIdentifier

knollfear:
### Fixes [BCDA-1232](https://jira.cms.gov/browse/BCDA-1232)
also resolves 1230 which had a strong dependency on this ticket.

CCLF8/9 files contain only a HICN (or MBI) as an identifier.  In order to get the correct record from Blue Button we need to make a request to the patient end point and provide a hash of the HICN.  THis PR creates a method for doing that.  It does not implement that method anywhere, that will be handled by BCDA-1233 
### Proposed changes:

- Added method to CCLFBeneficiary GetBlueButtonIdentifier
- Moved mockBlueButtonClient out of bcdaworker/maintest into testutils/client
- Added function to BB client to hash a string(HICN)
- Added Patient Struct to model response from BB for JSON unmarshalling

### Change Details

MockBlueButtonClient is a useful tool that will likely be used in other places so I felt it was reasonable to move it to testUtils.  I didn't like exposing MockBLueButtonClient.GetData because the method is private on actual bluebuttonclient, but it needs to be exported in order to work.

Patient Struct isn't strictly necessary, but I was curious how complicated it would be to cast BB responses as data structs and manipulate them.  Based on this work it is not particularly difficult and we don't need to shy away from it going forward.

GetBlueButtonIdentifier does a fair amount of work, but I wanted to make it foolproof for anyone trying to use it.

### Security Implications

This PR adds PBDKF2 which is covered under a previous [SIA](https://confluence.cms.gov/pages/viewpage.action?spaceKey=BCDA&title=PBKDF2+SIA). 

The hashing algorithm touches PII, but only for the purpose of concealing it in transit. 

### Acceptance Validation

This is a function that isn't used yet so it relies heavily on the tests for acceptance.
This is a call to BlueButton, we don't actually make those in unit testing, so it further relies on our library of canned responses 

### Feedback Requested

Design makes sense?
ideas for making mockBlueButtonClient.getData private


bcda/client/bluebutton.go
```diff
@@ -194,3 +205,12 @@ func GetDefaultParams() (params url.Values) {
 	params.Set("_format", "application/fhir+json")
 	return params
 }
+
+func HashHICN(toHash string) (hashedValue string) {
+	pepper, err := hex.DecodeString("b8ebdcc47fdd852b8b0201835c6273a9177806e84f2d9dc4f7ecaff08681e86d74195c6aef2db06d3d44c9d0b8f93c3e6c43d90724b605ac12585b9ab5ee9c3f00d5c0d284e6b8e49d502415c601c28930637b58fdca72476e31c22ad0f24ecd761020d6a4bcd471f0db421d21983c0def1b66a49a230f85f93097e9a9a8e0a4f4f0add775213cbf9ecfc1a6024cb021bd1ed5f4981a4498f294cca51d3939dfd9e6a1045350ddde7b6d791b4d3b884ee890d4c401ef97b46d1e57d40efe5737248dd0c4cec29c23c787231c4346cab9bb973f140a32abaa0a2bd5c0b91162f8d2a7c9d3347aafc76adbbd90ec5bfe617a3584e94bc31047e3bb6850477219a9")
```
dhgreene: If this needs to work in production with an actually secret pepper, we probably need a different mechanism for passing the pepper.

dhgreene: I bow to the explanation of the use case given out of band.

bcda/models/models.go
```diff
@@ -332,10 +334,53 @@ type CCLFFile struct {
...
+	if oldRecord.BlueButtonID != "" {
+		return oldRecord.BlueButtonID, nil
+	}
+
+	// didn't find a local value, need to ask BlueButton
+	jsonData, err := bb.GetBlueButtonIdentifier(client.HashHICN(cclfBeneficiary.HICN))
+	if err != nil {
+		return "", err
+	}
+	fmt.Println(jsonData)
```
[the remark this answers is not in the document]
knollfear: I think I nuked this one already.

bcda/models/models.go
```diff
@@ -25,7 +27,7 @@ func InitializeGormModels() *gorm.DB {
 	defer database.Close(db)
 
 	// Migrate the schema
-	// Add your new models here
+	// Add your new models herehttps://www.usajobs.gov/GetJob/ViewDetails/531221300
```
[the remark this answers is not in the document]
knollfear: fwiw, it is the job Jason was advertising.  If someone finds it and applies we are actually better off.  Also, I removed it already.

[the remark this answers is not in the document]
dhgreene: 👍 I think there might have been a couple commits not yet pushed as I was reviewing.

bcda/client/bluebutton.go
knollfear: Just going to post the convo here for everyone's benefit.

This value came from a confluence page and is duplicated in 2 gists linked from that page.  The point of this process is to define how a HICN should be hashed so that external partners and BlueButton are aligned.  Comparing an externally generated hash to an internally generated one requires that both parties follow the same recipe or process.  I believe that this necessitates exposing this value publicly in the same way one might expose a public key.

I am happy to keep this convo going here or on slack.

bcda/models/models.go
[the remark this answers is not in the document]
knollfear: I did a self PR and found some more things I missed.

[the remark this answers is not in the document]
StewGoin: If we get more applicants this way, it will be epic. XD

bcda/client/bluebutton.go
msnook: Should this be a constant?

bcda/client/bluebutton.go
```diff
@@ -28,6 +31,7 @@ type APIClient interface {
 	GetExplanationOfBenefitData(patientID, jobID string) (string, error)
 	GetPatientData(patientID, jobID string) (string, error)
 	GetCoverageData(beneficiaryID, jobID string) (string, error)
+	GetBlueButtonIdentifier(patientID string) (string, error)
```
msnook: patient ID or hashed HICN?

bcda/models/models.go
```diff
@@ -332,10 +334,52 @@ type CCLFFile struct {
...
+
+	// find another record with this HICN.  If it has a value use it.
+	// Find the first record with a matching HICN and an non empty BlueButtonID
+	db.Where("hicn = ? and blue_button_id <> ''", cclfBeneficiary.HICN).First(&oldRecord)
+	if oldRecord.BlueButtonID != "" {
+		return oldRecord.BlueButtonID, nil
+	}
+
+	// didn't find a local value, need to ask BlueButton
+	jsonData, err := bb.GetBlueButtonIdentifier(client.HashHICN(cclfBeneficiary.HICN))
```
msnook: I might be missing it, but shouldn't we be updating the DB with the info so it optimizes the request the next time?

bcda/client/bluebutton.go
```diff
@@ -96,6 +100,13 @@ func (bbc *BlueButtonClient) GetPatientData(patientID, jobID string) (string, er
 	return bbc.getData(blueButtonBasePath+"/Patient/", params, "")
 }
 
+func (bbc *BlueButtonClient) GetBlueButtonIdentifier(hashedHICN string) (string, error) {
```
msnook: I'm wondering if we should put the hashing functionality inside this function, so callers don't have to remember to do it?  So it would

bcda/client/bluebutton.go
tbellj: should we protect this salt value like a secret or is it okay like this?

knollfear: I firmly believe that this is a PUBLIC value that BB must share with anyone who is expected to request data from the hashedhicn endpoint.  Why do you think this should be protected?

bcda/client/bluebutton.go
knollfear: fixed in next commit.

bcda/client/bluebutton.go
knollfear: I kept it out so that if we needed to hash something else we had that option available to us.

bcda/client/bluebutton.go
msnook: I agree with that.  Can we make it a constant so its easy to update if the need arises?

knollfear: > Should this be a constant?

This fundamentally can not ever change or else every hash in BB back end would need to be recomputed.  I can certainly move it, but I don't think it will get us that much.

bcda/models/models.go
msnook: Thats fair.  Just wanted to make sure we weren't forgetting to do that somewhere.  Do we need a separate function to do that work?  Do you think its something that is out of scope of this ticket?  If so, we'll need a follow up ticket.

knollfear: I was planning on doing it in 1233 when I started actually getting the values.

msnook: I updated BCDA-1233 with a link to this convo, just so it doesn't get lost.

bcda/client/bluebutton.go
tbellj: Not saying it "should or shouldn't" be protected, just thought I'd raise the question here.

PR #253, opened 2019-05-01
BCDA-1242: Update Auth Function for Key Pairs

embh:
### Fixes [BCDA-1242](https://jira.cms.gov/browse/BCDA-1242)
BCDA needs a function in the auth package for generating a key pair for encryption, and assigning that key pair to an ACO.

### Proposed changes:
Add function to generate a key pair and assign it to a system, which can be linked to an ACO by client ID.

### Change Details
* Adds function `GenerateSystemKeyPair` that creates a 2048 bit RSA key pair, saves the public key in the database, and returns the private key.

### Security Implications
The newly added `GenerateSystemKeyPair` function uses Go's `rsa` library to create the key pair, as used elsewhere in the BCDA application. The new function is not currently called anywhere except for its tests.

### Acceptance Validation
Added two tests:
* `TestGenerateSystemKeyPair` - confirms that key pair is created successfully
* `TestGenerateSystemKeyPair_AlreadyExists` - confirms that function will return an error if encryption keys already exist for the system

### Feedback Requested
I know we wanted to put this in the `auth` package, but it was not clear where. Open to suggestions on where the function should live. And any other feedback.

### DIFF
--- a/bcda/auth/models.go
+++ b/bcda/auth/models.go
@@ -1,11 +1,16 @@
 package auth
 
 import (
-	"errors"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
 
 	"github.com/jinzhu/gorm"
 	_ "github.com/jinzhu/gorm/dialects/postgres"
 	"github.com/pborman/uuid"
+	"github.com/pkg/errors"
 
 	"github.com/CMSgov/bcda-app/bcda/database"
 	"github.com/CMSgov/bcda-app/bcda/models"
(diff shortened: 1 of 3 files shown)


bcda/auth/models_test.go
```diff
@@ -63,6 +66,81 @@ func (s *ModelsTestSuite) TestTokenCreation() {
...
+		s.FailNow(err.Error())
+	}
+	assert.Equal(s.T(), &privateKey.PublicKey, publicKey)
+
+	s.db.Unscoped().Delete(&pubEncrKey)
+	s.db.Unscoped().Delete(&system)
+	s.db.Unscoped().Delete(&group)
+}
+
+func (s *ModelsTestSuite) TestGenerateSystemKeyPair_AlreadyExists() {
```
dhgreene: This is a great test of the expected database behavior.  Can/should we test that a valid key pair is generated?

bcda/auth/models.go
```diff
@@ -63,3 +68,56 @@ func GetACOByClientID(clientID string) (models.ACO, error) {
 	}
 	return aco, err
 }
+
+/*
+ GenerateSystemKeyPair creates a keypair for a system, based on the provided client ID.
+ The public key is saved to the database and the private key is returned.
+*/
+func GenerateSystemKeyPair(clientID string) (string, error) {
```
dhgreene: To use the key, we seem to be having to convert the `string` to a `[]byte`.  Would that be a better return type?

embh: I was thinking about that. The CLI command function could convert the `[]byte` to `string` to print it out. 

Before I make the change, any preferences about the return type, @DeirdreHolub or @alexhova ?

alexhova: no preference here!

bcda/auth/models.go
```diff
@@ -63,3 +68,55 @@ func GetACOByClientID(clientID string) (models.ACO, error) {
...
+
+/*
+ GenerateSystemKeyPair creates a keypair for a system. The public key is saved to the database and the private key is returned.
+*/
+func GenerateSystemKeyPair(systemID uint) (string, error) {
+	db := database.GetGORMDbConnection()
+	defer database.Close(db)
+
+	var system models.System
+	err := db.Preload("EncryptionKeys").First(&system, systemID).Error
```
alexhova: i think `Find()` makes more sense semantically here but i know we've talked this to death already

PR #254, opened 2019-05-01
BCDA-1245 add func to revoke encryption keys

alexhova:
### Fixes [BCDA-1245](https://jira.cms.gov/browse/BCDA-1245)
We need a way to properly remove access of encryption keys for a system

### Proposed changes:
- add func to auth/models.go that soft deletes an encryption key belonging to a system
- add test to validate functionality

### Security Implications
No security implications with this PR yet.  It's just adding some functionality to inactivate a system's encryption keys.  These encryption keys are not being distributed or used yet.

### Acceptance Validation
I added a test to make sure that an encryption key is deleted without errors and that the deleted_at field is set (soft deletion)

### Feedback Requested
Why am I failing the assertion I wrote to make sure that the deleted_at column is not null?

### DIFF
--- a/bcda/auth/models.go
+++ b/bcda/auth/models.go
@@ -64,21 +64,20 @@ func GetACOByClientID(clientID string) (models.ACO, error) {
 }
 
 // RevokeSystemKeyPair soft deletes the specified encryption key so that it can no longer be used
-func RevokeSystemKeyPair(encryptionKeyID uint) (models.EncryptionKey, error) {
+func RevokeSystemKeyPair(encryptionKeyID uint) error {
 	db := database.GetGORMDbConnection()
 	defer database.Close(db)
 	encryptionKey := models.EncryptionKey{}
 
 	err := db.Find(&encryptionKey, encryptionKeyID).Error
 	if err != nil {
-		return encryptionKey, err
+		return err
 	}
 
 	err = db.Delete(&encryptionKey).Error
 	if err != nil {
-		return encryptionKey, err
+		return err
 	}
 
-	db.Unscoped().Find(&encryptionKey, encryptionKeyID)
-	return encryptionKey, nil
+	return nil
 }
--- a/bcda/auth/models.go
+++ b/bcda/auth/models.go
@@ -79,5 +79,6 @@ func RevokeSystemKeyPair(encryptionKeyID uint) (models.EncryptionKey, error) {
 		return encryptionKey, err
 	}
 
+	db.Unscoped().Find(&encryptionKey, encryptionKeyID)
 	return encryptionKey, nil
 }
--- a/bcda/auth/models.go
+++ b/bcda/auth/models.go
@@ -3,12 +3,11 @@ package auth
 import (
 	"errors"
 
+	"github.com/CMSgov/bcda-app/bcda/database"
+	"github.com/CMSgov/bcda-app/bcda/models"
 	"github.com/jinzhu/gorm"
 	_ "github.com/jinzhu/gorm/dialects/postgres"
 	"github.com/pborman/uuid"
-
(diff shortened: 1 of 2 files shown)


bcda/auth/models.go
```diff
@@ -63,3 +62,22 @@ func GetACOByClientID(clientID string) (models.ACO, error) {
...
+	db := database.GetGORMDbConnection()
+	defer database.Close(db)
+	encryptionKey := models.EncryptionKey{}
+
+	err := db.Find(&encryptionKey, encryptionKeyID).Error
+	if err != nil {
+		return encryptionKey, err
+	}
+
+	err = db.Delete(&encryptionKey).Error
```
dhgreene: This isn't updating `encryptionKey`, so when it's returned, it doesn't look like it's deleted.  You _could_ add this just before returning:

````
	// Update encryptionKey's deleted_at value
	err = db.Unscoped().Find(&encryptionKey, encryptionKeyID).Error
	if err != nil {
		return encryptionKey, err
	}
````

Comment (dhgreene) [reply bcda/auth/models.go]: Or maybe it's better just to return `error`

Comment (alexhova) [reply bcda/auth/models.go]: hmmm...  i'm going to return the deleted key like you showed above.  just in case we have an api or cli func that wants to return the deleted object

Comment (DeirdreHolub) [reply bcda/auth/models.go]: Why would we return a deleted object? No one can use it for any practical purpose.

Comment (alexhova) [reply bcda/auth/models.go]: yeah thats what i was originally doing.  honestly mostly did it for a higher fidelity unit test.  reverting now

Comment (DeirdreHolub) [reply bcda/auth/models.go]: > a higher fidelity unit test

I assume by this you meant that you wanted to make sure the thing was deleted? You could something along the lines of the code below to the test:
```
err = db.Unscoped().Find(&deletedKey, encryptionKeyID).Error
assert.Nil(err)
assert.NotNil(deletedKey)
assert.NotNil(deletedKey.DeletedAt)
```

**Comment by alexhova** [comment]
> The GORM documentation for delete has this at the top of it:
> 
> ```
> WARNING When deleting a record, you need to ensure its primary field has value, 
> and GORM will use the primary key to delete the record, if the primary key field is blank, 
> GORM will delete all records for the model
> ```
> 
> I think we should add a negative test to make sure the function fails when passed a EncryptionKey{} (nil value for the struct).

Yeah, i noticed that.  Haha that is really scary!  It just takes the id as the input though so it would be impossible for this to happen.  If it can't find a record with that id, it returns the error and doesn't delete

File: bcda/auth/models.go [hunk 2]
```diff
@@ -63,3 +62,22 @@ func GetACOByClientID(clientID string) (models.ACO, error) {
 	}
 	return aco, err
 }
+
+// RevokeSystemKeyPair soft deletes the specified encryption key so that it can no longer be used
+func RevokeSystemKeyPair(encryptionKeyID uint) error {
```
embh: The CLI would need an intermediate step to look up the key ID before calling this, right? Asking because the function for generating the key was meant to take the client ID, based on the mini tech spec. I wonder if it should be the same here.

alexhova: changed to systemID

bcda/auth/models_test.go
```diff
@@ -63,6 +63,19 @@ func (s *ModelsTestSuite) TestTokenCreation() {
 	assert.Equal(s.T(), tokenString, savedToken.TokenString)
 }
 
+func (s *ModelsTestSuite) TestRevokeSystemKeyPair() {
+	system := models.System{GroupID: "A00001"}
+	s.db.Save(&models.Group{GroupID: "A00001"})
+	s.db.Save(&system)
```
embh: On our earlier topic of unit test data, we might want to delete the created System and Group at the end of this test.

PR #256, opened 2019-05-07
BCDA-1194 Feature: Added ETL logging

tbellj:
Fixes [BCDA-1194](https://jira.cms.gov/browse/BCDA-1194)

Problem:
We need to have logging for every part of our cli ingesting process. We had some in place but needed to make sure we covered everything. This ticket should take care of that.

Proposed changes:
bcda/cli.go - added logging and print statements for our entire cclf ingestion workflow
bcda/cli_test.go - modified test case to successfully run

Security Implications:
No, there is no PII/PHI dealing with this change. However this ticket does log tasks that handle PII/PHI around our cclf ingestion process. There is **NO** logging of PHI/PII itself.

Acceptance Validation:
Yes

Screenshot:
Screen Shot 2019-05-13 at 11 43 07 AM

Feedback Requested:
Any and all

### DIFF
--- a/bcda/cli.go
+++ b/bcda/cli.go
@@ -128,14 +128,14 @@ func importCCLF0(fileMetadata *cclfFileMetadata) (map[string]cclfFileValidator, 
 	}
 
 	if _, ok := validator["CCLF8"]; !ok {
-		fmt.Printf("Failed to parse CCLF8 from CCLF0 file: %v.\n", fileMetadata)
-		err := fmt.Errorf("failed to parse CCLF8 from CCLF0 file: %v", fileMetadata)
+		fmt.Printf("Failed to parse CCLF8 from CCLF0 file: %#v.\n", fileMetadata)
+		err := fmt.Errorf("failed to parse CCLF8 from CCLF0 file: %#v", fileMetadata)
 		log.Error(err)
 		return nil, err
 	}
 	if _, ok := validator["CCLF9"]; !ok {
-		fmt.Printf("Failed to parse CCLF9 from CCLF0 file: %v.\n", fileMetadata)
-		err := fmt.Errorf("failed to parse CCLF9 from CCLF0 file: %v", fileMetadata)
+		fmt.Printf("Failed to parse CCLF9 from CCLF0 file: %#v.\n", fileMetadata)
+		err := fmt.Errorf("failed to parse CCLF9 from CCLF0 file: %#v", fileMetadata)
 		log.Error(err)
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


bcda/cli.go
```diff
@@ -62,6 +62,7 @@ const deleteThresholdHr = 8
 
 func importCCLF0(fileMetadata *cclfFileMetadata) (map[string]cclfFileValidator, error) {
 	if fileMetadata == nil {
+		fmt.Print("file CCLF0 not found")
```
embh: Like @knollfear mentioned in Slack, messages printed to stdout should be more like regular sentences (capitalized at the beginning at least)

bcda/cli.go
```diff
@@ -105,11 +109,15 @@ func importCCLF0(fileMetadata *cclfFileMetadata) (map[string]cclfFileValidator,
 					}
 					count, err := strconv.Atoi(string(bytes.TrimSpace(b[totalRecordStart:totalRecordEnd])))
 					if err != nil {
+						fmt.Printf("Failed to parse %s record count from CCLF0 file", filetype)
+						err = errors.Wrapf(err,"Failed to parse %s record count from CCLF0 file", filetype)
```
embh: "Failed..." message here is part of an error so we don't want that capitalized.

bcda/cli.go
```diff
@@ -120,17 +128,19 @@ func importCCLF0(fileMetadata *cclfFileMetadata) (map[string]cclfFileValidator,
...
 	if _, ok := validator["CCLF9"]; !ok {
+		fmt.Printf("failed to parse CCLF9 from CCLF0 file: %v", fileMetadata)
 		err := fmt.Errorf("failed to parse CCLF9 from CCLF0 file: %v", fileMetadata)
 		log.Error(err)
 		return nil, err
 	}
-	fmt.Printf("Imported CCLF0 file %s.\n", fileMetadata.name)
-	log.Infof("Imported CCLF0 file %s.", fileMetadata.name)
+	fmt.Printf("Successfully imported CCLF0 file %s.\n", fileMetadata.name)
+	log.Infof("Successfully imported CCLF0 file %s.", fileMetadata.name)
```
embh: Could it be unsuccessfully imported? Seems like this change from the old message might not be necessary.

bcda/cli.go
tbellj: doh! good catch

bcda/cli.go
tbellj: it's just more descriptive in my opinion especially since earlier in that method we have "Importing CCLF%d file".(Importing vs. imported) is now (Importing vs. Successfully imported)

bcda/cli.go
```diff
@@ -309,7 +345,9 @@ func importCCLFDirectory(filePath string) (success, failure, skipped int, err er
 	}
 
 	if len(cclfmap) == 0 {
-		return 0, 0, 0, errors.New("failed to find any CCLF files in directory")
+		err := errors.New("failed to find any CCLF files in directory")
+		log.Error(err)
```
embh: Would it be helpful to print this in addition to logging it?

embh:
Looking good! One request: could we add a little bit more information to the description? Just a sentence in Security Implications and a screenshot or example in Acceptance Validation would make the impact of the changes more visible.

bcda/cli.go
tbellj: yes and it does print because of the line afterwards. The return there gets us back to the invoking main.go where it bubbles up to stdout.

PR #258, opened 2019-05-13
Knollfear/bcda 1277 Refactor all code out of main.go

knollfear:
### Fixes [BCDA-1277](https://jira.cms.gov/browse/BCDA-1277)
Doing the work for 1277 got difficult because of the package organization and structure of our app.  This PR aims to refactor the codebase to put things that were previously in main into appropriate packages.  This will allow more reuse of code throughout the application.

### Proposed changes:

Move cli.go into CCLF package
Move router, middleware, and api to web package
create CLI package with most of the code from main.go
moved alpha token creation code to auth.models

### Change Details

THere sill be net no change to the functioning of the software, it will just be organized in a more maintainable way that should speed future development.

### Security Implications

There should be no changes in our security posture, this was purely a refactor, nothign is being done differently than it was before.

### Acceptance Validation

unit tests pass
organization scheme makes sense

### Feedback Requested

anything about how this is structured or could be better.


embh:
Could you update the title to our usual format? And thank you for the thorough description!

bcda/constants/constants.go
```diff
@@ -1,6 +1,10 @@
...
+package constants
 
 const DEVACOUUID = "0c527d2e-2e8a-4808-b11d-0fa06baf8254"
 const SMALLACOUUID = "3461C774-B48F-11E8-96F8-529269fb1459"
 const MEDIUMACOUUID = "C74C008D-42F8-4ED9-BF88-CEE659C7F692"
 const LARGEACOUUID = "8D80925A-027E-43DD-8AED-9A501CC4CD91"
+
+// App Name and usage.  Edit them here to prevent breaking tests
+const Name = "bcda"
+const Usage = "Beneficiary Claims Data API CLI"
```
embh: Because they're for the CLI (and not really used elsewhere), `Name` and `Usage` might make more sense in the `bcdacli` package.

knollfear: How strongly do feel about this change?  I moved them because I feel like we shouldn't be isolating our constants to just the scope where they are used.  By putting all constants in constants.go we create a single source of truth for all constants regardless of where they exist in the code and make it easier in the event they are used elsewhere in the future.

msnook: If we are moving them to a centralized location, maybe we should name them more specific?  A general constant named `Name` and `Usage` makes it hard to tell what it would be used for... if we want to keep these constants in a common file, maybe we should name them something like `App_Name` and `App_Usage` or `CLI_Name` and `CLI_Usage`?

embh: I'm not a fan in general of putting all constants in one place. To me it makes a lot more sense to have them in the package they apply to. No matter where they're used, the package works as a descriptor. E.g., `bcdacli.Name` is the name used by the CLI.

knollfear: @embh convinced me.  I have put it back in bcdacli

bcda/web/router_test.go
```diff
@@ -48,18 +48,18 @@ func (s *RouterTestSuite) TestDefaultRoute() {
...
 
 func (s *RouterTestSuite) TestDefaultProdRoute() {
-	err := os.Setenv("DEPLOYMENT_TARGET","prod")
+	err := os.Setenv("DEPLOYMENT_TARGET", "prod")
 	if err != nil {
-		s.FailNow("err in setting env var",err)
+		s.FailNow("err in setting env var", err)
 	}
-
+	// Need a new router because the one in the test setup is not right.
```
embh: Confused by this comment. Can we fix what's not right about it? If not, can we clarify the issue here?

knollfear: When the test is set up the variable is not set correctly because it gets set in the body of this test.  Therefore, in the running of this test after the variable is set it needs to create a new router.  This comment isn't strictly necessary, but github was not letting me resolve a conflict without making a change to the file so I elected to make this comment because it stumped Terrance and I why the test was failing and I thought it might be instructive to future devs.

embh:
```suggestion
	// Need a new router because the one in the test setup does not use the environment variable set it this test.
```

PR #259, opened 2019-05-13
BCDA-1285 Database: Add acos.public_key column

dhgreene:
### Fixes [BCDA-1285](https://jira.cms.gov/browse/BCDA-1285)
While we've created the tables we need to manage public keys for ACO's, for the time being we will be using a much simplified 1:1 relationship ACO:public_key.  Given:
* We also temporarily have a 1:1 relationship ACO:client_id and
* We're storing `client_id` and `secret` in the `acos` table

. . . it seems appropriate to create a `public_key` column in the `acos` table at least during our beta production phase.

Note that:
* In a separate ticket, we plan work to move the alternate tables to a branch, so that the code stays consistent.
* In a separate ticket, we plan to wire `aco.GetPublicKey()` to this new column

### Proposed changes:
* Add string column `acos.public_key`
* Test the new column

### Change Details
In practice, the column will almost certainly be unique.  But during testing, we are likely to use the same test key from time to time.

### Security Implications
This adds a public key to the database.  Since the public key is intended for public consumption, we don't care about its disclosure, just about its modification or loss.

### Feedback Requested
* Is the column named appropriately?
* Do we need additional constraints?
* Does the new test's dependency on and use of a database fixture create undue brittleness?


bcda/models/models_test.go
```diff
@@ -79,6 +81,38 @@ func (s *ModelsTestSuite) TestCreateACO() {
 	assert.NotNil(err)
 }
 
+func (s *ModelsTestSuite) TestACOPublicKeySave() {
+	assert := s.Assert()
+
+	// Setup ACO
+	cmsID := "A9994"
+	var aco ACO
+	s.db.Find(&aco, "aco_id = ?", cmsID)
```
DeirdreHolub: There is no column "aco_id" in the acos table. This find is actually returning an error that is not caught.

knollfear: I think you just need to do either of these things

`err := s.db.Find(&aco, "aco_id = ?", cmsID).Error`

`	if s.db.Find(&aco, "aco_id = ?", cmsID).RecordNotFound(){
		
	}`

It is all on the [gorm docs](http://gorm.io/docs/error_handling.html)

knollfear: This should also be First, Take or Last, all of which return a single record, not a list of all records (which may happen to be one record)

knollfear: On reading the rest of this I don't really see the benefit of looking up an existing ACO in the DB for this test.  Nothing about the state of the existing ACO is relevant to the test and it would just be cleaner to make a new one for this test, and delete it when the test ends.  That would simplify the test setup and ensure that there are no side effects like some other test loading up this ACO while it has the "wrong" key attached to it.

dhgreene: OK, going with `First` and a newly created ACO.

dhgreene:
@DeirdreHolub and @knollfear thanks for the careful oversight.  The tests are updated and ready for another review.

PR #260, opened 2019-05-14
Add Capability to Re-Generate and Update Client Credentials (in Alpha Mode)

msnook:
### Adds Capability to Re-Generate and Update Client Credentials (in Alpha Mode)

The `GenerateClientCredentials` method/capability is not supported in alpha mode.  We need to support that method, and expose it via CLI so that we can use existing ACOs for local/pipeline testing (and remove the reliance on `create-alpha-token` which creates a new disposable testing ACO each time it is called).

### Proposed changes:

Implement `GenerateClientCredentials`, expose it via CLI, and tie its usage into the local/Travis testing pipeline (replacing the usage of `create-alpha-token`).

### Change Details

- Implemented the previous unimplemented `GenerateClientCredentials` function for `auth` (alpha mode).
- Exposed `GenerateClientCredentials` via CLI.
- The `Makefile` was updated to use `generate-client-credentials` to dynamically acquire credentials (to be used for token generation) of a _pre-existing_ ACO (which defaults to `A9994` but can be changed dynamically based on environment variable).  This replaces the limitation of calling `create-alpha-token`, generating a new ACO each time we wanted to run a test.
- Added unit tests 

### Security Implications
PII/PHI is not affected here, however this is a change in `auth` (alpha mode) to support the regeneration of client credentials.  Please review carefully to see if there are any holes in this mechanism.

### Acceptance Validation

All tests are passing locally and in Travis.
Additionally, test execution where persistent tokens are utilized (AWS/Jenkins) is not affected.

### Feedback Requested
Please review.


bcda/main_test.go
```diff
@@ -237,6 +237,57 @@ func (s *MainTestSuite) TestCreateToken() {
 	buf.Reset()
 }
 
+func (s *MainTestSuite) TestGenerateClientCredentialsCLI() {
```
embh: Is there a benefit to having multiple test cases in one function? For me, it's clearer when there's one case per function. I know we haven't followed a consistent pattern for how we group tests though. Maybe a topic for the team to discuss.

msnook: I think its generally bad practice.  Ideally, we have one-to-one mapping of test case to test function.  I'd like to not re-arrange all of our tests as part of this PR; maybe a discussion for watercooler?

bcda/auth/alpha_test.go
```diff
@@ -112,10 +112,17 @@ func (s *AlphaAuthPluginTestSuite) TestDeleteClient() {
 }
 
 func (s *AlphaAuthPluginTestSuite) TestGenerateClientCredentials() {
-	r, err := s.p.GenerateClientCredentials("", 0)
-	assert.Empty(s.T(), r)
+        validClientID := "DBBD1CE1-AE24-435C-807D-ED45953077D3"
+        c, err := s.p.GenerateClientCredentials(validClientID, 0)
```
dhgreene: We've started caring more about side effects in our tests.  If we now care that this changes the credentials, we could add code to put things back the way they were.

bcda/main.go
```diff
@@ -300,25 +295,16 @@ func setUpApp() *cli.App {
...
 				aco, err := auth.GetACOByCMSID(acoCMSID)
                                 if err != nil {
                                         return err
                                 }                               
 
                                 // Generate new credentials
-				creds, err := auth.GetProvider().GenerateClientCredentials(aco.UUID.String(), ttlInt)
+				// TODO: GenerateClientCredentials interface to be updated as part of BCDA-1304
+				//       Passing in a ttl of 0 for now, since it is not used. 
+				creds, err := auth.GetProvider().GenerateClientCredentials(aco.UUID.String(), 0)
```
DeirdreHolub: To use the GenerateClientCredentials method, the ACO must have previously registered for and received credentials, and their client_id must have a value in their ACO record. So, we should use the aco.client_id here, not the aco.UUID.String(). The Okta plugin will error out here if we pass in the aco.UUID.String() here, because client_id for Okta is not the same as aco.UUID.

embh: In that case, do we still have a use for an internal ACO UUID?

msnook: Updated, thanks DJ.

bcda/auth/alpha_test.go
msnook: Agree, but this is not trivial since we're dealing with secrets in the DB.

PR #261, opened 2019-05-14
BCDA-1283 Feature: Update to router file to exclude user guide

tbellj:
Fixes [BCDA-1283](https://jira.cms.gov/browse/BCDA-1283)

Problem:
We had no way of limiting the user guide in production. Our desired state is to **not** have this shown at all when we launch to prod, this change does that.

Proposed changes:
bcda/router.go - disables access to the user guide in production
bcda/router_test.go - testing code that verifies a 404 response

Security Implications:
No, this change does not deal with PII/PHI at all.

Acceptance Validation:
I was able to fully test this and confirmed the 404 error.

Feedback Requested
Any and all

### DIFF
--- a/bcda/router.go
+++ b/bcda/router.go
@@ -15,9 +15,12 @@ func NewAPIRouter() http.Handler {
 	r := chi.NewRouter()
 	m := monitoring.GetMonitor()
 	r.Use(auth.ParseToken, logging.NewStructuredLogger(), HSTSHeader, ConnectionClose)
+
 	// Serve up the swagger ui folder
 	FileServer(r, "/api/v1/swagger", http.Dir("./swaggerui"))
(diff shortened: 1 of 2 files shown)


msnook:
I wonder if we should deploy this feature branch to `prod` now, just to see how it reacts (ensuring there are no side effects like health check or synthetic check failures we need to consider).  Remember, `prod` is locked down, so we can pretty much deploy what we want there now for testing purposes.

msnook:
> Assuming there is a ops repo ticket to set that env var appropriately LGTM!

@knollfear It is already set in all envs... here is the output of the env var in `prod`:
```
[ec2-user@ip-xx-xxx-xxx-xx ~]$ sudo cat /etc/sv/api/env/DEPLOYMENT_TARGET 
prod
```

PR #262, opened 2019-05-14
BCDA-1311 Feature: Timestamp fix for test

tbellj:
Fixes [BCDA-1311](https://jira.cms.gov/browse/BCDA-1311)

Problem:
Tests were failing due to timestamp precision errors. 

Proposed changes:
bcda/cli_test.go - added formatting to the timestamp checks

Security Implications
No, this change does **not** deal with PII/PHI

Acceptance Validation
yes

Feedback Requested
Any and all


bcda/cli_test.go
```diff
@@ -535,7 +535,7 @@ func (s *CLITestSuite) TestSortCCLFFiles() {
 	cclflist = cclfmap["A0001_18"]
```
embh: I think we need the time formatting up on line 522 also (GitHub isn't letting me put a comment there)

PR #264, opened 2019-05-15
Knollfear/bcda 1277 Import Synthetic CCLF Files

knollfear:
### Fixes [BCDA-1277](https://jira.cms.gov/browse/BCDA-1277)

To better simulate the production work flow we need a process to import a set of synthetic CCLF files.  This PR introduces a CLI to import a CCLF package and adds that process to the unit test execution.

### Proposed changes:

Created Synthetic CCLF files (3 per ACO size)
Created import process that zips source files and copies them to a temp directory
Created helper function to scan the directory tree for a path to assist in cases where relative paths are different between unit test execution and main app execution
Removed import Synthetic Beneficiary SQL file and replaced it with calls to import the sample files.

### Change Details

This is the next step towards making 1233 work which will put us at a place where we can do a full end to end test of our process from CCLF ingestion to file retrieval.

### Security Implications

This PR imports synthetic data that is already otherwise available, it just happens in a different way.

### Acceptance Validation

Unit tests now pass and do not rely on the fixture running

### Feedback Requested

Any


bcda/bcdacli/cli.go
```diff
@@ -361,8 +361,8 @@ func setUpApp() *cli.App {
 		},
 		{
 			Name:     "import-cclf-package",
```
msnook: Can you explain the reasoning behind using this new CLI instead of `import-cclf-directory`?  I think it would be most excellent if we were actively executing code internally/testing that will be used in real world.

knollfear: import-cclf-directory is an existing function that imports files that are already in a directory.  Under the covers this code calls that function, but it needs to do some external prep work as well, hence the different name.

msnook: Got it!  To avoid confusion, it might make sense to do a few things to ensure we aren't using this by accident in `prod`.
Can we rename to `import-synthetic-cclf-package`?  And maybe we could also even protect it so that its only usage in `DEBUG` mode?

knollfear: I can rename it.  I expect that it will in fact run in production so we can continue to test against production with data we can actually access.

msnook: Yeah, hoping we can make that happen.  The rename to include the term `synthetic` should be sufficient.  Thanks!

knollfear: Rename is done in the last commit.

PR #265, opened 2019-05-16
BCDA-1291 Feature: Get an ACO's Unique Public Key

dhgreene:
### Fixes [BCDA-1291](https://jira.cms.gov/browse/BCDA-1291)
Up to this point, we have been mocking requests for an ACO's public key by always returning the published public key (`shared_files/ATO_public.pem`).  Now that we have a place in the database to store the ACO's public key (PR #259 created `acos.public_key`), it's time to return it when requested.

### Proposed changes:
- Retrieve an ACO's public key on request
- Add public key to database fixtures that need a key.  Note that the same published public key is being used in fixtures that was being used in the mock.

### Change Details
- Assume the public key will be stored in PEM format
- To store the public key in database fixtures, add a column to `db/api.sql` and specify column names in `db/fixtures.sql`

### Database Migration Strategy
The sample public key will need to be added to the ACO used for smoke tests in `dev` and `test`  environments.  This will be done manually after the database has been migrated to include the `acos.public_key` column.

### Security Implications
To prevent data corruption or other issues, special care is taken to verify public key validity upon retrieval.  All phases of saving a public key (work to be done in further tickets) should also be safeguarded, from authz/authn to the formatting and usability of key being stored.

### Feedback Requested
- [ ]  Is there a better format for storing a public key in the database than PEM?
##### Advantages:
   - It's somewhat human-readable
   - It's likely the same format in which we want to require ACO's to present it to us 
##### Disadvantages:
   - It contains newlines, yet will not parse correctly if extra whitespace is introduced

- [ ] Should we be using multiple key pairs in testing?  If so, we need a mechanism for specifying which private key to use during smoke tests.

- [ ] Is there a better way of testing public keys than adding them to database fixtures?


karlmdavis:
I'd strongly suggest that you'll want to support more than one public key for an ACO, in the not-to-far future. That would allow them to rotate their keys without any downtime.

dhgreene:
> I'd strongly suggest that you'll want to support more than one public key for an ACO, in the not-to-far future. That would allow them to rotate their keys without any downtime.

Thanks, @karlmdavis .  Managing both a current key and a future key is definitely worth discussion.  _Everything_ about this PR is scoped for the current launch. including number of keys, location in the database, etc.

PR #266, opened 2019-05-16
BCDA-1233 Use CCLF data for Attribution

knollfear:
### Fixes [BCDA-1233](https://jira.cms.gov/browse/BCDA-1233)

Currently attribution uses acosbeneficiaries tsable which was just a stop gap we created while we sorted out how we would ingest files and use the CCLF8 data.  This PR switches attribution to using the CCLF Beneficiary table and records for the last imported CCLF8 file.

### Proposed changes:

rewrote aco.GetBeneficiaries to use cclf_Beneficiaries table
refactored job processing to resolve identifiers back to cclf_beneficiaries and fetch an identifier from Blue Button

### Change Details

Patched up several tests that were indiscriminately deleting files
added environment var to postman test container to allow for processing of bigger responses without crashing
fixed some errors in the synthetic beneficiary.sql file
Hardcoded Blue Button Identifiers on dev ACO in large fixture file

### Security Implications

This PR changes completely from using synthetic data loaded directly into the DB to data that was imported via a CCLF file.  This will drive how attribution works in production so if it is wrong there would be a big problem.

### Acceptance Validation

The local tests are not FULLY validated because the HICN's in the data do not appear to exist at the Blue Button server I am accessing.  I was able to complete the tests by harcoding BB identifiers, but I would prefer if these were dynamically pulled from BB instead.  I believe this will work in our deployed AWS infrastructure because it interacts with a different BB environment.

### Feedback Requested

Any, especially on test design.


bcda/models/models.go
```diff
@@ -166,20 +166,29 @@ type ACO struct {
...
-func (aco *ACO) GetBeneficiaryIDs() (beneficiaryIDs []string, err error) {
+func (aco *ACO) GetBeneficiaryIDs() (cclfBeneficiaryIDs []string, err error) {
+	if aco.CMSID == nil {
+		log.Errorf("No CMSID set for ACO: %s", aco.UUID)
+		return cclfBeneficiaryIDs, fmt.Errorf("no CMS ID set for this ACO")
+	}
 	db := database.GetGORMDbConnection()
 	defer database.Close(db)
+	var cclfFile CCLFFile
+	// should I put a filter here to make sure it isn't too old?
```
msnook: I don't know the answer to this question.  Should it become a `TODO` and a follow-up ticket in JIRA?

bcda/models/models.go
```diff
@@ -166,20 +166,29 @@ type ACO struct {
...
+	if db.Debug().Where("aco_cms_id = ? and cclf_num = 8", aco.CMSID).Order("timestamp desc").First(&cclfFile).RecordNotFound() {
```
msnook: Looks like `db.Debug()` might have been left in by accident.  Probably don't want that in production code.

bcda/models/models.go
knollfear: If we decide the answer is yes then we should.  If we don;t let's leave it as is and when the ticket comes up we can do it.

bcda/models/models.go
knollfear: fixed in incoming commit

bcdaworker/main.go
```diff
@@ -204,17 +204,28 @@ func writeBBDataToFile(bb client.APIClient, acoID string, beneficiaryIDs []strin
...
+		var cclfBeneficiary models.CCLFBeneficiary
+		db.First(&cclfBeneficiary, cclfBeneficiaryID)
+		blueButtonID, err := cclfBeneficiary.GetBlueButtonID(bb)
+		if err != nil {
+			log.Error(err)
+			errorCount++
+			appendErrorToFile(acoID, responseutils.Exception, responseutils.BbErr, fmt.Sprintf("Error retrieving BlueButton ID for cclfBeneficiary %s", cclfBeneficiaryID), jobID)
+		}
+		cclfBeneficiary.BlueButtonID = blueButtonID
+		db.Save(&cclfBeneficiary)
```
msnook: Just to confirm, this line fully address this comment, correct: 
https://github.com/CMSgov/bcda-app/pull/252/files#r280442154

Do we always need to call `db.Save()` here?  What if the `BlueButtonID` already exists for the `CCLFBeneficary`?  The extraneous call to save is harmless, but probably not necessary if the ID is already there.

msnook: Technically, we are finding the BB ID in the DB each time in current unit test scenario.  To ensure we don't have any surprises downstream, and to ensure we are testing both scenarios (ID exists and ID doesn't exist), we should definitely deploy this to AWS before we merge.  
To do this, would we need to update the `dev` DB with data in the new tables?

bcda/models/models.go
embh: I like the idea of making this a `TODO` and creating a JIRA ticket. Otherwise, let's take the comment out.

bcda/cclf/cclf_test.go
```diff
@@ -245,8 +259,12 @@ func (s *CCLFTestSuite) TestImportCCLF9() {
...
 	defer database.Close(db)
 
-	db.Unscoped().Delete(&models.CCLFBeneficiaryXref{})
-	db.Unscoped().Delete(&models.CCLFFile{})
+	var existngCCLFFiles []models.CCLFFile
+	db.Where("aco_cms_id = ?", "A0001").Find(&existngCCLFFiles)
+	for _, cclfFile := range existngCCLFFiles {
+		err := cclfFile.Delete()
+		assert.Nil(err)
+	}
```
embh: Can we make a cleanup function in the test file that does this work, maybe accepting the ACO ID as a parameter?

bcda/models/models.go
```diff
@@ -166,20 +166,29 @@ type ACO struct {
...
+		log.Errorf("Unable to find CCLF8 File for ACO: %v", *aco.CMSID)
+		return cclfBeneficiaryIDs, fmt.Errorf("unable to find cclfFile")
+	}
 
-	beneficiaryJoin := "inner join beneficiaries on acos_beneficiaries.beneficiary_id = beneficiaries.id"
-	if err = db.Table("acos_beneficiaries").Joins(beneficiaryJoin).Where("acos_beneficiaries.aco_id = ?", aco.UUID).Pluck("beneficiaries.blue_button_id", &beneficiaryIDs).Error; err != nil {
+	if err = db.Table("cclf_beneficiaries").Where("file_id = ?", cclfFile.ID).Pluck("ID", &cclfBeneficiaryIDs).Error; err != nil {
 		log.Errorf("Error retrieving ACO-beneficiaries for ACO ID %s: %s", aco.UUID.String(), err.Error())
```
embh: Might want to update the messaging here and below, since it's not just an ACO-beneficiary relationship anymore.
```suggestion
		log.Errorf("Error retrieving beneficiaries from latest CCLF8 file for ACO ID %s: %s", aco.UUID.String(), err.Error())
```

bcda/models/models.go
knollfear: I talked to Sara and this needs to be a ticket so I will make one and update.

bcda/cclf/cclf_test.go
knollfear: Sure.  Can we include that as part of the test cleanup epic?  I only did this so I could get through my tests and I don't want to get to wide ranging with this PR.

docker-compose.test.yml
```diff
@@ -18,7 +18,7 @@ services:
       - BB_SERVER_LOCATION=https://fhir.backend.bluebutton.hhsdevcloud.us
       - BB_CHECK_CERT=false
       - FHIR_PAYLOAD_DIR=../bcdaworker/data
-      - FHIR_STAGING_DIR=../bcdaworker/tmpdata
+      - FHIR_STAGING_DIR=data/test
```
msnook: I know we talked about this being the ultimate solution going forward, but I thought this was going to be cleaned up in a tech debt ticket?  Making this singular change in the compose stack could have integration test repercussions, as well as creates a sync issue with the other env vars that are pointing to `/bcdaworker/data`.  Can we be consistent and stick with modifying `FHIR_STAGING_DIR` in the test setup: https://github.com/CMSgov/bcda-app/blob/knollfear/BCDA-1233/bcdaworker/main_test.go#L38-L46
And then just create a ticket to coordinate/sync the compose stack with the internal unit test usage.
Note, I don't think any of this affects the production environment variables in S3.

msnook:
I suspect our performance tests will fail when this is introduced.  Do we need a ticket to update our performance tests to function properly once this change has been introduced to `master`?

docker-compose.test.yml
knollfear: The tests pass so this is a workable fix.  I don't understand the objection to this.  We can still write a ticket to do something else but doing it this way isn't going to change anything about our ability to do somethign else in the future.  Furthermore, this is just a kludge to fix the fact that SonarCube should just ignore this folder. I actually don't think we should ever be modifying the fhir staging dir in any tests unless that is specifically part of the test.  We don't flip flop that dir in production, we shouldn't be doing it in tests unless the test demands it.

msnook: Yeah, I wonder why the tests were updated to use that separate FHIR_STAGING_DIR in the first place?  Not sure if Sonar is to blame for that one, but ultimately we should investigate to see if Sonar has a .ignore file we can leverage.

knollfear: Those vars are also only set that way because the relative paths from the docker-compose file are not valid because the test runs from a fundamentally different path.  The app makes its own download directories on boot so setting it in docker compose should work regardless of the execution point (real vs tests)

bcda/models/models.go
```diff
@@ -395,10 +415,37 @@ func (cclfBeneficiary *CCLFBeneficiary) GetBlueButtonID(bb client.APIClient) (bl
...
+	var foundHicn = false
+	var foundBlueButtonID = false
 	blueButtonID = patient.Entry[0].Resource.ID
-	fullURL := patient.Entry[0].FullUrl
-	if !strings.Contains(fullURL, blueButtonID) {
-		err = fmt.Errorf("patient identifier not found in the URL")
+	for _, identifier := range patient.Entry[0].Resource.Identifier {
+		if strings.Contains(identifier.System, "hicn-hash") {
+			if identifier.Value != hashed_hicn {
+				foundHicn = false
```
tbellj: probably could rewrite this to only catch the true case since foundHicn is already set to false. but not a deal breaker

bcda/models/models.go
```diff
@@ -169,20 +169,29 @@ type ACO struct {
...
 
-	beneficiaryJoin := "inner join beneficiaries on acos_beneficiaries.beneficiary_id = beneficiaries.id"
-	if err = db.Table("acos_beneficiaries").Joins(beneficiaryJoin).Where("acos_beneficiaries.aco_id = ?", aco.UUID).Pluck("beneficiaries.blue_button_id", &beneficiaryIDs).Error; err != nil {
-		log.Errorf("Error retrieving ACO-beneficiaries for ACO ID %s: %s", aco.UUID.String(), err.Error())
+	if err = db.Table("cclf_beneficiaries").Where("file_id = ?", cclfFile.ID).Pluck("ID", &cclfBeneficiaryIDs).Error; err != nil {
+		log.Errorf("Error retrieving beneficiaries from latest CCLF8 file for ACO ID %s: %s", aco.UUID.String(), err.Error())
 		return nil, err
-	} else if len(beneficiaryIDs) == 0 {
+	} else if len(cclfBeneficiaryIDs) == 0 {
 		log.Errorf("Retrieved 0 ACO-beneficiaries for ACO ID %s", aco.UUID.String())
```
embh:
```suggestion
		log.Errorf("Found 0 beneficiaries from latest CCLF8 file for ACO ID %s", aco.UUID.String())
```

bcda/models/models.go
```diff
@@ -169,20 +169,29 @@ type ACO struct {
...
 		return nil, fmt.Errorf("retrieved 0 ACO-beneficiaries for ACO ID %s", aco.UUID.String())
```
embh:
```suggestion
		return nil, fmt.Errorf("found 0 beneficiaries from latest CCLF8 file for ACO ID %s", aco.UUID.String())
```

bcda/models/models.go
knollfear: I felt that this was more readable.  Also, if it is somehow in there both ways this will likely fail which I think we want.

bcda/models/models.go
```diff
@@ -384,7 +403,8 @@ func (cclfBeneficiary *CCLFBeneficiary) GetBlueButtonID(bb client.APIClient) (bl
 	}
 
 	// didn't find a local value, need to ask BlueButton
-	jsonData, err := bb.GetBlueButtonIdentifier(client.HashHICN(cclfBeneficiary.HICN))
+	hashed_hicn := client.HashHICN(cclfBeneficiary.HICN)
```
embh: Is snake case intentional or should this be `hashedHICN`?

knollfear: no, I have just been doing some JS on the side.  I will fix it.

bcda/models/models.go
```diff
@@ -395,10 +415,37 @@ func (cclfBeneficiary *CCLFBeneficiary) GetBlueButtonID(bb client.APIClient) (bl
 		return "", err
 	}
 
+	if len(patient.Entry) == 0 {
+		err = fmt.Errorf("patient identifier not found at BlueButton for cclfBenficiary ID: %v", cclfBeneficiary.ID)
+		log.Error(err)
+		return "", err
+	}
+	var foundHicn = false
```
embh: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

docker-compose.test.yml
embh: We can use `sonar.exclusions` to ignore that directory: https://docs.sonarqube.org/latest/project-administration/narrowing-the-focus/

bcda/models/models.go
knollfear: Changed.

PR #267, opened 2019-05-20
BCDA-1321 - Send additional information to Blue Button when making requests

knollfear:
### Fixes [BCDA-1321](https://jira.cms.gov/browse/BCDA-1321)

Blue Button back end has requested we supply the job ID and ACO CMS ID in the header when we make requests.  

### Proposed changes:

Changed API Client interface to just accept a string for all endpoints
Changed constructor for a BB Client to have a headerParams object that will drive the headers added to the requests
Altered client_test to be in the client package
Added test for header generation
Removed passing of JobID from several places

### Change Details

I decided that we should have a very flexible system for injecting these headers because it will reasonably change over time.  To do this I refactored how the variables were passed forward and made headerParams a concern of the client constructor.  This seemed to be the most efficient way of solving the problem.

### Security Implications

Job ID and CMSID are not PII.  In addition Jason blessed passing of this information over to Blue Button

### Acceptance Validation

Yes.  Tests passed and showed that headers are being set correctly.

### Feedback Requested

Any


bcda/client/bluebutton.go
```diff
@@ -27,16 +27,19 @@ var logger *logrus.Logger
 
 const blueButtonBasePath = "/v1/fhir"
 const blueButtonPepper = "b8ebdcc47fdd852b8b0201835c6273a9177806e84f2d9dc4f7ecaff08681e86d74195c6aef2db06d3d44c9d0b8f93c3e6c43d90724b605ac12585b9ab5ee9c3f00d5c0d284e6b8e49d502415c601c28930637b58fdca72476e31c22ad0f24ecd761020d6a4bcd471f0db421d21983c0def1b66a49a230f85f93097e9a9a8e0a4f4f0add775213cbf9ecfc1a6024cb021bd1ed5f4981a4498f294cca51d3939dfd9e6a1045350ddde7b6d791b4d3b884ee890d4c401ef97b46d1e57d40efe5737248dd0c4cec29c23c787231c4346cab9bb973f140a32abaa0a2bd5c0b91162f8d2a7c9d3347aafc76adbbd90ec5bfe617a3584e94bc31047e3bb6850477219a9"
+const JOBIDKEY = "BCDA-JOBID"
+const CMSIDKEY = "BCDA-CMSID"
```
embh: `jobIDKey` and `cmsIDKey` would follow Go convention. https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps

bcda/client/bluebutton.go
```diff
@@ -54,7 +57,7 @@ func init() {
 	}
 }
 
-func NewBlueButtonClient() (*BlueButtonClient, error) {
+func NewBlueButtonClient(headerParams map[string]string) (*BlueButtonClient, error) {
```
embh: It would be preferable for the headers to be associated with individual requests rather than the client. The client could be used for many requests with different header values. I don't think we want it tied to one job, for instance.

knollfear: It is already tied to one job.  The job makes its own client and that client exists until the job is processed and then it gets garbage collected.  We didn't implement a singleton pattern for client so they are not shared between processes and we will often have several floating around in memory.  If the design changes dramatically we can adjust, but given the current state I don't think it makes any sense to push this down to the individual request level.  If an individual request needs to send different headers it can set them on the client as needed, or simply create a new client.  Doing it this way lets the requests be as generic as possible and doesn't force future devs to handle a bunch of low level things on every request, because they are just defined on the client.  I actually think the client should implement a singleton default headers so we aren't having to make them from scratch each time and we can just append the ones that change for each request.

bcda/client/bluebutton.go
knollfear: fixed.

bcda/client/bluebutton.go
knollfear: On review header is not an object the way URL params is so a singleton isn't really going to work.  However, I still maintain that this is largely a client concern and not a request concern.

tbellj: After reading through this code, I see both points. From a clean code perspective, the headers probably should be at the request level like @embh suggested. But to @knollfear point the way our code is currently implemented we create a client per job, clients aren't shared so the need to have it at the request level is nice but doesn't really give us anything since they all have separate clients anyway. If we were to change our implementation and make each request have a shared client object then the headers should absolutely be at the request level. And using knoll's idea of a default header and having the clients overwrite/change for their specific calls is a good one, but that should be combined with @embh idea at the request level, if we decide to change our design. Like getDefaultHeaders or something and each request could just overwrite with their own stuff...I think :)

PR #268, opened 2019-05-21
BCDA-1322 Feature: Save ACO's Public Key

dhgreene:
### Fixes [BCDA-1322](https://jira.cms.gov/browse/BCDA-1322)
Now that we are encrypting files based on an ACO's unique public key, we need a way to store that key in the database.  

### Proposed changes:
This PR adds a method `aco.SavePublicKey()` that validates and saves the provided key.

### Change Details
- The key is passed as an `io.Reader` to facilitate uploading a file, but also allowing `string` or `[]byte` readers
- A new `rsautils` package is created with `/bcda/auth/rsautils/keytools.go` to factor out a `ReadPublicKey()` function

### Security Implications
This is saving user input to the database (though no callers yet exist for the new function).  Risk is mitigated by input tests against empty or unparse-able keys, or keys of insecure length.

### Feedback Requested
For `ReadPublicKey()`
- Is there a better location or name for the package?
- Is there a better way to specify a minimum key size (or should this be done closer to the key usage)?

For `aco.SavePublicKey()`
- Are there any missing test cases?
- Should we pass a `string` instead of `io.Reader`?


bcda/auth/rsautils/keytools.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+const RSAKEYMINBITS = 2048
+
+func ReadPublicKey (publicKey string) (*rsa.PublicKey, error) {
+	if publicKey == "" {
+		return nil, fmt.Errorf("empty string provided")
+	}
+
+	emptyPEMRegex := "-----BEGIN RSA PUBLIC KEY-----(\\W*)-----END RSA PUBLIC KEY-----"
```
DeirdreHolub: Don't pem.Decode() and/or x509.ParsePKIXPublicKey() validate the structure of the key, including the header and footer? Why are we duplicating that checking?

dhgreene: Good catch.  I was getting memory access violations with empty (header + footer with nothing in between) keys at one point, so I added the extra regex to filter them out, and tests to prevent regression.  But the tests pass without the extra regex filter, and without the empty string filter above it.  Useless code has now been removed.

PR #271, opened 2019-05-24
BCDA-1234: Update tests which rely on bene or acobene tables

embh:
### Fixes [BCDA-1234](https://jira.cms.gov/browse/BCDA-1234)
Synthetic data has been updated to follow an attribution model based on CCLF files, and the application needs to use that new attribution data model.

### Proposed changes:
Tests were updated along with changes to attribution logic in earlier work, and most CCLF tests are creating their own data rather than using fixtures. As a result, this PR is primarily cleanup. 

### Change Details
* Removes `ACOBeneficiary` and `Beneficiary` models
* Removes synthetic_beneficiaries.sql import from Makefile
* Adds helper function to cclf_test.go for deleting files by ACO
* Replaces test_synthetic_beneficiaries.sql import in unit_test.sh with test_synthetic_cclf_files_beneficiaries.sql

### Security Implications
No security implications. This work removes some models and updates tests.

### Acceptance Validation
Tests pass.

### Feedback Requested
Did I miss anything? Do we want to keep the old synthetic data SQL scripts (synthetic_beneficiaries.sql, test_synthetic_beneficiaries.sql)?


unit_test.sh
```diff
@@ -19,7 +19,7 @@ usql $DB_HOST_URL -c 'drop database if exists bcda_test;'
 usql $DB_HOST_URL -c 'create database bcda_test;'
 usql $TEST_DB_URL -f db/api.sql
 usql $TEST_DB_URL -f db/fixtures.sql
-usql $TEST_DB_URL -f db/test_synthetic_beneficiaries.sql
+usql $TEST_DB_URL -f db/test_synthetic_cclf_files_beneficiaries.sql
```
msnook: I think we want to intentionally leave this script out, due to the cool stuff happening here:
https://github.com/CMSgov/bcda-app/pull/271/files#diff-12c47e9f6d43cdddab176851f23f31dbR33

msnook:
I think we can remove these two scripts.  We shouldn't have any further need for them:
`synthetic_beneficiaries.sql` and `test_synthetic_beneficiaries.sql`

PR #272, opened 2019-05-24
BCDA-1259: Remove beneficiaries and acobeneficiaries table and associated structures/code

embh:
### Fixes [BCDA-1259](https://jira.cms.gov/browse/BCDA-1259)
Changes to the attribution process mean that the `beneficiaries` and `acos_beneficiaries` tables are no longer needed. Related models were removed with BCDA-1234.

### Proposed changes
Drop `beneficiaries` and `acos_beneficiaries` tables, and update functions that rely on them.

### Change details
* SQL script to drop `beneficiaries` and `acos_beneficiaries`
* `beneficiaries` and `acos_beneficiaries` removed from api.sql
* CLI `create-alpha-token` changed to require a CMS ID for the ACO in the format `T****` (`T` + four digits)
  * `AssignAlphaBeneficiaries()` removed due to reliance on dropped tables. To associate beneficiaries with an alpha ACO, synthetic CCLF files may be ingested.

### Security implications
No security impact. This script removes two tables that are not used and modifies the attribution process for an alpha ACO.

### Acceptance validation
Tests updated. Will deploy to dev environment if these changes are approved.

### Feedback requested
For the auth team: do we still need `create-alpha-token` and if so, does this approach work?

### DIFF
--- a/bcda/auth/models.go
+++ b/bcda/auth/models.go
@@ -159,8 +159,8 @@ func GenerateSystemKeyPair(systemID uint) (string, error) {
 
 	return string(privateKeyBytes), nil
 }
-func CreateAlphaToken(ttl int, acoSize string) (string, error) {
-	aco, err := createAlphaEntities(acoSize)
+func CreateAlphaToken(ttl int, acoCMSID string) (string, error) {
+	aco, err := createAlphaEntities(acoCMSID)
 	if err != nil {
 		return "", err
 	}
@@ -184,7 +184,7 @@ func CreateAlphaToken(ttl int, acoSize string) (string, error) {
 	return msg, nil
 }
 
-func createAlphaEntities(acoSize string) (aco models.ACO, err error) {
+func createAlphaEntities(acoCMSID string) (aco models.ACO, err error) {
 	db := database.GetGORMDbConnection()
 	defer database.Close(db)
 	tx := db.Begin()
@@ -201,13 +201,8 @@ func createAlphaEntities(acoSize string) (aco models.ACO, err error) {
 		return aco, tx.Error
 	}
 
-	aco, err = models.CreateAlphaACO(tx)
+	aco, err = models.CreateAlphaACO(acoCMSID, tx)
 	if err != nil {
-		tx.Rollback()
-		return aco, err
-	}
-
-	if err = models.AssignAlphaBeneficiaries(tx, aco, acoSize); err != nil {
 		tx.Rollback()
 		return aco, err
 	}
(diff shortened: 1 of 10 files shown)


bcda/auth/models.go
```diff
@@ -159,8 +159,8 @@ func GenerateSystemKeyPair(systemID uint) (string, error) {
 
 	return string(privateKeyBytes), nil
 }
-func CreateAlphaToken(ttl int, acoSize string) (string, error) {
-	aco, err := createAlphaEntities(acoSize)
+func CreateAlphaToken(ttl int, acoCMSID string) (string, error) {
```
dhgreene: I think we've replaced enough of `CreateAlphaToken`'s functionality to remove it now.  

Either way, because it's called by the [`tokengen` script](https://github.com/CMSgov/bcda-ops/blob/6052e9b26d09924993e02df0ca633cf4df464060/scripts/tokengen#L65), which in turn is used by a couple of playbooks, we'll want an ops ticket to either change the call or remove the `tokengen` script.

bcda/auth/models.go
```diff
@@ -184,7 +184,7 @@ func CreateAlphaToken(ttl int, acoSize string) (string, error) {
 	return msg, nil
 }
 
-func createAlphaEntities(acoSize string) (aco models.ACO, err error) {
+func createAlphaEntities(acoCMSID string) (aco models.ACO, err error) {
```
dhgreene: When we remove `createAlphaToken`, this function can also be removed.

bcda/models/models.go
```diff
@@ -286,22 +286,15 @@ func CreateUser(name string, email string, acoUUID uuid.UUID) (User, error) {
 }
 
 // CLI command only support; note that we are choosing to fail quickly and let the user (one of us) figure it out
-func CreateAlphaACO(db *gorm.DB) (ACO, error) {
+func CreateAlphaACO(acoCMSID string, db *gorm.DB) (ACO, error) {
```
dhgreene: Correct me if I'm wrong, but it seems like the only thing that calls this code is `createAlphaEntities()`, so this could go as well.

embh:
@dhgreene @DeirdreHolub @alexhova Made a new ticket for removing `create-alpha-token`: https://jira.cms.gov/browse/BCDA-1366 Are you okay with keeping the command for now, with the changes in this PR?

dhgreene:
> ... Are you okay with keeping the command for now, with the changes in this PR?

@embh Either way, there are breaking changes to the task I'm working on.  Your work seems sound, though I haven't run it.  I suggest another followup ticket to change our CMS ACO ID's in our tests and fixtures to the `T****` format you're using here.

PR #273, opened 2019-05-28
BCDA-1332 Feature: Changed to use UTF8 encoding to fix string formatting issues

tbellj:
Fixes [BCDA-1332](https://jira.cms.gov/browse/BCDA-1332)

Problem:
Our users were having a few problems with file decryption. Specifically the format of the data was not ideal, including a random 'b' in the beginning of the file and '\n' found throughout along with escaped single quotes.

Proposed changes:
decryption_utils/Python/decrypt.py - changed to use UTF8

Change Details
Decoded to use UTF-8.

Security Implications:
No, this change **does not** deal with PII/PHI.

Acceptance Validation:
Yes, the file now has the correct format.

Feedback Requested:
Any and all


decryption_utils/Python/decrypt.py
```diff
@@ -88,6 +89,8 @@ def valid_uuid(filename):
 
 def main():
     args = init()
+    if sys.version_info[0] < 3:
```
msnook: Was this solution tested on Python 2.x _and_ Python 3.x ?

tbellj: yes vigorously.

PR #274, opened 2019-05-29
BCDA-1358 New CCLF0 format based on sample files from EFT team.

knollfear:
### Fixes [BCDA-1358](https://jira.cms.gov/browse/BCDA-1358)

We slightly misinterpreted the spec for CCLF 0 and sample files provided by the EFT team weren;t working.  This fixes those files to be in the right format.

### Proposed changes:

Changed file position constants
replaced all CCLF0 sample files based on failed tests

### Security Implications

no PII is affected.  This is just a reformatting of existing synthetic files

### Acceptance Validation

yes.  THis is a true no op change.  As long as tests run it should be all fine.


bcda/cclf/cclf.go
```diff
@@ -81,9 +81,9 @@ func importCCLF0(fileMetadata *cclfFileMetadata) (map[string]cclfFileValidator,
 	defer r.Close()
 
 	const (
-		fileNumStart, fileNumEnd           = 0, 13
-		totalRecordStart, totalRecordEnd   = 35, 55
-		recordLengthStart, recordLengthEnd = 56, 69
+		fileNumStart, fileNumEnd           = 0, 7
+		totalRecordStart, totalRecordEnd   = 52, 63
+		recordLengthStart, recordLengthEnd = 64, 70
```
embh: With the start and end being 65 and 69 in the IP, should `recordLengthEnd` be 69?

knollfear: looks like I copied from Slack and got some whitespace that was extra.  I have changed it to 69.

PR #276, opened 2019-05-29
Create base structured logging pattern

DeirdreHolub:
### Fixes [BCDA-1356](https://jira.cms.gov/browse/BCDA-1356)

Updates logging in auth to use structured logging mechanism and record log entries as described in the Tech Spec.

### Proposed changes:
* creates a package global logger configured for structured logging
* adds logging functions that correspond to events from the tech spec
* adds consts appropriate for logging / events

### Approach
The tech spec discusses operations that emit events. This implementation assigns responsibility for doing the actual logging to individual functions that emit events. The data for those events is provided by the calling code that gathers the data. A tracking_id is among the fields of data; the intent with this id is to provide an identifier that can correlate a set of related events.

### Security Implications
*enhances security by providing an audit trail of events. Existing logging is largely focused on error detection.

### Acceptance Validation
Tests added for logging; all tests pass. All existing tests continue to pass as well.

### Feedback Requested
Any.

### DIFF
--- a/bcda/auth/alpha.go
+++ b/bcda/auth/alpha.go
@@ -8,7 +8,7 @@ import (
 
 	"github.com/CMSgov/bcda-app/bcda/database"
 	"github.com/CMSgov/bcda-app/bcda/models"
-	jwt "github.com/dgrijalva/jwt-go"
+	"github.com/dgrijalva/jwt-go"
 	"github.com/pborman/uuid"
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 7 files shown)


bcda/auth/logger.go
```diff
@@ -0,0 +1,98 @@
...
+			logger.SetOutput(file)
+		} else {
+			logger.Info("Failed to open Auth log file; using default stderr")
+		}
+	} else {
+		logger.Info("No Auth log location provided; using default stderr")
+	}
+}
+
+func mergeNonEmpty(data event) *logrus.Entry {
```
dhgreene: Based on my experience, I think you avoided future bugs by writing this method.  🐛 🔨 💯

bcda/auth/alpha.go
```diff
@@ -8,35 +8,47 @@ import (
...
+	"github.com/dgrijalva/jwt-go"
 	"github.com/pborman/uuid"
 	log "github.com/sirupsen/logrus"
 )
 
 type AlphaAuthPlugin struct{}
 
 func (p AlphaAuthPlugin) RegisterClient(localID string) (Credentials, error) {
-
+	regEvent := event{op: "RegisterClient", trackingID: localID}
```
dhgreene: I wish the struct was easier to work with.  There are some hints from the IDE about what keys might be possible, and when you command+hover over `event` it shows you a definition, but I'd like it even better if this was chain-able like:
`regEvent := event.New().op("RegisterClient").trackingID(localID)`
or functional like:
`regEvent := event.New(event.Op("RegisterClient"), event.TrackingID(localID))`

Maybe in a far-off refactor.

bcda/auth/alpha.go
```diff
@@ -8,35 +8,47 @@ import (
...
+	operationStarted(regEvent)
 	if localID == "" {
-		return Credentials{}, errors.New("provide a non-empty string")
+		// do we want to track usage errors?
+		regEvent.help = "provide a non-empty string"
+		operationFailed(regEvent)
```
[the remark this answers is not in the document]
DeirdreHolub: Yah, i think this is definitely an area for future refinement.

PR #278, opened 2019-05-30
BCDA-1304: Refactor: Remove TTL from auth.GenerateClientCredentials()

dhgreene:
### Fixes [BCDA-1304](https://jira.cms.gov/browse/BCDA-1304)
The auth providers have changed since the `auth.Provider` interface was originally defined.  We no longer use the `ttl` argument to `auth.GenerateClientCredentials()`.  All callers are currently using a zero value for `ttl`, so this is a minor refactoring.

### Proposed changes:
1) Change the `auth.GenerateClientCredentials()` signature and its implementation (and tests) in the auth providers:
- Alpha
- Okta
2) Remove the zero value sent by the one caller, the CLI command `generate-client-credentials`.

### Security Implications
No security impact should be felt; at this point this is entirely an internal refactoring.  While TTL was necessary when long-lived tokens were provided by this function, it is no longer relevant for client_id and client_secret.

### Feedback Requested
Are there any non-obvious impacts to these changes that have been missed?


bcda/bcdacli/cli.go
```diff
@@ -323,7 +323,7 @@ func setUpApp() *cli.App {
 				// Generate new credentials
 				// TODO: GenerateClientCredentials interface to be updated as part of BCDA-1304
```
alexhova: I suppose you could remove this `TODO` now

dhgreene: Good catch!

PR #283, opened 2019-06-03
Knollfear/bcda-1396 Fixing setting of the version after code refactor.

knollfear:
### Fixes [BCDA-1396](https://jira.cms.gov/browse/BCDA-1396)

Refactoring the code caused the version to not be set correctly.  

### Proposed changes:

Export the version variable in the bcdacli and web packages
set the version in the main.go init function

### Change Details

the variables were private, but I made them public so they could be set from main.go.

### Security Implications

no PII is affected.  This just exposes a version variable that had stopped getting set correctly

### Acceptance Validation

Yes.  Deployed to dev and validated on the version endpoint.

image

### Feedback Requested

any.


embh:
Instead of having version in multiple places, could we put it in `constants` and use that in cli.go and api.go? Building with something like `go build -ldflags "-X bcda/constants.Version=r1"`

knollfear:
> Instead of having version in multiple places, could we put it in `constants` and use that in cli.go and api.go? Building with something like `go build -ldflags "-X bcda/constants.Version=r1"`

Great suggestion.  I refactored it and will test it out on dev shortly.

knollfear:
> Instead of having version in multiple places, could we put it in `constants` and use that in cli.go and api.go? Building with something like `go build -ldflags "-X bcda/constants.Version=r1"`

@embh it is working now with setting the value in constants.  Please have another look when you get a chance.

bcda/constants/constants.go
```diff
@@ -4,3 +4,6 @@ const DEVACOUUID = "0c527d2e-2e8a-4808-b11d-0fa06baf8254"
 const SMALLACOUUID = "3461C774-B48F-11E8-96F8-529269fb1459"
 const MEDIUMACOUUID = "C74C008D-42F8-4ED9-BF88-CEE659C7F692"
 const LARGEACOUUID = "8D80925A-027E-43DD-8AED-9A501CC4CD91"
+
+// This is set during compilation.  See build_and_package.sh in the ops repo
```
embh: Great idea to put a comment here so we remember what changes this value 👍

PR #284, opened 2019-06-03
BCDA-1308: Create Automated Testing in Docker/Travis

embh:
### Fixes [BCDA-1308](https://jira.cms.gov/browse/BCDA-1308)
In Docker and Travis, fixtures are added to the database using SQL scripts. To more closely replicate the process of ingesting CCLF file data, the fixtures for CCLF files, beneficiaries, and ID cross-references should be imported from synthetic CCLF files.

### Proposed changes:
Remove current SQL script that inserts synthetic CCLF data and read it from synthetic CCLF files instead.

### Change Details
* Adds `load-synthetic-cclf-data` target to Makefile, which loads data from synthetic CCLF files using CLI command `import-synthetic-cclf-package`
* Adds `load-synthetic-cclf-data` as part of Makefile `docker-bootstrap` target
* Moves `docker-compose up` out of Travis and into Makefile target `docker-bootstrap`
* Adjusts worker logic to skip request to Blue Button if patient ID could not be retrieved
* Updates api.sql to contain all CCLF tables
* Removes db/synthetic_cclf_files_beneficiaries.sql
* Configures HICN hashing using values found here: https://github.com/CMSgov/bluebutton-ansible-playbooks-data-sandbox/blob/master/etl_server_redeploy.yml#L150-L151

### Security Implications
All data ingested is synthetic.

### Acceptance Validation
Tests pass with no changes.

### Feedback Requested
Is `import-synthetic-cclf-package` adequate for these tests, or should we zip the files and use the real `import-cclf-directory` command? The former simplifies the process, but happy to make the switch if anyone would prefer it.


Makefile
```diff
@@ -58,10 +58,16 @@ test:
...
-	docker-compose exec db psql "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -f /var/db/fixtures.sql
-	docker-compose exec db psql "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -f /var/db/synthetic_cclf_files_beneficiaries.sql
+	docker-compose run db psql "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -f /var/db/fixtures.sql
+
+load-synthetic-cclf-data:
+	docker-compose run api sh -c 'tmp/bcda import-synthetic-cclf-package --acoSize=dev --environment=test'
+	docker-compose run api sh -c 'tmp/bcda import-synthetic-cclf-package --acoSize=small --environment=test'
+	docker-compose run api sh -c 'tmp/bcda import-synthetic-cclf-package --acoSize=medium --environment=test'
+	docker-compose run api sh -c 'tmp/bcda import-synthetic-cclf-package --acoSize=large --environment=test'
+	docker-compose run api sh -c 'tmp/bcda import-synthetic-cclf-package --acoSize=extra-large --environment=test'
```
knollfear: Is this one strictly necessary?  If this is just for the travis test environment I think maybe it can be skipped.  There is no harm in leaving it, but it might be 30 seconds of processing we don't need to do.

Makefile
```diff
@@ -58,10 +58,16 @@ test:
 
 load-fixtures:
 	docker-compose up -d db
-	echo "Wait for db to be ready..."
+	echo "Wait for database to be ready..."
 	sleep 5
-	docker-compose exec db psql "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -f /var/db/fixtures.sql
-	docker-compose exec db psql "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -f /var/db/synthetic_cclf_files_beneficiaries.sql
+	docker-compose run db psql "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -f /var/db/fixtures.sql
```
knollfear: Instead of adding to fixtures, what about working in a sql-migrate execution like in unit_test.sh?  I think that is more in line with what we are doing in production since fixtures.sql isn't run there afaik.

.travis.yml
```diff
@@ -46,4 +46,5 @@ script:
   - make docker-bootstrap
   - docker-compose up -d
   - sleep 30
+  - make load-synthetic-cclf-data
```
msnook: Previously, loading the data was part of `make docker-bootstrap`, so it was not a step required by folks to run manually in their local setup.  Now, it is an explicitly separate target, which would have to be run separately in the local env setup.  
We'll need to either (1) update the[ readme to specify the order of operations required ](https://github.com/CMSgov/bcda-app/blob/master/README.md#build--start)to get a local env setup, or (2) setup up `docker-boostrap` to bring up the compose stack and load the synthetic data so that user's don't have to invoke those targets separately.

Makefile
embh: Good question. I was trying to replicate the behavior of the insert script we had before, which included all the sizes. Not sure that we really need it though. @msnook , @tbellj - thoughts?

msnook: @embh Does it affect performance of test execution?  If not, I don't think it hurts to keep it in.

Makefile
msnook: I think this is getting into the area of DB Migration strategy, which remains a topic of discussion for the whole team.

Makefile
knollfear: I believe it just takes the 30 or however many seconds importing 30,000 beneficiaries takes.  We don't import that many locally so I am not 100% certain how long it takes.

bcdaworker/main.go
```diff
@@ -224,16 +224,17 @@ func writeBBDataToFile(bb client.APIClient, acoID string, cclfBeneficiaryIDs []s
...
-		}
-		cclfBeneficiary.BlueButtonID = blueButtonID
-		db.Save(&cclfBeneficiary)
-		pData, err := bbFunc(blueButtonID, jobID)
-		if err != nil {
-			log.Error(err)
-			errorCount++
-			appendErrorToFile(acoID, responseutils.Exception, responseutils.BbErr, fmt.Sprintf("Error retrieving %s for beneficiary %s in ACO %s", t, blueButtonID, acoID), jobID)
 		} else {
-			fhirBundleToResourceNDJSON(w, pData, t, cclfBeneficiaryID, acoID, jobID)
```
msnook: Should this be moved out of this branch now that it is included in #289  ?

PR #285, opened 2019-06-05
BCDA-1306 Feature: added additional cclf0 validation

tbellj:
Fixes [BCDA-1306](https://jira.cms.gov/browse/BCDA-1306)

Problem:
Previously we had no way of detecting a duplication of cclf8 and or cclf9 file types in the cclf0 file. 

Proposed changes:
Added a check to make sure a duplication for cclf8 and or cclf9 file types causes an error.

Change Details:
bcda/cclf/cclf.go - logic to early detect a duplicate file type.
bcda/cclf/cclf_test.go  - added a test case.
shared_files/cclf0_MissingData/ - changed to the correct cclf0 format to include headers. Added an extra synthetic cclf0 file with duplicate entries.

Security Implications:
No, this change **does not** deal with PII/PHI. This work is for validating CCLF0 files, which contain statistics about other CCLF files but no PII or PHI.

Acceptance Validation:
Yes

Feedback Requested:
Any and all


StewGoin:
> Security Implications:
>No, this change does not deal with PII/PHI.

Is that accurate given that we're changing how we do error detection stuff for CCLF input files?

tbellj:
> > Security Implications:
> > No, this change does not deal with PII/PHI.
> 
> Is that accurate given that we're changing how we do error detection stuff for CCLF input files?

Since this solely dealt with the validation on a cclf0 file (which doesn't have any PII/PHI) I didn't see how this change would be a security concern. But the point you bring up might be valid, I'm interested to hear what the rest of the team thinks on this issue...

embh:
Maybe the description could explain why it doesn't deal with PII/PHI, e.g., "This work is for validating CCLF0 files, which contain statistics about other CCLF files but no PII or PHI."

bcda/cclf/cclf_test.go
```diff
@@ -57,6 +57,12 @@ func (s *CCLFTestSuite) TestImportCCLF0() {
 	cclf0metadata = &cclfFileMetadata{env: "test", acoID: "A0001", cclfNum: 0, timestamp: time.Now(), filePath: cclf0filePath, perfYear: 18}
 	_, err = importCCLF0(cclf0metadata)
 	assert.NotNil(err)
+
+	// duplicate file types from cclf0
+	cclf0filePath = BASE_FILE_PATH + "cclf0_MissingData/T.A0001.ACO.ZC0Y18.D181120.T1000013"
+	cclf0metadata = &cclfFileMetadata{env: "test", acoID: "A0001", cclfNum: 0, timestamp: time.Now(), filePath: cclf0filePath, perfYear: 18}
+	_, err = importCCLF0(cclf0metadata)
+	assert.NotNil(err)
```
embh: I know we're not doing this on the other tests in this function, but could you add an assertion here on the expected error message? We should have it on all the tests, but adding the rest may be out of scope for this ticket. These could be failing for reasons other than what we're expecting.

bcda/cclf/cclf.go
```diff
@@ -108,6 +108,14 @@ func importCCLF0(fileMetadata *cclfFileMetadata) (map[string]cclfFileValidator,
 					if validator == nil {
 						validator = make(map[string]cclfFileValidator)
 					}
+
+					if _, ok := validator[filetype]; ok {
+						fmt.Printf("Duplicate %v file type found from CCLF0 file.\n", filetype)
+						err := fmt.Errorf("Duplicate %v file type found from CCLF0 file.\n", filetype)
```
embh:
```suggestion
						err := fmt.Errorf("duplicate %v file type found from CCLF0 file.\n", filetype)
```

embh:
Thank you for fixing the tests! This looks awesome. Just one tiny error format change request.

bcda/cclf/cclf.go
tbellj: good eye!

PR #288, opened 2019-06-06
BCDA-1394 Transient Issue with Go Decryptor in Smoke Test

knollfear:
### Fixes [BCDA-1394](https://jira.cms.gov/browse/BCDA-1394)

The Go Decryptor has failed randomly a few times, but we didn't log enough information to let us pinpoint the cause.

### Proposed changes:

Added result print lines where panics were occuring.  
Changed os.exit codes to be unique based on which line failed.

### Change Details

also changed out 2 lines that had a nosec comment for the approved way of doing it.
removed --format standard-verbose from unit_test.sh

### Security Implications

T\his only concerns the decryptor which does affect PII, but the functionality is unchanged, just the error handling when things go wrong.

### Acceptance Validation

Decryptor tests should run as normal.

### Feedback Requested

Any


decryption_utils/Go/decrypt.go
```diff
@@ -66,51 +66,58 @@ func decryptFile(privateKey *rsa.PrivateKey, encryptedKey []byte, filename strin
...
-		log.Panic(err)
+		fmt.Println("Failed to read private key")
+		os.Exit(7)
 	}
 
 	data, _ := pem.Decode([]byte(pembytes))
 	err = pkFile.Close()
 	if err != nil {
-		log.Panic(err)
+		fmt.Println("Failed to decode private key")
```
embh: Should this be a message about `pkFile.Close()`?

knollfear: Good catch.  Fixed now.

PR #291, opened 2019-06-11
BCDA-1421: Decryption Utilities fail if path/to/filename.ndjson is passed in via --file

embh:
### Fixes [BCDA-1421](https://jira.cms.gov/browse/BCDA-1421)
Python decryption tool fails on Windows if the .ndjson file is referenced with a path.

### Proposed changes:
Use `os.path.split` to split the file path rather than hardcoded `/`, so it will work on any OS.

### Security Implications
This fixes a bug in our decryption script, but does not change the decryption process.

### Acceptance Validation
Ran successfully on macOS Mojave and Windows 7.

### Feedback Requested
Can you try decrypting with path included in the `--file` argument?

### DIFF
--- a/decryption_utils/Python/decrypt.py
+++ b/decryption_utils/Python/decrypt.py
@@ -82,7 +82,8 @@ def get_private_key(loc):
         return RSA.importKey(fh.read())
 
 def valid_uuid(filename):
-    uuid = filename.split("/")[-1].split(".")[0]
+    head, tail = os.path.split(filename)
+    uuid = tail.split(".")[0]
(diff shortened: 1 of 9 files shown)


tbellj:
not sure how to test this in windows?? but I'll try a path in the --file argument on my mac

decryption_utils/Python/decrypt.py
```diff
@@ -82,7 +82,8 @@ def get_private_key(loc):
         return RSA.importKey(fh.read())
 
 def valid_uuid(filename):
-    uuid = filename.split("/")[-1].split(".")[0]
+    head, tail = os.path.split(filename)
```
msnook: Huge nitpick, but reading this I was not sure what `head` and `tail` were supposed to represent.
Recommend renaming to something like `path` and `file` or similar.

PR #294, opened 2019-06-13
BCDA-1417: Fix BB client logging statements for responses

embh:
### Fixes [BCDA-1417](https://jira.cms.gov/browse/BCDA-1417)
Log entries for Blue Button responses are missing values. 

### Proposed changes:
Add missing values and remove fields where the value is unknown.

### Change Details
* Uses headers from the request to populate `bb_query_id`, `bb_query_ts`, `bb_uri`
* Removes `content_length` because Go will always set it to -1
* Adds `jobID` parameter to `getData` calls from `GetPatientData` and `GetCoverageData`
* Shortens messages to simply `request` or `response` (previously 'Blue Button request' and 'Blue Button response`) as these entries are part of the Blue Button request log and have other indicators that they are to/from BB
* Separates request and response logging into two functions, called before the request is made and after the response is received

### Security Implications
No new information has been added; all values being logged in responses are already logged under requests. The Blue Button URI may contain a Blue Button patient ID or hashed HICN.

### Acceptance Validation
Log entry examples of requests to and responses from BB dev environment. `beneficiary` ID is synthetic.
Screen Shot 2019-06-13 at 1 28 45 PM
Screen Shot 2019-06-14 at 12 19 02 PM

### Feedback Requested
Any

### DIFF
--- a/bcda/client/bluebutton.go
+++ b/bcda/client/bluebutton.go
@@ -32,7 +32,6 @@ type APIClient interface {
 	GetExplanationOfBenefitData(patientID, jobID string) (string, error)
 	GetPatientData(patientID, jobID string) (string, error)
 	GetCoverageData(beneficiaryID, jobID string) (string, error)
-
 	GetBlueButtonIdentifier(hashedHICN string) (string, error)
 }
 
--- a/bcda/client/bluebutton.go
+++ b/bcda/client/bluebutton.go
@@ -144,8 +144,11 @@ func (bbc *BlueButtonClient) getData(path string, params url.Values, jobID strin
 
 	addRequestHeaders(req, reqID)
 
+	go logRequest(req, jobID)
 	resp, err := bbc.httpClient.Do(req)
-	logRequest(req, resp, jobID)
+	if resp != nil {
+		logResponse(req, resp, jobID)
+	}
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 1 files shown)


embh:
Content length is removed here: https://github.com/CMSgov/bcda-app/pull/294/files#diff-ed36edc0ce96fd7e78c1bc475cca12c5L199

bcda/client/bluebutton.go
```diff
@@ -187,17 +187,16 @@ func logRequest(req *http.Request, resp *http.Response, jobID string) {
...
-	}).Infoln("Blue Button request")
+	}).Infoln("request")
 
 	if resp != nil {
 		logger.WithFields(logrus.Fields{
 			"resp_code":      resp.StatusCode,
-			"bb_query_id":    resp.Header.Get("BlueButton-OriginalQueryId"),
-			"bb_query_ts":    resp.Header.Get("BlueButton-OriginalQueryTimestamp"),
-			"bb_uri":         resp.Header.Get("BlueButton-OriginalUrl"),
+			"bb_query_id":    req.Header.Get("BlueButton-OriginalQueryId"),
```
tbellj: shouldn't these values be set from the resp param and not the req??

embh: Those headers used to get populated in the BB responses I think (which is why we had it set up that way originally) but don't anymore. So request is the only place where we can get them.

tbellj: oh okay, so BB changed the way they pass the data back to us. Gotcha

tbellj:
one more minor tiny change, but could the screenshot reflect both a call to getPatientData and getCoverageData??

PR #295, opened 2019-06-14
Establish a base code tree for SSAS

DeirdreHolub:
### Fixes [BCDA-1449](https://jira.cms.gov/browse/BCDA-1449) and [BCDA-1451](https://jira.cms.gov/browse/BCDA-1451)
We are setting up a new code tree for the System-to-System Authentication Service (SSAS) to live in so that it can be built and deployed as a standalone service or an embedded library. This PR is the first step in that process.

This PR establishes the tree and introduces it into the automated testing cycle in a separate docker container.

We have several tasks for refactoring the existing auth implementation in place prior to migrating the code (copying) the code to the SSAS tree. Once those are complete, the tree will grow rapidly.

### Security Implications
None. The purpose of this PR is to get the new tree configured for testing. 

### Acceptance Validation
All tests, including the simple test of main in new tree, run and pass.

### Feedback Requested
Any.


ssas/README.md
```diff
@@ -0,0 +1,60 @@
...
+    - _plugins_
+      - **alpha**
+        - alpha.go
+        - backend.go
+      - **okta**
+        - mokta.go
+        - okta.go
+        - oktaclient.go
+        - oktajwk.go
+    - provider.go
```
dhgreene: Based on my experience working in this proposed tree for a few hours, I'd recommend `provider.go` be put in the `/ssas/plugins` directory in a `plugins` package.  

It appears that separating the plugins into their own directories puts them in different packages, which `provider.go` will need to import.  Since we're wisely not allowing the `ssas` package to import its children, `provider.go` needs to be somewhere else.

DeirdreHolub: provider.go should be refactored to have only the interface in it.

PR #298, opened 2019-06-18
BCDA-1441: Additional error handling in smoke tests

embh:
### Fixes [BCDA-1441](https://jira.cms.gov/browse/BCDA-1441)
There are intermittent decryption errors when the smoke tests are run, but not enough information to determine the cause.

### Proposed changes:
Check for more scenarios that could cause decryption to fail and print errors to stdout.

### Change Details
* Go decryptor errors print to stdout
* bcda_client.go checks for `encryptedKey` and `privateKeyFile`, and prints an error and exits if either does not exist

### Security Implications
No security impact. These changes apply to smoke tests, which run with synthetic data.

### Acceptance Validation
A few examples of errors from the Go decryptor and test client.
Screen Shot 2019-06-17 at 2 37 42 PM
Screen Shot 2019-06-17 at 2 39 16 PM

### Feedback Requested
Any. Can you think of other places where the smoke tests or decryptors need to print more information?

### DIFF
--- a/decryption_utils/Go/decrypt.go
+++ b/decryption_utils/Go/decrypt.go
@@ -29,10 +29,21 @@ func init() {
 	flag.StringVar(&private, "pk", "", "location of private key to use for decryption of symmetric key")
 	flag.Parse()
 
-	if encryptedKey == "" || encryptedfilepath == "" || private == "" {
-		fmt.Println("missing argument(s)")
+	var missingArgs []string
+	if encryptedKey == "" {
+		missingArgs = append(missingArgs, "key")
+	}
+	if encryptedfilepath == "" {
+		missingArgs = append(missingArgs, "file")
+	}
+	if private == "" {
+		missingArgs = append(missingArgs, "pk")
+	}
+	if len(missingArgs) > 0 {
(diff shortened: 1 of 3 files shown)


decryption_utils/Go/decrypt.go
```diff
@@ -66,13 +77,13 @@ func decryptFile(privateKey *rsa.PrivateKey, encryptedKey []byte, filename strin
 	decryptedKey, err := rsa.DecryptOAEP(
 		sha256.New(), rand.Reader, privateKey, encryptedKey, []byte(base))
 	if err != nil {
-		fmt.Println("Failed to decrypt encrypted key")
+		fmt.Println("Failed to decrypt encrypted key. Error:", err)
 		os.Exit(3)
```
msnook: Is there value in returning different error codes now that we are properly reporting the exact error?  Does it make sense to switch it back to returning the standard exit code of 1 in the event of an error?

embh: Is there any chance we'd want to handle each exit code differently when the decryptor is called from bcda_client.go? I can't think of any reason to when we're printing the errors, so I agree that it makes sense to set them all to 1.

msnook: Agree!

PR #299, opened 2019-06-18
BCDA-1204 Remove tokens table

alexhova:
### Fixes [BCDA-1204](https://jira.cms.gov/browse/BCDA-1204)
Since we are no longer saving tokens to the db, we don't need a tokens table.  There is also some related code that will not be used

### Proposed changes:
- add migration to drop tokens table
- remove now unnecessary fields from the token model and some funcs using the token model as a receiver

### Security Implications
No security impact here.  We already aren't saving the tokens to the db so removing the tokens table does not affect security of the app

- [ ] This change adds new software dependencies
- [ ] This change alters security controls or supporting software
- [ ] This change introduces storage of new data
- [ ] A security checklist been completed for this change

### Acceptance Validation
Tests are passing and everything seems to work just as it did before removal

### Feedback Requested
Is there anything more that can be deleted?

### DIFF
--- a/bcda/auth/models.go
+++ b/bcda/auth/models.go
@@ -17,37 +17,18 @@ func InitializeGormModels() *gorm.DB {
 
 	// Migrate the schema
 	// Add your new models here
-	db.AutoMigrate(
-		&Token{},
-	)
-
-	// force manual deletion of foreign key and this related record (you can delete a Token, but not an aco with a token
-	db.Model(&Token{}).AddForeignKey("aco_id", "acos(uuid)", "RESTRICT", "RESTRICT")
+	db.AutoMigrate()
 
 	return db
 }
 
 type Token struct {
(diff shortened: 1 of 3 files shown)


bcda/auth/models.go
```diff
@@ -17,37 +17,18 @@ func InitializeGormModels() *gorm.DB {
 
 	// Migrate the schema
 	// Add your new models here
-	db.AutoMigrate(
-		&Token{},
-	)
-
-	// force manual deletion of foreign key and this related record (you can delete a Token, but not an aco with a token
-	db.Model(&Token{}).AddForeignKey("aco_id", "acos(uuid)", "RESTRICT", "RESTRICT")
+	db.AutoMigrate()
```
dhgreene: Is this a NOOP?  Should it still be called?

PR #303, opened 2019-06-20
BCDA-1427 Feature: Add function to create system

dhgreene:
### Fixes [BCDA-1427](https://jira.cms.gov/browse/BCDA-1427)
Context: The /auth/register endpoint will accept an ACO CMS ID and a public key, create a system in the database, and return credentials for that system.  The auth/system endpoint will also create a system.

Task: create a public function that will be called by either endpoint to create a system.  

### Proposed changes:
- Copy `bcda/auth/logger.go` to `ssas/logger.go`.  A future refactoring can remove the former file.
- Change `ssas/logger.go` functions to public functions, since callers each have their own package.
- Copy `auth.generateClientSecret()` to `ssas.GenerateClientSecret()` and temporarily place in `ssas/systems.go`.
- Add `ssas.GetSystemByClientID()` to facilitate looking up by the ID most commonly used for a System.

**Note that this code depends on PR #302 and should be reviewed after it has been merged with master or by viewing commit [9804271](https://github.com/CMSgov/bcda-app/commit/98042715e41f84066d2400e87b5ab58593e2a21d).**

### Security Implications
All of the arguments to `RegisterSystem()` will be coming as user input, and should be validated accordingly.

### Feedback Requested
- Since three tables are involved, should system creation be inside a transaction, even at the cost of complexity?
- In `ssas/systems.go` line 229, the question is raised whether when logging we care if `RegisterSystem()` was called from the CLI, the ACO via API, or ACO-MS via API.  What is your answer?
- Are there any code quality/testing suggestions?


alexhova:
>In ssas/systems.go line 229, the question is raised whether when logging we care if RegisterSystem() was called from the CLI, the ACO via API, or ACO-MS via API. What is your answer?

In my opinion, yes, let's log it.  it could help us in the future.  smart idea

alexhova:
>Since three tables are involved, should system creation be inside a transaction, even at the cost of complexity?

Good question.  I think this would be a good question for us to discuss to decide on a best practice in general.  I think it makes sense for this to be transactional (update all tables or do nothing).  But, we'll also want to think about db locks when we do that

dhgreene:
> > In ssas/systems.go line 229, the question is raised whether when logging we care if RegisterSystem() was called from the CLI, the ACO via API, or ACO-MS via API. What is your answer?
> 
> In my opinion, yes, let's log it. it could help us in the future. smart idea

OK, as discussed in standup we can take care of that logging from the caller.

dhgreene:
> > Since three tables are involved, should system creation be inside a transaction, even at the cost of complexity?
> 
> Good question. I think this would be a good question for us to discuss to decide on a best practice in general. I think it makes sense for this to be transactional (update all tables or do nothing). But, we'll also want to think about db locks when we do that

@alexhova and @DeirdreHolub this has been updated to use a transaction.

PR #304, opened 2019-06-20
BCDA-1321 Feature: Added additional tracking data to request headers

tbellj:
Fixes [BCDA-1321](https://jira.cms.gov/browse/BCDA-1321)

Problem:
Previously, BlueButton had no way of tracking our api calls within their own internal systems creating a traceability problem. Hopefully this alleviates that.

Proposed changes:
Code now grabs the cmsID in the processJob method and passes it the chosen function. The variable is later added as request header along with the current jobID. 
Keys used: BCDA-JOBID, BCDA-CMSID.

Change Details
bcda/client/bluebutton.go - added additional cmsID parameter in method signature. Added jobID and cmsID to request headers.
bcda/client/bluebutton_test.go- added test for request headers
bcda/testUtils/client.go - added cmsID to method parameters
bcdaworker/main.go - now passes the cms id into the bbFunc call
bcdaworker/main_test.go - added cmsID to test method calls
db/fixtures.sql - created a cmsID to our synthetic ACO Lorem Ipsum 

Security Implications:
No, this change does not deal with any PII/PHI.

Acceptance Validation:
Yes, all tests pass.

Screenshot:
Screen Shot 2019-06-25 at 11 54 12 AM

Feedback Requested:
Should we log the cmsID along with the jobID in our logRequest and logResponse methods?


bcda/client/bluebutton.go
```diff
@@ -182,14 +182,16 @@ func addRequestHeaders(req *http.Request, reqID uuid.UUID) {
 	req.Header.Add("BlueButton-OriginalUrl", req.URL.String())
 	req.Header.Add("BlueButton-OriginalQuery", req.URL.RawQuery)
 	req.Header.Add("BlueButton-BackendCall", "")
+	req.Header.Add("BCDA-JOBID", jobID)
+	req.Header.Add("BCDA-CMSID", cmsID)
 }
 
 func logRequest(req *http.Request, jobID string) {
```
embh: Can we drop the `jobID` parameter from `logRequest()` and `logResponse()` now that we're getting it from the header?

bcda/client/bluebutton.go
```diff
@@ -29,9 +29,9 @@ var logger *logrus.Logger
 const blueButtonBasePath = "/v1/fhir"
 
 type APIClient interface {
-	GetExplanationOfBenefitData(patientID, jobID string) (string, error)
-	GetPatientData(patientID, jobID string) (string, error)
-	GetCoverageData(beneficiaryID, jobID string) (string, error)
+	GetExplanationOfBenefitData(patientID, jobID string, cmsID string) (string, error)
```
embh:
```suggestion
	GetExplanationOfBenefitData(patientID, jobID, cmsID string) (string, error)
```
Extra `string` in there.

What do you think about renaming `cmsID` to `acoCMSID` (or `cmsACOID`) everywhere? I can imagine someone wondering if it means an ACO ID or perhaps some beneficiary ID.

bcda/client/bluebutton_test.go
```diff
@@ -212,6 +214,39 @@ func (s *BBTestSuite) TestHashHICN() {
...
+
+	assert.Equal(s.T(), req.URL.String(), req.Header.Get("BlueButton-OriginalUrl"))
+	assert.Equal(s.T(), req.URL.RawQuery, req.Header.Get("BlueButton-OriginalQuery"))
+	assert.Equal(s.T(), "", req.Header.Get("BlueButton-BackendCall"))
+
+	assert.Equal(s.T(), "543210", req.Header.Get("BCDA-JOBID"))
+	assert.Equal(s.T(), "A00234", req.Header.Get("BCDA-CMSID"))
+
+}
+
```
embh: Thank you for adding a test for that function!

bcda/testUtils/client.go
```diff
@@ -15,7 +15,7 @@ type BlueButtonClient struct {
 	HICN *string
 }
 
-func (bbc *BlueButtonClient) GetExplanationOfBenefitData(patientID string, jobID string) (string, error) {
+func (bbc *BlueButtonClient) GetExplanationOfBenefitData(patientID, jobID, cmsID string) (string, error) {
 	args := bbc.Called(patientID)
```
embh: Should the parameters on this `.Called()` be updated also? (I think you mentioned this in our discussion earlier)

bcdaworker/main.go
```diff
@@ -107,7 +113,7 @@ func processJob(j *que.Job) error {
 		}
 	}
 
-	fileUUID, err := writeBBDataToFile(bb, jobArgs.ACOID, jobArgs.BeneficiaryIDs, jobID, jobArgs.ResourceType)
+	fileUUID, err := writeBBDataToFile(bb, jobArgs.ACOID, *aco.CMSID, jobArgs.BeneficiaryIDs, jobID, jobArgs.ResourceType)
```
embh: This has me wondering yet again if we should keep our internal ACO UUIDs now that we have the CMS IDs. 🤔I'll add it to the brainstorming topics.

bcda/client/bluebutton.go
tbellj: couldn't agree more! I think we can safely rename the cmsID --> acoCMSID but I believe the header key has to stay the same (BCDA-CMSID). This is what BB is expecting from us

bcda/client/bluebutton.go
tbellj: yessss. Again you got the eyes

bcda/testUtils/client.go
tbellj: yes they should and the GetData method in this class also. Problem is updating will cause failures in our main_test.go, and I started to feel like the change was out of scope of what this ticket should be doing. But I do think a ticket should be created to not only update this but also make our main_test.go more consistent in either having all of the endpoints tested or none, similar to our earlier convo in slack.

bcda/client/bluebutton.go
tbellj: @embh actually after looking at that name change, we use cmsID everywhere, I found 12 different files. I think if we do it, then we should do it do it. What do you think about making that a separate ticket?

embh: I see some uses of `cmsID`, but they look to be on the ACO model or functions that are related to ACOs (e.g., `GetACOByCMSID()`. I don't think those need to change.

tbellj: oh okay, so just the new parameter additions??

embh: Yes, sorry about that! My original comment was confusing and should have said "everywhere *in this PR*." :)

tbellj: I actually think this is fine for now - 2 reasons mostly.
1) Based on your earlier comment: "This has me wondering yet again if we should keep our internal ACO UUIDs now that we have the CMS IDs." I think this is valid and the cmsID can just be our aco id or we can decide the name at a later date. I think to do the name change now from  cmsID --> acoCMSID/cmsACOID doesn't really make a lot of sense if this is something we're really considering.
2) Also the name change really needs to be done in all locations of our code not just this pr, lets be consistent and lock these names down, not just this one but also the beneficiary id vs. patient id thing. We should have a ticket to update all of this.

msnook: @tbellj Closing the loop here.  Can you make a ticket to address the out-of-scope concerns?

PR #305, opened 2019-06-20
BCDA-711: Introduce retry mechanism in data retrieval from BB

embh:
### Fixes [BCDA-711](https://jira.cms.gov/browse/BCDA-711)
When a request to Blue Button fails, an entry is added to the errors NDJSON file, with no way for users to re-request only the failed subset of data from BCDA. Some of these errors may be remedied by retrying failed BB requests.

### Proposed changes:
Add retry mechanism in Blue Button client.

### Change Details
* Adds retry logic in `BlueButtonClient.getData()`, using new environment variables:
  * `BB_REQUEST_MAX_TRIES` - Maximum attempts per Blue Button request
  * `BB_REQUEST_RETRY_INTERVAL_MS` - Length of pause, if any, that should happen between requests
* Splits bluebutton_test.go into two tests suites, to separate those that make requests from those that don't

### Security Implications
There is no change to how data is handled, but BCDA will now make more attempts to retrieve it from Blue Button if necessary.

### Acceptance Validation
Added tests with Blue Button client pointing to `httptest.Server` that returns only `500` responses.

### Feedback Requested
Any

### DIFF
--- a/bcda/client/bluebutton.go
+++ b/bcda/client/bluebutton.go
@@ -131,8 +131,6 @@ func (bbc *BlueButtonClient) getData(path string, params url.Values, jobID strin
 	txn := m.Start(path, nil, nil)
 	defer m.End(txn)
 
-	reqID := uuid.NewRandom()
-
 	bbServer := os.Getenv("BB_SERVER_LOCATION")
 
 	req, err := http.NewRequest("GET", bbServer+path, nil)
@@ -142,29 +140,27 @@ func (bbc *BlueButtonClient) getData(path string, params url.Values, jobID strin
 
 	req.URL.RawQuery = params.Encode()
 
-	addRequestHeaders(req, reqID)
+	addRequestHeaders(req, uuid.NewRandom())
 
-	go logRequest(req, jobID)
-	resp, err := bbc.httpClient.Do(req)
-	if resp != nil {
-		logResponse(req, resp, jobID)
-	}
-	if err != nil {
-		return "", err
-	}
+	tryCount := 0
+	maxTries := 3
+	retryInterval := 1000
 
-	defer resp.Body.Close()
+	for tryCount < maxTries {
+		tryCount++
+		if tryCount > 1 {
+			time.Sleep(time.Duration(retryInterval) * time.Millisecond)
+		}
 
-	if resp.StatusCode >= 400 {
-		return "", errors.New(resp.Status)
-	}
-
-	data, err := ioutil.ReadAll(resp.Body)
-	if err != nil {
-		return "", err
+		data, err := bbc.tryRequest(req, jobID)
+		if err != nil {
+			logger.Error(err)
+			continue
+		}
+		return data, nil
 	}
 
-	return string(data), nil
+	return "", fmt.Errorf("Blue Button request %s failed %d time(s)", req.Header.Get("BlueButton-OriginalQueryId"), tryCount)
 }
 
 func addRequestHeaders(req *http.Request, reqID uuid.UUID) {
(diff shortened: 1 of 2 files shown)


bcda/client/bluebutton.go
```diff
@@ -142,29 +140,29 @@ func (bbc *BlueButtonClient) getData(path string, params url.Values, jobID strin
...
-	}
-	if err != nil {
-		return "", err
-	}
+	queryID := uuid.NewRandom()
+	addRequestHeaders(req, queryID)
 
-	defer resp.Body.Close()
+	tryCount := 0
+	maxTries := utils.GetEnvInt("BB_REQUEST_MAX_TRIES", 3)
```
tbellj: BB_REQUEST_MAX_TRIES, BB_REQUEST_RETRY_INTERVAL_MS do these need to be env vars on the ops side?? Also do we want to add this to our docker.compose.yml files or are we going away from that?

bcda/client/bluebutton.go
```diff
@@ -142,29 +140,29 @@ func (bbc *BlueButtonClient) getData(path string, params url.Values, jobID strin
...
+	retryInterval := utils.GetEnvInt("BB_REQUEST_RETRY_INTERVAL_MS", 1000)
 
-	if resp.StatusCode >= 400 {
-		return "", errors.New(resp.Status)
-	}
+	for tryCount < maxTries {
+		tryCount++
+		if tryCount > 1 {
+			logger.Infof("Retrying Blue Button request %s in %d ms...", queryID, retryInterval)
```
tbellj: Would also be nice to know the number attempted so far in this log statement :)

bcda/client/bluebutton.go
embh: They do need to be in the ops repo once we decide those values, but it won't break anything to not set them yet because there are defaults. Not really necessary in the Docker configs for the same reason.

bcda/client/bluebutton.go
```diff
@@ -184,6 +182,29 @@ func addRequestHeaders(req *http.Request, reqID uuid.UUID) {
 	req.Header.Add("BlueButton-BackendCall", "")
 }
 
+func (bbc *BlueButtonClient) tryRequest(req *http.Request, jobID string) (string, error) {
+	go logRequest(req, jobID)
+	resp, err := bbc.httpClient.Do(req)
+	if resp != nil {
+		defer resp.Body.Close()
+		logResponse(req, resp, jobID)
```
tbellj: Do we want to continue logging the req in the logResponse method even though we just logged it a few lines earlier? The usage of the request parameter in both of those methods is kinda identical. But that could be worked in another ticket.

bcda/client/bluebutton.go
tbellj: true the defaults would work, but would also like to get some consistency on when we should create separate env vars and when the defaults are okay. This is a bigger discussion topic outside of this particular ticket, would like to hear @msnook opinion on this.

bcda/client/bluebutton.go
embh: Good question. As long as we have the request/query UUID, we could get the timestamp, URL, and job ID from the request log entry. So it could just be query ID and response code... but maybe we want all the fields for easier searching?

bcda/client/bluebutton.go
embh: Added! 
Screen Shot 2019-06-21 at 11 36 13 AM

bcda/client/bluebutton.go
msnook: I like sticking with the defaults for now.  We have the flexibility of changing if we need to, without having to redeploy new code.

PR #314, opened 2019-07-05
Move needed tooling code into SSAS; removed bcda dependencies

DeirdreHolub:
### Fixes [BCDA-1450](https://jira.cms.gov/browse/BCDA-1450)

To maintain SSAS as a standalone code base, it needs to have its own implementations of tooling that currently exists in bcda. This PR brings over much of that needed tooling.

### Proposed changes:
Uses of code in these bcda packages needed to be addressed:
- bcda/database
- bcda/auth/Hash
- bcda/auth/rsautils
- bcda/utils
- auth/Credentials

### Change Details
- the functions formerly provided in the database package are now part of the ssas package, and are implemented in connections.go
- the Hash struct and its functions are now part of the ssas package, and are implemented in hash.go
- the functions formerly provided in the rsautils package are now part of the ssas package, and are implemented in rsakeys.go
- the functions formerly provided in the utils package are now part of the cfg package, and are implemented in envv.go
- the Credentials struct is now in the ssas package, hanging out all by itself in credentials.go.

### Security Implications
None, given that this PR moves code that has already been reviewed from one code tree to another. The most significant code is the Hash structure, which has been thoroughly reviewed and was the subject of a SIA.

### Acceptance Validation
All tests run and pass.

### Feedback Requested
Consider these specific changes:
- moved connections.go into ssas instead of keeping it in a database package. I think it's too specific to our use and too small to justify a separate package.
- similarly, Hash, Credentials and rsautils are all in the ssas package. Perhaps credentials belongs in systems.go
- established a cfg package for configuration-related code, because I think this code will grow.

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -32,6 +32,9 @@ services:
       - AUTH_HASH_ITERATIONS=130000
       - AUTH_HASH_KEY_LENGTH=64
       - AUTH_HASH_SALT_SIZE=32
+      - SSAS_HASH_ITERATIONS=130000
+      - SSAS_HASH_KEY_LENGTH=64
+      - SSAS_HASH_SALT_SIZE=32
       - CCLF_IMPORT_STATUS_RECORDS_INTERVAL=10
     volumes:
       - .:/go/src/github.com/CMSgov/bcda-app
--- a/ssas/connection.go
+++ b/ssas/connection.go
@@ -1,0 +1,48 @@
+package ssas
+
+import (
+	"database/sql"
+	"log"
+	"os"
+	"runtime"
+
+	"github.com/jinzhu/gorm"
+	_ "github.com/jinzhu/gorm/dialects/postgres"
+	_ "github.com/lib/pq"
+)
+
+// Variable substitution to support testing.
+var LogFatal = log.Fatal
+
+func GetDbConnection() *sql.DB {
+	databaseURL := os.Getenv("DATABASE_URL")
+	db, err := sql.Open("postgres", databaseURL)
+	if err != nil {
+		LogFatal(err)
+	}
+	pingErr := db.Ping()
+	if pingErr != nil {
+		LogFatal(pingErr)
+	}
+	return db
+}
+
+func GetGORMDbConnection() *gorm.DB {
+	databaseURL := os.Getenv("DATABASE_URL")
+	db, err := gorm.Open("postgres", databaseURL)
+	if err != nil {
+		LogFatal(err)
+	}
+	pingErr := db.DB().Ping()
+	if pingErr != nil {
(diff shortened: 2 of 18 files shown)


ssas/connection.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/jinzhu/gorm"
+	_ "github.com/jinzhu/gorm/dialects/postgres"
+	_ "github.com/lib/pq"
+)
+
+// Variable substitution to support testing.
+var LogFatal = log.Fatal
+
+func GetDbConnection() *sql.DB {
+	databaseURL := os.Getenv("DATABASE_URL")
```
dhgreene: Is it too soon to specify a `ssas`-specific database URL?

ssas/rsakeys.go
```diff
@@ -0,0 +1,95 @@
...
+	"encoding/base64"
+	"encoding/json"
+	"encoding/pem"
+	"errors"
+	"fmt"
+	"math/big"
+)
+
+const RSAKEYMINBITS = 2048
+
```
dhgreene: Is now the time to add Godoc comments?

ssas/rsakeys.go
```diff
@@ -0,0 +1,95 @@
...
+func ReadPublicKey (publicKey string) (*rsa.PublicKey, error) {
```
dhgreene: Is now the time to add Godoc comments?

docker-compose.test.yml
```diff
@@ -32,6 +32,9 @@ services:
       - AUTH_HASH_ITERATIONS=130000
       - AUTH_HASH_KEY_LENGTH=64
       - AUTH_HASH_SALT_SIZE=32
```
embh: Do we need these `AUTH_...` vars now that there are SSAS versions?

ssas/credentials.go
```diff
@@ -0,0 +1,9 @@
+package ssas
+
+type Credentials struct {
```
dhgreene: I like your suggestion that this might belong better in `systems.go`.

docker-compose.test.yml
DeirdreHolub: We need to keep the existing code working whilst we work on our new version.

ssas/connection.go
DeirdreHolub: That one is more complicated to change, because it is used in many different places: launch scripts, makefiles, etc. I was saving it for later when we're done with writing new code and are working on integration.

ssas/credentials.go
DeirdreHolub: done.

ssas/rsakeys.go
DeirdreHolub: done.

PR #315, opened 2019-07-08
BCDA-1487 Feature: Returns an "un-nested" resource property in our ndjson responses.

tbellj:
Fixes [BCDA-1487](https://jira.cms.gov/browse/BCDA-1487)

Problem:
Prior to this change, our ndjson responses were returning values from an entries array as key value pairs of a resource map. We now just return the value portion of the resource map, as opposed to each individual entry.

Proposed changes:
Instead of returning each entry from our json object, we now return each resource value of that entry.

Change Details:
bcdaworker/main.go - grabs the resource value from the resource map.
bcdaworker/main_test.go - removes the "fullUrl" and "resource" strings from our tests. These are no longer included with the new way we drill down into the data. see here: http://hl7.org/fhir/STU3/bundle.html

Security Implications:
No, this change does not deal with any PII/PHI.

Acceptance Validation:
Yes all tests pass.

Screenshot:
Before: 
before
After: 
after

Feedback Requested:
Any and all

### DIFF
--- a/bcdaworker/main_test.go
+++ b/bcdaworker/main_test.go
@@ -104,6 +104,8 @@ func TestWriteEOBDataToFile(t *testing.T) {
 			var jsonOBJ map[string]interface{}
 			err := json.Unmarshal(scanner.Bytes(), &jsonOBJ)
 			assert.Nil(t, err)
+			assert.NotNil(t, jsonOBJ["status"], "JSON should contain a value for `status`.")
+			assert.NotNil(t, jsonOBJ["type"], "JSON should contain a value for `type`.")
 		}
 		assert.False(t, scanner.Scan(), "There should be only 66 entries in the file.")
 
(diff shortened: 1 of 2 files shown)


bcdaworker/main_test.go
```diff
@@ -104,8 +104,6 @@ func TestWriteEOBDataToFile(t *testing.T) {
 			var jsonOBJ map[string]interface{}
 			err := json.Unmarshal(scanner.Bytes(), &jsonOBJ)
 			assert.Nil(t, err)
-			assert.NotNil(t, jsonOBJ["fullUrl"], "JSON should contain a value for `fullUrl`.")
-			assert.NotNil(t, jsonOBJ["resource"], "JSON should contain a value for `resource`.")
```
embh: Are there other fields we could check for? Based on https://www.hl7.org/fhir/explanationofbenefit.html, there are some that should always be there, like `status` and `type`.

tbellj: yeah I was looking for something to key off of, but I would like to expand this check to our Coverage and Patient endpoints as well. Again tho, this file only tests our EOB stuff and not the others but we've already identified that in a future ticket. Will only implement this for EOB at the moment as suggested.

PR #316, opened 2019-07-09
BCDA-1144 add post group endpoint to admin routes

alexhova:
### Fixes [BCDA-1144](https://jira.cms.gov/browse/BCDA-1144)
This adds the endpoint to create a group via the auth api according to the ssas tech spec

### Proposed changes:
- add a function to the group models to create a group in the db
- add a function to api.go to take a request and create the group
- add a route to route a post request to the api func to create a group
- add tests

### Security Implications
No security concerns here.  While this is still part of the BCDA API the same security should be applied to this new endpoint

### Acceptance Validation
Tests pass.  New and old

### Feedback Requested
Any feedback is welcome.  Should we add additional validation to the contents of the JSON that is stored in the groups table of the db?

### DIFF
--- a/ssas/groups.go
+++ b/ssas/groups.go
@@ -4,8 +4,6 @@ import (
 	"encoding/json"
 	"log"
 
-	"github.com/CMSgov/bcda-app/ssas"
-
 	"github.com/jinzhu/gorm"
 	"github.com/jinzhu/gorm/dialects/postgres"
 )
@@ -33,12 +31,12 @@ type Group struct {
 
 func CreateGroup(gd GroupData) (Group, error) {
 	event := Event{Op: "CreateGroup", TrackingID: gd.ID}
-	ssas.OperationStarted(event)
+	OperationStarted(event)
 
 	gdBytes, err := json.Marshal(gd)
 	if err != nil {
 		event.Help = err
-		ssas.OperationFailed(event)
+		OperationFailed(event)
 		return Group{}, err
 	}
 
@@ -52,11 +50,11 @@ func CreateGroup(gd GroupData) (Group, error) {
 	err = db.Save(&g).Error
 	if err != nil {
 		event.Help = err
-		ssas.OperationFailed(event)
+		OperationFailed(event)
 		return Group{}, err
 	}
 
-	ssas.OperationSucceeded(event)
+	OperationSucceeded(event)
 	return g, nil
 }
 
--- a/ssas/groups.go
+++ b/ssas/groups.go
@@ -1,6 +1,7 @@
 package ssas
 
 import (
+	"encoding/json"
 	"log"
 
 	"github.com/jinzhu/gorm"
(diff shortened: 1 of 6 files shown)


ssas/groups.go
```diff
@@ -28,16 +29,38 @@ type Group struct {
 	Data    postgres.Jsonb `json:"data"`
 }
 
+func CreateGroup(gd GroupData) (Group, error) {
+	gdBytes, err := json.Marshal(gd)
+	if err != nil {
+		return Group{}, err
```
dhgreene: Is this a function that needs `ssas.OperationStarted()` and `ssas.OperationSucceeded()`/`ssas.OperationFailed()`?  If so, we should start the logging before our first return.

ssas/service/admin/api.go
```diff
@@ -3,10 +3,39 @@ package admin
...
 
+func createGroup(w http.ResponseWriter, r *http.Request) {
+	gd := ssas.GroupData{}
+	err := json.NewDecoder(r.Body).Decode(&gd)
+	if err != nil {
+		http.Error(w, "Failed to create group due to invalid request body", http.StatusBadRequest)
+		return
+	}
+
+	g, err := ssas.CreateGroup(gd)
```
dhgreene: I think we might want to log this [like Emily did in her recent PR](https://github.com/CMSgov/bcda-app/blob/ff40b28036af6e692384298fe2f1d08f207e16f8/ssas/service/admin/api.go#L25)

ssas/service/admin/api_test.go
```diff
@@ -25,6 +65,15 @@ func (s *APITestSuite) TearDownSuite() {
 	ssas.Close(s.db)
 }
 
+func (s *APITestSuite) TestCreateGroup() {
+	req := httptest.NewRequest("POST", "/group", strings.NewReader(sampleGroup))
+	handler := http.HandlerFunc(createGroup)
+	rr := httptest.NewRecorder()
+	handler.ServeHTTP(rr, req)
```
dhgreene: Should we delete the test group at the end of the test?  If so, consider using `ssas.CleanDatabase(group)`.

ssas/service/admin/router.go
```diff
@@ -1,13 +1,17 @@
...
-	"github.com/go-chi/chi"
 	"net/http"
+
+	"github.com/CMSgov/bcda-app/bcda/monitoring"
+	"github.com/go-chi/chi"
 )
 
 func NewRouter(middlewares ...func(http.Handler) http.Handler) http.Handler {
 	r := chi.NewRouter()
+	m := monitoring.GetMonitor()
```
dhgreene: IIRC we're intentionally not adding NewRelic monitoring at the moment.

ssas/groups.go
```diff
@@ -28,16 +29,38 @@ type Group struct {
...
+	}
+
+	g := Group{
+		GroupID: gd.ID,
```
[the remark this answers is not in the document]
alexhova: if this id is blank, it fails when it is saved to the db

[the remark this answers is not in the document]
dhgreene: When I test it, a blank gd.ID gets saved in the group_id column and succeeds.  We need that group_id not to be blank.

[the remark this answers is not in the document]
alexhova: my mistake!  i was thinking of a nil group_id.  it now fails with an error for a blank group_id

PR #317, opened 2019-07-11
BCDA-1499 Feature: Add Okta to SSAS

dhgreene:
### Fixes [BCDA-1499](https://jira.cms.gov/browse/BCDA-1499)
This PR sets up the basic environment (configuration, logging, and helper functions) for using Okta in the `/ssas` area.  It also implements calls to find Okta users and their multi-factor authentication (MFA) factors.

### Proposed changes:
* Put code needed by the coming MFA verification endpoints in `/ssas/service/public/okta.go`
* Put configuration variables and initialization in `/ssas/okta/okta.go`.  If/when we move the Okta auth provider into the `ssas` tree, these will be used by that code as well.

### Change Details
For unit testing, the `http.Client` is mocked so that we can fully control the responses.  Actual Okta responses for the test cases have been lightly scrubbed for test-case posterity.

This approach is fast and flexible for unit testing, but not effective for integration testing.  Bright ideas about how to do MFA integration testing in our CI/CD environment are  _very_ welcome.

### Security Implications
See the associated SIA

### Feedback Requested
* Code locations
* Missing test cases
* Brittle tests
* Other suggestions!


ssas/okta/okta.go
```diff
@@ -0,0 +1,89 @@
...
+	from https://hassansin.github.io/Unit-Testing-http-client-in-Go
+ */
+func (f RoundTripFunc) RoundTrip(req *http.Request) (*http.Response, error) {
+	return f(req), nil
+}
+
+/*
+	NewTestClient returns *http.Client with Transport replaced to avoid making real calls
+ */
+func NewTestClient(fn RoundTripFunc) *http.Client {
```
alexhova: this is a clever pattern!

PR #319, opened 2019-07-11
BCDA-1489 Feature: Block multiple concurrent attempts to endpoints

tbellj:
Fixes [BCDA-1489](https://jira.cms.gov/browse/BCDA-1489)

Problem:
Users of our API could potentially make multiple back-to-back calls to the same endpoint with no limitations. Not only does this stress our system resources causing things to slow, but it's also not useful for our users.

Proposed changes:
With these changes a user will no longer be able to make more than 1 request to the same endpoint when a job is in progress/pending. While this is happening, the same user is allowed to make a call to a different endpoint but subsequent calls after the first will result in a 429 too many requests error.

Change Details:
bcda/web/api.go - blocking code for subsequent calls to the same endpoint by the same ACO.
bcda/web/api_test.go - test code, different statuses, different endpoints, different ACO's.

Security Implications:
[No] This change does not deal with any PII/PHI. 
[No] This change adds new software dependencies
[No] This change alters security controls or supporting software
[No] This change introduces storage of new data
[No] A security checklist been completed for this change

But by limiting calls to our endpoints we do avoid the possibility of resource exhaustion from any malicious attackers. I believe this will improve our security posture going forward.

Acceptance Validation:
Yes, all tests cases pass.

Feedback Requested:
Any and all


bcda/web/api.go
```diff
@@ -110,16 +111,30 @@ func bulkRequest(t string, w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	db := database.GetGORMDbConnection()
+	defer database.Close(db)
+
 	acoID := ad.ACOID
+
+	var jobs []models.Job
+
+	// if we really do find this record
```
msnook: Any chance we can beef up the commentary here to describe what we're doing here, and why we're doing it?

PR #320, opened 2019-07-15
BCDA-1506: Create auth/system/:id:/credentials PUT endpoint

embh:
### Fixes [BCDA-1506](https://jira.cms.gov/browse/BCDA-1506)
The SSAS needs a mechanism for resetting credentials.

### Proposed changes:
Create `PUT /system/[id]/credentials`, which tells the service to disable the current secret for the system and create a new one. The new secret is returned.

### Change Details
* Creates router `PUT /system/[id]/credentials` path
* Adds `GetSystemByID()` and `ResetSecret()` to systems.go
* Adds `resetCredentials()` to admin/api.go

### Security Implications
This change resets the secret for a given system, which can help improve security.

### Acceptance Validation
Old and new tests pass.

### Feedback Requested
Any. A few questions:
* Should the logged event be called `ResetSecret` or `NewSecret`? My aim was to distinguish it from when a secret is created for the first time, but that might not be necessary.
* Is the response body format correct? Changes to content and/or naming? `{ "client_id": "1234", "client_secret": "secret" }`

### DIFF
--- a/ssas/systems.go
+++ b/ssas/systems.go
@@ -415,7 +415,7 @@ func GetSystemByID(id string) (System, error) {
 	)
 	defer Close(db)
 
-	if err := db.First(&system, id).Error; err != nil {
+	if err = db.First(&system, id).Error; err != nil {
 		err = fmt.Errorf("no System record found with ID %s", id)
 	}
 	return system, err
--- a/ssas/systems.go
+++ b/ssas/systems.go
@@ -415,6 +415,41 @@ func GenerateSecret() (string, error) {
 }
 
 /*
+	ResetSecret creates a new secret for the current system.
+*/
+func (system *System) ResetSecret(trackingID string) (string, error) {
+	db := GetGORMDbConnection()
+	defer Close(db)
+
+	newSecretEvent := Event{Op: "ResetSecret", TrackingID: trackingID, ClientID: system.ClientID}
+	OperationStarted(newSecretEvent)
+
+	secretString, err := GenerateSecret()
+	if err != nil {
+		newSecretEvent.Help = fmt.Sprintf("could not reset secret for clientID %s: %s", system.ClientID, err.Error())
+		OperationFailed(newSecretEvent)
+		return "", errors.New("internal system error")
+	}
+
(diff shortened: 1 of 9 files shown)


ssas/systems.go
```diff
@@ -414,6 +431,41 @@ func GenerateSecret() (string, error) {
...
+func (system *System) ResetSecret(trackingID string) (string, error) {
+	db := GetGORMDbConnection()
+	defer Close(db)
+
+	newSecretEvent := Event{Op: "ResetSecret", TrackingID: trackingID, ClientID: system.ClientID}
+	OperationStarted(newSecretEvent)
+
+	secretString, err := GenerateSecret()
+	if err != nil {
+		newSecretEvent.Help = fmt.Sprintf("could not reset secret for clientID %s: %s", system.ClientID, err.Error())
```
dhgreene: Presumably an error here or elsewhere in this function would have a message clear enough to distinguish what failed.  Do you think those that follow us might ever ask, "Did that unique cryptographic error bubble up from `GenerateSecret()` or `NewHash()`?"  I've often come down on the side of generating more unique error messages to combat that possibility, but what you have here has the upside of consistency.

ssas/systems.go
```diff
@@ -414,6 +431,41 @@ func GenerateSecret() (string, error) {
...
+		OperationFailed(newSecretEvent)
+		return "", errors.New("internal system error")
```
dhgreene: 👍 on returning a more generic error message, so that future public-facing endpoints can safely pass along the message.

dhgreene:
I like the unique name for the logged event.  The response body is great: it parallels the output from self-registration (which is specified by the RFC), but omits the unneeded system name.

ssas/systems.go
DeirdreHolub: The GenerateSecret() and NewHash() methods log their operation start, failure and success, so there should be no wonder here. In contrast to most approaches with error handling, which take a top down approach to logging (log the error at the last place it's handled), we are logging from the bottom up. We're doing this to provide useful diagnostic logs in production and also useful warning metrics.

PR #323, opened 2019-07-16
BCDA-1145 add update group endpoint

alexhova:
### Fixes [BCDA-1145](https://jira.cms.gov/browse/BCDA-1145)
This adds the PUT endpoint to update a group

### Proposed changes:
- add the route
- add the api func to decode the request and call the model func
- add the model func to update the group (except for the id and name)

### Security Implications
No security concerns on this one.  This is an admin endpoint and will only be accessible by the required systems

### Acceptance Validation
Tests pass!

### Feedback Requested
Any feedback is welcome


ssas/groups.go
```diff
@@ -66,6 +68,58 @@ func CreateGroup(gd GroupData) (Group, error) {
 	return g, nil
 }
 
+func UpdateGroup(id string, gd GroupData) (Group, error) {
```
dhgreene: If we're using the database ID as the group identifier for API calls (which is ok for uniqueness), we'll need to make sure we're returning it in `CreateGroup()`.  Right now that returns a 0 for the Group.ID.

ssas/groups.go
```diff
@@ -75,6 +129,23 @@ type GroupData struct {
...
+// Make the GroupData struct implement the sql.Scanner interface. This method
+// simply decodes a JSON-encoded value into the struct fields.
+func (gd *GroupData) Scan(value interface{}) error {
+	b, ok := value.([]byte)
+	if !ok {
+		return errors.New("type assertion to []byte failed")
+	}
+
+	return json.Unmarshal(b, &gd)
+}
```
embh: Could you clarify how `Value()` and `Scan()` are being used? My understanding is that one would create these methods if a field had a custom data type. So in this case, `Group.Data` would become type `GroupData` and be parsed automatically (not calling `Scan()` or `Value()` directly). https://github.com/jinzhu/gorm/issues/47

ssas/groups.go
alexhova: createGroup now returns the the created group so that the id can be used

ssas/groups_test.go
```diff
@@ -71,12 +71,35 @@ func (s *GroupsTestSuite) TestCreateGroup() {
...
 	_, err = CreateGroup(gd)
 	assert.EqualError(s.T(), err, "group_id cannot be blank")
 }
 
+func (s *GroupsTestSuite) TestUpdateGroup() {
+	groupBytes := []byte(SampleGroup)
+	gd := GroupData{}
+	err := json.Unmarshal(groupBytes, &gd)
+	assert.Nil(s.T(), err)
+	g, err := CreateGroup(gd)
```
embh: Would it be better to create the group directly? This test relies on `CreateGroup()`, which isn't being tested here.

ssas/service/admin/api.go
```diff
@@ -37,6 +38,36 @@ func createGroup(w http.ResponseWriter, r *http.Request) {
...
+	ssas.OperationCalled(ssas.Event{Op: "UpdateGroup", TrackingID: id, Help: "calling from admin.updateGroup()"})
+	g, err := ssas.UpdateGroup(id, gd)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Failed to update group. Error: %s", err), http.StatusBadRequest)
+		return
+	}
+
+	groupJSON, err := json.Marshal(g)
+	if err != nil {
+		http.Error(w, "Internal error", http.StatusInternalServerError)
```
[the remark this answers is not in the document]
dhgreene: Following up from a chat today: IMO the only change we'd need is an extra line here.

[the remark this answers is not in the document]
alexhova: happy to do that. i could also add it to the createGroup method similarly.  but then it _will_ log the success before logging the failure. so should i also add another operation failed to the next error check and put the operation succeeded just before the return of _this_ func?

ssas/service/admin/router.go
```diff
@@ -15,6 +15,7 @@ func init() {
 func Routes() *chi.Mux {
 	r := chi.NewRouter()
 	r.Post("/group", createGroup)
+	r.Put("/group/{id}", updateGroup)
```
dhgreene: Line 12 just above this (can't seem to tag it . . .) should somehow be updated to include the new endpoint.

alexhova: my understanding (from what we have already merged) was that the `/_info` endpoint just shows the resources that have methods in the admin namespace.  do you still think there is something to change here?

dhgreene: Yes, nothing to add without a refactor.  Probably fine for now.

PR #326, opened 2019-07-18
BCDA-1547: Create public-key endpoint

embh:
### Fixes [BCDA-1547](https://jira.cms.gov/browse/BCDA-1547)
BCDA needs a way to retrieve systems' public keys from SSAS.

### Proposed changes:
Create an endpoint for retrieving a system's public key.

### Change Details
* Adds `GET /system/[id]/key` route handled by new `getPublicKey()` function in api.go
* Adds `GetEncryptionKey()` to systems.go

### Security Implications
This new endpoint returns a system's public key, used by BCDA for encrypting data.
- This change **does not** add new software dependencies
- This change **does not** alter security controls or supporting software
- This change **does not** introduce storage of new data
- A security checklist **has not** been completed for this change

### Acceptance Validation
New tests added.

### Feedback Requested
Any


docker-compose.test.yml
```diff
@@ -20,6 +20,7 @@ services:
       - FHIR_PAYLOAD_DIR=../bcdaworker/data
       - FHIR_STAGING_DIR=../bcdaworker/tmpdata
       - FHIR_ARCHIVE_DIR=../bcdaworker/archive
+      - PENDING_DELETION_DIR=./tempCCLFDir
```
dhgreene: Is this line relevant to this PR?

embh: I added this after Deirdre pointed out the "invalid cross-device link" errors when the `import-synthetic-cclf-package` CLI command is used. Since it's not related to this ticket, I'll remove it and create a ticket to address this separately.

ssas/systems.go
```diff
@@ -134,14 +134,26 @@ func (system *System) GetSecret() (string, error) {
...
+		return encryptionKey, fmt.Errorf("cannot find key for clientID %s: %s", system.ClientID, err.Error())
+	}
+
+	return encryptionKey, nil
+}
+
+/*
+	GetPublicKey will retrieve the public key associated with the current system.
+*/
+func (system *System) GetPublicKey() (*rsa.PublicKey, error) {
```
dhgreene: If you don't need `GetPublicKey()` for the public-key endpoint, then you can delete this method.

embh: 👍 Done!

ssas/service/admin/api_test.go
```diff
@@ -161,6 +161,64 @@ func (s *APITestSuite) TestResetCredentials_InvalidSystemID() {
...
+	encrKey1 := ssas.EncryptionKey{
+		SystemID: system.ID,
+		Body:     key1Str,
+	}
+	err = s.db.Create(&encrKey1).Error
+	if err != nil {
+		s.FailNow(err.Error())
+	}
+
+	key2Str := "publickey2"
```
dhgreene: I might be missing why we need two public keys in this test.  There should only ever be one active per system.

embh: That was based on this criterion in the ticket: 
> If I submit two keys in a row (mimic-ing key rotation here), I get the second key I submitted, not the first

dhgreene: Ah!  I knew I was missing something.  😄 
I read that requirement as, "double-check that our soft-deletion key rotation strategy works."  So that would involve calling `ssas.SavePublicKey()` instead of putting the keys directly in the database.

embh:
Regarding @alexhova 's comment:
> Looks good. I just wonder if we should include logging similar to the other endpoints (`OperationCalled, Started, Succeeded, Failed`)

I'll add it, but I didn't originally because `SavePublicKey()` doesn't log events. Should we create a ticket for that?

embh:
> I think creating a ticket to just make sure logging across ssas in `api.go` and the model files is complete and consistent is a good idea. we can do our best along the way and circle back at the end and make sure it's all set.

👍 https://jira.cms.gov/browse/BCDA-1623

PR #327, opened 2019-07-18
BCDA-1500 Add /auth/authn/request endpoint

dhgreene:
### Fixes [BCDA-1500](https://jira.cms.gov/browse/BCDA-1500)
As the first of three MFA-related endpoints, `/auth/authn/request` will take a CMS identifier (for a user in Okta) and an MFA type (SMS, etc.) and return a message that the request was sent.  If everything goes well, the user will get a passcode (e.g. via SMS) to use with `/auth/authn/verify` (or if using the Push factor, the user's app will ask for approval and the next step will be `/auth/authn/transactions/{transaction_id}`).

Note that this PR should be reviewed after PR #317 as it builds on its code.

### Proposed changes:
* Add `PostFactorChallenge()` to send the request to Okta
* Add `RequestFactorChallenge()` to coordinate user ID and MFA factor ID lookups before calling `PostFactorChallenge()`
* Create an MFA provider interface and an ability to swap between live Okta communication and mocked responses (for integration testing--not yet used)
* Add a separate test suite for live Okta calls; this suite is protected by a build flag so that it is only run manually as needed
* Put `/auth/authn/request` in the public ssas router and add a Postman request for manual validation

### Security Implications
See associated SIA

### Acceptance Validation
Screen Shot 2019-07-18 at 4 58 10 PM

Screen Shot 2019-07-18 at 5 00 59 PM

### Feedback Requested
* Make sure to open all diffs if viewing in GitHub
* All comments are welcome!


ssas/README.md
```diff
@@ -56,13 +56,26 @@ Required values must be present in the docker-compose.*.yml files.
 | SSAS_HASH_ITERATIONS | Yes      | Controls how many iterations our secure hashing mechanism performs. Service will panic if this key does not have a value. |
 | SSAS_HASH_KEY_LENGTH | Yes      | Controls the key length used by our secure hashing mechanism. Service will panic if this key does not have a value. |
 | SSAS_HASH_SALT_SIZE  | Yes      | Controls salt size used by our secure hashing mechanism performs. Service will panic if this key does not have a value. |
+| SSAS_MFA_PROVIDER    | No       | Switches between mock Okta MFA calls and live calls.  Defaults to "Mock".
+| SSAS_MFA_CHALLENGE_REQUEST_MILLISECONDS | No | Minimum execution time for RequestFactorChallenge().  If not present, defaults to 1500.  In production, this should always be set longer than the longest expected execution time.  (Actual execution time is logged.)|
```
embh: I may have missed this-- where is the default set?

ssas/service/public/oktalive_test.go
```diff
@@ -0,0 +1,184 @@
...
+	factorVerification, err := s.oc.postFactorChallenge(s.userId, factor, trackingId)
+	if err == nil || factorVerification != nil {
+		s.FailNow("certificate signed by unexpected CA should abort TLS handshake")
+	}
+	okta.OktaCACertFingerprint = originalOktaCACertFingerprint
+}
+
+func (s *OktaLiveTestSuite) TestGetUserSuccess() {
+	fmt.Println("s.userId", s.userId)
+	fmt.Println("s.email", s.email)
```
embh: I suspect these were not intentionally left in :) (Or maybe they were!)

ssas/service/public/oktalive_test.go
```diff
@@ -0,0 +1,184 @@
...
+	s.oc = NewOktaMFA(okta.Client())
+}
+
+func (s *OktaLiveTestSuite) TestPostFactorChallengeSuccess() {
+	trackingId := uuid.NewRandom().String()
+	factor := Factor{Id: s.smsFactorId, Type: "sms"}
+
+	factorVerification, err := s.oc.postFactorChallenge(s.userId, factor, trackingId)
+	if err != nil || factorVerification == nil {
+		s.FailNow("factor result not parsed: " + err.Error())
```
embh: Thank you for putting messages on `FailNow()` calls! It's nice when a test actually reports why it failed. (And I am reminded to do that in tests I write!)

ssas/README.md
dhgreene: In a file that GitHub thinks has too many changes to show by default:
  https://github.com/CMSgov/bcda-app/blob/0576bb70472b6da50589862ebb97de73b8bf0da7/ssas/service/public/oktamfa.go#L63

ssas/service/public/oktalive_test.go
dhgreene: Good catch!

PR #331, opened 2019-07-23
BCDA-1146 implement delete group endpoint

alexhova:
### Fixes [BCDA-1146](https://jira.cms.gov/browse/BCDA-1146)
This PR implements the DELETE method for the Group resource

### Proposed changes:
- add the delete route
- add the api func to process the request
- add the model func to delete the group and its associated system and encryption keys from the db
- add tests

### Security Implications
No security implications at this point.  This will be an admin endpoint with security to ensure only authorized parties can affect their attributed groups

### Acceptance Validation
tests are passing!

### Feedback Requested
any feedback is welcome!

### DIFF
--- a/ssas/groups_test.go
+++ b/ssas/groups_test.go
@@ -79,34 +79,24 @@ func (s *GroupsTestSuite) TestCreateGroup() {
 }
 
 func (s *GroupsTestSuite) TestDeleteGroup() {
-	group := Group{GroupID: "groups-test-delete-group"}
+	group := Group{GroupID: "groups-test-delete-group-id"}
 	err := s.db.Create(&group).Error
 	if err != nil {
 		s.FailNow(err.Error())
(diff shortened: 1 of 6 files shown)


ssas/groups_test.go
```diff
@@ -77,6 +78,43 @@ func (s *GroupsTestSuite) TestCreateGroup() {
...
+	encrKey1 := EncryptionKey{
+		SystemID: system.ID,
+		Body:     key1Str,
+	}
+	err = s.db.Create(&encrKey1).Error
+	if err != nil {
+		s.FailNow(err.Error())
+	}
+
+	key2Str := "publickey2"
```
dhgreene: System:(active)EncryptKey should be a 1:1 relationship, so we probably shouldn't be testing with two active encryption keys.

PR #332, opened 2019-07-23
BCDA 1560: Add signing keys to server type

DeirdreHolub:
### Fixes [BCDA-1560](https://jira.cms.gov/browse/BCDA-1560)
Our servers, in order to mint tokens, need a token signing key. This PR adds a token signing key to the server type. This signing key is only used to sign tokens; it will not be used to secure communications channels for instances of the server type.

In addition, this PR adds a new invocation option to the executable. This new option runs the database migration script, but only if the DEBUG environment variable is set to true. The intended use for this option is to support integration testing with docker containers.

Finally, this PR adds the following commands to the makefile:

Command | Description
----------- | -----------
lint-ssas | lint only ssas files. The linter is silent unless it fails. Also runs the `gosec` scan, which reports each file as it examines it.
unit-test-ssas | run only ssas unit tests, using a separate shell script. You must have already started the db via `docker-compose up -d db` (or, if you want to see the logs, without the -d flag)
test-ssas | run lint-ssas and unit-test-ssas. Use in place of `make test` when your changes are only in the ssas tree.

### Proposed changes:
Add a server signing key to the server type. This signing key will be separate and distinct from the server's cert / key used for securing channels via ssl / tls / https.

### Change Details
- After creating the server, set the signing key from a pem file.

### Security Implications
- No PII
- A server without a valid signing key will fail to start. 
- Public key is produced directly from the private key, eliminating the possibility of mis-matched key files
- Enforces use of RSA keys
- Enforces minimum key size of 2048 bits
- Each server instance (in our case, admin and public) has its own separate signing key

### Acceptance Validation
All current tests pass.
Additional tests were added.
Postman tests run to confirm no breakage in API.

### Feedback Requested
Any feedback welcome.


ssas/rsakeys_test.go
```diff
@@ -1,25 +1,61 @@
...
+	assertT := assert.New(s.T())
+	var tests = []struct {
+		path   string
+		errMsg string
+	}{
+		{badBase64KeyPath, "decode"},
+		{notRSAKeyPath, "parse RSA"},
+		{tooSmallKeyPath, "insecure key length"},
+	}
+	for _, test := range tests {
```
dhgreene: This is my favorite part of the PR, with its terse code and thorough testing of keys.

DeirdreHolub:
> * Would it be just as easy to put a new Makefile in the `/ssas` directory?

Until this project is completely separated from BCDA, I think it is cleaner to keep the recipes in the same Makefile.

PR #333, opened 2019-07-23
BCDA-1550 Feature: Unknown file types to be removed within 24 hours.

tbellj:
Fixes [BCDA-1550](https://jira.cms.gov/browse/BCDA-1550)

Problem:
Problem: We previously dedicated a location for only cclf ingestions but now we have to account for an additional file type from 1-800 medicare for beneficiary suppressions. Our previous process would mark this file as an unknown file type and would immediately set it for deletion, we now hold on to this file (for no longer than a 24 hour period) for additional ETL processing.

Proposed changes:
In our cclf ingestion logic, we first check all files for metadata via the getCCLFFileMetadata method. If the filename doesn't align with what we expect, we then move that file to another location for deletion. However with these new changes, an additional check will be performed against the time we received the files. The unknown files will remain on our system for no longer than 24 hours as to not interfere with our security posture.

Change Details:
bcda/cclf/cclf.go - logic for checking suppression file types
bcda/cclf/cclf_test.go - test code for the suppression
bcda/bcdacli/cli_test.go - modified to account for an additional unknown file type
cclf_BadFileNames/T#EFT.ON.ACO.NGD1800.DPRF.D181120.T1000009 - added bene suppression file type 

Security Implications:
Yes, this code **will** handle files containing PII/PHI. We don't have the files quite yet but in the future this code will act on them.
This change **does not** add new software dependencies
This change **does not** alter security controls or supporting software
This change **does not** introduce storage of new data
A security checklist **has** been completed for this change

### Acceptance Validation
Yes, all tests pass.

Feedback Requested
Any and all


bcda/cclf/cclf_test.go
```diff
@@ -570,6 +570,14 @@ func (s *CCLFTestSuite) TestCleanupCCLF() {
 	testUtils.ResetFiles(s.Suite, BASE_FILE_PATH+"cclf/")
 }
 
+func (s *CCLFTestSuite) TestIsSuppressionFile() {
+	suppressionFilePath := BASE_FILE_PATH + "cclf_Suppression/T#EFT.ON.ACO.NGD1800.DPRF.D181120.T1000009"
```
msnook: Does `cclf_suppression` make sense?  Is this suppression file considered a CCLF file?

bcda/cclf/cclf.go
```diff
@@ -451,6 +459,12 @@ func sortCCLFFiles(cclfmap *map[string][]*cclfFileMetadata, skipped *int) filepa
 		metadata.filePath = path
 		metadata.deliveryDate = info.ModTime()
 		if err != nil {
+			if isSuppressionFile(info.Name()) {
```
msnook: Instead of checking for suppression file, lets just keep all files around if they are less than 24 hours.  On next execution (within 24 hours on cron), the file will be moved to pending deletion dir because we don't want to keep that sensitive data around for longer than 24 hours.

PR #334, opened 2019-07-24
BCDA-1593 Add /auth/authn endpoint

dhgreene:
### Fixes [BCDA-1593](https://jira.cms.gov/browse/BCDA-1593)
In order that ACO's can self-register systems, they need to be authenticated using multi-factor authentication (MFA).  This new API endpoint is the first in a series that, if successful, will yield a registration token that can be presented to `/auth/register`.

A later PR will add a token to the output, to be presented to other MFA endpoints to prove that password authentication was already successful.

### Proposed changes:
* Add `postPassword()` and `VerifyPassword()` to interact with Okta
* Write tests with the various responses expected from Okta
* Add a mocked `VerifyPassword()` to use in integration testing
* Add Postman request to our SSAS collection for the new endpoint
* Rename a struct used for return values to more accurately reflect its purpose

### Security Implications
See SIA

### Acceptance Validation
Screen Shot 2019-07-24 at 12 15 15 AM

### Feedback Requested
* Naming and consistency
* Opportunities for DRY-ing up the code
* Opportunities to add test coverage (currently at 71% for this package)
* Unnecessary structures (for instance, the utility of the `PasswordResponse` struct which has a single string value might be questionable)


ssas/okta/okta.go
```diff
@@ -35,26 +45,38 @@ func config() error {
 	if at != "" {
 		at = "[Redacted]"
 	}
-
 	OktaAuthString = fmt.Sprintf("SSWS %s", oktaToken)
+	OktaBaseUrl = os.Getenv("OKTA_CLIENT_ORGURL")
```
embh:
```suggestion
	OktaBaseURL = os.Getenv("OKTA_CLIENT_ORGURL")
```
https://github.com/golang/go/wiki/CodeReviewComments#initialisms

ssas/service/public/api.go
```diff
@@ -38,6 +38,190 @@ type RegistrationRequest struct {
 	JSONWebKeys JWKS   `json:"jwks"`
 }
 
+type MFARequest struct {
+	CmsID       string  `json:"cms_id"`
```
embh:
```suggestion
	CMSID       string  `json:"cms_id"`
```

ssas/service/public/router.go
```diff
@@ -35,6 +35,7 @@ func routes() *chi.Mux {
 	router := chi.NewRouter()
 	router.Use(service.NewAPILogger(), service.ConnectionClose)
 	router.Get("/token", token)
+	router.Post("/authn", VerifyPassword)
 	router.Post("/authn/request", RequestMultifactorChallenge)
```
DeirdreHolub: I think `request` is too generic for an endpoint. How about challenge? Then the sequence of requests would be:

POST /authn
POST /authn/challenge
POST /authn/verify
POST /authn/register

ssas/service/public/api.go
```diff
@@ -39,12 +39,69 @@ type RegistrationRequest struct {
...
+	trackingID = uuid.NewRandom().String()
+	event := ssas.Event{Op: "VerifyOktaPassword", TrackingID: trackingID, Help: "calling from public.VerifyPassword()"}
+	ssas.OperationCalled(event)
+	passwordResponse, err := GetProvider().VerifyPassword(passReq.CMSID, passReq.Password, trackingID)
+	if err != nil {
+		jsonError(w, "invalid_client_metadata", err.Error())
+		return
+	}
+
+	body, err := json.Marshal(passwordResponse)
```
DeirdreHolub: The password response just tells the user to `proceed to MFA verification`. The next step, authn/request, to do the mfa challenge requires a `factor_type`, but as a consumer of this api, I have no idea what the factor_types are.

dhgreene: Presumably, you know what factor type you registered with Okta, and originally we were keeping our cards close to our chest.  Now that we have password authentication completed before this endpoint is reached, we can give them a list in `POST /authn`.

DeirdreHolub:
FYI, I got the whole sequence to work with my Okta account and google auth in postman.

DeirdreHolub:
re
> Unnecessary structures (for instance, the utility of the PasswordResponse struct which has a single string value might be questionable)

What's the alternative? When it comes to parsing JSON in go, this is how it's most cleanly done. I prefer consistency in handling JSON as reducing the cognitive load for programmers.

ssas/service/public/api.go
```diff
@@ -39,12 +39,69 @@ type RegistrationRequest struct {
 }
 
 type MFARequest struct {
-	CmsID       string  `json:"cms_id"`
+	CMSID       string  `json:"cms_id"`
```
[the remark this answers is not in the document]
DeirdreHolub: Can we dodge that all by just calling LoginID?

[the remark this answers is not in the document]
dhgreene: Call `CMSID` `LoginID`?  OK.

ssas/service/public/api.go
dhgreene: BCDA-1624 has been created to remind us of this enhancement

PR #335, opened 2019-07-24
BCDA-1551 Feature: DO NOT MERGE (In testing) Added additional query parameter to patient endpoint to receive plaintext hicns

tbellj:
Fixes [BCDA-1551](https://jira.cms.gov/browse/BCDA-1551)

Problem:
Users need a way to identify claims data back to their patients and at the moment not sure how our api provides that. With this change a plaintext hicn will be provided via the patient endpoint giving our users a way to link payload data back to their patients.

Proposed changes:
This change will provide an un-hashed plaintext hicn for our users to use.

Change Details:
bcda/client/bluebutton.go - added query parameter to patient endpoint (behind feature flag)
bcda/client/bluebutton_test.go - testing logic for the above

Security Implications:
This change **will** deal with PII/PHI directly.
A security checklist has **not** been completed for this change

Acceptance Validation:
Yes, all tests pass

Feedback Requested:
Any and all


StewGoin:
> Security Implications:
No, this change does not deal with any PII/PHI at all.

I think this change *only* deals with PHI... if I read it correctly.

tbellj:
> > Security Implications:
> > No, this change does not deal with any PII/PHI at all.
> 
> I think this change _only_ deals with PII/PHI... if I read it correctly.

Okay, so let me ask the question. Are hicns PII/PHI and if so is simply seeing one a security violation?

StewGoin:
> Okay, so let me ask the question. Are hicns PII/PHI and if so is simply seeing one a security violation?
Yes, HICNs include an SSN as part of their value (could be the patient, could be the enrollee the patient is covered under). Aside from that health care ID's are PII/PHI as well in and of themselves.

https://en.wikipedia.org/wiki/Protected_health_information

Improper disclosure / breach /etc... would be just like any other PII/PHI.

rnagle:
@tbellj I see that we don't have a full security checklist completed for this change. Please take the time to put one together and update the PR description with a link when it's done.

In the meantime, please add a "DO NOT MERGE" to the PR title and/or close for now. I'd wager we still have some documentation to complete before this can make its way to `master`.

karlmdavis:
FYI: We're likely merging this same feature to BFS today or tomorrow: <https://github.com/CMSgov/bluebutton-data-server/pull/143>.

shea-parkes:
Comment from the peanut gallery:

I thought CMS was trying to fairly aggressively move from HICN to MBI.  The descriptions above all mention HICN (including discussing one of its weaknesses; that it included SSNs).  Here's one random source to provide some context: https://www.gormanhealthgroup.com/blog/hicn-to-mbi-what-your-organization-needs-to-know-about-the-2018-cms-change/

However, I did chase this upstream (CMSgov/bluebutton-data-server#143 plus a couple more) and I see notes that HICN and MBI were both included at one point of another.

So I'm guessing this is at least going to include MBI since the real logic appears to be upstream? (i.e. this is just modifying a query to the `CMSgov/bluebutton-data-server`)

And just to help refine the requirements: I believe all of the patient-level data received by Medicare ACOs is keyed on MBI (e.g. CCLF, patient-level assignment files).  I can't see JIRA, just GitHub, so I don't know how the upstream work item was worded.

PR #341, opened 2019-07-28
BCDA-1567: Modify BCDA aco.GetPublicKey() to call the appropriate endpoint in SSAS

embh:
### Fixes [BCDA-1567](https://jira.cms.gov/browse/BCDA-1567)
Currently, BCDA stores ACOs' public keys in its database. In the future, BCDA will retrieve these public keys from the System-to-System Authentication Service.

### Proposed changes:
Retrieve an ACO's public key from the SSAS when it is the auth provider for BCDA.

### Change Details
* Adds logic to `ACO.GetPublicKey()` to get the key using the SSAS client when environment variable `BCDA_AUTH_PROVIDER` is set to `ssas`
* Adds environment variables:
  * `SSAS_URL` - Base URL of SSAS server
  * `SSAS_USE_TLS` - Indicates whether TLS connection should be used for SSAS. Unless set to `false`, TLS is enabled. Note that the method of securing the connection to SSAS is still TBD. This was added for the purpose of disabling TLS for unit tests.

### Security Implications
This work prepares for a future change to use SSAS as the home of ACOs' public keys, which are used for encrypting the NDJSON data files exported from BCDA. Use of the SSAS client is behind a feature flag, and the client makes no connections at this time.

* This change **does not** add new software dependencies
* This change **does not** alter security controls or supporting software
* This change **does not** introduce storage of new data
* A security checklist **has not** been completed for this change
  * A security checklist has not been completed for this specific ticket. However, there is a checklist for SSAS admin connections. There will also be an SIA.

### Acceptance Validation
Tests have been added for `GetPublicKey()` in the SSAS client and for `ACO.GetPublicKey()` with `BCDA_AUTH_PROVIDER` set to `ssas`. They use `httptest.Server` to mock the expected responses from SSAS.

### Feedback Requested
Any


bcda/auth/client/ssas_test.go
```diff
@@ -14,12 +22,41 @@ func (s *SSASClientTestSuite) TestNewSSASClient() {}
 
 func (s *SSASClientTestSuite) TestCreateSystem() {}
 
-func (s *SSASClientTestSuite) TestGetPublicKey() {}
+func (s *SSASClientTestSuite) TestGetPublicKey() {
```
tbellj: this is a good method, but do we have test cases for the environment vars not being set? Like a check for the error strings that'd we expect in our logging.

PR #342, opened 2019-07-29
BCDA-1596 Feature: Added db model for suppression data

tbellj:
Fixes [BCDA-1596](https://jira.cms.gov/browse/BCDA-1596)

Problem:
We need to have our db models in place before implementing the 1-800 medicare suppression ETL.

Proposed changes:
Added db model for 1-800 medicare suppression data.

Change Details:
bcda/models/models.go - gorm db model code
db/api.sql - db schema updates

Security Implications:
Yes, this code **will** handle files containing PII/PHI.
A security checklist **has** been completed for this change

Acceptance Validation:
Yes, all tests pass

Feedback Requested:
Any and all

### DIFF
--- a/bcda/models/models.go
+++ b/bcda/models/models.go
@@ -357,7 +357,7 @@ type Suppression struct {
 	SAMHSASourceCode    string `gorm:"type:varchar(5)"`
 	SAMHSAEffectiveDt   time.Time
 	SAMHSAPrefIndicator string `gorm:"type:char(1)"`
-	ACOCMSID            string `gorm:"column:aco_cms_id"`
+	ACOCMSID            string `gorm:"column:aco_cms_id;type:char(5)"`
 	BeneficiaryLinkKey  int
 }
 
--- a/bcda/models/models.go
+++ b/bcda/models/models.go
@@ -351,10 +351,10 @@ type CCLFBeneficiary struct {
 type Suppression struct {
 	gorm.Model
 	HICN                string `gorm:"type:varchar(11);not null"`
-	SourceCode          string `json:"source_code"`
+	SourceCode          string `gorm:"type:varchar(5)"`
 	EffectiveDt         time.Time
 	PrefIndicator       string `gorm:"type:char(1)"`
-	SAMHSASourceCode    string `json:"samhsa_source_code"`
+	SAMHSASourceCode    string `gorm:"type:varchar(5)"`
 	SAMHSAEffectiveDt   time.Time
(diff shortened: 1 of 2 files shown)


bcda/models/models.go
```diff
@@ -347,6 +348,19 @@ type CCLFBeneficiary struct {
 	BlueButtonID string `gorm:"type: text"`
 }
 
+type Suppression struct {
+	gorm.Model
+	HICN                string `gorm:"type:varchar(11);not null"`
+	SourceCode          string `json:"source_code"`
```
embh: Should there be a length on the source code field? It has length 5 in the ICD (not clear if it's fixed or variable though).

bcda/models/models.go
```diff
@@ -347,6 +348,19 @@ type CCLFBeneficiary struct {
...
+	EffectiveDt         time.Time
+	PrefIndicator       string `gorm:"type:char(1)"`
+	SAMHSASourceCode    string `json:"samhsa_source_code"`
```
embh: Same question as above (length 5)

bcda/models/models.go
tbellj: good eyes! Yep I saw the same thing but the ICD also gives 2 examples of valid values...kinda contradicting their own ICD. They show 2 examples: "1800" and "1-800" one being length 4 and one being length 5, so I just went with a varchar...but maybe there's a better way. thoughts?

embh: Aha. In that case, `varchar(5)` probably works: variable length, max 5.

tbellj: nice!

bcda/models/models.go
```diff
@@ -347,6 +348,19 @@ type CCLFBeneficiary struct {
...
+type Suppression struct {
+	gorm.Model
+	HICN                string `gorm:"type:varchar(11);not null"`
+	SourceCode          string `gorm:"type:varchar(5)"`
+	EffectiveDt         time.Time
+	PrefIndicator       string `gorm:"type:char(1)"`
+	SAMHSASourceCode    string `gorm:"type:varchar(5)"`
+	SAMHSAEffectiveDt   time.Time
+	SAMHSAPrefIndicator string `gorm:"type:char(1)"`
+	ACOCMSID            string `gorm:"type:char(5)"`
```
embh: I just tried a GORM migration locally, and the column names all look correct except `ACOCMSID` (ends up as `acocms_id`). So we probably need to specify the name for that one.

tbellj: ooo nice, good check too. Will update

PR #352, opened 2019-08-05
BCDA-1597-b Feature: Moved delete function into utils package

tbellj:
Fixes [BCDA-1597-b](https://jira.cms.gov/browse/BCDA-1597-b)

Problem: This is simply a cleanup code procedure. No real problem existed but was suggested that this would be a better way.

Proposed changes:
Removed the delete directory function from the cclf package into it's own utility package. Deleting files shouldn't be related to any specific type of file.

Change Details:
bcda/bcdacli/cli.go - use of utils package
bcda/cclf/cclf.go - removed delete function
bcda/cclf/cclf_test.go - removed testing of delete function
bcda/cclf/testutils/cclfUtils.go - use of utils package
bcda/utils/etl.go - added delete function
bcda/utils/etl_test.go - added testing of delete function

Security Implications:
Yes, this change **does** deal with PII/PHI. It will be responsible for the clean up of both cclf files and 1-800 Medicare suppression files. This is existing code just relocated to another location.

This change does not add new software dependencies
This change does not alter security controls or supporting software
This change does not introduce storage of new data
A security checklist has not been completed for this change

Acceptance Validation:
Yes, all tests pass

Feedback Requested:
Any and all


bcda/utils/etl_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+type EtlTestSuite struct {
```
embh:
```suggestion
type ETLTestSuite struct {
```
https://github.com/golang/go/wiki/CodeReviewComments#initialisms

PR #363, opened 2019-08-12
BCDA-1663 add ssas to build and package rpm

mamurphy:
### Partially Fixes [BCDA-1663](https://jira.cms.gov/browse/BCDA-1663)
Add SSAS build and package to build job

### Proposed changes:
Build an RPM, place it in S3

### Change Details
This PR adds the building and packaging of the ssas rpm to the build_and_package script 
This is needed in addition to https://github.com/CMSgov/bcda-ops/compare/mamurphy/BCDA-1663

### Security Implications
This is part of a series of tickets to introduce Auth's SSAS to the API pipeline. Security checklists have been filled out.

[**NO**] This change adds new software dependencies
[**NO**] This change alters security controls or supporting software
[**NO**] This change introduces storage of new data
[**NO**] A security checklist been completed for this change

### Acceptance Validation
https://bcda-ci.adhocteam.us/job/BCDA%20-%20Build%20and%20Package/877/

### Feedback Requested
Please review


ops/build_and_package.sh
```diff
@@ -42,19 +42,34 @@ go build -ldflags "-X github.com/CMSgov/bcda-app/bcda/constants.Version=$VERSION
...
+fpm -v $VERSION -s dir -t rpm -n ssas ssas=/usr/local/bin/ssas
 
 #Sign RPMs
-WORKER_RPM="bcdaworker-*.rpm"
 echo "Importing GPG Key files"
 /usr/bin/gpg --batch --import $GPG_PUB_KEY_FILE
 /usr/bin/gpg --batch --import $GPG_SEC_KEY_FILE
 /usr/bin/rpm --import $GPG_PUB_KEY_FILE
 
+SSAS_RPM="ssas-*.rpm"
```
DeirdreHolub: I was thrown by the change in order here. That is, when building the packages, the order is (bcda, bcdaworker, ssas), but when signing the packages, the order is (ssas, bcdaworker, bcda). Is the change in order significant? Can we make the order consistent so no one asks this question again?

DeirdreHolub:
> VERSION=$1

The same version is used for bcda, bcda-worker and ssas. Is this really a release #, and therefore you want all the packages to have the same # to tie the release artifacts together? Should we (sometime in the future) support versions of independent artifacts included in a release?

ops/build_and_package.sh
mamurphy: The order was reversed before adding ssas; I suppose to avoid changing directories an extra time. I was just following what was already in place, but can update this if it makes sense to everyone.

msnook:
> > VERSION=$1
> 
> The same version is used for bcda, bcda-worker and ssas. Is this really a release #, and therefore you want all the packages to have the same # to tie the release artifacts together? Should we (sometime in the future) support versions of independent artifacts included in a release?
> @rnagle @msnook

BCDA and BCDA Worker will likely always be deployed together and thus always be on the same version.  
As for SSAS, one could make the argument that it is currently "part of" BCDA (even though we are making steps to separate, it is technically packaged and released with BCDA).  Therefore, it might make sense to version it the same as BCDA.  
In the future, when SSAS is released as a separate package entirely, then SSAS will own its versioning scheme.

rnagle:
This is looking good. For now, I think we should stick with the same `VERSION` across packages. Definitely see value in changing this in the future, though.

@mamurphy would it be worth sanity checking the rpm by copying to a `dev` ec2 instance and installing? I would guess that invoking `ssas` after install will just result in errors, but I think that's fine. Just want to know the package has a high likelihood of working properly before merging.

mamurphy:
> This is looking good. For now, I think we should stick with the same `VERSION` across packages. Definitely see value in changing this in the future, though.
> 
> @mamurphy would it be worth sanity checking the rpm by copying to a `dev` ec2 instance and installing? I would guess that invoking `ssas` after install will just result in errors, but I think that's fine. Just want to know the package has a high likelihood of working properly before merging.

See installation in this PR https://github.com/CMSgov/bcda-ops/pull/277

PR #364, opened 2019-08-13
BCDA-1490 Feature: Handling of requests during etl process

tbellj:
Fixes [BCDA-1490](https://jira.cms.gov/browse/BCDA-1490)

Problem:
As a request is being made it was possible that we could also receive new cclf files during that time window. CCLF file ingestion during api requests would be problematic as the data would be incomplete, which wouldn't be too good. This new code protects us against that.

Proposed changes:
Added an import status field to our cclf files and suppression files for querying a completed state when gathering beneficiaries.

Change Details:
bcda/cclf/cclf.go - sets import status field
bcda/cclf/cclf_test.go - tests for checking the additional import statuses
bcda/constants/constants.go - constants for import statuses
bcda/models/models.go - additional import status db fields, modified get beneficiaries query
bcda/models/models_test.go - tests for querying cclf files at different import stages/statuses
bcda/suppression/suppression.go - sets import status field
bcda/suppression/suppression_test.go - tests for checking different import statuses
db/api.sql - added new import_status fields
db/migrations/20190806-add-importstatus.sql - migration script to be ran in each environment
shared_files/suppressionfile_MissingData/T#EFT.ON.ACO.NGD1800.DPRF.D181120.T1000013 - modified to fail on import

Security Implications:
This change **will** deal with PII/PHI directly. This code will set an import status for incoming cclf and suppression files.

A security checklist has been completed for this change: 
https://confluence.cms.gov/pages/viewpage.action?spaceKey=BCDA&title=Sprint+4.4+ACO+API+-+Security+Checklist

Acceptance Validation:
Yes, all tests pass.

Feedback Requested:
Any and all


bcda/cclf/cclf.go
```diff
@@ -53,6 +54,7 @@ type cclfFileMetadata struct {
 	filePath     string
 	imported     bool
 	deliveryDate time.Time
+	fileId       uint
```
embh: Tiny rename here https://github.com/golang/go/wiki/CodeReviewComments#initialisms

bcda/suppression/suppression.go
```diff
@@ -26,6 +27,7 @@ type suppressionFileMetadata struct {
 	filePath     string
 	imported     bool
 	deliveryDate time.Time
+	fileId       uint
```
embh:
```suggestion
	fileID       uint
```

PR #366, opened 2019-08-19
BCDA-1653: Extensive synthetic 1-800-MEDICARE file creation

embh:
### Fixes [BCDA-1653](https://jira.cms.gov/browse/BCDA-1653)
We have only one synthetic 1-800-MEDICARE suppression data file.

### Proposed changes:
Create a script to generate synthetic 1-800-MEDICARE files for fixture ACO IDs and HICNs.

### Change Details
* Creates script `test/datagenerator/1800medicare.go` for creating synthetic 1-800-MEDICARE files with randomly generated data
  * Run with `go run`. ACOs that data will be created for can be changed in the file
  * Randomly selects HICNs from files of HICNs by ACO, creates a profile for each, and writes 0-5 suppression records for that profile
* Adds pared-down output from script (250 rows) `T#EFT.ON.ACO.NGD1800.DPRF.D190816.T0241390` to `shared_files/synthetic1800MedicareFiles/test/` so it is included in import tests

### Security Implications
No security implications. Mock files are created with entirely synthetic data.

### Acceptance Validation
Can be tested by running the script.

### Feedback Requested
Are there improvements you'd like to see? Taking command line arguments, filling in certain fields (some are always blank), etc.


test/datagenerator/1800medicare.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+}
+
+func profile(hicn string, acoID string) record {
+	dobMin, _ := time.Parse("2006-01-02", "1900-01-01")
+	dobMax := time.Now().Add(-65 * 365 * 24 * time.Hour)
+
+	p := record{
+		hicn:       field{l: 11, v: hicn},                     // HICN
+		blk:        field{l: 10},                              // Beneficiary link key
```
tbellj: is this blank? it's cool because that's valid just wanna make sure im reading this correctly

embh: Yes, any case where `v` isn't set will default to an empty string. If you think the bene link key would be useful, I'm happy to add it!

tbellj: nah I don't even know what it is :)

test/datagenerator/1800medicare.go
```diff
@@ -0,0 +1,161 @@
...
+	// Create 0-5 suppression records for this profile
+	ct := rand.Intn(6)
+	encDtMin, encDtMax := time.Now().Add(-7*24*time.Hour), time.Now()
+	effDtMin, effDtMax := time.Now().Add(-7*24*time.Hour), time.Now().Add(3*24*time.Hour)
+	records := []record{}
+
+	for i := 0; i < ct; i++ {
+		r := profile
+		r.encDate = field{l: 8, v: ccyymmdd(encDtMin, encDtMax)} // Encounter date
+		r.effDate = field{l: 8, v: ccyymmdd(effDtMin, effDtMax)} // Beneficiary data sharing effective date
```
tbellj: It'd be cool if we could throw a blank value for the effective date, just like the other fields since that's a valid option. It'd make our testing even more robust!

test/datagenerator/1800medicare.go
```diff
@@ -0,0 +1,161 @@
...
+		r.srcCode = field{l: 5, v: oneOfStr("1800", "")}         // Beneficiary data sharing source code
+		r.mechCode = field{l: 1, v: oneOfStr("T", "")}           // Beneficiary option data sharing decision mechanism code
+		r.prefInd = field{l: 1, v: oneOfStr("Y", "N", "")}       // Beneficiary data sharing preference indicator
+		r.saEffDate = field{l: 8}                                // Beneficiary substance abuse data sharing effective date
```
tbellj: same thought as the above comment, let's pass the date and a blank value for the r.saEffDate

test/datagenerator/1800medicare.go
```diff
@@ -0,0 +1,161 @@
...
+		panic(err)
+	}
+
+	_, err = outf.WriteString(fmt.Sprintf("HDR_BENEDATASHR%s\n", now.Format("20060102")))
+	if err != nil {
+		panic(err)
+	}
+
+	for _, acoID := range []string{"A9990", "A9991", "A9992", "A9993", "A9994"} {
+		inf, err := os.Open(fmt.Sprintf("%s-HICNs", acoID))
```
tbellj: Is this creating 0-5 suppression records for each hicn found from the synthetic data files A9990 --> A9994?

embh: 0-5 records for about half of the HICNs in those files, based on this line: https://github.com/CMSgov/bcda-app/pull/366/files#diff-31f06376cac09449837e97f31ce0172fR55

bcda/suppression/suppression_test.go
```diff
@@ -200,26 +238,35 @@ func (s *SuppressionTestSuite) TestGetSuppressionFileMetadata() {
...
 	modtimeAfter := time.Now().Truncate(time.Second)
-	for _, suppress := range suppresslist {
-		assert.Equal(modtimeBefore.Format("010203040506"), suppress.deliveryDate.Format("010203040506"))
-
-		// change the modification time for all the files
-		err := os.Chtimes(suppress.filePath, modtimeAfter, modtimeAfter)
-		if err != nil {
-			s.FailNow("Failed to change modified time for file", err)
-		}
+	assert.Len(suppresslist, 2)
```
tbellj: Oh was this broken out to know "which" file failed the test for the modification time change? Like better debugging if a failure happens??

test/datagenerator/1800medicare.go
tbellj: ahhh gotcha, I was confused on what that line was doing. Got it, slick solution for that.

bcda/suppression/suppression_test.go
embh: Yes. With the additional test suppression file, I updated this to check both. Also changed this assertion that was comparing a value to itself:
```
modtimeBefore := suppresslist[0].deliveryDate
...
assert.Equal(modtimeBefore.Format("010203040506"), suppress.deliveryDate.Format("010203040506"))
```

tbellj: oh crap! good find

PR #367, opened 2019-08-19
BCDA-1630 Feature: Updated cclf file name validation

tbellj:
Fixes [BCDA-1630](https://jira.cms.gov/browse/BCDA-1630)

Problem:
The filename format for cclf files will change and we need a way to simultaneously support both file formats. This will allow us to support both the current implementation and the new one. 

Proposed changes:
An update to our getCCLFFileMetadata method to parse data for both cclf file formats (with and without the BCD identifier)

Change Details:
bcda/cclf/cclf.go - additional method to handle new file format with the BCD identifier.
bcda/cclf/cclf_test.go - testing code to support the new function.
shared_files/cclf_BCD/T.BCD.ACO.ZC0Y18.D181120.T0001000 - new file formats
shared_files/cclf_BCD/T.BCD.ACO.ZC8Y18.D181120.T0001000 - new file formats
shared_files/cclf_BCD/T.BCD.ACO.ZC9Y18.D181120.T0001000 - new file formats

Security Implications:
This change does deal with PII/PHI directly. - This code will be responsible for ingesting cclf files (with the new BCD identifier) into our system.  
A security checklist has been completed for this change: https://confluence.cms.gov/pages/viewpage.action?spaceKey=BCDA&title=Sprint+4.4+ACO+API+-+Security+Checklist

Acceptance Validation
Yes, all tests pass.

Feedback Requested:
Any and all


bcda/cclf/cclf.go
```diff
@@ -352,6 +357,71 @@ func getCCLFFileMetadata(filePath string) (cclfFileMetadata, error) {
 	return metadata, nil
 }
 
+
+func getCCLFFileMetadataBCD(filePath string) (cclfFileMetadata, error) {
```
embh: Is it possible to combine this logic with the existing function? Maybe the regex can cover both, with variables for which position each field will be in within the regex matches. Or keep the existing check on the name containing `BCD` and use that to set the regex pattern?

bcda/cclf/cclf.go
```diff
@@ -300,6 +301,10 @@ func importCCLF(fileMetadata *cclfFileMetadata, importFunc func(uint, []byte, *g
 }
 
 func getCCLFFileMetadata(filePath string) (cclfFileMetadata, error) {
+	if strings.Contains(filePath,"BCD") {
+		return getCCLFFileMetadataBCD(filePath)
+	}
+
 	var metadata cclfFileMetadata
 	// CCLF filename convention for SSP: P.A****.ACO.ZC*Y**.Dyymmdd.Thhmmsst
```
embh: The `len(filenameMatches)` count below (line 308/313) should be fixed with this ticket. For the existing pattern, I believe it's 6 instead of 5.

bcda/cclf/cclf.go
tbellj: oh okay, are u saying like have the BCD check and then set the pattern string. And then have an if statement for cclfNumIndx..perfYearIdx etc to be used for the regex matches?

embh: Yep! Another option might be named groups in a regex pattern that matches the old and new filenames (instead of using indices). Probably lots of ways to do it :)

tbellj: interesting! I think I was leaning towards separate methods because of the actual logic change for getting the acoID. Its not a simple "filenameMatches[2]" like it used to be for the new format. And also was thinking that this is temporary since it'll just be one way in the future (whenever they decide to phase out the old way). This logic will just be the entire getCCLFFileMetadata. Also wanted to avoid a bunch of if statements lol. Let me think..

tbellj: I thought about this a little more last night (very little haha), But are we getting any technical advantage by making it all into one method? Either by using the named groups regex pattern thing or the indices? Im thinking that this is easier now and for the future when we stop supporting the old legacy way. Then it's just a copy and paste job from one method to the new one. No need to refactor anything. What do u think?

embh: That makes sense to me. 👍 Can always combine them later if it turns out we have to support both formats indefinitely.

PR #380, opened 2019-08-30
BCDA-1829: CLI modifications to support SSAS integration with both new and existing ACOs

embh:
### Fixes [BCDA-1829](https://jira.cms.gov/browse/BCDA-1829)
BCDA's CLI commands related to authentication require some changes to support new and existing ACOs with SSAS.

### Proposed Changes
* Rename `generate-client-credentials` to a more accurate name, `reset-client-credentials`
* Create new `generate-client-credentials` that registers a system
* Update `create-group` to require ACO ID (UUID or CMS ID)

### Change Details
* A hardcoded test public key is passed to `RegisterSystem` because a public key is required by SSAS. It is not used. Public keys will continue to be managed by BCDA.
* Test setup/teardown function names are corrected in /auth/client/ssas_test.go
* `create-aco` can no longer accept a group ID. The ACO must be created before its group so the ACO CMS ID can be included in the create group request's `xdata`
* Test script test/smoke_test/ssas_test.sh updated:
  * Setup includes, in order: Create ACO, create user, save public key, create group, generate credentials
  * Calls to bcda_client.go uncommented, making this a full test from ACO setup to decryption

### Security Implications

  - SSAS-BCDA integration SIA https://confluence.cms.gov/pages/viewpage.action?spaceKey=BCDA&title=SSAS+BCDA+Integration+SIA

### Acceptance Validation
Tests pass.

### Feedback Requested
Any


bcda/bcdacli/cli.go
```diff
@@ -309,6 +314,26 @@ func setUpApp() *cli.App {
...
+				if err != nil {
+					return err
+				}
+				fmt.Fprintln(app.Writer, msg)
+				return nil
+			},
+		},
+		{
+			Name:     "reset-client-credentials",
+			Category: "Authentication tools",
```
msnook: Should we clarify in comments/description/usage that the client ID is preserved, and only the client secret is updated as part of this command?

bcda/bcdacli/cli.go
```diff
@@ -158,7 +163,7 @@ func setUpApp() *cli.App {
 				},
 				cli.StringFlag{
 					Name:        "group-id",
-					Usage:       "Auth group ID (required for SSAS)",
+					Usage:       "(Optional) Auth group ID",
```
msnook: I like the flexibility in making `group-id` optional for `create-aco` and `aco-id` optional for `create-group`.  It allows us to execute the commands in any order we choose.
However, do we need any error handling for a scenario where we call `create-aco` with no `group-id` and then call `create-group` with no `aco-id`?  What would happen if an engineer ran the commands like that?

embh: The ACO and group would be created but not associated with each other. Relates to my question of if we need a command to tie existing groups and ACOs together. Do you have an idea for how we'd identify that there's an error? Should we require a certain order instead?

PR #384, opened 2019-09-03
BCDA-1707: SSAS: Fix testing issues/bugs

DeirdreHolub:
### Fixes [BCDA-1707](https://jira.cms.gov/browse/BCDA-1707)

Passing an empty string into GetSystemByID() was incorrectly finding the first entry in the table. It should have returned a not found error. This PR fixes that immediately. BCDA-1890 calls for additional follow up.

### Proposed Changes

Refactor query so that it fails when the id value is empty.
Add additional positive and negative tests.

### Security Implications

Code that inadvertently passed an empty string into this method could affect the wrong system.

### Acceptance Validation

Used correct database query function.
Added tests to validate.


ssas/systems.go
```diff
@@ -443,7 +443,8 @@ func GetSystemByID(id string) (System, error) {
 	)
 	defer Close(db)
 
-	if err = db.First(&system, id).Error; err != nil {
+	// must use the explicit where clause here because the id argument is a string
```
embh: Would it be beneficial to convert the string to `uint` first, to validate before hitting the database?

PR #392, opened 2019-09-10
BCDA-1899: Prioritize ACOs for CCLF ingestion based on environment variable

embh:
### Fixes [BCDA-1899](https://jira.cms.gov/browse/BCDA-1899)
When new ACOs are onboarded to BCDA, beneficiaries will be associated with them only after CCLF files have been ingested for that ACO. As BCDA optimizes the ingestion process, being able to prioritize certain ACOs' CCLF files for ingestion would help ensure that newly onboarded ACOs can retrieve data.

### Proposed Changes
Read a new environment variable that stores ACO identifiers and ingest those ACOs' CCLF data first.

### Change Details
* New environment variable `CCLF_PRIORITY_ACO_CMS_IDS` expects a list of ACO CMS IDs (`A####,A####`)
* After sorting files in CCLF directory, a new function, `orderACOs()`, creates a list of the ACO IDs in the map with any prioritized ACOs at the beginning
* For ingestion, files are selected from the map based on the order from `orderACOs()`
* `cclfMap` is now a map of maps: ACO ID to performance year to file metadata (previously compound ACO ID + performance year to metadata)
* New `containsString()` function in common utils that checks if an array contains a certain string

### Security Implications
No security implications. This is a way to optionally change the order of CCLF file ingestion.

### Acceptance Validation
Tests added for `orderACOs()` and for `ImportCCLFDirectory()` with `CCLF_PRIORITY_ACO_CMS_IDS` set.

### Feedback Requested
Any


bcda/cclf/cclf.go
```diff
@@ -469,17 +473,42 @@ func sortCCLFFiles(cclfmap *map[string][]*cclfFileMetadata, skipped *int) filepa
...
 		}
+
 		return nil
 	}
 }
 
+func orderACOs(cclfMap *map[string]map[int][]*cclfFileMetadata) []string {
+	var acoOrder []string
+
+	if priorityACOList := os.Getenv("CCLF_PRIORITY_ACO_CMS_IDS"); priorityACOList != "" {
```
msnook: Do we need an accompanying ops ticket to get this set in `prod`?

bcda/cclf/cclf_test.go
```diff
@@ -34,6 +34,44 @@ func TestCCLFTestSuite(t *testing.T) {
 	suite.Run(t, new(CCLFTestSuite))
 }
 
+func (s *CCLFTestSuite) TestImportCCLFDirectory_PriorityACOs() {
+	// The order they should be ingested in. 1 and 2 are prioritized; 3 is the other ACO in the directory.
+	var aco1, aco2, aco3 = "A9989", "A9988", "A0001"
```
msnook: I could be missing it, but do we have a test for a scenario where an ACO ID exists in our `CCLF_PRIORITY_ACO_CMS_IDS` env var but does not exist on the system?  I was thinking we should make sure that we don't bomb in that scenario, because it is definitely possible it could happen.

embh: This test shows that any ACO IDs in the priority list that aren't in the CCLF file metadata map will get ignored: https://github.com/CMSgov/bcda-app/pull/392/files#diff-71a7661f40e51b958b167a5dfb4fda57R489

bcda/cclf/cclf.go
embh: Indeed we do! https://jira.cms.gov/browse/BCDA-1951

PR #406, opened 2019-09-26
Update CI to remove redundant linting of SSAS code

msnook:
We are redundantly linting the SSAS code: once as part of API build and once as part of SSAS build.  The SSAS code should only be linted as part of the SSAS build.

### Proposed Changes

Exclude linting of SSAS code from both `gosec` and `golangci-lint`.

### Change Details

Updated the `Makefile` to explicitly skip the SSAS code when executing `gosec` and `golangci-lint` as part of the API lint target.
Updated the `test` and `test-ssas` targets so that `test` is API-specific and `test-ssas` is SSAS-specific.

### Security Implications

There are no security implications with this change.  This change simply tells our linting tool which directories to lint.

### Acceptance Validation

Build/Package is passing in Jenkins against this feature branch: https://bcda-ci.adhocteam.us/job/BCDA%20-%20Build%20and%20Package/1012/

### Feedback Requested

Please review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,8 +13,8 @@ package:
 	-v ${PWD}:/go/src/github.com/CMSgov/bcda-app packaging $(version)
 
 lint:
-	docker-compose -f docker-compose.test.yml run --rm tests golangci-lint run --deadline=2m
-	docker-compose -f docker-compose.test.yml run --rm tests gosec ./...
+	docker-compose -f docker-compose.test.yml run --rm tests golangci-lint run --deadline=2m --skip-dirs=./ssas
+	docker-compose -f docker-compose.test.yml run --rm tests gosec -exclude-dir=./ssas ./...
 
 lint-ssas:
 	docker-compose -f docker-compose.test.yml run --rm tests golangci-lint run ./ssas/...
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -13,8 +13,8 @@ package:
 	-v ${PWD}:/go/src/github.com/CMSgov/bcda-app packaging $(version)
 
 lint:
-	docker-compose -f docker-compose.test.yml run --rm tests golangci-lint run --deadline=2m
-	docker-compose -f docker-compose.test.yml run --rm tests gosec ./...
+	docker-compose -f docker-compose.test.yml run --rm tests golangci-lint run --deadline=2m --skip-dirs=ssas
+	docker-compose -f docker-compose.test.yml run --rm tests gosec -exclude-dir=ssas ./...
 
 lint-ssas:
```
embh: `lint-ssas` is called by `test-ssas`, but is `test-ssas` run automatically? Or only manually? Perhaps a question for @dhgreene, @alexhova, or @DeirdreHolub

msnook: Good question.  
Maybe we should just remove all SSAS executions from `make test` and drop them into `make test-ssas`.  
Then, we could explicitly call `make test-ssas` here (for Travis): https://github.com/CMSgov/bcda-app/blob/master/.travis.yml#L45
For Jenkins, is there an existing ticket for getting SSAS integration tests executed in the AWS pipeline?

PR #410, opened 2019-10-02
BCDA-1591 Feature: Fix for error files not correctly moving

tbellj:
Fixes [BCDA-1591](https://jira.cms.gov/browse/BCDA-1591)

Problem:
When a bulk data request generated error files (error.ndjson file), they weren't properly being moved from our staging directory. This change properly moves our error files into the correct location for cleanup.

Proposed Changes:
Moved error files from staging to the payload dir. 

Change Details:
bcda/models/models.go  - moved error files to the correct location
bcda/web/api.go - references the new location of the error files when polling a complete job status.
bcda/web/api_test.go - modified to test logic

Security Implications:
No this change does not deal with any PII/PHI data.

Acceptance Validation
Yes, all tests pass.

Screen Shot 2019-10-02 at 2 14 52 PM

Feedback Requested:
Any and all


msnook:
Regarding security implications... the content of the error file is not changing, right? The location of the file, however, is changing (to resolve the bug).  
Also, can you explain what the original root cause of the issue was?

tbellj:
> Regarding security implications... the content of the error file is not changing, right? The location of the file, however, is changing (to resolve the bug).
> Also, can you explain what the original root cause of the issue was?

root cause: Simply put, the error files weren't being moved from the staging directory to the payload/data directory. We were encrypting and moving the payload files while leaving the error files behind in staging. Also when polling job status (because of the incorrect location) they wouldn't populate the error array. Once I was able to move them to the correct location the job status calls for error files worked as expected. Also they were using the old standard of an ACO id in the filename which was changed after our (call chunking logic was introduced). It now uses the correct file name according to the generated UUID for that particular job.

PR #414, opened 2019-10-07
BCDA-1994: Increase Retention Threshold for Data Files on NFS

msnook:
### Fixes [BCDA-1994](https://jira.cms.gov/browse/BCDA-1994)

The most recent CCLF ETL process was the first execution of that process in which our software was required to accommodate all >500 ACOs (up until that point, we were only ingesting CCLF data for 4 ACOs).  We learned that with >500 ACOs, the ETL process takes >24 hours and the estimation is that it should take ~48 hours.  As a result, our file retention period (for CCLF files and 1-800-Medicare suppression files which enter the BCDA NFS via CMS EFT) must be increased from 24 to 72 hours.

### Change Details

- Updated the ETL code to remove files which are older than 72 hours instead of 24 hours.
- Updated the unit test code to accommodate the new time threshold.
- Modified several security documents (specific documents identified in the `Security Implications` section)

### Security Implications

This change means that CCLF files and 1-800-Medicare files (which contain sensitive information) will remain on the BCDA NFS for a longer period of time.  The files will be purged from the BCDA NFS after 72 hours (previously 24 hours).  Note that the BCDA NFS is within the data tier of the BCDA system architecture.

>Due to the security control modification, the following documents have been updated to indicate the 72-hour threshold:
(1) NFS/EFT SIA
(2) XLC Megadeck slides
(3) System Design Document 
(4) Privacy Impact Assessment

### Acceptance Validation

- Unit tests were modified to accommodate the new scenario, and those tests are passing.

### Feedback Requested

Please review.

### DIFF
--- a/bcda/cclf/cclf.go
+++ b/bcda/cclf/cclf.go
@@ -419,7 +419,7 @@ func sortCCLFArchives(cclfMap *map[string]map[int][]*cclfFileMetadata, skipped *
 			log.Errorf("Unknown file found: %s", metadata)
 			*skipped = *skipped + 1
 
-			deleteThreshold := time.Hour * time.Duration(utils.GetEnvInt("ARCHIVE_THRESHOLD_HR", 24))
+			deleteThreshold := time.Hour * time.Duration(utils.GetEnvInt("CCLF_ARCHIVE_THRESHOLD_HR", 72))
 			if metadata.deliveryDate.Add(deleteThreshold).Before(time.Now()) {
 				newpath := fmt.Sprintf("%s/%s", os.Getenv("PENDING_DELETION_DIR"), info.Name())
 				err = os.Rename(metadata.filePath, newpath)
@@ -587,7 +587,7 @@ func cleanUpCCLF(cclfMap map[string]map[int][]*cclfFileMetadata) error {
 				if !cclf.imported {
 					// check the timestamp on the failed files
 					elapsed := time.Since(cclf.deliveryDate).Hours()
-					deleteThreshold := utils.GetEnvInt("ARCHIVE_THRESHOLD_HR", 24)
+					deleteThreshold := utils.GetEnvInt("CCLF_ARCHIVE_THRESHOLD_HR", 72)
 					if int(elapsed) > deleteThreshold {
 						err := os.Rename(cclf.filePath, newpath)
 						if err != nil {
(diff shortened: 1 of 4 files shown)


bcda/cclf/cclf.go
```diff
@@ -419,7 +419,7 @@ func sortCCLFArchives(cclfMap *map[string]map[int][]*cclfFileMetadata, skipped *
 			log.Errorf("Unknown file found: %s", metadata)
 			*skipped = *skipped + 1
 
-			deleteThreshold := time.Hour * time.Duration(utils.GetEnvInt("ARCHIVE_THRESHOLD_HR", 24))
+			deleteThreshold := time.Hour * time.Duration(utils.GetEnvInt("BCDA_ETL_FILE_ARCHIVE_THRESHOLD_HR", 72))
```
dhgreene: I don't see this env var set elsewhere.  Will we be relying on the default, or setting the env var separately?

tbellj: Yeah, i assume we'll be relying on the default. I think that's standard practice at this point :)

msnook: @dhgreene Yes, standard practice has been to just rely on the default and only set/create an env var if we plan to override the default.

msnook:
> good stuff, i think there's a couple spots we want to include with this change. Did a grep and found some other references:
> bcda/web/api.go - GetJobTimeout method
> bcda/bcdacli/cli.go - archive-job-files cli method, this actually moves the files to the archive dir

I'm a little unclear on this request.  I don't think we want to increase the retention time of the data/files we have acquired from BB.  The scope is only to increase the retention time of the CCLF/Suppression files.  I could definitely be missing something though.  Interested in your thoughts.

tbellj:
> > good stuff, i think there's a couple spots we want to include with this change. Did a grep and found some other references:
> > bcda/web/api.go - GetJobTimeout method
> > bcda/bcdacli/cli.go - archive-job-files cli method, this actually moves the files to the archive dir
> 
> I'm a little unclear on this request. I don't think we want to increase the retention time of the data/files we have acquired from BB. The scope is only to increase the retention time of the CCLF/Suppression files. I could definitely be missing something though. Interested in your thoughts.

Yep totally makes sense, thanks for the clarification :)

PR #413, opened 2019-10-03
BCDA-2018: Separate SSAS from BCDA-API

PatriciaAnong:
### Fixes [BCDA-2018](https://jira.cms.gov/browse/BCDA-2018)

Separate the SSAS code from the ACO API repository. Create two new repositories to manage the code for BCDA SSAS - `bcda-ssas-ops` and `bcda-ssas-app`.

### Proposed Changes
Separate the SSAS code from the BCDA repositories to enable independent deployments of the SSAS code base.

### Change Details

Replicated SSAS code from the `bcda-ops` repo and created a new Jenkins job to package the SSAS RPM for use during deployments.

### Security Implications

- [ No ] new software dependencies

- [ No ] security controls or supporting software altered

- [ No ] new data stored or transmitted

- [ No ] security checklist is completed for this change

- [ ] requires more information or team discussion to evaluate security implications

### Acceptance Validation

Yes, I ran the build and package job on Jenkins with the modifications in this PR and the job ran successfully and deployed to Dev.
The smoke test was also successful.

https://bcda-ci.adhocteam.us/job/BCDA%20-%20Build%20and%20Package/1029/consoleFull
https://bcda-ci.adhocteam.us/job/BCDA%20-%20Test%20-%20Integration%20Smoke/998/console
### Feedback Requested

Please review

PR #420, opened 2019-11-07
BCDA-2212 Maintenance: Removing SSAS code and artifacts

dhgreene:
### Fixes [BCDA-2212](https://jira.cms.gov/browse/BCDA-2212)
Now that the SSAS service has it's [own repository](https://github.com/CMSgov/bcda-ssas-app), the SSAS code and files it depended on need to be removed.

All unit and integration tests should continue working after this PR.

### Proposed Changes
- Remove the entire `/ssas` directory
- Modify the `ssas` Docker image to get the code from GitHub rather than the local repository
- Update the method of migrating the SSAS database schema to match the latest code
- Remove references in code to SSAS, except through the `ssas` Docker image

### Security Implications

The changes in this PR are part of a shuffling of code between the `bcda-app` and `bcda-ssas-app` repositories, and should have no net change in security posture.

#### `make test`
Screen Shot 2019-11-06 at 8 48 10 PM
Screen Shot 2019-11-06 at 8 49 51 PM

#### `make smoke-test-ssas` 
(an integration test that belongs in the `bcda-app` repository)
Screen Shot 2019-11-06 at 8 53 52 PM

### Feedback Requested
- Should [`ops/build_and_package.sh`](https://github.com/CMSgov/bcda-app/compare/dhg/BCDA-2212?#diff-448684739eddd421f889bfa208d8616a) have been left alone, or altered in a different way?
- Are there other areas of the repo that can be removed/updated as part of this work?
- There are any number of ways to update [`Dockerfiles/Dockerfile.ssas`](https://github.com/CMSgov/bcda-app/compare/dhg/BCDA-2212?#diff-2fe7da7214a2780f2ef98f0b43a7dbb7); what further optimizations do you suggest?


Makefile
```diff
@@ -48,7 +44,6 @@ smoke-test:
 	BCDA_SSAS_CLIENT_ID=$(SSAS_ADMIN_CLIENT_ID) BCDA_SSAS_SECRET=$(SSAS_ADMIN_CLIENT_SECRET) CLIENT_ID=$(CLIENT_ID) CLIENT_SECRET=$(CLIENT_SECRET) docker-compose -f docker-compose.test.yml run --rm -w /go/src/github.com/CMSgov/bcda-app/test/smoke_test tests sh smoke_test.sh
 
 smoke-test-ssas:
```
msnook: Do we still need this?

Makefile
```diff
@@ -48,7 +44,6 @@ smoke-test:
 	BCDA_SSAS_CLIENT_ID=$(SSAS_ADMIN_CLIENT_ID) BCDA_SSAS_SECRET=$(SSAS_ADMIN_CLIENT_SECRET) CLIENT_ID=$(CLIENT_ID) CLIENT_SECRET=$(CLIENT_SECRET) docker-compose -f docker-compose.test.yml run --rm -w /go/src/github.com/CMSgov/bcda-app/test/smoke_test tests sh smoke_test.sh
```
msnook: Is `BCDA_SSAS_CLIENT_ID` and `BCDA_SSAS_SECRET` used in the standard smoke test?  I didn't think it was needed.

Dockerfiles/Dockerfile.ssas
```diff
@@ -11,11 +11,13 @@ RUN go get -u github.com/derekparker/delve/cmd/dlv
 RUN go get github.com/BurntSushi/toml
 RUN go get github.com/howeyc/fsnotify
 RUN go get github.com/mattn/go-colorable
+RUN go get github.com/CMSgov/bcda-ssas-app/ssas/service/main
 
-WORKDIR /go/src/github.com/CMSgov/bcda-app/ssas
-COPY . .
+COPY --from=migrate/migrate /migrate /usr/local/bin/migrate
```
msnook: Can we get away with the same technique we use in the `bcda-ssas-app` repo?  (i.e., a docker-compose file specifically for bringing `migrate/migrate` into the stack).   Then we don't have to unrealistically bake `migrate/migrate` into the SSAS service image.

Makefile
```diff
@@ -109,9 +104,8 @@ load-synthetic-suppression-data:
 	docker-compose run api sh -c 'tmp/bcda import-suppression-directory --directory=../shared_files/synthetic1800MedicareFiles'
 
 load-fixtures-ssas:
-	docker-compose up -d db
-	docker-compose run ssas sh -c 'tmp/ssas-service --migrate'
-	docker-compose run ssas sh -c 'tmp/ssas-service --add-fixture-data'
+	docker-compose run ssas migrate -database "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -path /go/src/github.com/CMSgov/bcda-ssas-app/db/migrations up
```
msnook: Can we run `migrate` the same way we do in the `bcda-ssas-app` repo?

msnook:
Just out of curiosity, what happens if we run with `BCDA_AUTH_PROVIDER=ssas`?  (That is ultimately the end goal, right?)

docker-compose.test.yml
```diff
@@ -43,24 +43,6 @@ services:
...
-      - SSAS_HASH_ITERATIONS=130000
-      - SSAS_HASH_KEY_LENGTH=64
-      - SSAS_HASH_SALT_SIZE=32
-      - SSAS_DEFAULT_SYSTEM_SCOPE=bcda-api
-      - SSAS_MFA_CHALLENGE_REQUEST_MILLISECONDS=0
-      - SSAS_MFA_TOKEN_TIMEOUT_MINUTES=60
-      - SSAS_MFA_PROVIDER=${SSAS_MFA_PROVIDER}
-      - SSAS_TOKEN_BLACKLIST_CACHE_CLEANUP_MINUTES=15
-      - SSAS_TOKEN_BLACKLIST_CACHE_TIMEOUT_MINUTES=1440
-      - SSAS_TOKEN_BLACKLIST_CACHE_REFRESH_MINUTES=5
```
msnook: What about the env vars that start with `AUTH_` and `OKTA_`?  Can any of those be removed?

msnook: Do we need a `bcda-ops` PR in addition to this PR to clean up any extraneous env vars?

Dockerfiles/Dockerfile.ssas
```diff
@@ -11,11 +11,13 @@ RUN go get -u github.com/derekparker/delve/cmd/dlv
 RUN go get github.com/BurntSushi/toml
 RUN go get github.com/howeyc/fsnotify
 RUN go get github.com/mattn/go-colorable
+RUN go get github.com/CMSgov/bcda-ssas-app/ssas/service/main
```
msnook: Interesting.  How does this work?  We're only pulling from `ssas/service/main`?  Is `go get` smart enough to pull everything needed?

Makefile
```diff
@@ -13,17 +13,13 @@ package:
 	-v ${PWD}:/go/src/github.com/CMSgov/bcda-app packaging $(version)
 
 lint:
-	docker-compose -f docker-compose.test.yml run --rm tests golangci-lint run --deadline=3m --skip-dirs=ssas
-	docker-compose -f docker-compose.test.yml run --rm tests gosec -exclude-dir=ssas ./...
+	docker-compose -f docker-compose.test.yml run --rm tests golangci-lint run --deadline=3m
+	docker-compose -f docker-compose.test.yml run --rm tests gosec ./...
 
-lint-ssas:
```
msnook: Does the `.PHONY` section need to be updated with the removal of some targets?

ops/build_and_package.sh
```diff
@@ -47,12 +47,6 @@ echo "Building bcdaworker..."
 go build
 echo "Packaging bcdaworker binary into RPM..."
 fpm -v $VERSION -s dir -t rpm -n bcdaworker bcdaworker=/usr/local/bin/bcdaworker
-cd ../ssas
```
msnook: This file will be modified as part of BCDA-2018.  This is fine in here for now, but ultimately we'll probably see some merge conflicts once BCDA-2018 is merged to `master` and you'll want to take from that as the "source of truth".

Makefile
dhgreene: Yes, it runs the ACO API->SSAS integration test we've talked about

Makefile
dhgreene: Good catch.  My initial tests seemed to show I should leave this line alone, but everything is fine now.  😄 Removed in latest commit.

Dockerfiles/Dockerfile.ssas
dhgreene: We can, but then we have to get the `.sql` migration files into that image.  Perhaps you would have better success than I did with that strategy.

What I like about this line is that it only copies one file.

Makefile
dhgreene: The main difference is that that in `bcda-ssas-app` we run `migrate` from a separate `migrate` image.  I'd be willing to give that strategy another try with some pairing time with you; here it would be more complicated because we'd be getting the migration files copied from a separate repo.

docker-compose.test.yml
dhgreene: Looking harder I did find 5 env vars beginning `OKTA_` that are removed in the latest commit.  The others, and those beginning with `AUTH_` are associated with code in `/bcda/auth/`.

dhgreene: Thankfully, no--these env vars are only set for SSAS.

Dockerfiles/Dockerfile.ssas
dhgreene: Yes, thankfully everything we need is pulled!  The `go get` command also builds the `main.go` in the specified directory, which is useful [in the Makefile](https://github.com/CMSgov/bcda-app/blob/e45bcf88e420002476d487269baa31c187aab603/Makefile#L108).  

Incidentally we might want a future ticket for renaming the `main` directory/package for SSAS to `ssas` so that `go get` builds a binary with a less generic name.

Makefile
dhgreene: It seems the SSAS-related targets never got put in the .PHONY section, so instead of being removed others need to be added.  I've done so in the latest commit, and sorted the `.PHONY` section alphabetically because it's getting a bit unwieldy.

Makefile
msnook: There is definitely some redundancy here, that could potentially be cleaned up.  Here are some thoughts/concerns:
(1) The `smoke-test` target ultimately runs `smoke_test.sh`.  But, the `smoke-test-ssas` target _also_ runs `smoke_test.sh` (after doing some initial SSAS-related setup).  Could we remove the `smoke-test-ssas` target, rename `ssas_test.sh` to `smoke_test.sh` (overwriting the current `smoke_test.sh`) since `ssas_test.sh`  is simply doing BCDA-related CLI calls with SSAS as the provider, and just keep one `smoke-test` Makefile target?
(2) Could we configure the `docker-compose` files so that SSAS is the default provider at all times, now that we've set up the stack with SSAS running?

Makefile
msnook: Right, I was hoping we'd use the `migrate/migrate` image to be consistent with how we do it in `bcda-ssas-app` repo _as well as_ how we do it in Jenkins.

Dockerfiles/Dockerfile.ssas
msnook: Do the `.sql` files exist in this image?  If so, we can just mount a volume when we spin up the container so that the dir with the `.sql` files in the container is shared with the host.

dhgreene: Latest commit uses the same technique as the `bcda-ssas-app` repo

Makefile
dhgreene: OK; figured it out in latest commit!

dhgreene:
> Just out of curiosity, what happens if we run with `BCDA_AUTH_PROVIDER=ssas`? 

A significant number of unit tests fail, and all the integration tests.

> That is ultimately the end goal, right?)

We _could_ choose that as the end goal, but based on our experience changing auth providers in the past this would be a significant effort, involving 1) changing the expectations of unit tests and 2) setting up credentials for integration tests and maybe some env vars.  I don't believe we have enough time set aside during this PI to complete this task.

msnook:
> > Just out of curiosity, what happens if we run with `BCDA_AUTH_PROVIDER=ssas`?
> 
> A significant number of unit tests fail, and all the integration tests.
> 
> > That is ultimately the end goal, right?)
> 
> We _could_ choose that as the end goal, but based on our experience changing auth providers in the past this would be a significant effort, involving 1) changing the expectations of unit tests and 2) setting up credentials for integration tests and maybe some env vars. I don't believe we have enough time set aside during this PI to complete this task.

Got it... but we _are_ running with SSAS provider in the integration tests here: https://github.com/CMSgov/bcda-app/blob/dhg/BCDA-2212/test/smoke_test/ssas_test.sh#L16

msnook:
Just took a closer look at the BCDA-2018 PR for this repo: https://github.com/CMSgov/bcda-app/pull/413 
Thinking about throwing it away in favor of the changes made as part of BCDA-2212.  Thoughts?

dhgreene:
This is ready for another look, @msnook.  The latest version includes changes we discussed offline.

test/smoke_test/smoke_test.sh
```diff
@@ -1,11 +1,34 @@
...
+        docker-compose exec db sh -c 'psql "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -c "DELETE FROM secrets WHERE system_id IN (SELECT id FROM systems WHERE group_id = '"'"'smoke-test-group'"'"')"'
+        docker-compose exec db sh -c 'psql "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -c "DELETE FROM systems WHERE group_id = '"'"'smoke-test-group'"'"'"'
+        docker-compose exec db sh -c 'psql "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -c "DELETE FROM groups WHERE group_id = '"'"'smoke-test-group'"'"'"'
+}
+trap cleanup EXIT
+
+docker-compose stop api
+SSAS_URL="http://ssas:3004" SSAS_PUBLIC_URL="http://ssas:3003" BCDA_AUTH_PROVIDER=ssas BCDA_SSAS_CLIENT_ID=$BCDA_SSAS_CLIENT_ID BCDA_SSAS_SECRET=$BCDA_SSAS_SECRET SSAS_ADMIN_CLIENT_ID=$BCDA_SSAS_CLIENT_ID SSAS_ADMIN_CLIENT_SECRET=$BCDA_SSAS_SECRET DEBUG=true docker-compose up -d api ssas
+
+echo "waiting for API to rebuild"
```
msnook: Recommendation: "waiting for API to rebuild with SSAS as auth provider"

test/smoke_test/smoke_test.sh
```diff
@@ -1,11 +1,34 @@
...
+CREDS=($(docker-compose exec api sh -c 'tmp/bcda generate-client-credentials --cms-id A9996' | tail -n2 | tr -d '\r'))
+CLIENT_ID=${CREDS[0]}
+CLIENT_SECRET=${CREDS[1]}
+
+CLIENT_ID=$CLIENT_ID CLIENT_SECRET=$CLIENT_SECRET docker-compose -f docker-compose.test.yml run --rm -w /go/src/github.com/CMSgov/bcda-app/test/smoke_test tests sh bulk_data_requests.sh
+docker-compose stop api ssas
+
+echo "waiting for API to rebuild"
+sleep 30
+docker-compose up -d api
```
msnook: Recommendation: "waiting for API to rebuild with alpha as auth provider"

test/smoke_test/smoke_test.sh
```diff
@@ -1,11 +1,34 @@
...
+SAVE_KEY_RESULT=$(docker-compose exec api sh -c 'tmp/bcda save-public-key --cms-id A9996 --key-file "../shared_files/ATO_public.pem"' | tail -n1)
+GROUP_ID=$(docker-compose exec api sh -c 'tmp/bcda create-group --id "smoke-test-group" --name "Smoke Test Group" --aco-id A9996' | tail -n1 | tr -d '\r')
+CREDS=($(docker-compose exec api sh -c 'tmp/bcda generate-client-credentials --cms-id A9996' | tail -n2 | tr -d '\r'))
+CLIENT_ID=${CREDS[0]}
+CLIENT_SECRET=${CREDS[1]}
+
+CLIENT_ID=$CLIENT_ID CLIENT_SECRET=$CLIENT_SECRET docker-compose -f docker-compose.test.yml run --rm -w /go/src/github.com/CMSgov/bcda-app/test/smoke_test tests sh bulk_data_requests.sh
+docker-compose stop api ssas
+
+echo "waiting for API to rebuild"
```
msnook: Recommendation: "waiting for API to rebuild with alpha as auth provider"

PR #422, opened 2019-11-11
BCDA-2282 Feature: Implemented Patient all endpoint

tbellj:
Fixes [BCDA-2282](https://jira.cms.gov/browse/BCDA-2282)

Problem:
In order to be more FHIR compliant, we needed to change our bulk data requests to match the query pattern for bulk Patient requests. This new pattern matches FHIR standards, see example here: /api/v1/Patient/$export?_type where _$export?_type_ will be followed by a comma delimited list of FHIR resource types.

Previously, requests were in the form: "/api/v1/ExplanationOfBenefit/$export"
With this current change it'll look like: "/api/v1/Patient/$export?_type=ExplanationOfBenefit"

Proposed Changes:
Modified all bulk data requests to match the FHIR standard for bulk Patient requests with the inclusion of query parameters to list different FHIR resource types. Also modified 429 logic to fit with this new workflow. This change encompasses both BCDA-2282 and BCDA-2283 work.

Change Details:
bcda/models/SwaggerStructs.go - added resource type parameter to query
bcda/database/connection.go - returns a queue database connection
bcda/models/models.go - added a resource type string field to the job key table
bcda/responseutils/writer.go - modified bulk requests urls to match new query pattern
bcda/web/api.go - modified bulk patient requests to match new query pattern, also changed 429 logic to work with these new changes
bcda/web/api_test.go - modified to work with the new bulk patient requests pattern
bcda/web/router.go - removed eob and coverage api routes
bcda/web/router_test.go - modified to work with the new bulk patient requests pattern, also removed unnecessary feature flags for outdated coverage and patient endpoint exposure.
bcdaworker/main.go - added a resource type to be saved with the job key
docker-compose.yml - added "sslmode=disable" to the queue database url
test/performance_test/performance.go - modified to work with the new bulk patient requests pattern
test/postman_test/BCDA-SSAS.postman_collection.json - modified to work with the new bulk patient requests pattern
test/postman_test/BCDA_Tests.postman_collection.json - modified to work with the new bulk patient requests pattern
test/postman_test/BCDA_Tests_Sequential.postman_collection.json - modified to work with the new bulk patient requests pattern
test/smoke_test/bcda_client.go - modified to work with the new bulk patient requests pattern

Security Implications:
Yes, this change does deal with PII/PHI. We are changing the way our users will receive sensitive data via the new query patterns for bulk patient requests outlined in the bulk FHIR spec. See here: http://hl7.org/fhir/uv/bulkdata/STU1/OperationDefinition-patient-export.html.

https://confluence.cms.gov/pages/viewpage.action?pageId=228920304

Acceptance Validation
Yes, all tests pass.

image
---------------------------------------------------------------------------------------------------
image

---------------------------------------------------------------------------------------------------
Swagger screenshots:
image

image

Feedback Requested:
Any and all


bcda/web/api.go
```diff
@@ -33,31 +33,6 @@ import (
...
-
-	Responses:
-		202: BulkRequestResponse
-		400: badRequestResponse
-		401: invalidCredentials
-		429: tooManyRequestsResponse
-		500: errorResponse
-*/
-
-func bulkEOBRequest(w http.ResponseWriter, r *http.Request) {
```
msnook: I think we are going to want to keep `/ExplanationOfBenefit` and `/Coverage` around for a little while, for backwards compatibility.

bcda/responseutils/writer.go
```diff
@@ -81,7 +81,7 @@ func CreateCapabilityStatement(reldate time.Time, relversion, baseurl string) *f
 					{
 						Name: "export",
 						Definition: &fhirmodels.Reference{
-							Reference: baseurl + "/api/v1/ExplanationOfBenefit/$export",
+							Reference: baseurl + "/api/v1/Patient/$export?_type=ExplanationOfBenefit",
```
msnook: Why explicitly indicating the `_type`?  The Capability is the `/Patient` endpoint.  What is the FHIR-compliant way of defining this endpoint in a capability statement.

bcda/web/api.go
```diff
@@ -204,6 +216,91 @@ func bulkRequest(t string, w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusAccepted)
 }
 
+func check429(jobs []models.Job, types []string, w http.ResponseWriter) ([]string, bool) {
```
msnook: This seems overly complex.  What if we give a `429` regardless of the type of job they have requested?
They have the freedom to request `/Patient` (ALL) and get everything with one request.

test/smoke_test/bcda_client.go
```diff
@@ -81,7 +81,7 @@ func getAccessToken() string {
 func startJob(resourceType string) *http.Response {
 	client := &http.Client{}
 
-	var url string = fmt.Sprintf("%s://%s/api/v1/%s/$export", proto, apiHost, resourceType)
+	var url string = fmt.Sprintf("%s://%s/api/v1/Patient/$export?_type=%s", proto, apiHost, resourceType)
```
msnook: What about a test for `/Patient` all?

bcda/database/connection.go
```diff
@@ -38,3 +38,18 @@ func GetGORMDbConnection() *gorm.DB {
 	}
 	return db
 }
+
+func GetQueueDbConnection() *sql.DB {
```
msnook: See comment below on `429`.  Maybe we can avoid requiring a connection to the queue db.

test/smoke_test/bcda_client.go
msnook: Yeah, we'll need a test for `/Patient` all so we can validate that the response format is correct.

bcda/web/api.go
```diff
@@ -279,9 +376,6 @@ func jobStatus(w http.ResponseWriter, r *http.Request) {
 			scheme = "https"
 		}
 
-		re := regexp.MustCompile(`/(ExplanationOfBenefit|Patient|Coverage)/\$export`)
```
msnook: Should we still have this check here, but for `/Patient` only?

msnook:
It is not clear to me where the response format has changed to accommodate a `/Patient` all request?  
It should align with this format . (see `Complete` section): http://hl7.org/fhir/uv/bulkdata/STU1/export/index.html#endpoint-1

bcda/responseutils/writer.go
tbellj: The spec doesn't really provide a "fhir compliant" way of showing this type of endpoint...at least i couldn't find it lol. But i also don't see why we wouldn't want to do this. (It's informing our users how to make the different calls we support via the new query parameters) But it's also possible I'm missing your point entirely. What is it specifically that concerns you??

test/smoke_test/bcda_client.go
tbellj: it was my understanding that we would add the new test cases (including patient all) in this future ticket here: BCDA-2284 and this pr was to retrofit all existing cases to work with the new patient query parameters. I could be wrong though, let me know if this is totally not the case.

bcda/web/api.go
tbellj: Nah we don't really need it with these changes. Its really only there to create a regex for getting the resource type from the job's request url to properly display the fhir resources in the output of the payload. That now comes from the resource type being stored in the jobKey table that gets populated during construction of the individual files themselves. So now the resource type and the filename both get stored together and retrieved together with this output.

bcda/web/api.go
tbellj: workin on that confluence page..stay tuned :)

bcda/database/connection.go
tbellj: good concerns, yeah I think putting that page together to explain the different scenarios will give us more of a chance for some good dialogue around this. I just NEED TO FINISH IT.

tbellj:
> It is not clear to me where the response format has changed to accommodate a `/Patient` all request?
> It should align with this format . (see `Complete` section): http://hl7.org/fhir/uv/bulkdata/STU1/export/index.html#endpoint-1

yup, fair enough. I should have added a screenshot for proof. Just added it.

test/smoke_test/bcda_client.go
msnook: Yes, all permutations can be addressed in BCDA-2284.  However, the standard `/Patient` request should probably be tested as part of this ticket.

msnook:
> > It is not clear to me where the response format has changed to accommodate a `/Patient` all request?
> > It should align with this format . (see `Complete` section): http://hl7.org/fhir/uv/bulkdata/STU1/export/index.html#endpoint-1
> 
> yup, fair enough. I should have added a screenshot for proof. Just added it.

Thanks, that response looks awesome.  Does it still work in scenario where there are multiple Patient, Coverage, and EOB files?  (This is probably the most common scenario).  Maybe we can test with ACO Medium or ACO Small?  Or, you could update the env vars locally so that each resource type only has like 20 beneficiaries per file (forcing multiple files locally)

test/smoke_test/bcda_client.go
tbellj: done!

tbellj:
> > > It is not clear to me where the response format has changed to accommodate a `/Patient` all request?
> > > It should align with this format . (see `Complete` section): http://hl7.org/fhir/uv/bulkdata/STU1/export/index.html#endpoint-1
> > 
> > 
> > yup, fair enough. I should have added a screenshot for proof. Just added it.
> 
> Thanks, that response looks awesome. Does it still work in scenario where there are multiple Patient, Coverage, and EOB files? (This is probably the most common scenario). Maybe we can test with ACO Medium or ACO Small? Or, you could update the env vars locally so that each resource type only has like 20 beneficiaries per file (forcing multiple files locally)

Added a screenshot showing the multiple (Patient, Coverage, EOB) files after changing the env vars to 20 benes.

test/smoke_test/bulk_data_requests.sh
```diff
@@ -8,4 +8,6 @@ go run bcda_client.go -host=api:3000 -clientID=$CLIENT_ID -clientSecret=$CLIENT_
 echo "Running Patient"
 go run bcda_client.go -host=api:3000 -clientID=$CLIENT_ID -clientSecret=$CLIENT_SECRET -endpoint=Patient
 echo "Running Coverage"
-go run bcda_client.go -host=api:3000 -clientID=$CLIENT_ID -clientSecret=$CLIENT_SECRET -endpoint=Coverage
\ No newline at end of file
+go run bcda_client.go -host=api:3000 -clientID=$CLIENT_ID -clientSecret=$CLIENT_SECRET -endpoint=Coverage
+echo "Running Patient All"
+go run bcda_client.go -host=api:3000 -clientID=$CLIENT_ID -clientSecret=$CLIENT_SECRET -endpoint=All
```
msnook: `All` is technically not a FHIR-compliant resource type.  I know this is just a test script, but it may cause confusion. 
 What if we make `-endpoint` optional?  If it is not passed in, then we assume it means we want all resource types.

test/smoke_test/bcda_client.go
```diff
@@ -80,8 +80,13 @@ func getAccessToken() string {
 
 func startJob(resourceType string) *http.Response {
 	client := &http.Client{}
+	var url string
 
-	var url string = fmt.Sprintf("%s://%s/api/v1/%s/$export", proto, apiHost, resourceType)
+	if resourceType != "All" {
```
msnook: See comment below.

test/smoke_test/bulk_data_requests.sh
msnook: Can we also add this to the Jenkinsfile for smoke tests, so that the default `/Patient` gets tested in AWS?

bcda/responseutils/writer.go
msnook: I guess the problem I see is that we should only need to include the endpoint we support (i.e., `/Patient`).  If we start adding all the `_type` options, then it opens up the door to tying us to include all permutations in the CapabilityStatement.

bcda/web/api.go
```diff
@@ -33,37 +32,12 @@ import (
...
-
-func bulkEOBRequest(w http.ResponseWriter, r *http.Request) {
-	bulkRequest("ExplanationOfBenefit", w, r)
-}
-
 /*
 	swagger:route GET /api/v1/Patient/$export bulkData bulkPatientRequest
 
-	Start patient data export
+	Start patient, explanation of benefit, and coverage data export
```
msnook: Maybe something like:
`Start data export for all supported resources types`

bcda/web/api.go
```diff
@@ -33,37 +32,12 @@ import (
...
 
-	Initiates a job to collect patient data from the Blue Button API for your ACO.
+	Initiates a job to collect patient, explanation of benefit, and coverage data from the Blue Button API for your ACO.
```
msnook: Maybe:
`Initiates a job to collect data from the Blue Button API for your ACO.  Support resource types are Patient, Coverage, and ExplanationOfBenefit.`

bcda/web/api.go
msnook: From Swagger, I am able to make a call to `/Patient`.  I think we should add the ability for the user to supply a `_type` parameter so that users can execute export jobs for individual resource types.  Does Swagger support that?

bcda/models/SwaggerStructs.go
```diff
@@ -132,6 +132,15 @@ type FileParam struct {
 	Filename string `json:"filename"`
 }
 
+// swagger:parameters bulkPatientRequest
+type ResourceTypeParam struct {
+	// Resource types requested
+	// in: query
+	// style: form
+	// explode: false
+	ResourceType []string `json:"_type"`
```
msnook: Shouldn't this be`string` not `[]string`?

PR #424, opened 2019-11-15
BCDA-2373: Improve Performance of Blue Button ID Acquisition

msnook:
### Fixes [BCDA-2373](https://jira.cms.gov/browse/BCDA-2373)

Queries to `cclf_beneficiaries` to find pre-existing Blue Button IDs are proving to be inefficient (3-4 second execution time).   Acquiring the Blue Button ID directly via lean `/Patient` call to Blue Button might speed up processing.

### Change Details

- Removed inefficient query.

### Security Implications

This is a performance optimization implemented by removing code.  No security implications are involved. 

### Acceptance Validation

Only local testing has been performed, due to DB/data size differences between `prod` and lower environments.

Additionally, the straight query was executed on `prod` and proven to take ~4 seconds in most cases (note that in the query below, the [HICN is synthetic](https://github.com/CMSgov/bcda-app/blob/master/shared_files/cclf/files/synthetic/test/small/ZC8#L2) 

```
time psql <SECRET_DB_CONN_STRING> -c "select * from cclf_beneficiaries where hicn = '1000003701' and blue_button_id='' limit 1;"

real    0m4.120s
```

### Feedback Requested

Please review.

### DIFF
--- a/bcda/models/models.go
+++ b/bcda/models/models.go
@@ -480,8 +480,6 @@ func (cclfBeneficiary *CCLFBeneficiary) GetBlueButtonID(bb client.APIClient) (bl
 	if cclfBeneficiary.BlueButtonID != "" {
 		return cclfBeneficiary.BlueButtonID, nil
 	}
-	db := database.GetGORMDbConnection()
-	defer db.Close()
 
 	// didn't find a local value, need to ask BlueButton
 	hashedHICN := client.HashHICN(cclfBeneficiary.HICN)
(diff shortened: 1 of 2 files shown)


bcda/models/models.go
```diff
@@ -480,17 +480,9 @@ func (cclfBeneficiary *CCLFBeneficiary) GetBlueButtonID(bb client.APIClient) (bl
 	if cclfBeneficiary.BlueButtonID != "" {
 		return cclfBeneficiary.BlueButtonID, nil
 	}
-	var oldRecord CCLFBeneficiary
 	db := database.GetGORMDbConnection()
```
tbellj: Not sure if we need a db connection at all with the query being removed??

msnook: Good eye!  Removed in latest commit.

PR #426, opened 2019-11-26
BCDA-2384 Feature: Updated regex for cclf zip files

tbellj:
Fixes [BCDA-2384](https://jira.cms.gov/browse/BCDA-2384)

Problem: 
Sometimes ACO-OS sends filenames in this format: P.BCD.ACO.ZC0Y19.D191106.T0001002, other times ACO-OS sends filenames in this format: P.BCD.ACOB.ZC0Y19.D191106.T0001002. Note the extra "B" appended to the ACO --> ACOB. We should be able to support both.

Proposed Changes:
Modified regex to support both filenames with "ACO" and "ACOB". 

Change Details:
bcda/cclf/cclf.go - changed to support both filename formats.
bcda/cclf/cclf_test.go - modified test code for new changes
shared_files/cclf/archives/valid/T.BCD.ACOB.ZC0Y18.D181120.T0001000 - new file format with "ACOB".

Security Implications:
Yes, this change **does** deal with PII/PHI. We've modified the regex used to validate which cclf files we import. The file contents themselves have not changed, we're importing the same data we've always imported. The filenames themselves is the only change.

https://confluence.cms.gov/pages/viewpage.action?spaceKey=BCDA&title=19-11-26%3A+Updated+regex+for+CCLF+file+names+-+Security+Checklist

Acceptance Validation:
All tests pass.

Feedback Requested:
Any and all


bcda/cclf/cclf.go
```diff
@@ -255,7 +255,7 @@ func importCCLF(fileMetadata *cclfFileMetadata, importFunc func(uint, []byte, *g
 func getCCLFArchiveMetadata(filePath string) (cclfFileMetadata, error) {
 	var metadata cclfFileMetadata
 	// CCLF filename convention for SSP with BCD identifier: P.BCD.ACO.ZC0Yyy.Dyymmdd.Thhmmsst (timestamp will include the ACO ID value)
-	filenameRegexp := regexp.MustCompile(`(T|P)\.BCD\.ACO\.ZC(0|8)Y(\d{2})\.(D\d{6})\.T(\d{4})\d{3}`)
+	filenameRegexp := regexp.MustCompile(`(T|P)\.BCD\.ACO.*\.ZC(0|8)Y(\d{2})\.(D\d{6})\.T(\d{4})\d{3}`)
```
msnook: `ACO.*` is quite wide open.
Should we update to `ACOB?` so that it matches either `ACO` or `ACOB`?

tbellj: Yeah I totally agree, I ended up reverting it back to what we had before though. We could support both just (ACO and ACOB) with something like 'ACO(|B)' but then the indexes would be different. I can do that no problem, just didn't wanna modify too much. What u think??

msnook: If you do it without a group then it will probably not affect your indexes... try `ACOB?`

tbellj: just pushed up, good stuff

PR #428, opened 2019-12-05
BCDA-1799 Feature: Added oauth endpoint to capability statement

tbellj:
Fixes [BCDA-1799](https://jira.cms.gov/browse/BCDA-1799)

Problem:
According to the fhir spec here: http://www.hl7.org/fhir/smart-app-launch/conformance/index.html#using-cs
oauth endpoints need to be included in our capability statement. Doing this change will make us more compliant with fhir standards.

Proposed Changes:
Added oauth token endpoint to our capability statement.

Change Details:
bcda/responseutils/writer.go - Added token endpoint to our capability statement.

Security Implications:
No this change **does not** deal with PII/PHI at all.

Acceptance Validation:
Yes, all tests pass and was able to see the new endpoint.

image

Feedback Requested:
Any and all. There's two items of concern for this fix.
1) The top level link provided in the extension url is broken. "http://fhir-registry.smarthealthit.org/StructureDefinition/oauth-uris" is provided in the example capability statement but doesn't look like it's being maintained.

2) The extension object itself is not being nested as shown in the oauth endpoint example here:
image

See our screenshot for the difference. Not quite sure if this is a big deal, but currently I don't see a way to add an extension object within another extension object given the strict nature of the fhir models: github.com/eug48/fhir/models. Seems like that image is not representative of their own model capabilities...

### DIFF
--- a/bcda/responseutils/writer.go
+++ b/bcda/responseutils/writer.go
@@ -57,15 +57,9 @@ func CreateCapabilityStatement(reldate time.Time, relversion, baseurl string) *f
 					Service: []fhirmodels.CodeableConcept{
 						{
 							Coding: []fhirmodels.Coding{
-								{Display: "OAuth", Code: "OAuth", System: "http://hl7.org/fhir/ValueSet/restful-security-service"},
-							},
-							Text: "OAuth",
-						},
-						{
-							Coding: []fhirmodels.Coding{
 								{Display: "SMART-on-FHIR", Code: "SMART-on-FHIR", System: "http://hl7.org/fhir/ValueSet/restful-security-service"},
 							},
-							Text: "SMART-on-FHIR",
+							Text: "OAuth2 using SMART-on-FHIR",
 						},
 					},
 				},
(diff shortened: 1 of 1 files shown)


dhgreene:
The broken URL still seems to be specified by the standard, so that's ok.

Not being able to nest the extension seems like a deal breaker to me.  The whole point is to be more compliant, and the `oauth_uris` extension expects the nesting, even in the [latest build of the spec](https://build.fhir.org/extension-oauth-uris.html).  I agree that `github.com/eug48/fhir` does not support nested extensions (just a value and a URL).  If this is important to us, I suggest making a PR for that project.  It might be as simple as a one-line change plus a unit test.

tbellj:
> The broken URL still seems to be specified by the standard, so that's ok.
> 
> Not being able to nest the extension seems like a deal breaker to me. The whole point is to be more compliant, and the `oauth_uris` extension expects the nesting, even in the [latest build of the spec](https://build.fhir.org/extension-oauth-uris.html). I agree that `github.com/eug48/fhir` does not support nested extensions (just a value and a URL). If this is important to us, I suggest making a PR for that project. It might be as simple as a one-line change plus a unit test.

ahhh good stuff David. That to me seems like a viable solution for this 👍

PR #429, opened 2019-12-10
BCDA-2405 Feature: Implemented Group endpoint

tbellj:
Fixes [BCDA-2405](https://jira.cms.gov/browse/BCDA-2405)

Problem:
In order to be more FHIR compliant, we needed to add an additional endpoint to our API. The group endpoint as described here: https://hl7.org/fhir/uv/bulkdata/export/index.html#endpoint---group-of-patients is included with these changes making us one step closer to full FHIR compliance.

The new endpoint matches this pattern:
/api/v1/Group/[id]/$export
where _$export? _type_ will be followed by a comma delimited list of FHIR resource types.

Proposed Changes:
Added a new endpoint for Group $export.

Change Details:
bcda/web/api.go - added new api endpoint for group export. (bulkGroupRequest function)
bcda/web/api_test.go - testing logic for validating request
bcda/web/router.go - added new group export route
bcda/web/router_test.go - added testing logic to support the group export route

Security Implications:
Yes, this change does deal with PII/PHI. We are changing the way our users will receive sensitive data via the new endpoint for bulk group requested outlined in the bulk FHIR spec. See here: http://hl7.org/fhir/uv/bulkdata/STU1/OperationDefinition-group-export.html

https://confluence.cms.gov/pages/viewpage.action?pageId=228920311

Acceptance Validation
Yes, all tests pass.

Feedback Requested:
Any and all.

### DIFF
--- a/bcda/models/SwaggerStructs.go
+++ b/bcda/models/SwaggerStructs.go
@@ -149,6 +149,17 @@ type BulkRequestHeaders struct {
 	Prefer string
 }
 
+// A BulkGroupRequest parameter model.
+//
+// This is used for operations that want the groupID of a group in the path
+// swagger:parameters bulkGroupRequest
+type GroupIDParam struct {
+	// ID of group export
+	// in: path
+	// required: true
+	GroupID string `json:"groupID"`
+}
+
 // JSON with a valid JWT
 // swagger:response tokenResponse
 type TokenResponse struct {
--- a/bcda/models/SwaggerStructs.go
+++ b/bcda/models/SwaggerStructs.go
@@ -132,7 +132,7 @@ type FileParam struct {
 	Filename string `json:"filename"`
 }
 
-// swagger:parameters bulkPatientRequest
+// swagger:parameters bulkPatientRequest bulkGroupRequest
 type ResourceTypeParam struct {
 	// Resource types requested
 	// in: query
@@ -141,7 +141,7 @@ type ResourceTypeParam struct {
 	ResourceType []string `json:"_type"`
 }
 
-// swagger:parameters bulkPatientRequest bulkEOBRequest bulkCoverageRequest
+// swagger:parameters bulkPatientRequest bulkGroupRequest
 type BulkRequestHeaders struct {
 	// required: true
 	// in: header
--- a/bcda/models/SwaggerStructs.go
+++ b/bcda/models/SwaggerStructs.go
@@ -157,7 +157,7 @@ type GroupIDParam struct {
 	// ID of group export
 	// in: path
 	// required: true
-	GroupID string `json:"groupID"`
+	GroupID string `json:"groupId"`
 }
 
 // JSON with a valid JWT
--- a/bcda/web/api.go
+++ b/bcda/web/api.go
@@ -65,7 +65,7 @@ func bulkPatientRequest(w http.ResponseWriter, r *http.Request) {
 }
 
 /*
-	swagger:route GET /api/v1/Group/{groupID}/$export bulkData bulkGroupRequest
+	swagger:route GET /api/v1/Group/{groupId}/$export bulkData bulkGroupRequest
 
     Start data export (for the specified group identifier) for all supported resource types
 
@@ -85,7 +85,7 @@ func bulkPatientRequest(w http.ResponseWriter, r *http.Request) {
 		500: errorResponse
 */
 func bulkGroupRequest(w http.ResponseWriter, r *http.Request) {
-	groupID := chi.URLParam(r, "groupID")
+	groupID := chi.URLParam(r, "groupId")
 	if groupID == groupAll {
 		resourceTypes, err := validateRequest(r); if err != nil {
 			responseutils.WriteError(err, w, http.StatusBadRequest)
--- a/bcda/web/api.go
+++ b/bcda/web/api.go
@@ -37,7 +37,7 @@ const (
 )
 
 /*
-	swagger:route GET /api/v1/Group/[id]/$export bulkData bulkGroupRequest
+	swagger:route GET /api/v1/Group/{id}/$export bulkData bulkGroupRequest
 
 	Start data export for all supported resource types
 
--- a/bcda/web/api.go
+++ b/bcda/web/api.go
@@ -252,7 +252,7 @@ func validateRequest (r *http.Request) ([]string, *fhirmodels.OperationOutcome) 
 			}
 		}
 	} else {
-		// patient all request, all resource types applied
+		// resource types not supplied in request; default to applying all resource types.
 		resourceTypes = append(resourceTypes, "Patient", "ExplanationOfBenefit", "Coverage")
 	}
 	return resourceTypes, nil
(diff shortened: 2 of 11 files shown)


bcda/web/api.go
```diff
@@ -32,8 +32,12 @@ import (
 
 var qc *que.Client
 
+const (
+	groupAll = "all"
+)
+
 /*
-	swagger:route GET /api/v1/Patient/$export bulkData bulkPatientRequest
+	swagger:route GET /api/v1/Group/[id]/$export bulkData bulkGroupRequest
```
msnook: I was talking to @epfrankle this morning; we should not make Swagger changes until the User Guide is updated.  This will allow us to release `/Group` without a dependency on the User Guide.
Can you add details to the User Guide ticket that updating Swagger should be part of the A/C?

msnook:
Looking good so far.  I love that we are able to reuse a lot of code here so we aren't reinventing the wheel.
We should include sufficient Postman tests, smoke tests (in the smoke test script), and smoke tests (in the Jenkins job).

bcda/web/api.go
tbellj: Yep thought about that, also the metadata endpoint should be part of that AC too. Its already added in the ticket, but I should have removed the actual comments here...good eyes.

tbellj:
> Looking good so far. I love that we are able to reuse a lot of code here so we aren't reinventing the wheel.
> We should include sufficient Postman tests, smoke tests (in the smoke test script), and smoke tests (in the Jenkins job).

Yep, take a look at the feedback requested section of this PR. We should probably have a discussion around the most efficient way to write those tests without duplicating a bunch of code. But i could be overthinking it lol

bcda/web/api.go
msnook: Should we add some details/commentary in the Swagger description for `/Group` regarding the use of the default group ID (i.e., `all`)?  Otherwise, the users won't know how to call the endpoint.
Should the group ID be passed into Swagger in a fashion similar to the job number?

test/smoke_test/bcda_client.go
```diff
@@ -40,8 +40,9 @@ func init() {
 	flag.StringVar(&clientSecret, "clientSecret", "", "client secret for retrieving an access token")
 	flag.StringVar(&apiHost, "host", "localhost:3000", "host to send requests to")
 	flag.StringVar(&proto, "proto", "http", "protocol to use")
-	flag.StringVar(&endpoint, "endpoint", "", "endpoint to test")
+	flag.StringVar(&resourceType, "resourceType", "", "resourceType to test")
 	flag.IntVar(&timeout, "timeout", 300, "amount of time to wait for file to be ready and downloaded.")
+	flag.StringVar(&endpointBase, "endpointBase", "", "base type of request endpoint")
```
msnook: Can we keep the name `endpoint` instead of `endpointBase`?

msnook:
Do we have a postman test for `/Group` with an invalid group id?

tbellj:
> Do we have a postman test for `/Group` with an invalid group id?

Good catch will add

test/smoke_test/bcda_client.go
tbellj: done

test/smoke_test/bcda_client.go
```diff
@@ -78,14 +79,18 @@ func getAccessToken() string {
 	return t.AccessToken
 }
 
-func startJob(resourceType string) *http.Response {
+func startJob(endpoint, resourceType string) *http.Response {
 	client := &http.Client{}
 	var url string
 
+	if endpoint != "Patient" {
```
msnook: I think it might be better to take `endpoint` as it is passed in.  Allows for the test client to be more flexible... i.e., add more endpoints, test error cases (passing in bad endpoints), etc.
Suggest doing something like:
```
if endpoint == "Group" {
     groupId = "all"
}
```

test/performance_test/performance.go
```diff
@@ -63,7 +64,16 @@ func main() {
 }
 
 func makeTarget(accessToken string) vegeta.Targeter {
-	url := fmt.Sprintf("%s://%s/api/v1/Patient/$export?_type=%s", proto, apiHost, endpoint)
+	if endpoint != "Patient" {
```
msnook: I think it might be better to take endpoint as it is passed in. Allows for the test client to be more flexible... i.e., add more endpoints, test error cases (passing in bad endpoints), etc.
(see similar comment below in `bcda_client.go`

bcda/web/api.go
```diff
@@ -213,6 +231,33 @@ func check429(jobs []models.Job, types []string, w http.ResponseWriter) ([]strin
...
+					resourceMap[p] = true
+					resourceTypes = append(resourceTypes, p)
+				} else {
+					oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, "Repeated resource type", responseutils.RequestErr)
+					return nil, oo
+				}
+			}
+		}
+	} else {
+		// patient all request, all resource types applied
```
msnook: this might not be a `/Patient` request.  the new `/Group` comes through this path too, right?
maybe this comment could be updated to say something like:
```
No resource types supplied in request; default to applying all resource types.
```

test/smoke_test/bcda_client.go
tbellj: yeah lol this is better. will do 👍

bcda/web/api.go
tbellj: Matt with the eyes!!

bcda/models/SwaggerStructs.go
```diff
@@ -141,14 +141,25 @@ type ResourceTypeParam struct {
...
 
+// A BulkGroupRequest parameter model.
+//
+// This is used for operations that want the groupID of a group in the path
+// swagger:parameters bulkGroupRequest
+type GroupIDParam struct {
+	// ID of group export
+	// in: path
+	// required: true
+	GroupID string `json:"groupID"`
```
msnook: Can this be renamed to `groupId`?

bcda/web/api.go
```diff
@@ -53,32 +57,46 @@ var qc *que.Client
...
-			}
+	resourceTypes, err := validateRequest(r); if err != nil {
+		responseutils.WriteError(err, w, http.StatusBadRequest)
+		return
+	}
+	bulkRequest(resourceTypes, w, r)
+}
+
+/*
+	swagger:route GET /api/v1/Group/{groupID}/$export bulkData bulkGroupRequest
```
msnook: `groupId`

bcda/web/api.go
```diff
@@ -53,32 +57,46 @@ var qc *que.Client
...
+
+	Responses:
+		202: BulkRequestResponse
+		400: badRequestResponse
+		401: invalidCredentials
+		429: tooManyRequestsResponse
+		500: errorResponse
+*/
+func bulkGroupRequest(w http.ResponseWriter, r *http.Request) {
+	groupID := chi.URLParam(r, "groupID")
```
msnook: `groupId`

PR #430, opened 2019-12-11
BCDA-2414 Add indexes to cclf_beneficiaries

dhgreene:
### Fixes [BCDA-2414](https://jira.cms.gov/browse/BCDA-2414)
The `aco.GetBeneficiaries()` method is slower than we'd like.  In particular, https://github.com/CMSgov/bcda-app/blob/c5c04b84b08fd1a1416dd665e167016b412101e9/bcda/models/models.go#L265 has been taking about 11 seconds to complete.

After adding indexes on two columns, it is sped up 100-fold.

### Proposed Changes
- Add two indexes to the `cclf_beneficiaries` table
    - `hicn`
    - `file_id` 

### Security Implications

The addition of these indexes has a small impact on availability.  The intended (and tested) impact is a significant increase in speed for some queries.  The possible unintended impact would be a slowdown for writes to the table, but 1) the chance is low and 2) this would affect automated processes only, not user queries.

### Migration Strategy
This is already active in `prod` and `dev`.  GORM will update the `test` during the next deployment, and we will run a Jenkins job to update `opensbx`.

### Acceptance Validation
Sample query execution time
- Before indexes: ~11s
- After indexes: ~.1s

### Feedback Requested
- Are other queries affected?
- Is there a different index type anyone would recommend?

### DIFF
--- a/db/migrations/20191211-index-cclf-imports.sql
+++ b/db/migrations/20191211-index-cclf-imports.sql
@@ -1,0 +1,2 @@
+CREATE INDEX idx_cclf_beneficiaries_file_id ON public.cclf_beneficiaries USING btree (file_id);
+CREATE INDEX idx_cclf_beneficiaries_hicn ON public.cclf_beneficiaries USING btree (hicn);
--- a/db/migrations/20191211-index-cclf-imports.sql
+++ b/db/migrations/20191211-index-cclf-imports.sql
@@ -1,2 +1,3 @@
 CREATE INDEX idx_cclf_beneficiaries_file_id ON public.cclf_beneficiaries USING btree (file_id);
 CREATE INDEX idx_cclf_beneficiaries_hicn ON public.cclf_beneficiaries USING btree (hicn);
+CREATE INDEX idx_cclf_beneficiaries_mbi ON public.cclf_beneficiaries USING btree (mbi);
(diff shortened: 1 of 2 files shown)


dhgreene:
> Can we add the indexes to the model here: https://github.com/CMSgov/bcda-app/blob/master/bcda/models/models.go#L436-L443

👍 

> Additionally, should we create an index on the `MBI` field as well (to future-proof this solution)?

OK, that seems logical.  Both changes present in the latest commit.

db/migrations/20191211-index-cclf-imports.sql
```diff
@@ -0,0 +1,3 @@
+CREATE INDEX idx_cclf_beneficiaries_file_id ON public.cclf_beneficiaries USING btree (file_id);
```
msnook: Can this file be removed now that the indexes are incorporated in the `gorm` migration?

dhgreene: Don't we need it in order to migrate the other AWS environments via Jenkins?

msnook: I was hoping the `gorm` auto-migrate capability would take care of it for us, for free.  Adding the index to the model should achieve this for us.
Maybe you can deploy this branch to `dev` to see if the auto-migrate functionality adds the indexes automatically.

dhgreene: You called it--I was not expecting changes of any kind for an existing column, not even an index.  💯 Script deleted.

PR #432, opened 2019-12-19
BCDA-2478 Feature: Configure BB request around mbiHASH or hicnHASH

tbellj:
Fixes [BCDA-2478](https://jira.cms.gov/browse/BCDA-2478)

Problem:
In order to support the transition from hicn to mbi we've adjusted the hashing to be configurable based on a new env var: PATIENT_IDENTIFIER_MODE. It will now use this new configuration to determine the correct mode for requesting a BB ID.

Proposed Changes:
Use of an environment variable to determine which identifier to use for hashing and making BB ID request.

Change Details:
bcda/client/bluebutton.go - modified methods to be more generic
bcda/client/bluebutton_test.go - modified to work with new method signature
bcda/models/models.go - modified to use the correct identifier (hicn or mbi) for hashing and making BB ID request
bcda/models/models_test.go - modified to work with new logic
bcda/testUtils/client.go - modified to work with new method signature
bcdaworker/main_test.go - modified to work with new method signature

Security Implications:

Yes, this change **does** deal with PII/PHI. It now determines which identifier to use (hicn/mbi), for hashing and making BB ID request.

Acceptance Validation

Yes, all current tests pass. Tests at the moment are real minimum, mostly just retrofitted to use the new logic. More tests to come..

Feedback Requested:
This is just an initial first cut, so lets get all the eyes we can.


bcda/client/bluebutton.go
```diff
@@ -103,10 +103,17 @@ func (bbc *BlueButtonClient) GetPatient(patientID, jobID, cmsID string) (string,
...
 }
 
-func (bbc *BlueButtonClient) GetPatientByHICNHash(hashedHICN string) (string, error) {
+func (bbc *BlueButtonClient) GetPatientByIdentifierHash(hashedIdentifier string) (string, error) {
 	params := GetDefaultParams()
+
+	identifier := "hicnHash"
+	// will probably get rid of this by passing in an extra param
+	if utils.FromEnv("PATIENT_IDENTIFIER_MODE","HICN_MODE") == "MBI_MODE" {
+		identifier = "hicnMbi"
```
msnook: Identifier should be `mbiHash` when in MBI mode.  See here: https://github.com/CMSgov/beneficiary-fhir-data/pull/172/files

bcda/client/bluebutton.go
```diff
@@ -238,7 +245,7 @@ func GetDefaultParams() (params url.Values) {
 	return params
 }
 
-func HashHICN(toHash string) (hashedValue string) {
+func HashIdentifier(toHash string) (hashedValue string) {
 	blueButtonPepper := os.Getenv("BB_HASH_PEPPER")
```
msnook: Need to confirm if these values will be different between `hicn-hash` and `mbi-hash`.

bcda/client/bluebutton_test.go
```diff
@@ -236,16 +236,16 @@ func (s *BBRequestTestSuite) TestGetMetadata_500() {
 }
 
 // Sample values from https://confluence.cms.gov/pages/viewpage.action?spaceKey=BB&title=Getting+Started+with+Blue+Button+2.0%27s+Backend#space-menu-link-content
-func (s *BBTestSuite) TestHashHICN() {
+func (s *BBTestSuite) TestHashIdentifier() {
```
msnook: We should have a unit test for `mbi-hash` _and_ `hicn-hash` since we'll be using both mechanisms in our code.

bcda/models/models.go
```diff
@@ -482,9 +482,19 @@ func (cclfBeneficiary *CCLFBeneficiary) GetBlueButtonID(bb client.APIClient) (bl
 		return cclfBeneficiary.BlueButtonID, nil
 	}
 
+	patientIdMode := "HICN"
+	identifier :=  cclfBeneficiary.HICN
+	systemMode := "hicn-hash"
+	if utils.FromEnv("PATIENT_IDENTIFIER_MODE","HICN_MODE") == "MBI_MODE"  {
+		patientIdMode = "MBI"
+		identifier = cclfBeneficiary.MBI
+		systemMode = "hicn-mbi"
```
msnook: This should be `mbi-hash`.  See https://github.com/CMSgov/beneficiary-fhir-data/pull/172/files

bcda/client/bluebutton.go
tbellj: received 01/06/2020:
"Received confirmation from BFD that the hashing algorithm will be the same for HICN and MBI. No need to change iterations, or pepper, etc"

bcda/client/bluebutton_test.go
tbellj: agree but i don't know about adding a unit test for this particular method for that since the HashIdentifier function itself only takes in a string. It doesn't care if its a hicn or mbi, it's just a string value. I could be misunderstanding

msnook: Cool, I think I was just missing where we were unit testing lookup by `mbi-hash`.  If that doesn't exist, we definitely need to add one.

bcda/client/bluebutton.go
tbellj: good eye!

bcda/models/models.go
tbellj: good eye again!

bcda/client/bluebutton_test.go
tbellj: yep, we can do it on the TestGetBlueButtonID method but more so on the test client itself TestGetPatientByIdentifierHash function but we'll need the synthetic data load from bfd for the latter. U were right, this is gonna have to sit until we're ready for that data.

msnook: @tbellj All of these unit tests appear to use the mock BlueButtonClient in `testUtils`.  No dependency on BFD is necessary in the unit tests.  I think we should pretty much duplicate all of these tests, but call them with an MBI so we are sure that we have that coverage.  Maybe an approach to take is to set the `PATIENT_IDENTIFIER_MODE` before each test to identify which mode we are testing in.

msnook: Also, all the synthetic MBI data is now loaded into our local env if you sync this branch with `master`

msnook:
I think there are quite a few conflicts due to new changes introduced in the past 2-3 weeks.  Can you sync your branch with `master`?

msnook:
@tbellj 
Was this comment addressed?  https://github.com/CMSgov/bcda-app/pull/432/files#r370441006
I'm worried about the unit test coverage for both modes.  Need to support both modes in the unit testing (now that we have test data for both) by setting the mode within the test.

bcda/client/bluebutton.go
```diff
@@ -103,10 +103,16 @@ func (bbc *BlueButtonClient) GetPatient(patientID, jobID, cmsID string) (string,
 	return bbc.getData(blueButtonBasePath+"/Patient/", params, jobID, cmsID)
 }
 
-func (bbc *BlueButtonClient) GetPatientByHICNHash(hashedHICN string) (string, error) {
+func (bbc *BlueButtonClient) GetPatientByIdentifierHash(hashedIdentifier, patientIdMode string) (string, error) {
 	params := GetDefaultParams()
+
+	identifier := "hicnHash"
```
msnook: This should change to `hicn-hash` according to: https://github.com/CMSgov/beneficiary-fhir-data/blob/0877232340983fda8dbcd202012e991b8b531ac5/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/TransformerConstants.java#L199

bcda/client/bluebutton.go
```diff
@@ -103,10 +103,16 @@ func (bbc *BlueButtonClient) GetPatient(patientID, jobID, cmsID string) (string,
...
+	if patientIdMode == "MBI_MODE" {
+		identifier = "mbiHash"
```
msnook: This should change to `mbi-hash` according to: https://github.com/CMSgov/beneficiary-fhir-data/blob/0877232340983fda8dbcd202012e991b8b531ac5/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/TransformerConstants.java#L213

bcda/client/bluebutton.go
```diff
@@ -103,10 +103,16 @@ func (bbc *BlueButtonClient) GetPatient(patientID, jobID, cmsID string) (string,
...
+	}
+
 	// FHIR spec requires a FULLY qualified namespace so this is in fact the argument, not a URL
-	params.Set("identifier", fmt.Sprintf("http://bluebutton.cms.hhs.gov/identifier#hicnHash|%v", hashedHICN))
+	params.Set("identifier", fmt.Sprintf("http://bluebutton.cms.hhs.gov/identifier#%s|%v", identifier, hashedIdentifier))
```
msnook: This should change to `https://bluebutton.cms.gov/resources/identifier/%s|%v` according to: https://github.com/CMSgov/beneficiary-fhir-data/blob/0877232340983fda8dbcd202012e991b8b531ac5/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/TransformerConstants.java#L78

bcda/models/models.go
tbellj: oh it looks like u actually said it here, oops my bad

bcda/client/bluebutton.go
tbellj: so should this be hicnHash or hicn-hash. The link above says hicn-hash but in slack(https://adhoc.slack.com/archives/CK6Q7BBPC/p1580850698011000)  it says hicnHash ??

bcda/client/bluebutton.go
tbellj: also does this same stuff apply to the system mode we have in the models.go here https://github.com/CMSgov/bcda-app/blob/03491102295a8c6b250028a7b94e057268e24d66/bcda/models/models.go#L502

bcda/client/bluebutton.go
tbellj: done will be in next commit

bcda/client/bluebutton.go
tbellj: > @tbellj
> Was this comment addressed? https://github.com/CMSgov/bcda-app/pull/432/files#r370441006
> I'm worried about the unit test coverage for both modes. Need to support both modes in the unit testing (now that we have test data for both) by setting the mode within the test.

nope, u right never addressed mostly because not sure what is needed honestly, we can talk tomorrow tho

bcda/client/bluebutton.go
msnook: Yeah, it should be `hicn-hash`.  It will work with `hicnHash` too, but if you look at the `TransformerConstants` file you'll see that things get complicated for us if we choose to continue to use `hicnHash`.
https://github.com/CMSgov/beneficiary-fhir-data/blob/0877232340983fda8dbcd202012e991b8b531ac5/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/TransformerConstants.java

bcda/client/bluebutton.go
msnook: > bcda-app/bcda/models/models.go

Yup, need to use `hicn-hash` and `mbi-hash` according to the BFD code here: https://github.com/CMSgov/beneficiary-fhir-data/blob/0877232340983fda8dbcd202012e991b8b531ac5/apps/bfd-server/bfd-server-war/src/main/java/gov/cms/bfd/server/war/stu3/providers/TransformerConstants.java

PR #433, opened 2019-12-23
BCDA-2491 Feature: Filter CCLF files to valid date range

dhgreene:
### Fixes [BCDA-2491](https://jira.cms.gov/browse/BCDA-2491)
If for some reason an old CCLF file is presented, or a post-dated CCLF file, we should not attempt to ingest the data. 

### Proposed Changes
- Ignore CCLF files in the future
- Ignore CCLF files more the 45 days in the past

### Change Details
- For testing purposes, allow for an optional environmental variable to set the file comparison date, which defaults to `time.Now()`

### Security Implications

If anything, this change improves security by reducing the range of unexpected and probably incorrect data that will be used by BCDA.

### Acceptance Validation
- Existing and new tests passing

### Feedback Requested
- Is there any downside to using the environmental variable for selecting a file comparison date?

### DIFF
--- a/bcda/cclf/cclf.go
+++ b/bcda/cclf/cclf.go
@@ -296,8 +296,9 @@ func getCCLFArchiveMetadata(filePath string) (cclfFileMetadata, error) {
 	}
 
 	// Files must be no older than 45 days
-	filesBefore := refDate.Add(time.Duration(int64(time.Hour) * int64(24)))
-	if t.Before(filesBefore) || t.After(time.Now()) {
+	filesNotBefore := refDate.Add(-1 * time.Duration(int64(time.Hour) * int64(24 * 45)))
+	filesNotAfter := refDate
+	if t.Before(filesNotBefore) || t.After(filesNotAfter) {
 		fmt.Printf("Date '%s' from file %s is out of range; comparison date %s\n", date, filePath, refDate.Format("060102"))
 		err = errors.New(fmt.Sprintf("date '%s' from file %s out of range; comparison date %s", date, filePath, refDate.Format("060102")))
 		log.Error(err)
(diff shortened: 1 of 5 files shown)


bcda/cclf/cclf.go
```diff
@@ -253,6 +253,12 @@ func importCCLF(fileMetadata *cclfFileMetadata, importFunc func(uint, []byte, *g
 }
 
 func getCCLFArchiveMetadata(filePath string) (cclfFileMetadata, error) {
+	refDateString := os.Getenv("CCLF_REF_DATE")
```
msnook: Just a though...
From a configuration simplification perspective, I feel like `CCLF_REF_DATE` should remain static at `time.Now()` and the lookback value (45, in this case) should be the config value.  In other words, we _always_ reference the date comparison against current time, but the lookback value could potentially change from the default value of 45 if we need to.

dhgreene: Wouldn’t that be more complicated?  To use our test files in the repo, the lookback value would need to increase over time.

msnook: Yup, good point.  I didn't think about the test files.

bcda/cclf/cclf.go
```diff
@@ -289,6 +295,16 @@ func getCCLFArchiveMetadata(filePath string) (cclfFileMetadata, error) {
 		return metadata, err
 	}
 
+	// Files must be no older than 45 days
+	filesNotBefore := refDate.Add(-1 * time.Duration(int64(time.Hour) * int64(24 * 45)))
```
msnook: Should the lookback value be an env var that defaults to 45?

PR #439, opened 2020-01-08
BCDA-2506 Feature: Implement Blue Button ID support for suppression

tbellj:
Fixes [BCDA-2506](https://jira.cms.gov/browse/BCDA-2506)

Problem:
We need a way to continually support suppression of beneficiary data once we switch over from HICN --> MBI. The cclf files themselves will no longer contain a hicn value and the data response we get from BFD will only contain MBI's. Our source for suppression data (NGD) has yet to nail down a date to support this hicn to mbi transition but suppression of data is 100% still necessary to prevent a potential data leak. This ticket provides a workable solution until NGD has fully transitioned to the MBI crossover. 

Proposed Changes:
Instead of relying on HICN values to support the suppression of beneficiary data, we now make a request using the hashed HICN values to BlueButton in order to retrieve BB Ids. These ids are now being stored in our suppression table and used against the list of cclf beneficiary ids for the correct attribution of data.

Change Details:
bcda/bcdacli/cli.go - added a cli to call the blue button request logic.
bcda/models/models.go - changed from a collection of hicns used for suppression --> to a collection of bluebutton ids. added a blue button id column to the suppression data model. Added a GetBlueButtonID function on the suppression object.
bcda/models/models_test.go - supporting logic for the above change.
bcda/suppression/suppression.go - added method that saves the BB id to each individual suppression record.
bcdaworker/main.go - Added extra suppression check before requesting payload data from BB.
db/api.sql - added blue button id field to the suppression data model, more so to keep in line with the gorm changes.

Security Implications:
Yes this change **does** deal with PII/PHI data. It uses hashed hicn values to make request for BB ids. 

- [x] new data stored or transmitted
  We are now storing BB ids in our suppression table. That data is **not** considered PII/PHI.

Acceptance Validation:
Tests pass, but there are a lot more tests to come.

Failure to find a Blue Button ID:
image

Completed output:
image

Feedback Requested:
Any and all


bcda/bcdacli/cli.go
```diff
@@ -471,6 +471,17 @@ func setUpApp() *cli.App {
 				return err
 			},
 		},
+		{
+			Name:     "retrieve-suppression-hicn-to-bbid",
```
msnook: This will need to be called after 1-800-Medicare-Import.
Instead of a new Jenkins job, I recommend running the command immediately following the ETL: https://github.com/CMSgov/bcda-ops/blob/master/jenkins_files/Jenkinsfile.import_1800medicare#L57

It will also need to be called as part of our smoke tests:
https://github.com/CMSgov/bcda-ops/blob/master/jenkins_files/Jenkinsfile.smoke_test#L273

db/api.sql
```diff
@@ -84,6 +84,7 @@ create table suppression_files (
 create table suppressions (
     id serial primary key,
     file_id integer not null,
+    blue_button_id text,
```
msnook: I updated BCDA-2562 (the ticket for `r43` release) to indicate that a DB migration is required after deployment.

bcda/bcdacli/cli.go
```diff
@@ -471,6 +471,17 @@ func setUpApp() *cli.App {
...
+			Category: "Data import",
```
msnook: I updated BCDA-2562 (the ticket for `r43` deployment) to indicate that we'll need to trigger 1-800-Medicare ingestion as soon we deploy so that the BB ID is populated.

bcda/models/models.go
```diff
@@ -248,22 +250,21 @@ func (aco *ACO) GetBeneficiaries(includeSuppressed bool) ([]CCLFBeneficiary, err
...
-				WHERE effective_date <= NOW() AND preference_indicator != '' 
-				GROUP BY hicn
+				SELECT blue_button_id, MAX(effective_date) max_date
+				FROM suppressions
+				WHERE effective_date <= NOW() AND preference_indicator != ''
+				GROUP BY blue_button_id
 			) h
-			JOIN suppressions s ON s.hicn = h.hicn and s.effective_date = h.max_date
-			WHERE preference_indicator = 'N'`).Pluck("hicn", &suppressedHICNs)
+			JOIN suppressions s ON s.blue_button_id = h.blue_button_id and s.effective_date = h.max_date
```
msnook: We'll need to add an index on `blue_button_id` in the `cclf_beneficiaries` table.

bcda/models/models.go
```diff
@@ -530,6 +536,55 @@ func (cclfBeneficiary *CCLFBeneficiary) GetBlueButtonID(bb client.APIClient) (bl
 	return blueButtonID, nil
 }
 
+// This method will ensure that a valid BlueButton ID is returned.
+// If you use suppressionBeneficiary.BlueButtonID you will not be guaranteed a valid value
+func (suppressionBeneficiary *Suppression) GetBlueButtonID(bb client.APIClient) (blueButtonID string, err error) {
```
msnook: Should refactor this so that both flavors of `GetBlueButtonID()` use the same code... i.e., the one called on the `cclfbeneficiary` model and the one called on the `suppressionbeneficiary` model.

bcdaworker/main.go
```diff
@@ -204,9 +204,24 @@ func writeBBDataToFile(bb client.APIClient, acoID string, acoCMSID string, cclfB
...
+		for _, suppressedBBID := range suppressedList {
+			if blueButtonID == suppressedBBID {
+				foundSuppressed = true
+				break
+			}
+		}
+
+		// skip over this cclf beneficiary if their blue button id is suppressed
+		if foundSuppressed {
+			continue
```
msnook: Can we log this for auditing purposes?

bcdaworker/main.go
```diff
@@ -204,9 +204,24 @@ func writeBBDataToFile(bb client.APIClient, acoID string, acoCMSID string, cclfB
 	totalBeneIDs := float64(len(cclfBeneficiaryIDs))
 	failThreshold := getFailureThreshold()
 	failed := false
+	suppressedList := models.GetSuppressedBlueButtonIDs()
```
msnook: I don't think this list is going to be populated with what you want, since the return from this function is a list that was populated when the request was made (hence, no BB IDs in `cclf_beneficiaries` table yet).

bcda/models/models.go
tbellj: good catch!

bcda/models/models.go
tbellj: agreed! Like a common base function called by 2 different methods. Nice

bcdaworker/main.go
```diff
@@ -204,9 +204,24 @@ func writeBBDataToFile(bb client.APIClient, acoID string, acoCMSID string, cclfB
...
 
 	for _, cclfBeneficiaryID := range cclfBeneficiaryIDs {
 		blueButtonID, err := beneBBID(cclfBeneficiaryID, bb)
+		foundSuppressed := false
+
+		for _, suppressedBBID := range suppressedList {
```
msnook: Instead of iterating, maybe just a query is all we need; each query should be fast, due to the indexes.  Something like below (it might need a little work, but this is the general idea):

```
SELECT blue_button_id
FROM suppressions
WHERE blue_button_id=? AND effective_date <= NOW() AND preference_indicator != '';
```

The `?` above would be the blueButtonID returned from `beneBBID`.

If we have a match, then we skip this bene (since the bene is suppressed).  Otherwise, we continue.

bcdaworker/main.go
msnook: Need to fix race condition (as we discussed on slack).

bcda/suppression/suppression.go
```diff
@@ -420,3 +421,40 @@ func updateImportStatus(m *suppressionFileMetadata, status string) {
...
+	}()
+
+	bb, err := client.NewBlueButtonClient()
+	if err != nil {
+		err = errors.Wrap(err, "could not create Blue Button client")
+		log.Error(err)
+		return 0, 0, err
+	}
+
+	// add lots of logging
```
embh: Can probably remove this comment.

bcdaworker/main.go
```diff
@@ -204,9 +204,24 @@ func writeBBDataToFile(bb client.APIClient, acoID string, acoCMSID string, cclfB
...
+				foundSuppressed = true
+				break
+			}
+		}
+
+		// skip over this cclf beneficiary if their blue button id is suppressed
+		if foundSuppressed {
+			continue
+		}
+
```
embh: If we're checking for suppression at this stage, does it still need to happen when the job is created? https://github.com/CMSgov/bcda-app/pull/439/files#diff-efd18a7d1940026f39dee80f50bb3bebR253 I'm wondering if the performance hit of redundant behavior is worth trying to keep the files equal(ish) in terms of beneficiary count.

db/api.sql
embh: Maybe `varchar(15)`? https://bluebutton.cms.gov/resources/variables/bene_id/

bcdaworker/main.go
msnook: @embh 
Great comment.  We've been wrestling with this one... lots of discussions on this topic.  
One reason for the redundant check is that if this is an ACO's first pull after a new CCLF import, we won't have the BBIDs yet, so that first check won't actually remove suppressed benes (which is why we need the second check done by the workers).  If this is not the first pull after a new CCLF import, then we will have the BB ID, so the second check is indeed redundant.
We decided to keep the redundant because in March/April timeframe (we hope), we expect the suppression files to be compliant (i.e., use MBI instead of HICN).  When that occurs, we'll rip out the check done by the workers and revert back to the original code (but based on MBI not HICN).  
I agree there is a performance hit - not sure which is the more sound approach.  Definitely open to a lot more discussion.  Appreciate your thoughts.

embh: I forgot about the CCLF beneficiary records not having the BB IDs initially. That makes sense. Could a 1-800 file be ingested while export jobs are running? It might be preferable to leave the suppression logic later in the process so the most up-to-date sharing preferences are applied.

msnook: > I forgot about the CCLF beneficiary records not having the BB IDs initially. That makes sense. Could a 1-800 file be ingested while export jobs are running? It might be preferable to leave the suppression logic later in the process so the most up-to-date sharing preferences are applied.

Yup, it is possible for the suppression data to be ingested while export jobs are running.  However, there is an XX hour period of time after a sharing preference is updated (I believe it is 72 hours), which we'd be well within if we considered the previous day's suppression data in the event new data came in while a job was running.

bcdaworker/main.go
embh: What if we had `models.GetSuppressedBlueButtonIDs()` return a `map` with the BB IDs as keys? That should be faster than iterating. https://www.darkcoding.net/software/go-slice-search-vs-map-lookup/

db/api.sql
tbellj: We could do this but Im unsure of any unintended consequences of changing this field type..(thinking db migration stuff). The same thing also applies to our cclf_beneficiaries table too, (where it was copied from). And also unsure if this change should happen in the gorm model itself?? 2 sources of truth..ahhhh

msnook: @embh Thoughts?

PR #447, opened 2020-01-23
BCDA-2509 Feature: Implemented strategy to support CCLF zip changes

tbellj:
Fixes [BCDA-2509](https://jira.cms.gov/browse/BCDA-2509)

Problem:
ACO-OS is planning a new file format for all cclf files delivered to us. The new format delivery will be a top level zipped folder with all cclf files located inside. No longer will there be individual zip folders for each cclf file type.

Proposed Changes:
Changed our ingestion logic to support the new cclf file format changes from ACO-OS.

Change Details:
bcda/cclf/cclf.go - changed our ingestion process for new cclf file format changes. 
testfiles to come...(this will be updated)

Security Implications:
Yes, this change **does** deal with PII/PHI. It handles the ingestion of CCLF file data to be persisted in our database. 

https://confluence.cms.gov/display/BCDA/2020-01-23%3A+CCLF+zip+file+changes+Security+Checklist

Acceptance Validation:
Yes, was able to fully test the original cli command for importing cclf files with the new format changes. Used test files delivered from ACO-OS.

Screenshots:
image

image

Feedback Requested:
Any and all


msnook:
As part of this change with this new format, HICN can be null now, right?  Does this need to change?
https://github.com/CMSgov/bcda-app/blob/master/bcda/models/models.go#L417

bcda/cclf/cclf.go
```diff
@@ -417,53 +405,74 @@ func sortCCLFArchives(cclfMap *map[string]map[int][]*cclfFileMetadata, skipped *
 			return nil
 		}
 
-		zipReader, err := zip.OpenReader(path)
+		zipReader, err := zip.OpenReader(filepath.Clean(path))
+		defer zipReader.Close()
+
 		if err != nil {
+			// this skipped count is not entirely accurate.
```
msnook: Why?   We should only count skips for the top-level zips; not the individual files within the zips.  That would get very noisy, since we automatically skip 10 files in each archive.

bcda/cclf/cclf.go
```diff
@@ -509,87 +518,96 @@ func validate(fileMetadata *cclfFileMetadata, cclfFileValidator map[string]cclfF
...
 }
 
-func validateFileName(fileName string) error {
-	filenameRegexp := regexp.MustCompile(`(T|P).*\.((?:A|T)\d{4})\.ACO.*\.ZC(0|8)Y(\d{2})\.(D\d{6}\.T\d{6})\d`)
-	valid := filenameRegexp.MatchString(fileName)
-	if !valid {
-		fmt.Printf("Invalid filename for file: %s.\n", fileName)
-		err := fmt.Errorf("invalid filename for file: %s", fileName)
+func validateCCLFFolderName(folderName string) error {
+	// add comment for new folder format string
```
msnook: I don't understand this comment.

bcda/cclf/cclf.go
```diff
@@ -509,87 +518,96 @@ func validate(fileMetadata *cclfFileMetadata, cclfFileValidator map[string]cclfF
...
-					err := fmt.Errorf("maximum record count reached for file %s (expected: %d, actual: %d)", key, validator.totalRecordCount, count)
-					log.Error(err)
-					return err
-				}
-			} else {
-				fmt.Printf("Incorrect record length for file %s, Expected record length: %d, Actual record length: %d.\n", key, validator.maxRecordLength, bytelength)
-				err := fmt.Errorf("incorrect record length for file %s (expected: %d, actual: %d)", key, validator.maxRecordLength, bytelength)
+	}
+
+	rc, err := rawFile.Open()
```
msnook: It is unclear to me what is happening here.  Which file is this opening?  The top-level zip or the text file within the zip?

msnook:
Interested to see the test files.  Will pull down your branch and test with the new files once they are committed!

msnook:
The security checklist referenced in the description is the incorrect security checklist.  I don't think one exists for the new CCLF zip bundle.

msnook:
Looks good!
A couple things we'll need to do before merging...
(1) Merge `master` into this branch
(2) Deploy the feature branch to to NFS, and test with the corresponding changes to the smoke test in bcda-ops: https://github.com/CMSgov/bcda-ops/pull/402

bcda/cclf/cclf.go
```diff
@@ -427,43 +447,61 @@ func sortCCLFArchives(cclfMap *map[string]map[int][]*cclfFileMetadata, skipped *
...
-			if (*cclfMap)[metadata.acoID][metadata.perfYear] != nil {
-				(*cclfMap)[metadata.acoID][metadata.perfYear] = append((*cclfMap)[metadata.acoID][metadata.perfYear], &metadata)
+		for _, f := range zipReader.File {
+			metadata, err := getCCLFFileMetadata(f.Name)
+			metadata.filePath = path
+			metadata.deliveryDate = info.ModTime()
+
+			if err != nil {
+				// skipping files with a bad name.  An unknown file in this dir isn't a blocker
+				fmt.Printf("Unknown file found: %s.\n", f.Name)
```
msnook: Do we have to log this?  We're going to get TONS of these log statements due to skipping CCLF3, 4, 5, 6, ... etc.

bcda/cclf/cclf.go
```diff
@@ -67,22 +67,27 @@ func importCCLF0(fileMetadata *cclfFileMetadata) (map[string]cclfFileValidator,
...
-		log.Infof("Reading file #%d from archive %s", i, fileMetadata)
-		if err := validateFileName(f.Name); err != nil {
-			fmt.Printf("Unknown file name when validating file: %s.\n", f.Name)
-			err = errors.Wrapf(err, "unknown file name when validating file: %s.\n", f.Name)
-			log.Error(err)
-			return nil, err
+	var rawFile *zip.File
+
+	for _, f := range r.File {
+		// needed for split files, otherwise this could have been an '==' check
```
msnook: What are "split files"?  Why can't this be an `==` check?

bcda/cclf/cclf.go
```diff
@@ -67,22 +67,27 @@ func importCCLF0(fileMetadata *cclfFileMetadata) (map[string]cclfFileValidator,
...
+
+		if rawFile == nil {
+			fmt.Printf("File %s not found in archive %s.\n", fileMetadata.name, fileMetadata.filePath)
+			err = errors.Wrapf(err, "file %s not found in archive %s", fileMetadata.name, fileMetadata.filePath)
+			log.Error(err)
 			return nil, err
 		}
-		rc, err := f.Open()
+
+		rc, err := rawFile.Open()
```
msnook: Trying to understand what this is doing... is it opening a zip (i.e., CCLF0) within the top-level zip?  Or is this how we are intending to access the plaintext file within the top-level zip?  Remember, according to the tech spec (and according to the files that have been distributed via ACO-OS), only the top-level zip is in zip format and all files within the zip (including CCLF0 and CCLF8) are in plaintext format.

bcda/cclf/testutils/cclfUtils.go
```diff
@@ -61,13 +60,14 @@ func ImportCCLFPackage(acoSize, environment string) (err error) {
...
-		archiveName := fmt.Sprintf("T.BCD.ACO.%sY%s.T%s%s", file.Name(), dateStr, acoIDNum, "000")
-		err = zipTo(fmt.Sprintf("%s/%s", sourcedir, file.Name()), filename, fmt.Sprintf("%s/%s", DestDir, archiveName))
-		if err != nil {
-			return err
-		}
+	for i, file := range files {
+		timeStr := time.Now().Add(time.Minute * time.Duration(i-1)).Format("1504059")
+			archiveName := fmt.Sprintf("T.BCD.A%s.ZCY%s.T%s", acoIDNum, dateStr, timeStr)
+			filename := fmt.Sprintf("T.BCD.A%s.%sY%s.T%s", acoIDNum, file.Name(), dateStr, "0000000")
+			err = zipTo(fmt.Sprintf("%s/%s", sourcedir, file.Name()), filename, fmt.Sprintf("%s/%s", DestDir, archiveName))
```
msnook: I ran your branch locally to try to understand why the tests are running successfully in Travis but failing in `dev`.  It looks like these test files are configured incorrectly.  This code creates 3 zips - one for cclf0, one for cclf8, and one for cclf9.  We can't use `zipTo()` here since that creates a new zip file for every file passed to it... we need to instead create an archive and add files to it.  An example can be found here: https://golangcode.com/create-zip-files-in-go/

bcda/cclf/cclf.go
tbellj: Split file logic was there before, i personally didn't understand why we needed it (like if we can't trust them to deliver files...), but i only focused on changing the file format acceptance logic and nothing else.

bcda/cclf/cclf.go
tbellj: this is opening up the actual file

bcda/cclf/cclf.go
tbellj: we don't have to log this if we don't think we need it

bcda/cclf/cclf.go
tbellj: Opening up the file within the zip

bcda/cclf/cclf.go
tbellj: yup u right, changed :)

bcda/cclf/cclf.go
tbellj: wanted to add something like this: 
// CCLF foldername convention for SSP with BCD identifier: P.BCD.A****.ZCY**.Dyymmdd.Thhmmsst
You'll see this in the next commit

PR #450, opened 2020-02-05
BCDA-2628 Maintenance: Encrypt local/Travis BFD variables

dhgreene:
### Fixes [BCDA-2628](https://jira.cms.gov/browse/BCDA-2628)
The source of our synthetic data for testing in development environments and in Travis CI is changing.  The new data source requires that we keep access details secret.  This PR provides for those values to be secret; it will be followed by a PR for BCDA-2650 which will provide new encrypted access details.

### Proposed Changes
- Encrypt relevant files/settings using Ansible Vault
- Store the password for Vault in a place Travis CI can access it (encrypted with our repo's Travis CI public key, stored in `.travis.yml`)
- Take all appropriate precautions against accidentally committing secrets, including
   - Borrowing a Vault helper script from the ops repo
   - Setting both Git and the Vault helper script to ignore files without an `.encrypted` extension 
   - Borrowing a well-tested git pre-commit hook from the ops repo to balk at committing any files with an `.encrypted` extension if they are not encrypted by Vault

### Security Implications
Ansible Vault is now part of this public repository.

This change follows the security model proposed [by the BFD team](https://confluence.cms.gov/pages/viewpage.action?spaceKey=BB&title=Update+to+SIA+to+include+request+to+open+prod-sbx+to+general+internet+access+for+developer+usage)

#### Travis runs using secrets
Screen Shot 2020-02-05 at 1 07 25 AM

### Feedback Requested
- All feedback welcome, especially questions regarding security and usability
- Is this the right opportunity to refactor more instances of "BB" to "BFD"?
- Does the README make sense?
- Does this process of managing secrets make life much harder for the software engineers?


docker-compose.yml
```diff
@@ -26,6 +26,8 @@ services:
     build:
       context: .
       dockerfile: Dockerfiles/Dockerfile.bcda
+    env_file:
+      - ./shared_files/config/local.env
```
msnook: I like this approach, but I'm feeling like it obfuscates the configuration a little bit.  For example, if I were looking at the `docker-compose` file only, I would have to dig to determine where/how certain vars like `BB_HASH_ITER` and `BB_HASH_PEPPER` are being set.  Maybe we can have comments in the `docker-compose` indicating exactly which vars are being set from `local.env`?

shared_files/config/.gitignore
```diff
@@ -0,0 +1,3 @@
+*
+!*.encrypted
```
msnook: I didn't know you could use `!` in `.gitignore` 💯 
Does this mean that all files except `README.md` and `*.encrypted` are ignored by `git`?

ops/copy.encrypted
```diff
@@ -0,0 +1,18 @@
+#! /bin/bash -
```
msnook: I totally understand why we have it and why we need it.  But its an annoying step, so I'm trying to think of alternatives.

Throwing this out there...
What if we no longer store the vars with a `.encrypted` extension, and instead keep them as-is (like we do in `bcda-ops`)?  Instead, we create a directory called `encrypted_config` and use the directory name to key off of for managing secrets, instead of extension?  (Or, we just decide that `config` is where we keep all encrypted vars and keep it named as-is).
Then, the encryption/decryption process happens to the file in-place, just like in `bcda-ops`.  No need for this script, and our `pre-commit` hook will ensure we don't commit unencrypted vars.

msnook:
Once this is merged to `master`, lets make an announcement so everyone on the team pulls in the pre-commit hook.
Also, maybe it makes sense to add the pre-commit hook (for both repos) to the BCDA HW to ensure everyone on the team has it locally?

msnook:
What follow-up is necessary in BCDA-2650?

shared_files/config/.gitignore
dhgreene: Yes!  I would have preferred to keep all the settings in the .gitignore at the root of the repo, but this strategy works best in its own file.

ops/copy.encrypted
dhgreene: A separate directory might slightly reduce confusion, but we would still need a version of this script (at least for Travis) to copy the files to their destination once they've been unencrypted.  

As for having a separate step: it's totally possible to embed this in the `secrets` script with a `--copy` argument.  But as I was working on implementing that, it became clear that the `secrets` script is more general purpose (it doesn't care what files it's working on or how they're named), and this script cares about naming (or if we follow your suggestion, about knowing a destination).

Do you like it better if we remove this script and create:
- A dedicated `shared_files/config/encrypted/` subdirectory
- An enhancement to the `secrets` script that allows for `ops/secrets --decrypt shared_files/config/encrypted/ --copy shared_files/config/`

msnook: > but we would still need a version of this script (at least for Travis) to copy the files to their destination once they've been unencrypted.
Wouldn't the `copy` that Travis needs just be a one-liner in the `.travis.yml` file?

docker-compose.yml
dhgreene: Great idea.  I've done this in the latest commit.

ops/copy.encrypted
msnook: > A dedicated `shared_files/config/encrypted/` subdirectory

We could keep it named as is `/shared_files/config` and update the `README` within that directory to indicate that only encrypted data lives in this directory.  Then, the env vars could be named exactly what we need them to be and no unnecessary copying is required.

> An enhancement to the secrets script that allows for `ops/secrets --decrypt shared_files/config/encrypted/ --copy shared_files/config/`

I don't think we need to change the `secrets` script.

dhgreene: OK, I think the latest commits following your suggestion make for a better workflow.

msnook:
This looks really good to me.  Interested to see if @nickrobison-usds or @mamurphy have any feedback.

.travis.yml
```diff
@@ -13,6 +13,10 @@ language: go
 go:
   - 1.11.5
 
+# Needed for Ansible Vault
+python:
+  - 2.7
```
nickrobison-usds: Does Ansible not support Python 3? I think 2 is EOL this year, right?

.travis.yml
```diff
@@ -38,10 +42,21 @@ before_script:
...
+
 
 # script always runs to completion (set +e). If we have linter issues AND a
 # failing test, we want to see both. Configure golangci-lint with a
 # .golangci.yml file at the top level of your repo.
 script:
+  # Put unencrypted secrets in the expected place
+  - echo $VAULT_PW > /home/travis/gopath/src/github.com/CMSgov/bcda-app/.vault_password
+  - bash /home/travis/gopath/src/github.com/CMSgov/bcda-app/ops/secrets --decrypt
+  - cp -fv /home/travis/gopath/src/github.com/CMSgov/bcda-app/shared_files/encrypted/* /home/travis/gopath/src/github.com/CMSgov/bcda-app/shared_files/decrypted/
```
nickrobison-usds: Should this be a move rather than a copy, to avoid leaving unencrypted files in the directory?

.travis.yml
dhgreene: Good catch.  Moving to the Python 3 RPM instead.

.travis.yml
dhgreene: Changed in the latest commit.

.travis.yml
dhgreene: @nickrobison-usds I could use a fresh approval, thanks!

PR #454, opened 2020-02-11
BCDA-1551 Feature: Change for recieving plaintext MBI

tbellj:
Fixes [BCDA-1551](https://jira.cms.gov/browse/BCDA-1551)

Problem:
As currently implemented our users have no way of identifying patients from the downloaded payload. The identifier values were previously hashed, but with this change values will be presented as usable plain text.

Proposed Changes:
Added the necessary header values to our BFD request to provide a response with a plaintext MBI.

Change Details:
bcda/client/bluebutton.go - added necessary header value
bcda/client/bluebutton_test.go - testing logic for new header value
bcda/models/models.go - changed to validate the plaintext identifier when in MBI mode.
bcda/testUtils/client.go - changed to use synthetic MBI from static data file
shared_files/synthetic_beneficiary_data/Patient_MBI - changed to a synthetic plaintext MBI

Security Implications:
Yes, this change **does** deal with PII/PHI data. We are now receiving plaintext MBI's in our data responses from BFD.

https://confluence.cms.gov/pages/viewpage.action?pageId=263566369

Acceptance Validation:
Yes all tests pass

Feedback Requested:
Any and all

### DIFF
--- a/bcda/models/models.go
+++ b/bcda/models/models.go
@@ -500,7 +500,7 @@ func (suppressionBeneficiary *Suppression) GetBlueButtonID(bb client.APIClient) 
 func GetBlueButtonID(bb client.APIClient, modelIdentifier, patientIdMode, reqType string, modelID uint) (blueButtonID string, err error) {
 	systemMode := "hicn-hash"
 	if patientIdMode == "MBI_MODE" {
-		systemMode = "mbi-hash"
+		systemMode = "us-mbi"
 	}
 
 	// didn't find a local value, need to ask BlueButton
@@ -528,9 +528,14 @@ func GetBlueButtonID(bb client.APIClient, modelIdentifier, patientIdMode, reqTyp
 	var foundIdentifier = false
 	var foundBlueButtonID = false
 	blueButtonID = patient.Entry[0].Resource.ID
+	// once this is set to true, what happens with future iterations??
 	for _, identifier := range patient.Entry[0].Resource.Identifier {
 		if strings.Contains(identifier.System, systemMode) {
-			if identifier.Value == hashedIdentifier {
+			hashedOrRawID := hashedIdentifier
+			if patientIdMode == "MBI_MODE" {
+				hashedOrRawID = modelIdentifier
+			}
+			if identifier.Value == hashedOrRawID {
 				foundIdentifier = true
 			}
 		} else if strings.Contains(identifier.System, "bene_id") {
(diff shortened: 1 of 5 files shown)


bcda/models/models.go
```diff
@@ -528,9 +528,14 @@ func GetBlueButtonID(bb client.APIClient, modelIdentifier, patientIdMode, reqTyp
 	var foundIdentifier = false
 	var foundBlueButtonID = false
 	blueButtonID = patient.Entry[0].Resource.ID
+	// once this is set to true, what happens with future iterations??
```
msnook: What does this comment mean?

bcda/testUtils/client.go
```diff
@@ -64,7 +64,8 @@ func (bbc *BlueButtonClient) GetData(endpoint, patientID string) (string, error)
 			}
 			cleanData := strings.Replace(string(fData), "20000000000001", patientID, -1)
 			if bbc.MBI != nil {
-				cleanData = strings.Replace(cleanData, "c945694486322cfcf792a4d1c69ea88bf0c134edaf9f1d68a18466f9dc6ec2fe", client.HashIdentifier(*bbc.MBI), -1)
+				// no longer hashed, but this is only a test file with synthetic test data
+				cleanData = strings.Replace(cleanData, "-1Q03Z002871", *bbc.MBI, -1)
```
msnook: Does this mean the unhashed MBI that is coming back is a negative value?  I'm not sure that is correct.  Looking at the list of MBIs loaded into the `prod-sbx`, I expect this to be non-negative.

bcda/models/models.go
tbellj: nothing false alarm, was over engineering in my head

bcda/testUtils/client.go
tbellj: yeah i just put the negative on it to signify a purely synthetic value, i can take it off. But this code really only touches the synthetic patient static file found here: shared_files/synthetic_beneficiary_data/Patient_MBI

msnook: Ah!  That makes sense.  Thanks for clarifying.  I think it is perfect as-is.  No need to change.

bcda/models/models.go
msnook: Can this comment be removed?

tbellj: done

PR #455, opened 2020-02-13
BCDA-2738: Accommodate removal of HICN hash from BFD's Patient resource

tbellj:
### Fixes [BCDA-2738](https://jira.cms.gov/browse/BCDA-2738)

Once BFD removes the hashed hicn from payload responses our code would no longer be able to find the identifier (hicn-hash). This would cause a processing problem in our code if not fixed, so this ticket is too alleviate that problem in the future.

### Change Details

- bcda/models/models.go - changed identifier verifying logic.

- bcda/testUtils/client.go - changed to only get data from 1 synthetic file instead of two. Hashed hicns are going away :)

- shared_files/synthetic_beneficiary_data/Patient_MBI - removed the _MBI identifier on the filename as it's no longer needed.

- shared_files/synthetic_beneficiary_data/Patient_HICN - deleted, no longer needed

### Security Implications

Yes this change **does** deal with PII/PHI data. It handles the processing from BFD for verifying the patient identifier (MBI) in the received payload response.

### Acceptance Validation

Local testing has passed.

### Feedback Requested

Please review.


bcda/models/models.go
```diff
@@ -529,12 +524,13 @@ func GetBlueButtonID(bb client.APIClient, modelIdentifier, patientIdMode, reqTyp
 	var foundBlueButtonID = false
 	blueButtonID = patient.Entry[0].Resource.ID
 	for _, identifier := range patient.Entry[0].Resource.Identifier {
-		if strings.Contains(identifier.System, systemMode) {
-			hashedOrRawID := hashedIdentifier
+		if strings.Contains(identifier.System, "us-mbi") {
```
msnook: Looks like we're only checking for the identifier, which is a bit worrisome.  We used to check for the actual hash value.
What if we go back to doing something like this:  https://github.com/CMSgov/bcda-app/blob/a98e431349ece5cb2dd95ac7c731fe5c496ea769/bcda/models/models.go#L504-L512
(but, of course, morph it slightly so we handle the various modes)

tbellj: which actual hashed value can we check for when we make a request via hashed hicn (due to suppression). We only get back the a hashed mbi and plaintext mbi, which is why the logic has that if statement. If you know a value we can check for when in hicn mode im all ears.

msnook: Can't check for any value in HICN mode, due to this impeding change by BFD.  But we can in MBI mode, right?

msnook: This looks good to me now.

msnook: This looks good to me now.  Confirmed that the actual value is being checked (not just the identifier).

PR #459, opened 2020-02-26
BCDA-2090: Add support for `_since` parameter

msnook:
### Fixes [BCDA-2090](https://jira.cms.gov/browse/BCDA-2090)

The API lacks support for FHIR-compliant `_since` parameter.

### Change Details

- Provided ability for user to pass `_since` parameter to `/Patient` and `/Group` bulk data endpoints.
- Introduced validation against `string` data that is passed via `_since` (validated using FHIR DateTime)
- Added technique to acquire the BFD `lastUpdated` metadata for the entire `Bundle`.  This is done upon every single job request, and written to the `Jobs` table as the `transactionTime`.  This date is used as the `transactionTime` contained in the response of the Job status endpoint, and also used as the upper bound to BFD (using `lastUpdated`) regardless of whether or not `_since` is being used.
- Propagated `_since` data through the system and used to populate the `lastUpdated` parameter when requesting data from Beneficiary FHIR Data Server.  The date passed using the `_since` parameter is used as the lower bound in `lastUpdated` when making requests to BFD.
- Implemented validation so that the API returns an error if the date specified in `_since` is prior to the first BFD data load.
- Added unit tests and Postman tests to fully execute positive and negative scenarios with `/Patient` and `/Group`

### Security Implications

When a bulk data request is made using BCDA, PHI/PII is pulled from the Beneficiary FHIR Data Server and aggregated into `ndjson` files that are provided to the caller.  Usage of the `_since` parameter provides the ability for users to isolate the amount of data returned based on timeframe.

> The security checklist can be found [here](https://confluence.cms.gov/pages/viewpage.action?spaceKey=BCDA&title=20-02-24%3A+_Since+Parameter+Security+Checklist).

### Acceptance Validation

In addition to extensive testing via Postman, manual testing was also performed using Swagger in local environment and `dev` environment.

The screenshot below is an example test scenario in the `dev` environment, where the `_since` was parameter was used to specify a data _after_ the date the data was loaded in BFD `prod-sbx` (resulting in 0 data returned).

Screen Shot 2020-02-25 at 4 17 05 PM

### Feedback Requested

Please review.


bcda/web/api_test.go
```diff
@@ -517,49 +664,54 @@ func validateRequestHelper(endpoint string, s *APITestSuite) {
...
+		sinceRequestString = fmt.Sprintf("&_since=%s", since)
+	}
+
 	if endpoint == "Patient" {
 		if resourceType != "" {
-			requestUrl = fmt.Sprintf("/api/v1/%s/$export?_type=%s", endpoint, resourceType)
+			requestUrl = fmt.Sprintf("/api/v1/%s/$export?_type=%s%s", endpoint, resourceType, sinceRequestString)
 		} else {
-			requestUrl = fmt.Sprintf("/api/v1/%s/$export", endpoint)
+			requestUrl = fmt.Sprintf("/api/v1/%s/$export%s", endpoint, sinceRequestString)
```
dhgreene: We probably don't want to omit the `?` and produce `/api/v1/Patient/$export&_since={date}`

bcda/web/api_test.go
```diff
@@ -568,9 +720,9 @@ func bulkRequestHelper(endpoint, resourceType string) (string, func(http.Respons
 	} else {
 		endpoint = "Group/all"
 		if resourceType != "" {
-			requestUrl = fmt.Sprintf("/api/v1/%s/$export?_type=%s", endpoint, resourceType)
+			requestUrl = fmt.Sprintf("/api/v1/%s/$export?_type=%s%s", endpoint, resourceType, sinceRequestString)
 		} else {
-			requestUrl = fmt.Sprintf("/api/v1/%s/$export", endpoint)
+			requestUrl = fmt.Sprintf("/api/v1/%s/$export%s", endpoint, sinceRequestString)
```
dhgreene: Same as above: no `?` to begin the query

test/postman_test/BCDA_Tests_Sequential.postman_collection.json
```diff
@@ -46,8 +50,8 @@
 					"script": {
 						"id": "00a9ceb7-a32e-4a21-a689-201af70162bf",
 						"exec": [
-							"pm.test(\"Response time is less than 200 ms\", function() {",
-							"    pm.expect(pm.response.responseTime).to.be.below(200);",
+							"pm.test(\"Response time is less than 100 ms\", function() {",
+							"    pm.expect(pm.response.responseTime).to.be.below(100);",
```
dhgreene: Will this really get reverted in the merge?

bcda/client/bluebutton.go
```diff
@@ -116,16 +117,18 @@ func (bbc *BlueButtonClient) GetPatientByIdentifierHash(hashedIdentifier, patien
 	return bbc.getData(blueButtonBasePath+"/Patient/", params, "", "")
 }
 
-func (bbc *BlueButtonClient) GetCoverage(beneficiaryID, jobID, cmsID string) (string, error) {
+func (bbc *BlueButtonClient) GetCoverage(beneficiaryID, jobID, cmsID, since string) (string, error) {
 	params := GetDefaultParams()
 	params.Set("beneficiary", beneficiaryID)
+        UpdateParamWithLastUpdated(&params, since)
```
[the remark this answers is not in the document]
msnook: Ran `go fmt` on the file.

bcda/web/api_test.go
msnook: Good catch!

bcda/web/api_test.go
msnook: Good catch!

test/postman_test/BCDA_Tests_Sequential.postman_collection.json
msnook: Yikes!  I wonder why the automatic merge took my version of the file over the version coming from `master`!   Fixed.

bcda/client/bluebutton.go
```diff
@@ -256,3 +259,10 @@ func HashIdentifier(toHash string) (hashedValue string) {
 	}
 	return hex.EncodeToString(pbkdf2.Key([]byte(toHash), pepper, blueButtonIter, 32, sha256.New))
 }
+
+func UpdateParamWithLastUpdated(params *url.Values, since string) {
+	// only set the parameter if it exists and begins with "gt" (to align with what is expected in _lastUpdated)
```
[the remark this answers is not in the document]
msnook: I agree with this;  changing to `ge`.

bcda/models/SwaggerStructs.go
```diff
@@ -134,13 +134,21 @@ type FileParam struct {
...
+	// (Optional) Resource types requested
 	// in: query
 	// style: form
 	// explode: false
 	ResourceType []string `json:"_type"`
 }
 
+// swagger:parameters bulkPatientRequest bulkGroupRequest
+type SinceParam struct {
+        // (Optional) Only include resource versions that were created at or after the given instant in time.  Format of string must align with the FHIR dateTime datatype.
```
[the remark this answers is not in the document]
msnook: Good idea.  Added.

bcda/models/SwaggerStructs.go
```diff
@@ -143,7 +143,7 @@ type ResourceTypeParam struct {
 
 // swagger:parameters bulkPatientRequest bulkGroupRequest
 type SinceParam struct {
-        // (Optional) Only include resource versions that were created at or after the given instant in time.  Format of string must align with the FHIR dateTime datatype.
+        // (Optional) Only include resource versions that were created at or after the given instant in time.  Format of string must align with the FHIR dateTime datatype.  Example: `2020-02-13T08:00:00.000-05:00`
```
dhgreene: Can this be a Swagger example?
- [Spec](https://swagger.io/docs/specification/adding-examples/)
- [SSAS precedence](https://github.com/CMSgov/bcda-ssas-app/blob/c404741a27b48667768c24166d86b1ff020cdf75/ssas/swagger_structs.go#L183) (though admittedly not a param example)

msnook: The `example` annotation is not working in this repo.  I wonder if its set up differently in `bcda-ssas-app`?

dhgreene: Looks like it's simply different for parameters.

msnook:
> It's super exciting to see this PR. I'm a big fan.
> 
> So, there are a couple of subtle bugs dealing with the need to avoid gaps and duplicates in sequences of _since jobs and the FHIR specification. The goal is if a client makes a sequence of _since calls, in each call using the transactionTime returned from the previous call, the client will receive all updated resources just once. The way that FHIR spec handles this no duplicates problem is that it puts an upper bound in the lastUpdated of the resources returned.
> 
> > transactionTime indicates the server's time when the query is run. The response SHOULD NOT include any resources modified after this instant, and SHALL include any matching resources modified up to and including this instant.
> 
> So, since BFD loads resources continuously and the requests that a job makes to BFD happens over a period of time, every request to the BFD should include the job's `transacationTime` as an upper bound for `lastUpdated`.
> 
> An even subtler point, the `transactionTime` for a job should come from the BFD before the job starts. If you keep it the create time of the job, and BFD writes occur during the job, then there will be gaps in the resources returned by BCDA for some beneficiaries in the job.

Thanks for the feedback @RickHawesUSDS .  An upper bound has been added.

bcda/client/bluebutton.go
```diff
@@ -256,3 +259,13 @@ func HashIdentifier(toHash string) (hashedValue string) {
...
 }
+
+func UpdateParamWithLastUpdated(params *url.Values, since string, jobCreation time.Time) {
+	// only set the lower bound parameter if it exists and begins with "ge" (to align with what is expected in _lastUpdated)
+	if len(since) > 0 && strings.HasPrefix(since, "ge") {
+		params.Set("_lastUpdated", since)
+
+		// now add the upper bound based on job creation
+		params.Add("_lastUpdated", "lt"+jobCreation.Format(time.RFC3339Nano))
+	}
```
RickHawesUSDS: Does the BFD accept a nanosecond precision value? FHIR is a millisecond precision specification. HAPI which is the BFD validating library for FHIR is pretty particular about this. I've tried other precisions and found that it is not accepted.

bcda/client/bluebutton.go
```diff
@@ -256,3 +259,13 @@ func HashIdentifier(toHash string) (hashedValue string) {
 	}
 	return hex.EncodeToString(pbkdf2.Key([]byte(toHash), pepper, blueButtonIter, 32, sha256.New))
 }
+
+func UpdateParamWithLastUpdated(params *url.Values, since string, jobCreation time.Time) {
+	// only set the lower bound parameter if it exists and begins with "ge" (to align with what is expected in _lastUpdated)
+	if len(since) > 0 && strings.HasPrefix(since, "ge") {
+		params.Set("_lastUpdated", since)
+
```
RickHawesUSDS: I'm suspicious that this part of the code is using a string instead of the time type. It looks this string comes from the query parameter of the request if I read the code correctly. If so, it seems like this code would allow SQL injection attacks on the BFD.

bcda/client/bluebutton.go
```diff
@@ -256,3 +259,13 @@ func HashIdentifier(toHash string) (hashedValue string) {
 	}
 	return hex.EncodeToString(pbkdf2.Key([]byte(toHash), pepper, blueButtonIter, 32, sha256.New))
 }
+
+func UpdateParamWithLastUpdated(params *url.Values, since string, jobCreation time.Time) {
+	// only set the lower bound parameter if it exists and begins with "ge" (to align with what is expected in _lastUpdated)
+	if len(since) > 0 && strings.HasPrefix(since, "ge") {
+		params.Set("_lastUpdated", since)
```
RickHawesUSDS: The lower bound should be a `gt`. Here's the relevant FHIR specification 

> Resources will be included in the response if their state has changed after the supplied time (e.g. if Resource.meta.lastUpdated is later than the supplied _since time).

bcda/client/bluebutton.go
RickHawesUSDS: Same type of comment on the upper bound. The upper bound should be a `le`.

msnook: BFD is accepting our nanosecond requests.  This also appears to align with the FHIR `dateTime` format:
https://www.hl7.org/fhir/datatypes.html#dateTime

Relevant portion of the regex is `(\.[0-9]+)`

bcda/client/bluebutton.go
msnook: Yup!  Updated.

bcda/client/bluebutton.go
msnook: Updated.

bcda/client/bluebutton.go
msnook: Input validation is done here: https://github.com/CMSgov/bcda-app/pull/459/files#diff-9b652353711032572a163389de40e045R263

msnook:
> It's super exciting to see this PR. I'm a big fan.
> 
> So, there are a couple of subtle bugs dealing with the need to avoid gaps and duplicates in sequences of _since jobs and the FHIR specification. The goal is if a client makes a sequence of _since calls, in each call using the transactionTime returned from the previous call, the client will receive all updated resources just once. The way that FHIR spec handles this no duplicates problem is that it puts an upper bound in the lastUpdated of the resources returned.
> 
> > transactionTime indicates the server's time when the query is run. The response SHOULD NOT include any resources modified after this instant, and SHALL include any matching resources modified up to and including this instant.
> 
> So, since BFD loads resources continuously and the requests that a job makes to BFD happens over a period of time, every request to the BFD should include the job's `transacationTime` as an upper bound for `lastUpdated`.
> 
> An even subtler point, the `transactionTime` for a job should come from the BFD before the job starts. If you keep it the create time of the job, and BFD writes occur during the job, then there will be gaps in the resources returned by BCDA for some beneficiaries in the job.

Thanks for the feedback.  Per our discussions, the `transactionTime` is now being acquired from BFD (using a non-existent patient in the request to retrieve the Bundle's `lastUpdated` metadata).

msnook:
@RickHawesUSDS @dhgreene This is now ready for another review!

PR #464, opened 2020-03-09
BCDA-2730 Feature: Disable TRACE and OPTIONS HTTP methods

jeffcorpuz:
### Fixes [BCDA-2730](https://jira.cms.gov/browse/BCDA-2730)

Address OPTIONS and TRACE vulnerability. 

### Proposed Changes

Disable OPTIONS and TRACE HTTP methods by just commenting them out until addressed.

### Security Implications

### Acceptance Validation
- https://bcda-ci.adhocteam.us/job/BCDA%20-%20Build%20and%20Package/1354/

### Feedback Requested

Any feedback concerning commenting out the methods vs removing them flat out right.

### DIFF
--- a/vendor/github.com/soheilhy/cmux/matchers.go
+++ b/vendor/github.com/soheilhy/cmux/matchers.go
@@ -44,13 +44,13 @@ func prefixByteMatcher(list ...[]byte) Matcher {
 }
 
 var defaultHTTPMethods = []string{
-	"OPTIONS",
+	// "OPTIONS",
 	"GET",
 	"HEAD",
 	"POST",
 	"PUT",
 	"DELETE",
-	"TRACE",
+	// "TRACE",
 	"CONNECT",
 }
 
--- a/vendor/github.com/soheilhy/cmux/matchers.go
+++ b/vendor/github.com/soheilhy/cmux/matchers.go
@@ -44,13 +44,13 @@ func prefixByteMatcher(list ...[]byte) Matcher {
 }
 
 var defaultHTTPMethods = []string{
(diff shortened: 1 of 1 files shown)


vendor/github.com/soheilhy/cmux/matchers.go
```diff
@@ -44,13 +44,13 @@ func prefixByteMatcher(list ...[]byte) Matcher {
 }
 
 var defaultHTTPMethods = []string{
-	"OPTIONS",
+	// "OPTIONS",
```
dhgreene: This change might just work, but we need to limit our changes to our own code base--nothing from the vendor directory.

jeffcorpuz: It is the only place where OPTIONS or TRACE are declared so I figure this might be the place

msnook: Agree, we should not be updating vendor code.

msnook:
I did some snooping...
First, I attempted to run `OPTIONS` and `TRACE` methods against our server, and received a `405` in both cases.  
Second, I found this source code from Zap.  I think this is the scan that was done for this scenario.  https://github.com/zaproxy/zap-extensions/blob/master/addOns/ascanrulesBeta/src/main/java/org/zaproxy/zap/extension/ascanrulesBeta/ProxyDisclosureScanner.java
Really confused how we are failing this test in Zap :(

jeffcorpuz:
Closing PR for this as this test case as running TRACE and OPTIONS results in a 405

PR #473, opened 2020-03-30
Add gofmt test to pre-commit git hook

orenfromberg:
### Fixes [BCDA-2778](https://jira.cms.gov/browse/BCDA-2778)

This branch adds a script to the `pre-commit` git hook that checks if any staged go files that are added, copied, or modified need to be formatted with `gofmt`.

A companion PR will be created for `bcda-ssas-app`.

### Proposed Changes

Adding some logic to `ops/pre-commit` based on [this](https://github.com/golang/go/blob/release-branch.go1.1/misc/git/pre-commit) and making the script executable.

### Change Details

* Ran `chmod +x ops/pre-commit`

### Security Implications

### Acceptance Validation
- [x] the Git pre-commit hooks in bcda-app includes a test for gofmt

It isn't necessary to test this change in dev because this is only for local dev environments.

I tested this branch by copying `ops/pre-commit` to `.git/hooks` with:
```bash
$ cp ops/pre-commit .git/hooks
```
Then I created an unformatted file with:
```bash
$ cat <<EOF > bcda/bar.go
package bar
func bar() int {
return      1
}
EOF
```
Then I staged the file and tried to commit:
```bash
$ git add bcda/bar.go
$ git commit -m "test"
Go files must be formatted with gofmt. Please run:
  gofmt -w /home/oren/dev/bcda-app/bcda/bar.go
```

I then ran the `gofmt` command and committed successfully, then undid the commit:
```bash
$ gofmt -w /home/oren/dev/bcda-app/bcda/bar.go
$ git commit -m "test"
[oren/BCDA-2778-gofmt-git-hook 411a6b7] test
 1 file changed, 5 insertions(+)
 create mode 100644 bcda/bar.go
$ git reset --soft HEAD~1
```
### Feedback Requested

All feedback is welcome.


msnook:
This is awesome.
Some questions/comments:
(1) Do we need to format every single file in our repo, so future usage of this pre-commit hook doesn't introduce false positives?  It is definitely possible that, since we didn't have this hook before, there are mis-formatted files in our codebase.  If we decide to do this, should it be part of this ticket or a separate ticket?
(2) How will we broadcast to the team that a new version of the pre-commit hook needs to be copied to folks' local?

orenfromberg:
@msnook thank you!
1. This pre-commit hook will only check if any staged files that have been added, copied, or modified in any way are not formatted according to `gofmt` and will disregard other files that are not within that set. The pre-commit hook will gradually ensure (as it is used) that any file that is added, copied, or modified in the future is formatted properly, so I don't think it would be necessary to run `gofmt` broadly now. If that is something we want to do, it would be safe to do it as part of another ticket.
1. Good question. Would it make sense to send out a message to the bcda dev community on our Slack as well as CMS slack? Do we want to add a step to `make docker-bootstrap` to automatically copy `ops/pre-commit` to `.git/hooks`? A combination of both?

msnook:
> @msnook thank you!
> 
> 1. This pre-commit hook will only check if any staged files that have been added, copied, or modified in any way are not formatted according to `gofmt` and will disregard other files that are not within that set. The pre-commit hook will gradually ensure (as it is used) that any file that is added, copied, or modified in the future is formatted properly, so I don't think it would be necessary to run `gofmt` broadly now. If that is something we want to do, it would be safe to do it as part of another ticket.
> 2. Good question. Would it make sense to send out a message to the bcda dev community on our Slack as well as CMS slack? Do we want to add a step to `make docker-bootstrap` to automatically copy `ops/pre-commit` to `.git/hooks`? A combination of both?

1. Yup, the usage makes sense.  I'm worried that we already have an existing number of files that aren't formatted properly, and I was thinking we should baseline the code with proper formatting so all future users' changes are ONLY related to lines they change, and not caused by `go fmt`.  We can either baseline/format all files as part of this PR, or once this is merged, we can run `go fmt ./...` on the repo in a separate feature branch and ensure everything is formatted properly for future developers.
2. Sending out a message to the BCDA dev community once this PR is merged is sufficient.  Adding the step to `make docker-bootstrap` is a bonus and I'll leave that up to you.

PR #474, opened 2020-03-31
Addressing GoSec linting rule for unsafe defer

orenfromberg:
### Fixes [BCDA-2804](https://jira.cms.gov/browse/BCDA-2804)

This branch removes the `nosec` comments that ignore the linting error of trying to defer a function call that may return an error.

I used some of the guidance from [here](https://www.joeshaw.org/dont-defer-close-on-writable-files/) and [here](https://blog.learngoprogramming.com/5-gotchas-of-defer-in-go-golang-part-iii-36a1ab3d6ef1).

The problem is that this code:
```go
	defer fileToZip.Close()
```
will cause this linting error:
```
[/go/src/github.com/CMSgov/bcda-app/bcda/cclf/testutils/cclfUtils.go:118] - G307 (CWE-): Deferring unsafe method "*os.File" on type "Close" (Confidence: HIGH, Severity: MEDIUM)
  > defer fileToZip.Close()
```
This situation occurs in different scenarios in our own code:
1. assign error to a named return value
1. explicitly return error
1. log error and return a tuple with the error
1. log error
1. panic with error

Because each of these situations varied, I decided to refactor calls to `.Close()` by deferring a closure that would handle the error according to how it is handled in the context of the function it is found in.

An alternative strategy would be writing a common utility for closing files, but that would involve refactoring the functions that use the common utility to match how it handles both errors as return values and the way it logs/handles the error.

### Proposed Changes

* remove `nosec` comments
* replacing defer calls to `f.Close()` with a closure that handles the error according to the context of the function it is found in
* fix linter warnings about function documentation and variable names

### Security Implications

### Feedback Requested
All feedback is welcome, especially with regard to the approach I took here. Does it make more sense to use a common close method instead of this approach?


bcda/cclf/testutils/cclfUtils.go
```diff
@@ -67,15 +66,20 @@ func ImportCCLFPackage(acoSize, environment string) (err error) {
...
+		sourceFilename := fmt.Sprintf("%s/%s__%s", sourcedir, file.Name(), filename)
+		fileList = append(fileList, sourceFilename)
 	}
 
 	newZipFile, err := os.Create(fmt.Sprintf("%s/%s", DestDir, archiveName))
 	if err != nil {
 		return err
 	}
-	defer newZipFile.Close() // #nosec G307
+	defer func() {
```
msnook: I see an opportunity to be more `DRY` here.  Can we define a helper utility for closing files?  Maybe we can define it [here](https://github.com/CMSgov/bcda-app/blob/master/bcda/utils/common.go).

bcda/cclf/testutils/cclfUtils.go
```diff
@@ -67,15 +66,20 @@ func ImportCCLFPackage(acoSize, environment string) (err error) {
...
+		if ferr := newZipFile.Close(); ferr != nil {
+			err = ferr
```
msnook: Should we at least log the error?

bcda/cclf/testutils/cclfUtils.go
```diff
@@ -101,26 +105,32 @@ func ImportCCLFPackage(acoSize, environment string) (err error) {
...
+// AddFileToZip adds the file to a zip archive
+func AddFileToZip(zipWriter *zip.Writer, filename string) (err error) {
+	sourceData := strings.Split(filename, "__")
 	src := sourceData[0]
 	filename = sourceData[1]
 
 	fileToZip, err := os.Open(filepath.Clean(src))
 	if err != nil {
-		return err
+		return
```
msnook: Is there a reason for removing the `err` from the `return` statement?

test/smoke_test/bcda_client.go
```diff
@@ -134,7 +134,12 @@ func writeFile(resp *http.Response, filename string) {
 	if err != nil {
 		panic(err)
 	}
-	defer out.Close() // #nosec G307
+	defer func() {
+		if ferr := out.Close(); ferr != nil {
+			panic(ferr)
```
msnook: Maybe just log instead of panic here.  I think we'll get that for free, though, if we use a common utility function.

bcda/cclf/testutils/cclfUtils.go
orenfromberg: I'll give it a shot!

bcda/cclf/testutils/cclfUtils.go
orenfromberg: the error isn't being ignored, but it is overwriting any error that is being returned by the outer function.

I'm going to rewrite this to use the helper method.

bcda/cclf/testutils/cclfUtils.go
orenfromberg: I removed `err` because it wasn't strictly necessary after modifying the function signature to use a named return value. Using a named return value allows me to set the return value in the closure method that is deferred. This is testing my golang knowledge. Can you return a value from a function that is being called in a defer statement?

orenfromberg:
@msnook I tried creating a couple of helper functions but not sure they are worth the effort. Please let me know what you think and if you have any improvements.

bcda/cclf/testutils/cclfUtils.go
orenfromberg: This is testing my golang skills. The logic here is checking if the error exists and if the returned error is nil, and then setting the named return value to the error. This is to prevent overwriting any other error being returned by the method.

bcda/cclf/testutils/cclfUtils.go
msnook: In this particular case, the `err` that is being returned is related to the call to `os.Open()`, which is completely separate from the usage of `defer`.  So, by removing the `err` from the return statement, we'd lose the error reason if the file fails to open.  I think this can be considered a global comment.

bcda/utils/common.go
```diff
@@ -72,3 +72,19 @@ func IsUUID(s string) bool {
...
+
+// CloseSafely will close a file and log any errors
+func CloseSafely(f *os.File) {
+	if ferr := f.Close(); ferr != nil {
+		logrus.Error(ferr)
+	}
+}
+
+// CloseSafelyAndReturnError will close a file, log, and return any errors
+func CloseSafelyAndReturnError(f *os.File) (err error) {
```
msnook, line 84: Is this function necessary?  Previously, we were completely losing the error when we called `defer f.Close()`.  No need to return an error now, unless I'm missing a scenario

bcda/utils/common.go
```diff
@@ -72,3 +72,19 @@ func IsUUID(s string) bool {
 	re := regexp.MustCompile("^[a-fA-F0-9]{8}(?:-[a-fA-F0-9]{4}){3}-[a-fA-F0-9]{12}$")
 	return re.MatchString(s)
 }
+
+// CloseSafely will close a file and log any errors
+func CloseSafely(f *os.File) {
```
msnook, line 77: Naming functions is hard :)
What exactly do we mean by "safely"?  And what exactly are we closing?
Recommend something like `CloseFile()`

bcda/cclf/testutils/cclfUtils.go
orenfromberg: Oops. Thanks for bearing with me as I get familiar with golang :smile: 



Thanks for bearing with me as I get myself familiarized with golang :smile:

bcda/utils/common.go
orenfromberg: No, it isn't necessary. If we are losing the error then no need to accommodate this function. I was trying to do a lot more than was called for!

PR #484, opened 2020-05-08
BCDA-2981: Fix critical code smells

PatKelsh:
### Fixes [BCDA-2981](https://jira.cms.gov/browse/BCDA-2981)

Fixed some critical SonarQube code smells. Mostly duplicated string literals and some addressing code complexity

### Security Implications

While there should be no logical changes to the code, some of these changes are in the authentication components and should be double checked to make sure none of the changes effect how authentication functions

### Feedback Requested

Look over the changes to the code for best practices and see if the changes make sense for how the team would like to write code. Does following the standard rules make sense or should we write new rules?

### DIFF
--- a/bcda/auth/alpha.go
+++ b/bcda/auth/alpha.go
@@ -14,12 +14,14 @@ import (
 
 type AlphaAuthPlugin struct{}
 
+const NonEmptyString = "provide a non-empty string"
+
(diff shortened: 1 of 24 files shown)


PatKelsh:
Review of these changes shows that by trying to address the code complexity warnings from SonarQube we actually make that code more complex and harder to read. As these changes don't effect security or logic but only make aesthetic changes while making the code harder to read this ticket is being closed

PR #486, opened 2020-05-14
Re-arrange order of Postman tests to avoid 429 race condition

msnook:
### Fixes Race Condition in Postman Tests

In `prod`, if an ACO requests the same resource type repeatedly, they will be presented with a `429` response.  We have several Postman tests which validate this behavior.  However, in some scenarios, our tests might execute too quickly and inadvertently request the same resource type before the previous one has completed by the asynchronous worker job (this can occur in `prod` when existing jobs are in the queue; for example, from a real ACO requesting data).
An example of this occurring can be viewed [here](https://bcda-ci.adhocteam.us/job/BCDA%20-%20Test%20-%20Integration%20Smoke/1928/console).

### Change Details

- Re-arrange tests slightly to ensure that we don't incur a `429` when we don't expect it (due to tests executing too quickly).

### Security Implications

### Acceptance Validation

Smoke tests executed against `dev` using this feature branch with [successful result](https://bcda-ci.adhocteam.us/job/BCDA%20-%20Test%20-%20Integration%20Smoke/1930/).
Smoke tests executed against `prod` using this feature branch with [successful result](https://bcda-ci.adhocteam.us/job/BCDA%20-%20Test%20-%20Integration%20Smoke/1931/).

### Feedback Requested

Please review.

### DIFF
--- a/test/postman_test/BCDA_Tests_Sequential.postman_collection.json
+++ b/test/postman_test/BCDA_Tests_Sequential.postman_collection.json
@@ -1648,86 +1648,6 @@
 			"response": []
 		},
 		{
-			"name": "Start repeated type for /Group EOB export",
-			"event": [
-				{
-					"listen": "test",
-					"script": {
-						"id": "ea1530bd-2243-4f35-8e2c-f4eeb9269c71",
-						"exec": [
-							"pm.test(\"Status code is 400\", function() {",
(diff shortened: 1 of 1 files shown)


msnook:
Had to close/re-open to trigger a rebuild in Travis.  Ugh.

PR #496, opened 2020-06-05
BCDA-3140: Priority weighting given to jobs

PatKelsh:
### Fixes [BCDA-3140](https://jira.cms.gov/browse/BCDA-3140)

Applies a priority weight to every job based on what the parameters of query. The lower the priority number, the sooner the job will be completed relative to other jobs in the queue

### Proposed Changes

The `Job` model has a new field `Priority` that is used by que-go when determining the order to start jobs. The value for priority is generated when jobs are added to the queue

### Change Details

* Added priority to the job enqueuing function
* Updated tests to also test for priority

### Security Implications

### Acceptance Validation

Deployed to dev: https://bcda-ci.adhocteam.us/job/BCDA%20-%20Build%20and%20Package/1565/

### Feedback Requested

Is the best place for the logic to determine a job priority? Should it be with the model or somewhere else? Any other tests that should be part of this?

### DIFF
--- a/bcda/models/models.go
+++ b/bcda/models/models.go
@@ -65,6 +65,7 @@ type Job struct {
 	JobCount          int
 	CompletedJobCount int
 	JobKeys           []JobKey
+	Priority          int // job priority based on request args
 }
 
 func (job *Job) CheckCompletedAndCleanup(db *gorm.DB) (bool, error) {
@@ -149,6 +150,7 @@ func (job *Job) GetEnqueJobs(resourceTypes []string, since string, newBeneficiar
 					ResourceType:    rt,
 					Since:           since,
 					TransactionTime: job.TransactionTime,
+					Priority:        job.Priority,
 				})
 				if err != nil {
 					return nil, err
@@ -640,4 +642,5 @@ type jobEnqueueArgs struct {
 	ResourceType    string
 	Since           string
 	TransactionTime time.Time
+	Priority        int
 }
--- a/bcda/models/models.go
+++ b/bcda/models/models.go
@@ -65,7 +65,7 @@ type Job struct {
 	JobCount          int
 	CompletedJobCount int
 	JobKeys           []JobKey
-	Priority          int // job priority based on request args
+	// Priority          int // job priority based on request args
 }
 
 func (job *Job) CheckCompletedAndCleanup(db *gorm.DB) (bool, error) {
@@ -150,7 +150,8 @@ func (job *Job) GetEnqueJobs(resourceTypes []string, since string, newBeneficiar
 					ResourceType:    rt,
 					Since:           since,
 					TransactionTime: job.TransactionTime,
-					Priority:        job.Priority,
+					// Priority:        job.Priority,
+					Priority: setJobPriority(job.ACOID.String(), rt, len(since) != 0),
 				})
 				if err != nil {
 					return nil, err
@@ -168,6 +169,30 @@ func (job *Job) GetEnqueJobs(resourceTypes []string, since string, newBeneficiar
 		}
 	}
 	return enqueJobs, nil
+}
+
+func setJobPriority(acoID string, resourceType string, sinceParam bool) int {
+	var priority int
+	if isSyntheticACO(acoID) {
+		priority = 10 // priority level for jobs for sythetic ACOs that are used for smoke testing
+	} else if resourceType == "Patient" || resourceType == "Coverage" {
+		priority = 20 // priority level for jobs that only request smaller resources
+	} else if sinceParam {
+		priority = 30 // priority level for jobs that only request data for a limited timeframe
+	} else {
+		priority = 100 // default priority level for jobs
+	}
+	return priority
+}
+
+func isSyntheticACO(acoID string) bool {
+	testACOs := []string{"A9990", "A9991", "A9992", "A9993", "A9994"}
+	for _, testACO := range testACOs {
+		if testACO == acoID {
+			return true
+		}
+	}
+	return false
 }
 
 func (j *Job) StatusMessage() string {
--- a/bcda/models/models.go
+++ b/bcda/models/models.go
@@ -150,8 +150,7 @@ func (job *Job) GetEnqueJobs(resourceTypes []string, since string, newBeneficiar
 					ResourceType:    rt,
 					Since:           since,
 					TransactionTime: job.TransactionTime,
-					// Priority:        job.Priority,
-					Priority: setJobPriority(job.ACOID.String(), rt, len(since) != 0),
+					Priority:        setJobPriority(job.ACOID.String(), rt, len(since) != 0),
 				})
 				if err != nil {
 					return nil, err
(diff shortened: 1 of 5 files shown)


bcda/models/models.go
```diff
@@ -149,6 +149,7 @@ func (job *Job) GetEnqueJobs(resourceTypes []string, since string, newBeneficiar
 					ResourceType:    rt,
 					Since:           since,
 					TransactionTime: job.TransactionTime,
+					Priority:        setJobPriority(job.ACOID.String(), rt, len(since) != 0),
```
msnook, line 152: Nifty!

bcda/models/models.go
```diff
@@ -168,6 +169,30 @@ func (job *Job) GetEnqueJobs(resourceTypes []string, since string, newBeneficiar
...
+	} else if sinceParam {
+		priority = 30 // priority level for jobs that only request data for a limited timeframe
+	} else {
+		priority = 100 // default priority level for jobs
+	}
+	return priority
+}
+
+func isSyntheticACO(acoID string) bool {
+	testACOs := []string{"A9990", "A9991", "A9992", "A9993", "A9994"}
```
msnook, line 187: Hard-coding these might lead to maintenance issues.  Should we make this configurable via env var?  
We have something very similar (env var representing a list of ACOs) [here](https://github.com/CMSgov/bcda-app/blob/master/bcda/cclf/cclf.go#L508) (`CCLF_PRIORITY_ACO_CMS_IDS` env var which is defined [here](https://github.com/CMSgov/bcda-ops/blob/master/config/nfs/dev/CCLF_PRIORITY_ACO_CMS_IDS)).  
Would need to create a separate env var for `dev`, `test`, `opensbx`, and `prod`.

Or is there another option?

bcda/models/models.go
```diff
@@ -168,6 +169,30 @@ func (job *Job) GetEnqueJobs(resourceTypes []string, since string, newBeneficiar
 	return enqueJobs, nil
 }
 
+func setJobPriority(acoID string, resourceType string, sinceParam bool) int {
+	var priority int
```
msnook, line 173: Maybe a comment in here to indicate that low number equals high priority.

bcda/models/models.go
PatKelsh: no, that makes the most sense, having them be defined in environment variables

bcda/models/models.go
PatKelsh: done

bcda/models/models.go
```diff
@@ -168,6 +169,36 @@ func (job *Job) GetEnqueJobs(resourceTypes []string, since string, newBeneficiar
...
+		priority = 30 // priority level for jobs that only request data for a limited timeframe
+	} else {
+		priority = 100 // default priority level for jobs
+	}
+	return priority
+}
+
+// Checks to see if an ACO is synthetic based on a list of sythetic ACOS provided by an
+// environment variable.
+func isSyntheticACO(acoID string) bool {
```
msnook, line 190: Instead of using the term "synthetic ACO", can we instead replace with "high priority ACO"?  That way, its a bit more flexible - we can add other ACOs which aren't synthetic in the list.

bcda/models/models_test.go
```diff
@@ -585,12 +590,49 @@ func (s *ModelsTestSuite) TestGetEnqueJobs_Coverage() {
...
 		assert.True(len(jobArgs.BeneficiaryIDs) <= 5)
 		assert.Equal(since, jobArgs.Since)
+		assert.Equal(20, jobArgs.Priority)
 		assert.NotNil(jobArgs.TransactionTime)
 	}
 	assert.Equal(50, enqueuedBenes)
 	os.Unsetenv("BCDA_FHIR_MAX_RECORDS_COVERAGE")
 }
 
+func (s *ModelsTestSuite) TestGetEnqueJobs_WithSyntheticACOs() {
```
msnook, line 600: See comment above... can be called `TestGetEnqueJobs_WithHighPriorityACOs`

bcda/models/models_test.go
```diff
@@ -585,12 +590,49 @@ func (s *ModelsTestSuite) TestGetEnqueJobs_Coverage() {
...
+		// ACOID:      uuid.Parse("ACO9991"),
+		ACOID:      uuid.Parse(constants.DevACOUUID),
+		RequestURL: "/api/v1/Patient/$export?_type=Patient",
+		Status:     "Pending",
+	}
+	s.db.Save(&j)
+	defer s.db.Delete(&j)
+
+	since := ""
+	err := os.Setenv("SYNTHETIC_ACO_IDS", constants.DevACOUUID)
```
msnook, line 613: Can this be set in `docker-compose` as a comma-delimited list, so it represents exactly what will be defined in AWS?
Can look something like this in `docker-compose`:
```
A9990, A9991, A9992, A9993, A9994
```

docker-compose.test.yml
```diff
@@ -45,6 +45,7 @@ services:
       - SSAS_PUBLIC_URL=http://ssas:3003
       - PATIENT_IDENTIFIER_MODE=MBI_MODE
       - BCDA_ENABLE_NEW_GROUP=true
+      - PRIORITY_ACO_IDS=A9990,A9991,A9992,A9993
```
msnook, line 48: Should we also place this in `docker-compose.yml` so the system runs locally with the list in place?  These env vars are just set in the context of the test container.

bcda/models/models.go
msnook: Should `isSyntheticACO` be renamed to `isPriorityACO`?  
I think we shouldn't have the term "synthetic ACO" in any of this - instead, use "priority ACO".

docker-compose.test.yml
msnook, line 48: Do we need `A9994` too?

bcda/models/models.go
```diff
@@ -149,6 +149,7 @@ func (job *Job) GetEnqueJobs(resourceTypes []string, since string, newBeneficiar
 					ResourceType:    rt,
 					Since:           since,
 					TransactionTime: job.TransactionTime,
+					Priority:        setJobPriority(*aco.CMSID, rt, len(since) != 0),
```
msnook, line 152: According to [this documentation](https://godoc.org/github.com/bgentry/que-go#Job), `Priority` should not be set as part of the `Args` field.  Instead, it is its own field and should be set independently [here](https://github.com/CMSgov/bcda-app/blob/master/bcda/models/models.go#L157-L160).

bcda/models/models.go
```diff
@@ -65,6 +65,7 @@ type Job struct {
 	JobCount          int
 	CompletedJobCount int
 	JobKeys           []JobKey
+	Priority          int
```
msnook, line 68: This is unnecessary, right?  The `Priority` should be on the `que.Job` table already, which you are already setting correctly below.

msnook:
Tested locally and it works great!

PR #506, opened 2020-07-01
BCDA-2751 Deduplicate patient data coming from CCLF

PatKelsh:
### Fixes [BCDA-2751](https://jira.cms.gov/browse/BCDA-2751)

Beneficiaries can show up multiple times in the same CCLF file which causes difficulties for end users who have to shift through the attribution data we provide to remove the duplicates. This branch will de-duplicate the attribution data when responding to a bulk data request

### Proposed Changes

Updated the query to filter out duplicate beneficiaries and test data to include duplicates

### Security Implications

Ensures redundant data is not being transmitted to the client

### Acceptance Validation

Deployed change to 'dev'


bcda/models/models_test.go
```diff
@@ -948,6 +948,12 @@ func (s *ModelsTestSuite) TestGetBeneficiaries_Unsuppressed() {
 		s.FailNow("Failed to save CCLF file", err.Error())
 	}
 	defer s.db.Unscoped().Delete(&cclfFile)
+	// cclfFile2 := CCLFFile{CCLFNum: 7, ACOCMSID: acoCMSID, ImportStatus: constants.ImportComplete}
```
msnook, line 951: Is this dead code?

msnook:
In the description for `Security Implications`, this PR is in fact dealing with PHI/PII.  Maybe that should be un-checked and something like this should be included in the `Security Implications` section:
"Ensures redundant data is not being transmitted to the client"

msnook:
Can the description be beefed up to indicate how this situation arises in  CCLF8?  Basically, we receive a list of beneficiaries in CCLF8 and if a beneficiary has been in/out of hospice, they may show up multiple times.  This scenario results in our attribution list containing multiple entries for the same beneficiary, which results in the client receiving redundant data.  To eliminate this scenario, we are de-duplicating the beneficiary list at the time the client makes a bulk data request.

PR #509, opened 2020-07-06
BCDA-3112: Address FHIR compliance for Patient-level export

PatKelsh:
### Fixes [BCDA-3112](https://jira.cms.gov/browse/BCDA-3112)

Fixing the FHIR conformance issues related to the Patient resource

### Proposed Changes

* Added validation for the `_outputFormat` query paramter
* Expanded validation for the `_since` query parameter to reject parameters with excess characters and reject dates that are in the future
### Security Implications

### Acceptance Validation
Deployed to dev and tested using the "BCDA-FHIR Scan" job in Jenkins to scan for compliance.
NOTE: Not all resources are FHIR compliant at this time but reading the log of the scan the patient level tests will all be passing


bcda/web/api.go
```diff
@@ -305,12 +305,26 @@ func validateRequest(r *http.Request) ([]string, *fhirmodels.OperationOutcome) {
...
+		if err != nil || sinceDate.After(time.Now()) {
+			oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, responseutils.FormatErr, "Invalid date format supplied in _since parameter. Date must be a date that has already passed")
+			return nil, oo
+		}
+	}
+
+	//validate "_outputFormat" parameter
+	params, ok = r.URL.Query()["_outputFormat"]
+	if ok {
+		if params[0] != "ndjson" && params[0] != "application/fhir+ndjson" && params[0] != "application/ndjson" {
```
msnook, line 324: Can we add a unit test for this?

msnook:
Looks like the Postman tests are failing here: https://travis-ci.org/github/CMSgov/bcda-app/builds/705560498

It is possible that it is failing because of the call to `new Date()` in the test (which is very unrealistic anyway) and it is treating it as a future date?  Maybe we need to make it a `var yesterday = (Date.now() - 86400000)` or something like what is offered [here](https://stackoverflow.com/questions/5511323/calculate-the-date-yesterday-in-javascript)?

bcda/web/api.go
```diff
@@ -305,12 +305,26 @@ func validateRequest(r *http.Request) ([]string, *fhirmodels.OperationOutcome) {
 	params, ok = r.URL.Query()["_since"]
 	if ok {
 		// Regex pattern for FHIR Instant format found at https://www.hl7.org/fhir/datatypes.html#primitive
-		instantRegex := `([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1])T([01][0-9]|2[0-3]):[0-5][0-9]:([0-5][0-9]|60)(\.[0-9]+)?(Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00))`
-		validSince, err := regexp.MatchString(instantRegex, params[0])
-		if err != nil || !validSince {
+		instantRegex := `^([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1])T([01][0-9]|2[0-3]):[0-5][0-9]:([0-5][0-9]|60)(\.[0-9]+)?(Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00))$`
```
msnook, line 308: Does this change also address https://jira.cms.gov/browse/BCDA-3292 ?

PatKelsh: Yes

test/postman_test/BCDA_Tests_Sequential.postman_collection.json
```diff
@@ -1351,7 +1351,9 @@
 					"script": {
 						"id": "b822f300-2ead-4f25-b623-22c68df6b017",
 						"exec": [
-							"let timestamp = new Date().toJSON();",
+							"let date = new Date()",
```
msnook, line 1354: Looks like this is NOT the problem.  This can probably be reverted.  The issue is with the `Z` in the timestamp.

bcda/web/api.go
```diff
@@ -305,12 +305,26 @@ func validateRequest(r *http.Request) ([]string, *fhirmodels.OperationOutcome) {
...
+		validSinceFormat, err := regexp.MatchString(instantRegex, params[0])
+		if err != nil || !validSinceFormat {
 			oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, responseutils.FormatErr, "Invalid date format supplied in _since parameter.  Date must be in FHIR Instant format.")
 			return nil, oo
 		}
+		sinceDate, err := time.Parse("2006-01-02T15:04:05.000-07:00", params[0])
```
msnook, line 314: The FHIR Instant format supports a timestamp that looks like this, but also a timestamp that contains varying decimal places in the seconds field, or a `Z` in place of a timezone.  I think the parsing is failing if the timestamp contains a `Z`.

msnook: Can use `time.RFC339` format here, maybe?
But still need to make sure the num decimal places is supported.

mtrang1263: You can also use `time.RFC3339Nano` It'll handle the millis for you.

Ex: https://play.golang.org/p/Gbe1aQ5eXs8

bcda/web/api.go
mtrang1263: Since the FHIR#instant type appears to the same as RFC3339 format, can we remove the regex check and instead rely on the error being returned by the `time.Parse` call?

msnook: Unfortunately, it is NOT the same as RFC3339.  FHIR Instant allows more than 10 decimal places for "seconds", RFC3339 only allows 9. :(

mtrang1263: Ah darn! Picosecond resolution is pretty impressive :)

Inside the doc, it does list the following description:

```

An instant in time in the format YYYY-MM-DDThh:mm:ss.sss+zz:zz (e.g. 2015-02-07T13:28:17.239+02:00 or 2017-01-01T00:00:00Z).
--
```

Should the documentation (and regex) be updated to reflect the format only goes up to milliseconds?

msnook: The regex for fhir instant was taken straight from the spec: https://www.hl7.org/fhir/datatypes.html#primitive
I think we should support that exactly as-is (which, unfortunately, is even more precise than RFC3339)

mtrang1263: Gotcha. It also matches the spec for [xsd:datetime](https://www.w3.org/TR/xmlschema-2/#dateTime)

bcda/web/api.go
mtrang1263: Since the spec allows for unbounded number of subsecond digits, we'll be able to handle micros/nanos if the user provided that (if we use RFC3339Nano)

test/postman_test/BCDA_Tests_Sequential.postman_collection.json
```diff
@@ -1351,7 +1351,7 @@
 					"script": {
 						"id": "b822f300-2ead-4f25-b623-22c68df6b017",
 						"exec": [
-							"let timestamp = new Date().toJSON();",
+							"let timestamp = date.toJSON();",
```
msnook, line 1354: should it be `new Date().toJSON()` like it was before?

bcda/web/api.go
msnook: Did we decide on removing this regex and instead using `RFCNano3339` ?

bcda/web/api_test.go
```diff
@@ -796,7 +968,13 @@ func validateRequestHelper(endpoint string, s *APITestSuite) {
 	assert.Equal(s.T(), "Repeated resource type", err.Issue[0].Details.Coding[0].Display)
 }
 
-func bulkRequestHelper(endpoint, resourceType, since string) (string, func(http.ResponseWriter, *http.Request), *http.Request) {
+type RequestParams struct {
```
msnook, line 971: Smart!  Thanks for adding this and making our test code better!

bcda/web/api_test.go
```diff
@@ -810,11 +988,14 @@ func bulkRequestHelper(endpoint, resourceType, since string) (string, func(http.
...
+	if testRequestParams.resourceType != "" {
+		q.Set("_type", testRequestParams.resourceType)
+	}
+	if testRequestParams.since != "" {
+		q.Set("_since", testRequestParams.since)
 	}
-	if since != "" {
-		q.Set("_since", since)
+	if testRequestParams.outputFormat != "" {
+		q.Set("_outputFormat", testRequestParams.since)
```
msnook, line 998: Should this be `testRequestParams.outputFormat`?

bcda/web/api_test.go
```diff
@@ -761,33 +907,59 @@ func validateRequestHelper(endpoint string, s *APITestSuite) {
...
-	_, _, req = bulkRequestHelper(endpoint, "Coverage", "")
+	requestParams = RequestParams{resourceType: "Coverage"}
+	_, _, req = bulkRequestHelper(endpoint, requestParams)
 	resourceTypes, err = validateRequest(req)
 	assert.Nil(s.T(), err)
 	assert.Equal(s.T(), 1, len(resourceTypes))
 	assert.Contains(s.T(), resourceTypes, "Coverage")
 
-	_, _, req = bulkRequestHelper(endpoint, "Practitioner", "")
+	// requestParams = RequestParams{resourceType: "Coverage", outputFormat: "ndjson"}
```
msnook, line 931: What is all of this test code commented out?

bcda/web/api_test.go
```diff
@@ -196,6 +196,51 @@ func (s *APITestSuite) TestBulkCoverageRequestInvalidSinceFormatOnlyInvalidDate(
...
+
+func (s *APITestSuite) TestBulkCoverageRequestInvalidSinceFutureDate() {
+	futureDate := time.Now().AddDate(0, 0, 1).Format(time.RFC3339Nano)
+	bulkCoverageRequestInvalidSinceDateHelper("Patient", futureDate, s)
+	s.TearDownTest()
+	s.SetupTest()
+	bulkCoverageRequestInvalidSinceDateHelper("Group/all", futureDate, s)
+}
+
+func (s *APITestSuite) TestBulkCoverageRequestInvalidOutputFormatTextHTML() {
```
msnook, line 223: Do we need three more test functions to test the positive cases where `_outputFormat` is set correctly?  (i.e., `ndjson`, `application/fhir+ndjson`, and `application/ndjson`)?

PR #515, opened 2020-07-17
BCDA-3383: Pre-seed postgres db used for unit testing

mtrang1263:
### Proposed Changes

To improve unit test performance, we want to initialize the Postgres DB in the most time efficient way. We should be using tooling supplied by Postgres, instead of running update calls through the bcda-app.

### Change Details

1. Created new Postgres db container used exclusively for testing (db-test).
    - db-test is automatically seeded on startup via `docker-entrypoint-initdb.d`
2. Added new command to Makefile to spin up db-test instance.
3. Added new command to Makefile to update seed data.

### Security Implications

### Acceptance Validation

CI tests pass

### Feedback Requested
Review


unit_test.sh
```diff
@@ -10,47 +10,11 @@ timestamp=`date +%Y-%m-%d_%H-%M-%S`
 mkdir -p test_results/${timestamp}
 mkdir -p test_results/latest
 
-echo "Setting up test DB (bcda_test)..."
```
msnook, line 13: I'm worried, that if we eliminate this setup script, it will be difficult for a developer to re-configure/maintain the state of the unit test database.  I wonder if we should create a unit test migration script, which can be used as a starting point for the image...  that way, if anyone needs to change it, they can start with this script.
Ideally, we'd move to a mocking strategy where we mock the DB.  But in the meantime, this will be a good improvement.

docker-compose.test.yml
```diff
@@ -48,9 +47,18 @@ services:
       - PRIORITY_ACO_IDS=A9990,A9991,A9992,A9993,A9994
     volumes:
       - .:/go/src/github.com/CMSgov/bcda-app
+  db-test:
```
msnook, line 50: Should this be called `db-unit-test` to be explicit, so that folks aren't confused when running postman/integration tests?  This DB is only used for unit tests, and not used for any other tests, and I think we need to make that clear.

README.md
```diff
@@ -93,6 +93,18 @@ make test
 make performance-test
 ```
 
+### Updating seed data for unit tests
```
msnook, line 96: Might need to add some clarity here... see this comment:  https://github.com/CMSgov/bcda-app/pull/515/files#r456796847

docker-compose.test.yml
mtrang1263: 👍

unit_test.sh
mtrang1263: Let me update the comment in the README. The `dump.pgdata` and `01-restore.sh` files are intended as the starting point of the image. The pg_restore command in the script encompasses all of the `usql` and `import-synthetic-cclf-package` commands.

README.md
mtrang1263: Let me know if this comment clarifies the behavior of the dump.pgdata file.

README.md
```diff
@@ -93,6 +93,22 @@ make test
...
+
+After the user has finished updating the Postgres db used for unit testing with the new data, the user can update
+the seed data by running the following comamnd:
+```sh
+make unit-test-db-snapshot
+```
+
+This script will update `./db/testing/docker-entrypoint-initdb.d/dump.pgdata` file.
+Thise file is used to initialize the Postgres db with all of the necessary data needed for the various unit tests.
```
msnook, line 105: Typo... `thise`

unit_test.sh
msnook: Yup, I totally get that the packaged `pgdata` file encompasses all of the `usql` and `import-synthetic-cclf-package` commands.  I'm just a little concerned from a historical perspective... future maintainers might not know how to make adjustments to the data using the legacy mechanism.  Using the `pgrestore` makes this process much faster from a CI perspective, but it might cause challenges from a DB-maintenance perspective.  Is there any possible way, for maintenance purposes, to add the legacy mechanism for creating the base DB image?  Maybe we can add the legacy script using `usql` and `import-synthetic-cclf-package` as a giant comment section in `01-restore.sh`?

mtrang1263: That makes sense. I'll go ahead and update the restore script to capture the arguments.

PR #516, opened 2020-07-17
BCDA-3384: Update test behaviors to allow for local testing

mtrang1263:
### Fixes [BCDA-3384](https://jira.cms.gov/browse/BCDA-3384)

For developer velocity, we'd like the ability to run certain unit tests locally, without needing to spin up a Docker container.

### Change Details

1. When comparing Expiry headers use time comparisons instead of string comparisons. When running within the Docker container, timezones were consistent (UTC), so string comparisons worked.
2. Push up test setup behaviors into the corresponding TestSuite. This affects the following test files: cli_test.go, cclf_test.go, suppression_test.go
3. Use temporary directories instead of hard-coded paths. The hard-coded path was using `/go/...` which is restricted on a local file system.
4. Update docker-compose.test to only mount the test_results directory. This will isolate changes made in the Docker container from spilling out to the local file system.

### Security Implications

### Acceptance Validation

Local CI and Travis CI tests are passing.

### Feedback Requested

Please review


mtrang1263:
There's still one open task to update the README to describe how to get the local env setup working.

To do this, I'll want to have https://github.com/CMSgov/bcda-app/pull/515 merged. That way the README can leverage some newly introduced Make commands.

docker-compose.test.yml
```diff
@@ -47,7 +47,7 @@ services:
       - BCDA_ENABLE_NEW_GROUP=true
       - PRIORITY_ACO_IDS=A9990,A9991,A9992,A9993,A9994
     volumes:
-      - .:/go/src/github.com/CMSgov/bcda-app
+      - ./test_results:/go/src/github.com/CMSgov/bcda-app/test_results
```
msnook, line 50: The spillage is acceptable, I think.  I believe we originally set it up this way so that a developer could make changes to a test file locally, and then running `make unit-test` would pick up those changes.  When I pulled down this branch,  and ran `make unit-test` after making a change to a test file, it did not pick up my changes.

msnook:
I created this ticket for a future enhancement: https://jira.cms.gov/browse/BCDA-3398

docker-compose.test.yml
mtrang1263: Ah, that's fair point. Since we mounted the entire directory, when we spun up the container, we would have the entire directory available to us. I'm going to update the Makefile to rebuild the image so we can pick up local changes. That way we can pick up local changes (as expected) and still allow the Docker container to be self contained (i.e. using resources when the image was built).

msnook: Good idea!

PR #519, opened 2020-07-23
BCDA-3370 Add tests to verify Blue Button client uses gzip encoding

mtrang1263:
### Fixes [BCDA-3370](https://jira.cms.gov/browse/BCDA-3370)

Ensure that we're making requests to Blue Button API with `Accept-Encoding: gzip` set.

### Change Details

We had initially thought that we need to explicitly add the `Accept-Encoding: gzip` when making requests to Blue Button API. After looking deeper in the implementation, we need to ensure that we **do not** set this header. If the header is present, then the HTTP client needs to explicitly handle the decompression of the gzip data.

The underlying `http.Transport` will handle this for us if we have `DisableCompression: false`. It transparently decompresses the data and the HTTP client is given the decompressed data.

Actual changes:
1. Adding tests to verify that we are handling gzip responses from the server successfully (i.e. the response does not need to be explicitly de-compressed).
2. Add more comments on why we do not (and should not) set the `Accept-Encoding: gzip` header in our request.
3. Refactored our header test to ensure that all of our requests to Blue Button API have the correct headers set. This refactoring uses [Table Drive Tests](https://github.com/golang/go/wiki/TableDrivenTests)

### Security Implications

### Acceptance Validation

Local tests + CI build
### Feedback Requested

Please review. If you can look at the table driven test, I think it's a good pattern to start using elsewhere.

### DIFF
--- a/bcda/client/bluebutton_test.go
+++ b/bcda/client/bluebutton_test.go
@@ -1,6 +1,8 @@
 package client_test
 
 import (
+	"compress/gzip"
+	"encoding/json"
 	"fmt"
 	"math/rand"
 	"net/http"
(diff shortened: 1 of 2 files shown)


bcda/client/bluebutton_test.go
```diff
@@ -328,35 +332,103 @@ func (s *BBRequestTestSuite) TearDownAllSuite() {
...
-	assert.Equal(s.T(), "543210", req.Header.Get("BCDA-JOBID"))
-	assert.Equal(s.T(), "A00234", req.Header.Get("BCDA-CMSID"))
-	assert.Equal(s.T(), "mbi", req.Header.Get("IncludeIdentifiers"))
+func (s *BBRequestTestSuite) TestValidateRequestHeaders() {
+	tests := []struct {
+		name          string
+		funcUnderTest func(bbClient *client.BlueButtonClient, jobID, cmsID string) (string, error)
+	}{
+		{
+			"ExplanationOfBenefit",
```
msnook, line 341: Should this be `GetExplanationOfBenefit` (for consistency)?

bcda/client/bluebutton_test.go
```diff
@@ -328,35 +332,103 @@ func (s *BBRequestTestSuite) TearDownAllSuite() {
...
+		{
+			"GetPatientByIdentifierHash",
+			func(bbClient *client.BlueButtonClient, jobID, cmsID string) (string, error) {
+				return s.bbClient.GetPatientByIdentifierHash("hashedIdentifier", "patientIdMode")
+			},
+		},
+	}
 
+	for _, tt := range tests {
+		s.T().Run(tt.name, func(t *testing.T) {
```
msnook, line 367: Maybe, in the future as part of BCDA-3398, we can mark this as `t.Parallel()`?

bcda/client/bluebutton_test.go
mtrang1263: 👍

bcda/client/bluebutton_test.go
mtrang1263: Yeah that's a great idea. We'll definitely quite a bit of speed up by running package tests concurrently, and then we can add in parallelization within each file test.

PR #520, opened 2020-07-28
BCDA-1327: Add cutoff time parameter when retrieving CCLF files

mtrang1263:
### Fixes [BCDA-1327](https://jira.cms.gov/browse/BCDA-1327)

Adding cutoff time parameter when retrieving CCLF files for the getBeneficiaries call

### Proposed Changes

1. Adding cutoff time parameter to the query to retrieve CCLF files
2. Refactor out GetNewAndExistingBeneficiaries/GetBeneficiaries call to use new service/repository design.

### Change Details

1. Cutoff parameter will act as a lower bound for CCLF files. We **will not** return CCLF files that are older than a particular date.
2. To help decouple our code from the postgres, we need to move out the query logic that was in models.go into a separate file/package.
    - Service: Contains all of the business logic needed to retrieve the particular domain object. The service contains most of the logic that was captured in models.go
    - Repository: Contains all of the logic needed to query the datastore (Postgres). All of the db calls that were in the models.go are now in the repository.go.
3. Since we do not want to refactor all of models.go to decouple the db, we need to have access to the service instance. We treat the service as a singleton and models.go will call the service directly. Once we refactor all of the business logic out of models.go,  we can remove the service reference.

### Security Implications

We added [go-sqlmock](https://github.com/DATA-DOG/go-sqlmock). These mocks are only used within test scope.

### Acceptance Validation

1. CI Passes
2. Code coverage remains the same (or is improved since we can now test corner cases)
3. Deploy to dev and run smoke tests/possibly run BDT.


bcda/models/postgres/repository.go
```diff
@@ -0,0 +1,113 @@
...
+
+	const (
+		// this is used to get unique ids for de-duplicating MBIs that are listed multiple times in the CCLF8 file
+		idQuery = "SELECT id FROM ( SELECT max(id) as id, mbi FROM cclf_beneficiaries WHERE file_id = ? GROUP BY mbi ) as id"
+	)
+	var beneficiaries []*models.CCLFBeneficiary
+
+	// NOTE: We changed the query that was being used for "old benes"
+	// By querying by IDs, we really should not need to also query by the corresponding MBIs as well
+	query := r.db.Where("id in (?)", r.db.Raw(idQuery, cclfFileID).SubQuery())
```
mtrang1263, line 82: The old version of this query used `mbi IN (oldMBIs) AND id IN (uniqueIDs)` for old benes and `mbi NOT IN (oldMBIs) and ids IN (uniqueIDs)`.

By querying for just `id IN (uniqueIDs)`, we'll roundtrip both old and new benes in a single query and filter them on the host.

mtrang1263:
Wanted to upload the code coverage reports before (first image) and after this refactor (second). With the introduction of the service interface and repository interface, we can have more tests cases that catch the edge conditions. Like CCLF file not found, verifying we're handling suppressed MBIs correctly.


beforeRefactor


afterRefactor

bcda/models/models.go
```diff
@@ -29,6 +28,24 @@ const BCDA_FHIR_MAX_RECORDS_EOB_DEFAULT = 200
...
 const BCDA_FHIR_MAX_RECORDS_COVERAGE_DEFAULT = 4000
 
+// NOTE: This should be temporary, we should get to the point where this file only contains data models. Once that happens,
+// we no longer have the need for the data models tor produce other data models and we can remove reference to the service.
+var (
+	serviceInstance Service // Singleton service instance
+	once            sync.Once
+)
+
+// GetService returns the singleton instance of Service. It creates the service if it has not been created before.
```
msnook, line 38: Curious how this works in a multi-threaded env, like we have in AWS.

bcda/models/models_test.go
```diff
@@ -833,46 +740,47 @@ func (s *ModelsTestSuite) TestGetMaxBeneCount() {
...
-	acoUUID = uuid.Parse(constants.LargeACOUUID)
-	err = s.db.Find(&largeACO, "UUID = ?", acoUUID).Error
-	assert.Nil(err)
-	beneficiaries, err = largeACO.GetBeneficiaries(true)
-	assert.Nil(err)
-	assert.NotNil(beneficiaries)
-	assert.Equal(100, len(beneficiaries))
-
-}
+// TODO: Implement these tests in our service_test.go
```
msnook, line 743: Lots of commented out tests... have they been moved elsehwere?

mtrang1263: Whoops, forgot to remove these tests. Yes they've all been implemented in service_test.go

bcda/models/models.go
mtrang1263: sync.Once wraps the exclusive lock (and I think an atomic bool?) to ensure it only runs once. So we should be safe here.

bcda/models/models_test.go
```diff
@@ -833,280 +742,6 @@ func (s *ModelsTestSuite) TestGetMaxBeneCount() {
 	assert.EqualError(err, "invalid request type")
 }
 
-func (s *ModelsTestSuite) TestGetBeneficiaries() {
```
mtrang1263, line 836: All of the GetBeneficiary tests are now replaced by tests in service_test.go and repository_test.go.

These tests were verifying three main things.

We can get beneficiary data successfully -> replaced by service_test.go (and updated models_test.go)
We do not retrieve beneficiary data for CCLF files that are not marked as "Complete"
- We replaced this test by using mocks in both service_test.go and repository_test.go
- For service_test.go, we mocked out the repository call go GetCCLFFile such that it only will respond to the call iff the import status of the CCLF is complete (constants.ImportComplete). So if we happened to be using the incorrect status in our query (ex: constants.ImportFail), we would've caught in the service_test because the expectations of the mock were not satisifed.
- For repository.go, we validated that we have the import_status = ? clause in our query for CCLF files.
We do not retrieve beneficiaries that are in the suppressions list. The same approach was done as 2. Combination of service_test and repository_test.go
- For service test, we validated that we are passing in the result of GetSuppressedMBIs into the repository query for GetCCLFBeneficiaries call (by asserting that the mock was called with the correct MBIs). This means that we're passing in the expected suppressed MBIs to the repository call.
- For the repository test, we verify that we have the following condition present iff the caller supplies suppressed (ignored) MBIs. AND ("cclf_beneficiaries"."mbi" NOT IN

bcda/models/mock_repository_test.go
```diff
@@ -0,0 +1,110 @@
+// Code generated by mockery v0.0.0-dev. DO NOT EDIT.
```
msnook, line 1: This is auto-generated?  Should we provide instructions somewhere in our `README` to inform developers how to generate these files, in order to ensure satisfactory maintainability?

bcda/web/api.go
```diff
@@ -31,12 +33,27 @@ import (
...
 	groupAll = "all"
 )
 
+func init() {
+	// Ensure that models.go is properly initialized with the service reference.
+	// As we refactor more of the code, we should be able to remove the initialization
+	// from models.go
+	cutoffDuration := time.Duration(utils.GetEnvInt("CCLF_CUTOFF_DATE_DAYS", 45)*24) * time.Hour
+	db := database.GetGORMDbConnection()
+	db.DB().SetMaxOpenConns(utils.GetEnvInt("BCDA_DB_MAX_OPEN_CONNS", 25))
```
msnook, line 50: Why 25, 25,  and. 5 for these values?

bcda/models/mock_repository_test.go
mtrang1263: Great idea, let update the readme with this information

bcda/web/api.go
mtrang1263: I was looking at some posts about perf benefits and came across this one: https://www.alexedwards.net/blog/configuring-sqldb that seemed to match our size of reqeusts.

PR #521, opened 2020-07-28
BCDA-2751: Optimize Beneficiary De-Dup Query to Use a Subquery

msnook:
### Fixes [BCDA-2751](https://jira.cms.gov/browse/BCDA-2751)

This is a follow-up to the first iteration of de-duplicating beneficiaries that might show up in an ACO's attribution list.  The first iteration can be found here: https://github.com/CMSgov/bcda-app/pull/506

In this follow-up iteration, the de-dup query has been optimized to use a sub-query.  This ensures that a large array is not used in the SQL computation, which could result in the following Postgres error:
```
got xxxxxx parameters but PostgreSQL only supports 65535 parameters
```

### Proposed Changes

Details on proposed guidance acquired from [here](https://stackoverflow.com/questions/52712022/what-is-the-most-performant-way-to-rewrite-a-large-in-clause).

### Change Details

Updated de-dup logic to use `id IN (subquery)` instead of `id IN (value [, ...])`

### Security Implications

The updated/refactored query handles PHI/PII data.  It has been tested thoroughly through automated regression tests and manual scenarios.

### Acceptance Validation

- All existing and un-modified unit/integration tests pass, which means this query refactoring has not regressed.
- Locally, I created a scenario where an ACO had >65535 beneficiaries, and was able to use the code in its original state (prior to implementing this change) to result in the above mentioned Postgres error when a bulk request was made.  Then, I applied the changes from this PR and confirmed that a bulk request can be made successfully.

### Feedback Requested

Please review.

### DIFF
--- a/bcda/models/models.go
+++ b/bcda/models/models.go
@@ -286,7 +286,6 @@ type ACO struct {
 // the first array contains all NEW beneficaries that were added to CCLF since the date supplied
 // the second array contains all EXISTING benficiaries that have existed in CCLF since prior to the date supplied
 func (aco *ACO) GetNewAndExistingBeneficiaries(includeSuppressed bool, since string) (newBeneficiaries, beneficiaries []CCLFBeneficiary, err error) {
-	var cclfBeneficiariesOld []string
 
 	if aco.CMSID == nil {
 		log.Errorf("No CMSID set for ACO: %s", aco.UUID)
@@ -310,7 +309,8 @@ func (aco *ACO) GetNewAndExistingBeneficiaries(includeSuppressed bool, since str
 	// Get all the beneficiaries from the older CCLF for this ACO, to use as a basis for comparison
 	// If the older CCLF was not found, then all beneficiaries will be considered new
 	if foundCclfFileOld {
-		err = db.Table("cclf_beneficiaries").Where("file_id = ?", cclfFileOld.ID).Pluck("mbi", &cclfBeneficiariesOld).Error
+		cclfBeneficiariesOldQuery := fmt.Sprintf("SELECT mbi FROM cclf_beneficiaries WHERE file_id = %d", cclfFileOld.ID)
+		err = db.Raw(cclfBeneficiariesOldQuery).Error
 		if err != nil {
 			log.Errorf("Error retrieving beneficiaries from previous CCLF8 for ACO ID %s: %s", aco.UUID.String(), err.Error())
 			return nil, nil, err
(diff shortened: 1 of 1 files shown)


bcda/models/models.go
```diff
@@ -322,14 +322,13 @@ func (aco *ACO) GetNewAndExistingBeneficiaries(includeSuppressed bool, since str
...
 
 		// this is used to get unique ids for de-duplicating MBIs that are listed multiple times in the CCLF8 file
-		var uniqueIds []int64
-		db.Raw("SELECT id FROM ( SELECT max(id) as id, mbi FROM cclf_beneficiaries WHERE file_id = ? GROUP BY mbi ) as id", cclfFileNew.ID).Pluck("id", &uniqueIds)
+		uniqueIdQuery := fmt.Sprintf("SELECT id FROM ( SELECT max(id) as id, mbi FROM cclf_beneficiaries where file_id = %d GROUP BY mbi ) as id", cclfFileNew.ID)
 
 		// Populate new beneficiaries collection
 		if suppressedMBIs != nil {
-			err = db.Not("mbi", suppressedMBIs).Not("mbi", cclfBeneficiariesOld).Where("id IN (?)", uniqueIds).Find(&newBeneficiaries, "file_id = ?", cclfFileNew.ID).Error
+			err = db.Not("mbi", suppressedMBIs).Not("mbi", cclfBeneficiariesOld).Where("id IN (?)", db.Raw(uniqueIdQuery).SubQuery()).Find(&newBeneficiaries, "file_id = ?", cclfFileNew.ID).Error
```
mtrang1263, line 329: Should we also consider adding this protection in the following spots?

1. Not("mbi", suppressedMBIs)
2. Not("mbi", cclfBeneficiariesOld)
3. Where("mbi IN (?)", cclfBeneficiariesOld)

msnook: Absolutely necessary.   Great catch @mtrang1263 !

msnook: The `suppressedMBIs` arrays are bound here, so I don't think there is a pressing need to refactor those.  If you think strongly, we can add a separate ticket to address.  
I definitely agree that the `cclfBeneficiariesOld` arrays should be refactored to use a SubQuery.  Great catch @mtrang1263 !  I have updated in latest commit.

bcda/models/models.go
```diff
@@ -310,7 +309,8 @@ func (aco *ACO) GetNewAndExistingBeneficiaries(includeSuppressed bool, since str
 	// Get all the beneficiaries from the older CCLF for this ACO, to use as a basis for comparison
 	// If the older CCLF was not found, then all beneficiaries will be considered new
 	if foundCclfFileOld {
-		err = db.Table("cclf_beneficiaries").Where("file_id = ?", cclfFileOld.ID).Pluck("mbi", &cclfBeneficiariesOld).Error
+		cclfBeneficiariesOldQuery := fmt.Sprintf("SELECT mbi FROM cclf_beneficiaries WHERE file_id = %d", cclfFileOld.ID)
+		err = db.Raw(cclfBeneficiariesOldQuery).Error
```
mtrang1263, line 313: We don't need this query, right? We're just dropping the results and requerying at lines 329 and 331

msnook: I was thinking we could "error fast", like we did previously, so as not to affect existing flow.

mtrang1263: Gotcha, I can see the benefit of the early exit.

PR #523, opened 2020-07-29
BCDA-3366 FHIR compliance scan uses original BDT tool

orenfromberg:
### Fixes [BCDA-3366](https://jira.cms.gov/browse/BCDA-3366)

This branch moves us to using the original BDT tool instead of having to maintain our own fork.

### Proposed Changes

Adding a subdirectory called `bdt` that includes assets that are used to build the docker image for the tool.

### Change Details

* Using the original BDT tool [here](https://github.com/smart-on-fhir/bdt/)
* When the docker image is built, assets are copied from the `bdt` directory that contain the configuration and script for running the tool

### Security Implications

not a new software dependency because we were dependent on it indirectly, and we are removing our dependency on the fork that we were maintaining [here](https://github.com/orenfromberg/bdt-bcda).

### Acceptance Validation
- [x] tested with sandbox and dev environments.

### Feedback Requested

All feedback is welcome. Do we want to modify the way the tool is executed so that it is not excluding any tests?


Dockerfiles/Dockerfile.bdt
```diff
@@ -2,10 +2,13 @@ FROM node:current-slim
 
 WORKDIR '/'
 
-RUN apt update -y && apt install curl jq git -y
```
mtrang1263, line 5: Would be cool to see what dependencies/tools are available in a base image!

msnook: Yup!  The base image optimization strategy is going to be fun!

PR #524, opened 2020-07-31
BCDA-3371 - Add paging support on bluebutton client

mtrang1263:
### Fixes [BCDA-3371](https://jira.cms.gov/browse/BCDA-3371)

In order to handle data requests for beneficiaries with large claims data, we need to be able to chunk up the request into smaller, more predictable sizes. To do this, we'll want to leverage bluebutton's paging support

### Proposed Changes

Introduce client that can handle paging requests. This client must be able to fully consume the resource so that the bundle resource is fully hydrated and equivalent response if the caller had made a single non-paged request.

### Change Details

1. Introduce FHIR Client that can handle page requests. NOTE: Paging requests is turned off by default.
2. Refactored bluebutton client to return concrete types (Bundle) instead of string objects. This changed was needed to incrementally add new Entries to the bundle response when using paging. 

### Security Implications

1. github.com/cenkalti/backoff - Adding open source backoff library. This library is a bit more feature rich than our current approach.

### Acceptance Validation
CI Passes
Deploy to dev with paging enabled and verify everything works as expected.


bcda/client/bluebutton_test.go
```diff
@@ -2,8 +2,8 @@ package client_test
 
 import (
```
mtrang1263, line 3: This test file changed because our client now returns a Bundle response. Refactored the tests so that they validate the response from the server.

Since we cannot validate the path in these tests, updated `TestValidateRequest` to verify the query parameters are set as expected

bcda/client/testdata/Patient.json
```diff
@@ -0,0 +1,62 @@
+{
```
msnook, line 1: Are these new test files redundant/duplicates of the test files located here: https://github.com/CMSgov/bcda-app/tree/master/shared_files/synthetic_beneficiary_data

mtrang1263: Yup they are, thinking more about it. Doesn't really buy us anything to keep the `testdata` directory. I'll go ahead and remove it.

mtrang1263:
> Looks good! Just a question/comment regarding the new test file data.
> Also, should we have a ticket in the backlog for turning this on once we have done further testing with BFD?

Yup! Great idea, let me go ahead and create the turn/test ticket for prod work.

PR #527, opened 2020-08-04
 BCDA-3304 - Remove unnecessary dependency between queue-db and documentation

mtrang1263:
### Fixes [BCDA-3304](https://jira.cms.gov/browse/BCDA-3304)

Removes a failure point around the documentation Docker container.

```
ERROR: for documentation  You cannot remove a running container DOCUMENTATION_ID. Stop the container before attempting removal or force remove
```

This was a result of each Make (lint, package, unit-test) command re-creating the documentation container.

### Proposed Changes

Update docker-compose files to allow for concurrent execution of multiple stages without needing to sleep between stages.

### Change Details

1. Remove unneeded dependency between the queue-db container and the documentation container.

### Security Implications

### Acceptance Validation

Successfully ran build & deploy job

Screen Shot 2020-08-04 at 3 06 46 PM

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,8 +10,6 @@ services:
       - "5433:5432"
     volumes:
       - ./db/worker.sql:/docker-entrypoint-initdb.d/schema.sql
-    depends_on:
-      - documentation
   db:
     image: postgres
     environment:
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ package:
 LINT_TIMEOUT ?= 3m
 lint:
 	docker-compose -f docker-compose.test.yml build tests
-	docker-compose -f docker-compose.test.yml run --rm tests golangci-lint run --deadline=$(LINT_TIMEOUT) --verbose --concurrency=1
+	docker-compose -f docker-compose.test.yml run --rm tests golangci-lint run --deadline=$(LINT_TIMEOUT) --verbose
 	docker-compose -f docker-compose.test.yml run --rm tests gosec ./...
 
 # The following vars are available to tests needing SSAS admin credentials; currently they are used in smoke-test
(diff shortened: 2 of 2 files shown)


docker-compose.yml
```diff
@@ -10,8 +10,6 @@ services:
       - "5433:5432"
     volumes:
       - ./db/worker.sql:/docker-entrypoint-initdb.d/schema.sql
-    depends_on:
```
mtrang1263, line 13: Only call explicitly made to the documentation container is done in the `make package`

mtrang1263:
> I think we need a dependency on `documentation` somewhere, right? Otherwise, we run the risk of bringing up the stack with Swagger unavailable.

I was thinking about updating `api` and listing documentation as a dependency of it, but we would still arrive at the same problem (concurrent Make commands are going to conflict).

Then I started thinking about what are the actual requirements of the services in our docker-compose file. I don't think we need to have swagger available to use api, worker, db, or queue. It's a nice side effect that when we run `docker-compose up api`, we're also building and updating the swagger docs, but I wonder if it's better to put these layers inside the Makefile instead of the docker-compose file.

So maybe we have a `make api` which runs `docker-compose build documentation && docker-compose build api && docker-compose up -d api`?

msnook:
> > I think we need a dependency on `documentation` somewhere, right? Otherwise, we run the risk of bringing up the stack with Swagger unavailable.
> 
> I was thinking about updating `api` and listing documentation as a dependency of it, but we would still arrive at the same problem (concurrent Make commands are going to conflict).
> 
> Then I started thinking about what are the actual requirements of the services in our docker-compose file. I don't think we need to have swagger available to use api, worker, db, or queue. It's a nice side effect that when we run `docker-compose up api`, we're also building and updating the swagger docs, but I wonder if it's better to put these layers inside the Makefile instead of the docker-compose file.
> 
> So maybe we have a `make api` which runs `docker-compose build documentation && docker-compose build api && docker-compose up -d api`?

Good idea.  Maybe we don't need it after all, in order to de-couple the services as you suggested.  After all, a simple `docker-compose up -d` will bring all services up anyway... no need to tie `documentation` service to any of the other elements.   Approving as-is.

PR #530, opened 2020-08-12
BCDA-3428 - Adding timing metrics to CCLF Ingest process

mtrang1263:
### Implements [BCDA-3428](https://jira.cms.gov/browse/BCDA-3428)

To get better insights into how we can improve our CCLF file ingest performance, we need to time the various methods. 

### Change Details

Adding tracing to the following calls:

1. sortCCLFFiles
2. ImportCCLF0
3. ValidateCCLF8
4. ImportCCLF8
5. CleanupCCLFFiles

### Security Implications

Brought in `github.com/newrelic/go-agent/_integrations/nrlogrus`. Only used in test scope to inject a logger to validate our usages of NewRelic transaction/segments.

### Acceptance Validation
1. CI passes
2. Able to ingest CCLF data.
3. Timing metrics reported to NewRelic.

Screen Shot 2020-08-13 at 4 56 03 PM
Screen Shot 2020-08-13 at 4 55 15 PM

### DIFF
--- a/bcda/cclf/metrics/metrics.go
+++ b/bcda/cclf/metrics/metrics.go
@@ -4,6 +4,9 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"time"
+
+	"github.com/CMSgov/bcda-app/bcda/utils"
 
 	newrelic "github.com/newrelic/go-agent"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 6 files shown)


bcda/cclf/metrics/metrics.go
```diff
@@ -0,0 +1,96 @@
+package metrics
```
msnook, line 1: Interesting.  I wonder if we'd get the same benefit, with less code, if we used the New Relic transaction monitoring?  Is something like this useful? https://github.com/CMSgov/bcda-app/pull/111

mtrang1263: Unfortunately, the slick monitoring that we get for free from NewRelic really times RPC calls (HTTP, DB calls). To do in-process timing, like we're doing for CCLF ingestion, we need to orchestrate the transactions/segments.

We definitely should use this approach when timing the Postgres calls! If we can upgrade to 3.x, they have out of the box monitoring for Postgres, like we're using for HTTP tracing.

msnook: Gotcha.  I'd love to see, as part of the `Acceptance Validation`, some results you've pulled out of New Relic after deploying this feature branch to `dev` and running the smoke test.

mtrang1263: Yup, will do!

orenfromberg: @mtrang1263 Sorry for being pedantic, when you said "orchestrate", did you mean "instrument"?

PR #532, opened 2020-08-14
BCDA-3428 - Adding timing metrics for cleanupCCLF

mtrang1263:
### Proposed Changes

1. Need to emit timing metrics for cleaning up CCLF files.

### Change Details

1. Create a new parent timer to measure the time it takes to clean up all CCLF files. This was failing before because the supplied context did not have a timer associated with it.

### Security Implications

### Acceptance Validation
Validated that cleanupCCLF metrics are appearing on NewRelic

Screen Shot 2020-08-14 at 2 37 23 PM
Screen Shot 2020-08-14 at 2 36 59 PM

### DIFF
--- a/bcda/cclf/cclf.go
+++ b/bcda/cclf/cclf.go
@@ -427,8 +427,11 @@ func ImportCCLFDirectory(filePath string) (success, failure, skipped int, err er
 		}()
 	}
 
-	err = cleanUpCCLF(ctx, cclfMap)
-	if err != nil {
+	if err = func() error {
+		ctx, c := metrics.NewParent(ctx, "ImportCCLFDirectory#cleanupCCLF")
+		defer c()
+		return cleanUpCCLF(ctx, cclfMap)
+	}(); err != nil {
 		log.Error(err)
 	}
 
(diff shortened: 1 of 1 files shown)


bcda/cclf/cclf.go
```diff
@@ -651,14 +654,11 @@ func validateCCLFFolderName(folderName string) error {
...
 	errCount := 0
-	close := metrics.NewChild(ctx, "cleanUpAllCCLF")
-	defer close()
-
 	for _, perfYearCCLFFileList := range cclfMap {
 		for _, cclfFileList := range perfYearCCLFFileList {
 			for _, cclf := range cclfFileList {
 				func() {
-					close := metrics.NewChild(ctx, "cleanUpIndividualCCLF")
+					close := metrics.NewChild(ctx, fmt.Sprintf("cleanUpCCLF%d", cclf.cclfNum))
```
msnook, line 661: The CCLF num is static across ACOs.  I wonder if it makes sense to instead incorporate the ACO ID for the specific CCLF here?  Or, is that already being captured somewhere?
`cclf.acoID`

mtrang1263: I think for a metric name, it'd make sense to use more generic fields (like CCLF numbers).

What's not captured here, and we may want to explore, is adding custom tags to the metrics. I think if we add tags to segments/transactions, we'd be able to see what specific ACO is associated with a slower transaction, while still being able to use generic aggregation.

msnook: Makes sense!

PR #534, opened 2020-08-18
BCDA-3413 Added authentication client version to /_auth endpoint 

PatKelsh:
### Fixes [BCDA-3413](https://jira.cms.gov/browse/BCDA-xxx)

We want to be able to quickly determine which version of the SSAS or other authentication resource is deployed to a given environment

### Proposed Changes

* Added additional values to `/_auth` endpoint which will display the version of the authentication server if that information is available

### Change Details

* Added version information to the endpoint `/_auth`
* Added new `GetVersion` function to the provider interface
* Implemented `GetVersion` function for SSAS client class

### Security Implications

### Acceptance Validation

Deployed to dev
Screen Shot 2020-08-18 at 3 57 54 PM


bcda/auth/client/ssas.go
```diff
@@ -378,3 +378,33 @@ func (c *SSASClient) setAuthHeader(req *http.Request) error {
...
+	if err := c.setAuthHeader(req); err != nil {
+		return "", err
+	}
+	resp, err := c.Do(req)
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+
+	ssasLogger.Info(resp.Body)
```
mtrang1263, line 398: Curious about this line, do need to log the response body for these requests at the INFO level? Maybe DEBUG or not logging?

And I had thought we could not read the response body multiple times. I believe that this is working now since we're actually stringifying the io.Reader (and not necessarily the contents).

Here's the log output:

```
{"file":"/Users/martintrang/go/src/github.com/CMSgov/bcda-app/bcda/auth/client/ssas.go:398","func":"github.com/CMSgov/bcda-app/bcda/auth/client.(*SSASClient).GetVersion","level":"info","msg":"\u0026{0x1142570 0xc000394040 0x14970e0}","time":"2020-08-18T09:24:42-06:00"}
```

bcda/auth/client/ssas.go
```diff
@@ -378,3 +378,33 @@ func (c *SSASClient) setAuthHeader(req *http.Request) error {
...
+	type SSASVersion struct {
```
mtrang1263, line 399: Nit - can we switch this to be lowercase to make in explicit that this field is not exported/public?

bcda/web/api.go
```diff
@@ -594,6 +594,10 @@ func healthCheck(w http.ResponseWriter, r *http.Request) {
 func getAuthInfo(w http.ResponseWriter, r *http.Request) {
 	respMap := make(map[string]string)
 	respMap["auth_provider"] = auth.GetProviderName()
+	version, err := auth.GetProvider().GetVersion()
+	if err != nil {
```
mtrang1263, line 598: Should this condition be `err == nil`? Or we can do an early return like what's on line 603.

bcda/auth/client/ssas_test.go
```diff
@@ -277,6 +278,54 @@ func (s *SSASClientTestSuite) TestGetToken() {
 	assert.Equal(s.T(), tokenString, string(respKey))
 }
 
+func (s *SSASClientTestSuite) TestGetVersionPassing() {
```
mtrang1263, line 281: I wonder if this can be written as a table driven test.

Your test struct could have the following:

- Response to return
- Expected version
- Expected error

That way the server set up and env var stuff is taken care of for you.

bcda/auth/client/ssas.go
```diff
@@ -378,3 +378,33 @@ func (c *SSASClient) setAuthHeader(req *http.Request) error {
...
+func (c *SSASClient) GetVersion() (string, error) {
+	req, err := http.NewRequest("GET", fmt.Sprintf("%s/_version", c.baseURL), nil)
+	if err != nil {
+		return "", err
+	}
+
+	if err := c.setAuthHeader(req); err != nil {
+		return "", err
+	}
+	resp, err := c.Do(req)
```
mtrang1263, line 392: Should we add some checks on the response code as well? If we get a 4xx or 5xx, we return that error code up?

bcda/web/api.go
```diff
@@ -595,7 +595,7 @@ func getAuthInfo(w http.ResponseWriter, r *http.Request) {
 	respMap := make(map[string]string)
 	respMap["auth_provider"] = auth.GetProviderName()
 	version, err := auth.GetProvider().GetVersion()
```
mtrang1263, line 597: Should we be returning an error if we fail to get the auth provider?

bcda/auth/client/ssas.go
```diff
@@ -393,18 +393,20 @@ func (c *SSASClient) GetVersion() (string, error) {
 	if err != nil {
 		return "", err
 	}
+	if resp.StatusCode != 200 {
+		return "", errors.New("SSAS server failed to return version ")
+	}
 	defer resp.Body.Close()
```
mtrang1263, line 399: We should move this defer ahead of the resp.StatusCode check. We'll want to ensure that response body is always closed.

bcda/web/api.go
```diff
@@ -594,6 +594,12 @@ func healthCheck(w http.ResponseWriter, r *http.Request) {
 func getAuthInfo(w http.ResponseWriter, r *http.Request) {
 	respMap := make(map[string]string)
 	respMap["auth_provider"] = auth.GetProviderName()
+	version, err := auth.GetProvider().GetVersion()
+	if err == nil {
+		respMap["version"] = version
+	} else {
+		respMap["error message"] = err.Error()
```
mtrang1263, line 601: Sorry Pat, I was more thinking of doing something like the following:

```
http.Error(w, err.Error(), http.StatusInternalServerError)
```

What do you think? Might be more clear to return 500 over returning 200 with the error in the body.
```

PatKelsh: @mtrang1263 i'm not sure. this is at the end of the day a status page and users are gonna want to know what the set  "auth_provider" is. i don't think throwing a 500 error for this is the right move

mtrang1263: Good point! Better to render something for the status page versus an empty page/empty response.

PR #535, opened 2020-08-20
BCDA 3520 Maintenance: Update Golang to 1.15

jeffcorpuz:
### Fixes [BCDA-3520](https://jira.cms.gov/browse/BCDA-3520)

- New version of Go has been released : https://blog.golang.org/go1.15
- Need to update Dockerfiles and travis.yml to account for this

### Proposed Changes

- Update Dockerfiles, travis.yml to use latest and greatest Go

### Security Implications


.travis.yml
```diff
@@ -11,7 +11,7 @@ language: go
 # updates. Any older versions be considered deprecated. Don't bother testing
 # with them.
 go:
-  - 1.14.1
+  - 1.15.0
```
msnook, line 14: Should this be `1.15` instead of `1.15.0`?

jeffcorpuz: yes, why did I think Travis was smart enough to handle this but actually you'd have to declare it as "1.15" not 1.15 ... 

example:
```Note that, in order to choose Go 1.10, you must use go: "1.10" (a string), not go: 1.10 (a float). Using a float results in the use of Go 1.1.
```

PR #538, opened 2020-08-25
BCDA-3429 - Use COPY FROM to bulk load CCLF beneficary data

mtrang1263:
### Fixes [BCDA-3429](https://jira.cms.gov/browse/BCDA-3429)

We want to improve CCLF ingestion process by optimizing how we insert data into our Postgres instance.

### Proposed Changes

Change ingest pattern from single INSERT call per beneficiary to use COPY FROM that allows us to stream all of the beneficiary records.

### Change Details

When using single INSERT per beneficiary, we're paying overhead on opening and committing the transaction for every single INSERT call.

Postgres documentation recommends using COPY FROM statement to perform this bulk loading. With this approach, we open a single transaction to encompass the entire CCLF ingest process for a given CCLF file. Once all beneficiaries have been inserted, we'll commit the transaction. This process is repeated for every CCLF file that needs to be ingested.

The transaction encompasses several statements. Each statement will be capped at 100k beneficiaries (by default). Even though this should affect Postgres (it should be able to handle millions of insert calls in the statement), we wanted to provide an upper bound to avoid issues with large data sets. Once a statement hits the 100k count, we'll flush the statement and recreate a new statement to contain the remaining beneficiaries.

Since the entire CCLF file ingest is captured within a single transaction, we will no longer have partial beneficiary data inserted into our database. It'll be all or nothing.

Implementation based on: https://pkg.go.dev/github.com/lib/pq?tab=doc#hdr-Bulk_imports

### Security Implications

### Acceptance Validation

1. Added unit tests to verify that statements are committed once the maximum is reached.
2. Run CCLF ingest on dev to confirm functionality.

Screen Shot 2020-08-25 at 11 08 50 AM

3. Verified beneficiary counts are equivalent using `INSERT ...` and `COPY FROM...`

### DIFF
--- a/bcda/cclf/cclf.go
+++ b/bcda/cclf/cclf.go
@@ -203,7 +203,6 @@ func importCCLF(ctx context.Context, fileMetadata *cclfFileMetadata, importer im
 
 	db := database.GetGORMDbConnection()
 	defer database.Close(db)
-	db.LogMode(true)
 
 	err = db.Create(&cclfFile).Error
 	if err != nil {
--- a/bcda/cclf/cclf.go
+++ b/bcda/cclf/cclf.go
@@ -214,7 +214,7 @@ func importCCLF(ctx context.Context, fileMetadata *cclfFileMetadata, importer im
 
 	fileMetadata.fileID = cclfFile.ID
 
-	importStatusInterval := utils.GetEnvInt("CCLF_IMPORT_STATUS_RECORDS_INTERVAL", 1000)
+	importStatusInterval := utils.GetEnvInt("CCLF_IMPORT_STATUS_RECORDS_INTERVAL", 10000)
 	importedCount := 0
 	var rawFile *zip.File
 
--- a/bcda/cclf/cclf.go
+++ b/bcda/cclf/cclf.go
@@ -149,7 +149,7 @@ func importCCLF0(ctx context.Context, fileMetadata *cclfFileMetadata) (map[strin
 func importCCLF8(ctx context.Context, fileMetadata *cclfFileMetadata) error {
 	importer := &cclf8Importer{
 		logger:            log.StandardLogger(),
-		maxPendingQueries: utils.GetEnvInt("STATEMENT_EXEC_COUNT", 100000),
+		maxPendingQueries: utils.GetEnvInt("STATEMENT_EXEC_COUNT", 200000),
 	}
 
 	err := importCCLF(ctx, fileMetadata, importer)
--- a/bcda/cclf/cclf.go
+++ b/bcda/cclf/cclf.go
@@ -13,7 +13,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/jinzhu/gorm"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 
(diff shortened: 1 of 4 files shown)


msnook:
Does this awesome performance enhancement invalidate this ticket: https://jira.cms.gov/browse/BCDA-3553
With the performance being <30 seconds (vs ~7 mins), I'd like to see us exercise CCLF Ingestion on each test for awhile.

bcda/cclf/cclf.go
```diff
@@ -148,27 +147,12 @@ func importCCLF0(ctx context.Context, fileMetadata *cclfFileMetadata) (map[strin
...
-			fmt.Println("Could not create CCLF8 beneficiary record.")
-			err = errors.Wrap(err, "could not create CCLF8 beneficiary record")
-			log.Error(err)
-			return err
-		}
-		return nil
-	})
+	importer := &cclf8Importer{
+		logger:            log.StandardLogger(),
+		maxPendingQueries: utils.GetEnvInt("STATEMENT_EXEC_COUNT", 100000),
```
msnook, line 152: Can we bump this to 200,000?  This will ensure that all current CCLFs get loaded in one shot; and if we run into problems, we can tweak this value.

mtrang1263: > Does this awesome performance enhancement invalidate this ticket: https://jira.cms.gov/browse/BCDA-3553
> With the performance being <30 seconds (vs ~7 mins), I'd like to see us exercise CCLF Ingestion on each test for awhile.

I think it's a great idea to defer the smoke test changes for a little while at least. I still think we'll get some good benefits from reducing the scope of the ingest work. Definitely agreed that it's valuable since we've made pretty drastic changes to the CCLF ingest process and we should leave it in place for now.

mtrang1263:
> Does this awesome performance enhancement invalidate this ticket: https://jira.cms.gov/browse/BCDA-3553
> With the performance being <30 seconds (vs ~7 mins), I'd like to see us exercise CCLF Ingestion on each test for awhile.

I think it's a great idea to defer the smoke test changes for a little while at least. I still think we'll get some good benefits from reducing the scope of the ingest work. Definitely agreed that it's valuable since we've made pretty drastic changes to the CCLF ingest process and we should leave it in place for now.

bcda/cclf/cclf.go
```diff
@@ -258,26 +242,48 @@ func importCCLF(ctx context.Context, fileMetadata *cclfFileMetadata, importFunc
...
+			log.Error(err)
+			return err
 		}
-	}
 
-	successMsg := fmt.Sprintf("Successfully imported %d records from CCLF%d file %s.", importedCount, fileMetadata.cclfNum, fileMetadata)
-	fmt.Println(successMsg)
-	log.Infof(successMsg)
+		importedCount++
+		if importedCount%importStatusInterval == 0 {
```
msnook, line 283: Is this still needed, now that we've significantly increased the speed of ingestion?  If we do think its needed, should we increase the `importStatusInterval` to maybe 5000 or 10000?

mtrang1263: :+1: - I'll bump it up to 5k for now. I think it's nice signal to give us an idea that something is going on the instance.

bcda/cclf/importer_test.go
```diff
@@ -0,0 +1,209 @@
...
+		logger:            logrus.New(),
+		maxPendingQueries: 10,
+	}
+	assert.NoError(s.T(), importer.flush(context.Background()))
+}
+
+func getBeneficiary(fileID uint) *models.CCLFBeneficiary {
+	return &models.CCLFBeneficiary{
+		FileID: fileID,
+		// We expect 11 byte sfor HICN and MBI - we'll ensure that each of the values are AT LEAST 11 bytes
```
msnook, line 205: Minor typo - `sfor`

PR #539, opened 2020-08-25
BCDA-3510 - Only create job iff all of its dependent information was populated

mtrang1263:
### Fixes [BCDA-3510](https://jira.cms.gov/browse/BCDA-3510)

During the job creation process, we can encounter situations that cause a particular job to get stuck in the `Pending` or `In Progress` state.

For example, a job is created that depends on 10 queuejobs. We update our Job details (in the DB) to expect 10 queuejobs to be complete. After updating the job, we successfully insert 9 queuejobs, but fail to insert the last queuejob. The created job would process 9 jobs and be stuck indefinitely in the `In Progress` state.

Because of this edge case, we need to ensure that jobs are only created once all of its dependent information is present. This includes having all of the queuejobs successfully added.

Since we're adding queuejobs prior to adding the actual parent job, we can encounter a situation where the queue worker fails to find its parent. We've added some conditional logic to handle this situation. It will allow for transient errors (we picked up the queuejob before we've been able to insert the parent job) as well as the situation where we've added queuejobs and failed to add its parent job.

Since the queuejobs and job may exist in different databases, we cannot use a transaction to encompass all of these inserts.
 
### Change Details
1. Change API implementation to add the parent Job **after** computing all of its fields (job count, transaction time) and **after** successfully adding the necessary queuejobs to the queue.
2. Updated worker logic to retry a missing parent Job a certain number of times (defaulted to 3) before it assumes that the queuejob is orphaned and removes itself from the queue.

### Security Implications

### Acceptance Validation

Updated unit tests to verify jobs are not created when we encounter issues during job initialization process.

### DIFF
--- a/bcda/web/api.go
+++ b/bcda/web/api.go
@@ -141,6 +141,22 @@ func bulkRequest(resourceTypes []string, w http.ResponseWriter, r *http.Request,
 		return
 	}
 
+	if qc == nil {
+		err = errors.New("queue client not initialized")
+		log.Error(err)
+		oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, responseutils.Processing, "")
+		responseutils.WriteError(oo, w, http.StatusInternalServerError)
+		return
+	}
+
+	bb, err := client.NewBlueButtonClient()
+	if err != nil {
+		log.Error(err)
+		oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, responseutils.Processing, "")
+		responseutils.WriteError(oo, w, http.StatusInternalServerError)
+		return
+	}
+
 	db := database.GetGORMDbConnection()
 	defer database.Close(db)
 
@@ -172,31 +188,7 @@ func bulkRequest(resourceTypes []string, w http.ResponseWriter, r *http.Request,
 		RequestURL: fmt.Sprintf("%s://%s%s", scheme, r.Host, r.URL),
 		Status:     "Pending",
 	}
-	if result := db.Save(&newJob); result.Error != nil {
-		log.Error(result.Error.Error())
-		oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, responseutils.DbErr, "")
-		responseutils.WriteError(oo, w, http.StatusInternalServerError)
-		return
-	}
 
-	// After successfully creating the job, we need to mark the job as FAILED if we had any other issues
-	// with populating data associated with the job.
-	defer func() {
-		if err != nil {
-			if err1 := db.Model(&newJob).Update("status", "Failed").Error; err1 != nil {
-				err1 = errors.Wrap(err1, "error when updating job status to \"Failed\"")
-				log.Error(err1)
-			}
-		}
-	}()
-
-	bb, err := client.NewBlueButtonClient()
-	if err != nil {
-		log.Error(err)
-		oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, responseutils.Processing, "")
-		responseutils.WriteError(oo, w, http.StatusInternalServerError)
-		return
-	}
 	// request a fake patient in order to acquire the bundle's lastUpdated metadata
 	b, err := bb.GetPatient("FAKE_PATIENT", strconv.FormatUint(uint64(newJob.ID), 10), acoID, "", time.Now())
 	if err != nil {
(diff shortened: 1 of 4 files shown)


bcda/web/api.go
```diff
@@ -179,6 +179,17 @@ func bulkRequest(resourceTypes []string, w http.ResponseWriter, r *http.Request,
 		return
 	}
 
+	// After successfully creating the job, we need to mark the job as FAILED if we had any other issues
+	// with populating data associated with the job.
+	defer func() {
+		if err != nil {
+			if err1 := db.Model(&newJob).Update("status", "Failed").Error; err1 != nil {
```
msnook, line 186: Should we retry here?  A job in `pending` state blocks ACOs from making any requests (resulting in `429` responses).
Also, does it make sense to create a ticket to monitor for jobs in `Pending` state?  I think this would be good insight... it could warn us of thinks like: (1) workers not keeping up with jobs in the queue, (2) jobs stuck in `Pending`, etc

mtrang1263: Ah totally forgot how bad this would be if the clean up fails (ACO will be blocked indefinitely). I was thinking it's a minor inconvenience and the caller can retry. 

Since it's that critical that we have a fully formed job (and that has several error scenarios like we noted). What are your thought about putting the job creation into a transaction? We can pre-compute quite a bit of the information before opening the transaction. And then we have guarantees that all jobs that are in the pending state are valid. That way we don't get into the situation where an ACO is blocked indefinitely.

mtrang1263: > Also, does it make sense to create a ticket to monitor for jobs in `Pending` state? I think this would be good insight... it could warn us of thinks like: (1) workers not keeping up with jobs in the queue, (2) jobs stuck in `Pending`, etc

Would our current JobQueueCount alert be sufficient to cover these scenarios? Assuming that all `Pending` jobs are valid.

msnook: Transaction makes sense to me, if you think its feasible/in-scope for this ticket.  And yes, good point about `JobQueueCount` alert... that should be sufficient here!

bcda/web/api_test.go
```diff
@@ -427,6 +427,11 @@ func bulkEOBRequestNoBeneficiariesInACOHelper(endpoint string, s *APITestSuite)
 	handler.ServeHTTP(s.rr, req)
 
 	assert.Equal(s.T(), http.StatusInternalServerError, s.rr.Code)
+
+	// NOTE: This relies on the fact that this caller was the last
```
msnook, line 431: I don't fully understand this new Note.  Can you elaborate?

mtrang1263: This might be moot if we end up including the job creation in a transaction, but I don't have any hooks on retrieving our proposed jobID, so I need to search for jobs based on ACOs.

If there's another test running in parallel in this one that happens to be inserting another job for the same ACO, this test may fail because we'll pick up the other job (not the one we had created) and that job may not be failed.

msnook: Ahhh, makes sense.

bcdaworker/main.go
```diff
@@ -86,9 +86,23 @@ func processJob(j *que.Job) error {
...
-		return errors.Wrap(err, "could not retrieve job from database")
+	result := db.First(&exportJob, "ID = ?", jobArgs.ID)
+
+	if result.RecordNotFound() {
+		// Based on the current backoff delay (j.ErrorCount^4 + 3 seconds), this should've given
+		// us plenty of headroom to ensure that the parent job will never be found.
+		maxNotFoundRetries := int32(utils.GetEnvInt("BCDA_WORKER_MAX_JOB_NOT_FOUND_RETRIES", 3))
+		if j.ErrorCount >= maxNotFoundRetries {
+			log.Errorf("No job found for ID: %d acoID: %s. Retries exhausted. Removing job from queue.", jobArgs.ID, jobArgs.ACOID)
+			return nil
```
mtrang1263, line 97: Comment that returning no error will remove this from the queue.

bcda/web/api.go
```diff
@@ -240,9 +274,6 @@ func bulkRequest(resourceTypes []string, w http.ResponseWriter, r *http.Request,
 			return
 		}
 	}
-
-	w.Header().Set("Content-Location", fmt.Sprintf("%s://%s/api/v1/jobs/%d", scheme, r.Host, newJob.ID))
```
mtrang1263, line 244: Needed to pull this into the defer call. If we encounter an issue on the commit, we can notify the caller of this failure.

PR #540, opened 2020-08-26
BCDA-3522 - Prioritize active ACOs during CCLF ingest process

mtrang1263:
### Fixes [BCDA-3522](https://jira.cms.gov/browse/BCDA-3522)

To improve the experience for onboarded ACOs, we want to prioritize CCLF ingestion for onboarded ACOs.

### Change Details

1. Query database for active ACOs. The active ACOs will have their CCLF ingest prioritized.
2. If the query fails or we do not have any active ACOs, we'll default to a random order.
3. Update our unit-test-db snapshot to have the data staged to verify our queries are working as expected.

### Security Implications

### Acceptance Validation

* Added unit tests to verify functionality.
* CI Passes.


msnook:
This is a great idea.  Should we have an accompanying PR in `bcda-ops` to remove the env var from the applicable environments?

bcda/cclf/cclf.go
```diff
@@ -559,6 +561,30 @@ func orderACOs(cclfMap *map[string]map[int][]*cclfFileMetadata) []string {
...
 
+func getPriorityACOs(db *gorm.DB) []string {
+	const query = `
+	SELECT trim(both '["]' from g.x_data::json->>'cms_ids') "aco_id" 
+	FROM systems s JOIN groups g ON s.group_id=g.group_id 
+	WHERE s.deleted_at IS NULL AND g.group_id IN (SELECT group_id FROM groups WHERE x_data LIKE '%A%' and x_data NOT LIKE '%A999%') AND
+	s.id IN (SELECT system_id FROM secrets WHERE deleted_at IS NULL);
+	`
+	
+	defaultPriority := strings.Split(os.Getenv("CCLF_PRIORITY_ACO_CMS_IDS"), ",")
```
msnook, line 572: Is this even necessary anymore if we are doing the check in real time?  Might be too many knobs having both capabilities.  I'd say we default to random order if the query fails (and log it so its in splunk).

mtrang1263: Good idea, I was thinking that a fallback would be nice, but agreed that we may have too many options at our controls. Will remove.

Since we're pulling out this value, I'll go ahead and put up a PR for the ops change as well.

mtrang1263:
`db/testing/docker-entrypoint-initdb.d/dump.pgdata` has been updated to include ACO data. It allows `TestImportCCLFDirectory_PriorityACOs` and `TestGetPriority` tests to still work.

PR #542, opened 2020-08-31
BCDA-3398 - Support running unit tests in parallel

mtrang1263:
### Fixes [BCDA-3398](https://jira.cms.gov/browse/BCDA-3398)

To improve developer efficiencies, we want to run our unit tests in parallel. We could not do this before due to 429 responses when submitting jobs.

### Change Details
1. Switch over api_test.go to use Small ACO. It was using Dev ACO, which can cause collisions with other tests.
2. Update our job update commands in bulkConcurrentRequestHelper to no longer assume that the latest created job was created by api_test.go. This assumption is no longer true since we're running tests in parallel.
3. Fix race found after running tests in a parallel.

### Security Implications

### Acceptance Validation

1. Ran unit tests locally to ensure that they pass reliably.
2. CI Passes.

### DIFF
--- a/bcda/client/bluebutton.go
+++ b/bcda/client/bluebutton.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"context"
 	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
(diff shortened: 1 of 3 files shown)


msnook:
This looks good. Since we're dealing with race-like conditions here, I wonder if we should run Build/Package a few times on Jenkins to see if we have a stable execution (with no race conditions).  To expedite, you can choose `Do Not Deploy` since all we care about is the unit test execution.

mtrang1263:
> This looks good. Since we're dealing with race-like conditions here, I wonder if we should run Build/Package a few times on Jenkins to see if we have a stable execution (with no race conditions). To expedite, you can choose `Do Not Deploy` since all we care about is the unit test execution.

Ran the build & package stages 3 times to confirm everything is OK. 

Screen Shot 2020-09-01 at 11 43 46 AM

PR #544, opened 2020-09-03
BCDA-3145|3398 - Fixing more failures around test parallelization

mtrang1263:
### Change Details

1. Remove global transaction from bcdaworker. Since multiple threads are calling processjobs, we need to ensure that we do not overwrite the transaction. This should remove some error logs we've seen on the worker logs.

```
{"file":"/go/src/github.com/CMSgov/bcda-app/bcdaworker/main.go:356","func":"main.fhirBundleToResourceNDJSON","level":"error","msg":"improper segment use: the Transaction must be used in a single goroutine and segments must be ended in \"last started first ended\" order: see https://github.com/newrelic/go-agent/blob/master/GUIDE.md#segments","time":"2020-09-02T11:56:46-04:00"}
```

2. Use different test data for cli_test#TestImportSuppressionDirectory. This ensures isolation from the suppression_test.go.

### Security Implications

### Acceptance Validation

Ran tests locally and several times on dev to resolve race condition and fix issue with suppression test.

### DIFF
--- a/bcdaworker/main.go
+++ b/bcdaworker/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bufio"
+	"context"
 	"database/sql"
 	"encoding/json"
 	"fmt"
@@ -30,8 +31,7 @@ import (
 )
 
 var (
-	qc  *que.Client
-	txn newrelic.Transaction
+	qc *que.Client
 )
 
 type jobEnqueueArgs struct {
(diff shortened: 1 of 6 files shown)


bcdaworker/main.go
```diff
@@ -477,6 +478,14 @@ func updateJobQueueCountCloudwatchMetric() {
 	}
 }
 
+func getSegment(ctx context.Context, name string) newrelic.Segment {
```
msnook, line 481: does this fix this issue?
https://jira.cms.gov/browse/BCDA-3145

mtrang1263: Yes it does! I didn't know that we had a ticket for this. Will update the PR with more details

msnook (approved):
👍

PR #545, opened 2020-09-03
BCDA-3589 - Consolidate JobEnqueueArgs definition

mtrang1263:
### Fixes [BCDA-3589](https://jira.cms.gov/browse/BCDA-3589)

There were two definitions of JobEnqueueArgs, in bcda/models.go and bcdaworker/main.go. These definitions were equivalent.

Since we were using JSON serialization AND the field name/types are equivalent, we do not need to worry about any data migration from the old definitions.

### Change Details

1. Allow models.go JobEnqueueArgs definition to be exported and used in other packages (including bcdaworker).
2. Remove the struct definition from the bcdaworker code.

### Security Implications

### Acceptance Validation

CI and unit tests are sufficient. Since the definition is not changed, we do not need to perform any data migration.


msnook (approved):
🎉

orenfromberg (approved):

PR #547, opened 2020-09-16
BCDA-3551 - Update performance tests commands to use clientID/clientSecret

mtrang1263:
### Fixes [BCDA-3551](https://jira.cms.gov/browse/BCDA-3551)

Performance tests were silently failing with authentication issues. Need to update auth mechanism and have failures to notify developers when issues arise with the test process.

### Change Details
1. Update CLI to accept clientID and clientSecret. These are used to retrieve a token.

### Security Implications

### Acceptance Validation
1. Ran positive test with successful requests being made to BCDA backend.
2. Ran negative test to verify performance test fails when it receives non 2xx response codes.

### DIFF
--- a/test/performance_test/performance.go
+++ b/test/performance_test/performance.go
@@ -17,8 +17,11 @@ import (
 
 // might add a with metrics bool option
 var (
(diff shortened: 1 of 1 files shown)


msnook (approved):
Thanks for fixing this.

PR #553, opened 2020-09-21
BCDA-3711 - Add support for gzip encoded file responses.

mtrang1263:
### Fixes [BCDA-3711](https://jira.cms.gov/browse/BCDA-3711)

We noticed that the NDJSON files that we serve from our ` /data/{jobId}/{filename}` endpoint was quite large (~133M) and we were spending 4-5 minutes to complete the request. To improve the performance of our system, we want to be able to serve up these files faster, while not changing the content-type (leaving it as fhir+ndjson).

### Change Details

1. Update our serveData API to check if the user supplied "Accept-Encoding: gzip". If this header is provided, we serve out gzip encoded content. We also return "Content-Type: gzip" to notify the user on how they should unpack the response.
2. Update swagger definitions to call out the new "Accept-Encoding" header.

### Security Implications

Security Checklist can be found [here](https://confluence.cms.gov/display/BCDA/20+-+09+-+22%3A++Serving+compressed+bulk+data+requests).
- [x] requires more information or team discussion to evaluate security implications

CMS wanted us to verify that it's safe for us to implement this form of HTTP compression. Since we're serving a static asset and the payload does not contain any secret information, we should not be concerned with the vulnerabilities that target this method.

### Acceptance Validation

1. Added unit tests to api.go to verify that we can serve and unpack gzip encoded content.
2. Manually ran requests against dev to verify the performance improvements. With gzip-encoding:  1.578s/3M. Without gzip-encoding: 16.213s/124M


msnook:
This is awesome.   Does it make sense to also provide explicit direction to users in the form of (a) updates to user guide, and (b) updates to our smoke test to show the usage in action?  I think both would be valuable and would be easy to point to in the event we get questions/comments.  The user guide updates could be a one or two liner indicating the capability exists;  and the updates to the smoke test might actually speed up our [weekly tests that run against a larger ACO](https://github.com/CMSgov/bcda-ops/blob/master/jenkins_files/Jenkinsfile.smoke_test#L101).
Not sure if those updates belong as part of BCDA-3711 or if we should make separate ticket(s).

mtrang1263:
> This is awesome. Does it make sense to also provide explicit direction to users in the form of (a) updates to user guide, and (b) updates to our smoke test to show the usage in action? I think both would be valuable and would be easy to point to in the event we get questions/comments. The user guide updates could be a one or two liner indicating the capability exists; and the updates to the smoke test might actually speed up our [weekly tests that run against a larger ACO](https://github.com/CMSgov/bcda-ops/blob/master/jenkins_files/Jenkinsfile.smoke_test#L101).
> Not sure if those updates belong as part of BCDA-3711 or if we should make separate ticket(s).

I created a ticket for the user guide/comms to help drive adoption: https://jira.cms.gov/browse/BCDA-3751

I think it makes sense to update smoke tests to use this for BCDA-3711. I'll go ahead and work on that.

test/smoke_test/bcda_client.go
```diff
@@ -125,13 +129,34 @@ func get(location string) *http.Response {
...
+
 	/* #nosec */
 	out, err := os.Create(filename)
 	if err != nil {
 		panic(err)
 	}
 	defer utils.CloseFileAndLogError(out)
-	num, err := io.Copy(out, resp.Body)
+
+	/* #nosec - not concerned with OOM since we're using this in a controlled setting */
```
msnook, line 158: Do we want to have a general `nosec` here, or maybe something more specific so we're only skipping a single rule?  Example can be found [here](https://github.com/CMSgov/bcda-app/blob/master/bcda/models/postgres/repository.go#L98) with documentation [here](https://github.com/securego/gosec#annotating-code).

msnook (approved):
This is great!

PatKelsh (approved):

PR #559, opened 2020-09-28
BCDA-3692 - Add CEC/NGACO requests to smoke test

mtrang1263:
### Change Details

1. Add requests for CEC ACO (`E9996`) and NGACO (`V996`) for our smoke tests.
2. Introduced a lighter version of our requests (`bulk_data_requests_lite.sh`) for usage in the new ACO types.

### Security Implications

### Acceptance Validation
CI passes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,8 +37,6 @@ SSAS_ADMIN_CLIENT_SECRET := $(shell docker-compose run --rm ssas sh -c 'main --r
 #    export ACO_CMS_ID=A9999; make postman env=local
 # or
 #    ACO_CMS_ID=A9999 make postman env=local
-#
-# TODO: as part of BCDA-3692, generate credentials for E9994 and V994
 ACO_CMS_ID ?= A9994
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -37,8 +37,6 @@ SSAS_ADMIN_CLIENT_SECRET := $(shell docker-compose run --rm ssas sh -c 'main --r
 #    export ACO_CMS_ID=A9999; make postman env=local
 # or
 #    ACO_CMS_ID=A9999 make postman env=local
-#
-# TODO: as part of BCDA-3692, generate credentials for E9994 and V994
```
mtrang1263, line 41: Remove this TODO without any changes because the caller can supply E9994/V994 in the MAKE command. The resulting CLIENT_ID and CLIENT_SECRET are used in other tests (BDT and Postman). But we're only validating a single ACO at time.

If we'd like to run Postman + BDT scan with the CEC/NGACO, we should update our travis.yml, but I do not think that it's necessary to repeat those tests. We're already verifying functionality in the smoke tests.

msnook: Makes sense to me!

msnook (approved):
⚡

PR #561, opened 2020-09-29
BCDA-3692 - Fix cms_id types to account for varying length IDs

mtrang1263:
### Fix found by [BCDA-3692](https://jira.cms.gov/browse/BCDA-3692)

With the addition of NGACOs, we have a CMS ID that is less than 5 characters (V***). As a result, we need to change our a schemas that expected a 5 character ID.

The new data type should accept UP TO 5 characters but SHOULD NOT pad the ID with extra spaces.

### Change Details

1. Update our automigrate process to alter the acos#cms_id column to varchar(5).
2. Update our automigrate process to alter the cclf_files#aco_cms_id column to varchar(5).
3. Adding tests to verify behavior of new columns.

### Security Implications

### Acceptance Validation

1. CI Passes
2. Run smoke tests on dev with new ACO types and confirm everything is working as expected.


mtrang1263 (commented):
db/testing/docker-entrypoint-initdb.d/dump.pgdata was changed to reflect the new schema (using varchar(5) instead of char(5))

msnook (approved):
Nice work sticking with this one!

PR #562, opened 2020-09-30
BCDA-3618 - Fixing SwaggerStruct definition for serveData response

mtrang1263:
### Fixes [BCDA-3618](https://jira.cms.gov/browse/BCDA-3618)

During our code generation experiment, we encountered an issue with the generated go-client returning an empty response body. The root cause was traced to an issue with our FileNDJSON response struct.

During swagger generation, it assumes that we have a declared type for the response body. This usually takes the form of a JSON structure. Since our response for the data request is the raw NDJSON string (and not a JSON struct)  our previous definition did not work as expected.

### Change Details

1. Declared NDJSON response type which is of type string. This satisfies the swagger type requirement.
2. Cleaned up response header definition to be included in the FileNDJSON response struct.

### Security Implications

### Acceptance Validation
1. Validated swagger docs are generated as expected.
2. Validated generated Go client code can retrieve the data from the `/data/...` endpoint.

Screen Shot 2020-09-30 at 2 03 07 PM


msnook (approved):
Nice!

PR #563, opened 2020-10-02
BCDA-3715 - Update db migration strategy to match bcda-ssas-app

mtrang1263:
### Fixes [BCDA-3715](https://jira.cms.gov/browse/BCDA-3715)

To improve our DB schema migration strategy, we wanted to adopt the structure used in bcda-ssas-app. These DB migrations include an up/down script to allow us to proceed/rollback and migrations. It also allows to better audit what our schema looked like at various points in time.

### Change Details

1. Introduced baseline schema definition for bcda and bcda_queue databases.
_NOTE: bcda schema only includes tables owned by bcda-app_
2. Added tests to verify migration scripts initialize the database as expected.
3. Removed any usages to gorm.Automigrate.
4. Removed our initial schema definitions (`./db/api.sql` and `./db/worker.sql`) when spinning up the Postgres containers. Instead, we'll use the migration tool to set up the database.
### Security Implications

Added [migrate](https://github.com/golang-migrate/migrate) tool. This is the same migration tool used in bcda-ssas-app

### Acceptance Validation

CI. Since we're relying on the migration tool to properly set up the bcda and bcda_queue databases (instead of the init.sql), we should be confident that the db migration is working as expected. If there was an issue in the migration script, we would not properly set up the database and the CI tests would fail.


bcda/bcdacli/cli.go
```diff
@@ -72,9 +72,6 @@ func setUpApp() *cli.App {
 				api.SetQC(qc)
 
 				fmt.Fprintf(app.Writer, "%s\n", "Starting bcda...")
-				if os.Getenv("DEBUG") == "true" {
-					autoMigrate()
```
msnook, line 76: Good riddance!

db/migrations/migrations_test.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	// Expected output:
+	// <VERSION>
+	// If there's a failure (i.e. dirty migration)
+	// <VERSION> (dirty)
+	out, err := exec.Command("migrate", "-database", m.dbURL, "-path",
+		m.migrationPath, "version").CombinedOutput()
+	if err != nil {
+		t.Errorf("Failed to retreive version information %s", err.Error())
```
msnook, line 196: Spelling error.

msnook:
Does this need to change at all (at the very least, there is a reference to `sql-migrate`)?
https://github.com/CMSgov/bcda-app/blob/master/db/testing/docker-entrypoint-initdb.d/01-restore.sh#L10

mtrang1263:
> Does this need to change at all (at the very least, there is a reference to `sql-migrate`)?
> https://github.com/CMSgov/bcda-app/blob/master/db/testing/docker-entrypoint-initdb.d/01-restore.sh#L10

Good find! Let me update with the up to date instructions

mtrang1263:
> > Does this need to change at all (at the very least, there is a reference to `sql-migrate`)?
> > https://github.com/CMSgov/bcda-app/blob/master/db/testing/docker-entrypoint-initdb.d/01-restore.sh#L10
> 
> Good find! Let me update with the up to date instructions

Spoke more about it offline, and we're OK removing this section. Rationale is that we don't have a real purpose of creating the test db from scratch. Any updates are incremental and applied on top of the existing DB.

msnook (approved):
🥇

dhgreene (commented):
Some nice improvements in `migrations_test.go`!

Dockerfiles/Dockerfile.tests
```diff
@@ -9,6 +9,10 @@ RUN go get -u github.com/xo/usql
 RUN go get -u github.com/securego/gosec/cmd/gosec
 RUN go get -u gotest.tools/gotestsum
 RUN go get -u github.com/tsenart/vegeta
+RUN go get -u -d github.com/golang-migrate/migrate/cmd/migrate; \
+    cd $GOPATH/src/github.com/golang-migrate/migrate/cmd/migrate; \
+    git checkout v4.10.0; \
```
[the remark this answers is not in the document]
mtrang1263: Wanted to keep `bcda-app` and `bcda-ssas-app` on the same versions. We'll have a ticket to move to them both to the latest version.

PatKelsh (commented):
Are there any changes we have to make to the unit-test or load-fixtures jobs?

mtrang1263:
> Are there any changes we have to make to the unit-test or load-fixtures jobs?

Nothing for the unit-tests. It has it's own DB instance that it spins up and seeds with expected data. 

For the load-fixtures, we made some changes here to ensure we build the db in its expected state: https://github.com/CMSgov/bcda-app/pull/563/files#diff-b67911656ef5d18c4ae36cb6741b7965R83

PatKelsh (approved):

dhgreene (approved):
Exciting work!  🥇

PR #564, opened 2020-10-02
BCDA-3739: Validate OpenAPI Schema Generation 

msnook:
### Fixes [BCDA-3739](https://jira.cms.gov/browse/BCDA-3739)

We recently encountered an [issue](https://github.com/CMSgov/bcda-static-site/issues/79) where the OpenAPI schema was not generated properly.  We should fail the build/package if the OpenAPI schema is invalid.

### Change Details

- Added a simple validation step to ensure OpenAPI schema was generated correctly at build time

### Security Implications

### Acceptance Validation

Performed a negative test locally, and the build was halted.

### Feedback Requested

Please review.

### DIFF
--- a/ops/swagger_to_openapi.sh
+++ b/ops/swagger_to_openapi.sh
@@ -9,4 +9,7 @@ java -jar -DswaggerUrl=swagger.yaml /swagger-converter/jetty-runner.jar /swagger
 sleep 30
 curl -X POST -d "@swaggerui/swagger.json" localhost:8080/api/convert -H "Content-Type: application/json" -o temp.json
 cat temp.json | jq > swaggerui/openapi.json
-grep "Beneficiary Claims Data API" swaggerui/openapi.json
+
+# Remove openapi.json file if it is invalid
+# Note that build_and_package.sh script validates that openapi.json exists; so RPM packaging will fail if there is an invalid openapi.json
+grep "Beneficiary Claims Data API" swaggerui/openapi.json || rm swaggerui/openapi.json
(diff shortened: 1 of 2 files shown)


ops/swagger_to_openapi.sh
```diff
@@ -8,4 +8,8 @@ set -o pipefail
 java -jar -DswaggerUrl=swagger.yaml /swagger-converter/jetty-runner.jar /swagger-converter/server.war &
 sleep 30
 curl -X POST -d "@swaggerui/swagger.json" localhost:8080/api/convert -H "Content-Type: application/json" -o temp.json
-cat temp.json | jq > swaggerui/openapi.json
\ No newline at end of file
+cat temp.json | jq > swaggerui/openapi.json
+
+# Remove openapi.json file if it is invalid
+# Note that build_and_package.sh script validates that openapi.json exists; so RPM packaging will fail if there is an invalid openapi.json
+grep "Beneficiary Claims Data API" swaggerui/openapi.json || rm swaggerui/openapi.json
```
mtrang1263, line 15: If we need the openapi.json file to exist for the RPM stage to succeed, should we fail this script instead of removing the file? That way, we would catch the failure earlier in the process (here vs. at the RPM stage)

ops/swagger_to_openapi.sh
```diff
@@ -8,4 +8,8 @@ set -o pipefail
 java -jar -DswaggerUrl=swagger.yaml /swagger-converter/jetty-runner.jar /swagger-converter/server.war &
 sleep 30
 curl -X POST -d "@swaggerui/swagger.json" localhost:8080/api/convert -H "Content-Type: application/json" -o temp.json
```
mtrang1263, line 10: I wonder if we can inspect a response code to determine failures? 

I understand what line 15 is doing, but we could get some failures if we had some slight name changes.

msnook: The problem which motivated this change occurred as a result of a successful request (`200`), but an invalid `openapi.json` file returned.  If we `grep` for the proper text to exist in the returned file (if a file is returned), do we cover our bases?

mtrang1263: Ah I see. Didn't know that we were receiving a 200 OK from the converter.

mtrang1263 (approved):
LGTM!

PR #565, opened 2020-10-06
BCDA-3719 - Update CCLFFile definition to include type field

mtrang1263:
### Fixes [BCDA-3719](https://jira.cms.gov/browse/BCDA-3719)

To support runouts, we want to add a new type parameter into the CCLF file definition. This type parameter will allow us to distinguish which CCLF files are associated with runouts thereby letting us figuring out the set of runout beneficiaries.

### Change Details

1.  Add new type parameter to our CCLFFile definition
2. Add new migration script to add the column to our databases.

### Security Implications

### Acceptance Validation

CI passes


bcda/models/models_test.go
```diff
@@ -836,3 +836,37 @@ func (s *ModelsTestSuite) TestCMSID() {
 	assert.Equal(s.T(), 1, len(actualCMSID))
 	assert.Equal(s.T(), cmsID, actualCMSID[0])
 }
+
+func (s *ModelsTestSuite) TestCCLFFileType() {
+	noType := &CCLFFile{
```
msnook, line 841: Should we add one more test to include explicitly defining the type `FileTypeDefault`?

db/migrations/migrations_test.go
```diff
@@ -98,6 +98,57 @@ func (s *MigrationTestSuite) TestBCDAMigration() {
 				}
 			},
 		},
+		{
+			"Add type column to cclf_files",
+			func(t *testing.T) {
+				migrator.runMigration(t, "2")
+				noType := &models.CCLFFile{
```
msnook, line 105: Should we also support a test case where we explicitly define `FileTypeDefault`?

bcda/models/models_test.go
mtrang1263: > Should we add one more test to include explicitly defining the type `FileTypeDefault`?

The `noType` here should verify the `FileTypeDefault` case (since the zero value for CCLFFile is `FileTypeDefault`).

db/migrations/migrations_test.go
mtrang1263: > Should we also support a test case where we explicitly define `FileTypeDefault`?

The `noType` here should verify the `FileTypeDefault` case (since the zero value for CCLFFile is `FileTypeDefault`). But I can add an explicit one if you think that behavior should be called out.

msnook: I was thinking an explicit test case would be a good idea (3 tests in total), but I'm ok if we just use the `noType` case.

msnook (approved):
👍

PatKelsh (approved):
lgtm

PR #566, opened 2020-10-07
BCDA-3681 - Add IP option when registering system

mtrang1263:
### Part of [BCDA-3681](https://jira.cms.gov/browse/BCDA-3681)

We want to add the ability to associate a set of IP addresses when registering a new system.
Right now, this IP association is only valid for SSAS auth providers.

### Change Details

1. Update our Provider#RegisterSystem method to accept a list of IP address. Using a variadic input for IPs to allow users to exclude the IP arguments when calling RegisterSystem.

The following snippets are all valid:
```
cred, err := RegisterSystem("id", "key", "groupID")
cred, err := RegisterSystem("id", "key", "groupID", "1.2.3.4", "5.6.7.8")
cred, err := RegisterSystem("id", "key", "groupID", []string{"10.11.12.13"}...)
```

2. Update our cli command for generating system credentials to accept list of IP addresses.

### Security Implications

### Acceptance Validation
1. Added test cases to verify that we're passing the IP address argument correctly when making the associated SSAS call.
2. Will have an associated change in bcda-ops where we can verify that the IP address are being added when the caller makes the generate-creds call.

Successfully validate 2. See: https://github.com/CMSgov/bcda-ops/pull/577 for more details.

### DIFF
--- a/bcda/auth/alpha.go
+++ b/bcda/auth/alpha.go
@@ -17,7 +17,7 @@ type AlphaAuthPlugin struct{}
 // validates that AlphaAuthPlugin implements the interface
 var _ Provider = AlphaAuthPlugin{}
 
-func (p AlphaAuthPlugin) RegisterSystem(localID, publicKey, groupID string) (Credentials, error) {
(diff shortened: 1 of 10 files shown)


msnook (approved):
🚀

PR #567, opened 2020-10-08
BCDA-3720 - Update CCLF ingest to support runout files

mtrang1263:
### Fixes [BCDA-3720](https://jira.cms.gov/browse/BCDA-3720)

Runout files have a slightly different format versus regular CCLF files. For the ZIP file, we can expect `ZCR` instead of `ZCY`. For individual CCLF files, we can expect `ZC0R,ZC8R` instead of `ZC0Y,ZC8Y`

### Change Details

1. Update regex found in fileprocessor.go to support the new expectations from runout files.
2. Update our metadata to include filetype. If we have `R`, we assume runout. If we have `Y`, we assume a regular CCLF file.

### Security Implications
With this change, we will be able to ingest runout CCLF files. The data stored will be exactly the same as regular CCLF files.

### Acceptance Validation
CI passes - new tests written to validate that we can process runout files as expected.

### DIFF
--- a/bcda/cclf/cclf.go
+++ b/bcda/cclf/cclf.go
@@ -238,7 +238,7 @@ func importCCLF(ctx context.Context, fileMetadata *cclfFileMetadata, importer im
 	rc, err := rawFile.Open()
 	if err != nil {
 		fmt.Printf("Could not read file %s for CCLF%d in archive %s.\n", cclfFile.Name, fileMetadata.cclfNum, fileMetadata.filePath)
-		err = fmt.Errorf("could not read file %s for CCLF%d in archive %s", cclfFile.Name, fileMetadata.cclfNum, fileMetadata.filePath)
+		err = errors.Wrapf(err, "could not read file %s for CCLF%d in archive %s", cclfFile.Name, fileMetadata.cclfNum, fileMetadata.filePath)
 		log.Error(err)
 		return err
 	}
--- a/bcda/cclf/cclf.go
+++ b/bcda/cclf/cclf.go
@@ -34,6 +34,7 @@ type cclfFileMetadata struct {
 	imported     bool
 	deliveryDate time.Time
 	fileID       uint
+	fileType     models.CCLFFileType
 }
 
 type cclfFileValidator struct {
@@ -199,6 +200,7 @@ func importCCLF(ctx context.Context, fileMetadata *cclfFileMetadata, importer im
 		Timestamp:       fileMetadata.timestamp,
 		PerformanceYear: fileMetadata.perfYear,
 		ImportStatus:    constants.ImportInprog,
+		Type:            fileMetadata.fileType,
 	}
 
 	db := database.GetGORMDbConnection()
@@ -228,7 +230,7 @@ func importCCLF(ctx context.Context, fileMetadata *cclfFileMetadata, importer im
 
 	if rawFile == nil {
 		fmt.Printf("File %s not found in archive %s.\n", fileMetadata.name, fileMetadata.filePath)
-		err = errors.Wrapf(err, "file %s not found in archive %s", fileMetadata.name, fileMetadata.filePath)
+		err = fmt.Errorf("file %s not found in archive %s", fileMetadata.name, fileMetadata.filePath)
 		log.Error(err)
 		return err
 	}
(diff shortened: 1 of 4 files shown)


bcda/cclf/cclf.go
```diff
@@ -228,15 +230,15 @@ func importCCLF(ctx context.Context, fileMetadata *cclfFileMetadata, importer im
 
 	if rawFile == nil {
 		fmt.Printf("File %s not found in archive %s.\n", fileMetadata.name, fileMetadata.filePath)
-		err = errors.Wrapf(err, "file %s not found in archive %s", fileMetadata.name, fileMetadata.filePath)
+		err = fmt.Errorf("file %s not found in archive %s", fileMetadata.name, fileMetadata.filePath)
```
mtrang1263, line 233: No longer call `errors.Wrap(err,...)` because the `err == nil` at this point. Calling `errors.Wrap(nil,...)`, returns a nil error.

See: https://github.com/pkg/errors/blob/master/errors.go#L185

bcda/cclf/cclf_test.go
```diff
@@ -351,6 +354,47 @@ func (s *CCLFTestSuite) TestGetPriorityACOs() {
...
+				cclfNum:   8,
+				name:      cclfName,
+				acoID:     cmsID,
+				timestamp: time.Now(),
+				perfYear:  20,
+				fileType:  tt.fileType,
+				filePath:  fileName,
+			}
+
+			s.NoError(importCCLF(context.Background(), metadata, noopImporter{}))
```
mtrang1263, line 389: Using noopImporter because we need to validate the the fileType is being saved as expected.

bcda/cclf/fileprocessor.go
```diff
@@ -113,8 +113,8 @@ func (p *processor) handleArchiveError(path string, info os.FileInfo, cause erro
 }
 
 func getCMSID(name string) (string, error) {
-	// CCLF foldername convention with BCD identifier: P.BCD.<ACO_ID>.ZCY**.Dyymmdd.Thhmmsst
-	exp := regexp.MustCompile(`(?:T|P)\.BCD\.(.*)\.ZCY\d{2}\.D\d{6}\.T\d{7}`)
+	// CCLF foldername convention with BCD identifier: P.BCD.<ACO_ID>.ZC[Y|R]**.Dyymmdd.Thhmmsst
+	exp := regexp.MustCompile(`(?:T|P)\.BCD\.(.*)\.ZC(?:Y|R)\d{2}\.D\d{6}\.T\d{7}`)
```
msnook, line 117: Should this be `[Y|R]` instead of `?:Y|R`?

bcda/cclf/fileprocessor_test.go
```diff
@@ -144,6 +146,7 @@ func TestGetCMSID(t *testing.T) {
 		cmsID    string
 	}{
 		{"validSSPPath", "path/T.BCD.A0001.ZCY18.D181120.T1000000", false, "A0001"},
```
msnook, line 148: Could we have a test that looks like `ZC20` (missing `Y` and `R`)?

bcda/cclf/fileprocessor_test.go
```diff
@@ -144,6 +146,7 @@ func TestGetCMSID(t *testing.T) {
...
+		{"validSSPRunoutPath", "path/T.BCD.A0002.ZCR18.D181120.T1000000", false, "A0002"},
```
msnook, line 149: Should this line:
```
{"missingBCD", "path/T.A0001.ACO.ZC8Y18.D18NOV20.T1000009", true, ""},
```
Look like this:
```
{"missingBCD", "path/T.A0001.ACO.ZCY18.D18NOV20.T1000009", true, ""},
```

bcda/cclf/fileprocessor_test.go
```diff
@@ -217,6 +222,19 @@ func TestGetCCLFMetadata(t *testing.T) {
 				acoID:     sspID,
 				timestamp: validTime,
 				perfYear:  20,
+				fileType:  models.FileTypeDefault,
+			},
+		},
+		{
+			"Runout SSP file", sspID, sspRunoutFile, "",
```
msnook, line 229: Does it make sense to test runout files for the other ACO types?

bcda/cclf/fileprocessor.go
mtrang1263: Was trying to figure out a non-capturing OR expression!

bcda/cclf/fileprocessor_test.go
mtrang1263: I was going back and forth on this. Since we've only guaranteed that we'll receive the runout files for SSP ACOs, I decided to leave them off.

Once we have the conversations with CEC/NGACO (and other types), we  can add those tests in. Just in case those have a different format.

msnook (approved):
🎸

PR #570, opened 2020-10-22
BCDA-3582 - Update BCDA to support querying v1/v2 BFD endpoints

mtrang1263:
### Fixes [BCDA-3582](https://jira.cms.gov/browse/BCDA-3582)

We need to have a mechanism where callers can retrieve data from BFD v1 endpoints (STU3) and v2 endpoints (R4).

### Change Details
1. Update our JobEnqueueArgs to accept a basePath parameter. This parameter will be used to allow our worker to make v1/v2 BFD calls.
2. Update v1/v2 APIs to contain a list of supported resourceTypes. This information is passed into requests.go.
3. Update requests.go to contain a handler struct to capture resourceTypes and basePath parameters. v1 and v2 API will instantiate the handler with the expected parameters.
4. Refactor validation tests into request_test.go (from v1/api_test.go)

### Security Implications

This PR **does not** call BFD v2 endpoints since they are not yet available. We will have follow up PR that switches the basePath. As such, no new data is stored/transmitted.

### Acceptance Validation
CI passes

### DIFF
--- a/bcda/api/requests.go
+++ b/bcda/api/requests.go
@@ -273,6 +273,68 @@ func (h *Handler) bulkRequest(resourceTypes []string, w http.ResponseWriter, r *
 	}
 }
 
+func (h *Handler) validateRequest(r *http.Request) ([]string, *fhirmodels.OperationOutcome) {
+
+	// validate optional "_type" parameter
+	var resourceTypes []string
+	params, ok := r.URL.Query()["_type"]
+	if ok {
+		resourceMap := make(map[string]bool)
+		params = strings.Split(params[0], ",")
+		for _, p := range params {
+			if !resourceMap[p] {
+				resourceMap[p] = true
+				resourceTypes = append(resourceTypes, p)
+			} else {
+				oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, responseutils.RequestErr, "Repeated resource type")
+				return nil, oo
+			}
+		}
+	} else {
+		// resource types not supplied in request; default to applying all resource types.
+		resourceTypes = append(resourceTypes, "Patient", "ExplanationOfBenefit", "Coverage")
+	}
+
+	for _, resourceType := range resourceTypes {
+		if _, ok := h.supportedResources[resourceType]; !ok {
+			oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, responseutils.RequestErr,
+				fmt.Sprintf("Invalid resource type %s. Supported types %s.", resourceType, h.supportedResources))
+			return nil, oo
+		}
+	}
+
+	// validate optional "_since" parameter
+	params, ok = r.URL.Query()["_since"]
+	if ok {
+		sinceDate, err := time.Parse(time.RFC3339Nano, params[0])
+		if err != nil {
+			oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, responseutils.FormatErr, "Invalid date format supplied in _since parameter.  Date must be in FHIR Instant format.")
+			return nil, oo
+		} else if sinceDate.After(time.Now()) {
+			oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, responseutils.FormatErr, "Invalid date format supplied in _since parameter. Date must be a date that has already passed")
+			return nil, oo
+		}
+	}
+
+	//validate "_outputFormat" parameter
+	params, ok = r.URL.Query()["_outputFormat"]
+	if ok {
+		if params[0] != "ndjson" && params[0] != "application/fhir+ndjson" && params[0] != "application/ndjson" {
+			oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, responseutils.FormatErr, "_outputFormat parameter must be application/fhir+ndjson, application/ndjson, or ndjson")
+			return nil, oo
+		}
+	}
+
+	// we do not support "_elements" parameter
+	_, ok = r.URL.Query()["_elements"]
+	if ok {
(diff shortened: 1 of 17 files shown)


bcda/api/requests_test.go
```diff
@@ -133,6 +117,85 @@ func (s *RequestsTestSuite) TestRunoutDisabled() {
 	s.Contains(string(body), "Invalid group ID")
 }
 
+func (s *RequestsTestSuite) TestInvalidRequests() {
```
mtrang1263, line 120: These tests replace all of the removed tests from bcda/api/v1/api_test.go

bcda/api/v2/api.go
```diff
@@ -15,6 +15,14 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+var h *api.Handler
+
+func init() {
+	// TODO (BCDA-3582) - Since v2 APIs are not available (as of 2020-10-21), we're still
+	// routing requests to the v1 BFD endpoints
+	h = api.NewHandler([]string{"Patient", "Coverage"}, "/v1/fhir")
```
mtrang1263, line 23: Hard coding the supported resource types (instead of env var). We'll want to do some verification on BCDA end before enabling a new resource type.

bcda/bcdacli/cli.go
```diff
@@ -47,40 +42,46 @@ func setUpApp() *cli.App {
 	app.Usage = Usage
 	app.Version = constants.Version
 	var acoName, acoCMSID, acoID, accessToken, threshold, acoSize, filePath, dirToDelete, environment, groupID, groupName, ips string
+	var httpPort, httpsPort int
 	app.Commands = []cli.Command{
 		{
 			Name:  "start-api",
 			Usage: "Start the API",
+			Flags: []cli.Flag{
```
mtrang1263, line 50: Allows us to verify that we can launch the API in our unit tests.

bcda/api/requests.go
```diff
@@ -6,6 +6,7 @@ import (
 	"strconv"
 
 	"github.com/go-chi/chi"
+	"github.com/jackc/pgx"
```
msnook, line 9: New dependency; might consider a security checklist to make sure we aren't introducing a dependency that might be troublesome.

bcda/api/requests.go
```diff
@@ -345,6 +335,43 @@ func ValidateRequest(r *http.Request) ([]string, *fhirmodels.OperationOutcome) {
 	return resourceTypes, nil
 }
 
+func check429(jobs []models.Job, types []string, w http.ResponseWriter) ([]string, bool) {
```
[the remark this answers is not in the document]
mtrang1263: It doesn't consider basePath when checking for the 429. I think that we would still limit the end user to a single concurrent resource request at a time though. Even though they are different FHIR response types, they will still put the same pressure on our API/BFD, so it would make sense to still limit?

bcda/bcdacli/cli.go
mtrang1263: Nope! It'll still default to 3001 for http and 3000 for https

bcda/api/requests.go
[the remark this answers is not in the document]
msnook: I was thinking about a scenario where they want to compare v1 payload vs v2 payload... they'd have to make the calls in sequence instead of in parallel.  I'm sure either approach is fine - just throwing this out as a known limitation.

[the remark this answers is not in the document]
mtrang1263: That's a great use case. I'll go ahead and update it to distinguish between v1/v2 calls

bcda/api/requests.go
```diff
@@ -156,9 +164,16 @@ func (h *Handler) bulkRequest(resourceTypes []string, w http.ResponseWriter, r *
 	// Overall, this will prevent a queue of concurrent calls from slowing up our system.
 	// NOTE: this logic is relevant to PROD only; simultaneous requests in our lower environments is acceptable (i.e., shared opensbx creds)
 	if (os.Getenv("DEPLOYMENT_TARGET") == "prod") && (!db.Find(&jobs, "aco_id = ?", acoID).RecordNotFound()) {
-		if types, ok := check429(jobs, resourceTypes, w); !ok {
-			w.Header().Set("Retry-After", strconv.Itoa(utils.GetEnvInt("CLIENT_RETRY_AFTER_IN_SECONDS", 0)))
-			w.WriteHeader(http.StatusTooManyRequests)
+		if types, err := check429(jobs, resourceTypes, version); err != nil {
```
mtrang1263, line 167: Using a custom error type instead of the boolean to avoid multiple writes the Http.ResponseWriter.

If we encountered a non duplicate error (i.e. url parsing issue), we would need to do an early return AFTER we've written an operational outcome with status code.

By returning a custom error type, we can handle the response writing in the bulkRequest function and do not have to pass in the writer.

bcda/api/requests.go
```diff
@@ -335,17 +350,32 @@ func (h *Handler) validateRequest(r *http.Request) ([]string, *fhirmodels.Operat
...
+// Returns the unworkedTypes (if any)
+func check429(jobs []models.Job, types []string, version string) ([]string, error) {
 	var unworkedTypes []string
 
 	for _, t := range types {
 		worked := false
 		for _, job := range jobs {
 			req, err := url.Parse(job.RequestURL)
 			if err != nil {
-				log.Error(err)
```
mtrang1263: Since we weren't doing an early return here, line 351 would've panicked because req is nil in this case.

If we did an early return (when we only had boolean type), we would still write the 429 response AFTER we've written the OperationalOutcome.

msnook (commented):
This looks great!  Thanks for adding the additional `429` implementation and testing.

PatKelsh (approved):

PR #573, opened 2020-10-29
BCDA-3920 - Update 429 check to only use in-progress/pending jobs

mtrang1263:
### Fixes [BCDA-3920](https://jira.cms.gov/browse/BCDA-3920)

With this [PR](https://github.com/CMSgov/bcda-app/pull/570), we started to parse the job's requestURL to determine if we are servicing a V1 or V2 request.

We had expected that all jobs were of type Group or Patient. However, there are older jobs for other resources (ex: Coverage). This caused our regex to fail and prevented smoke tests from passing in the prod environment.

### Change Details

Only return in progress OR pending jobs from the database when applying our 429 check. Completed, Failed, Expired, etc. should not be used. This matches our business logic in the check429 function. We only consider in progress and pending jobs.

### Security Implications

### Acceptance Validation

Added new unit test to verify the functionality.

### DIFF
--- a/bcda/api/requests.go
+++ b/bcda/api/requests.go
@@ -157,14 +157,15 @@ func (h *Handler) bulkRequest(resourceTypes []string, w http.ResponseWriter, r *
 	defer database.Close(db)
 	acoID := ad.ACOID
 
-	var jobs []models.Job
+	var pendingAndInProgressJobs []models.Job
 	// If we really do find this record with the below matching criteria then this particular ACO has already made
 	// a bulk data request and it has yet to finish. Users will be presented with a 429 Too-Many-Requests error until either
 	// their job finishes or time expires (+24 hours default) for any remaining jobs left in a pending or in-progress state.
 	// Overall, this will prevent a queue of concurrent calls from slowing up our system.
 	// NOTE: this logic is relevant to PROD only; simultaneous requests in our lower environments is acceptable (i.e., shared opensbx creds)
-	if (os.Getenv("DEPLOYMENT_TARGET") == "prod") && (!db.Find(&jobs, "aco_id = ?", acoID).RecordNotFound()) {
-		if types, err := check429(jobs, resourceTypes, version); err != nil {
+	if (os.Getenv("DEPLOYMENT_TARGET") == "prod") &&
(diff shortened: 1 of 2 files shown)


bcda/api/requests.go
```diff
@@ -157,14 +157,15 @@ func (h *Handler) bulkRequest(resourceTypes []string, w http.ResponseWriter, r *
...
+	var pendingAndInProgressJobs []models.Job
 	// If we really do find this record with the below matching criteria then this particular ACO has already made
 	// a bulk data request and it has yet to finish. Users will be presented with a 429 Too-Many-Requests error until either
 	// their job finishes or time expires (+24 hours default) for any remaining jobs left in a pending or in-progress state.
 	// Overall, this will prevent a queue of concurrent calls from slowing up our system.
 	// NOTE: this logic is relevant to PROD only; simultaneous requests in our lower environments is acceptable (i.e., shared opensbx creds)
-	if (os.Getenv("DEPLOYMENT_TARGET") == "prod") && (!db.Find(&jobs, "aco_id = ?", acoID).RecordNotFound()) {
-		if types, err := check429(jobs, resourceTypes, version); err != nil {
+	if (os.Getenv("DEPLOYMENT_TARGET") == "prod") &&
+		(!db.Find(&pendingAndInProgressJobs, "aco_id = ? AND status IN (?, ?)", acoID, "In Progress", "Pending").RecordNotFound()) {
```
msnook, line 167: Is there some redundancy here?
https://github.com/CMSgov/bcda-app/blob/master/bcda/api/requests.go#L391

mtrang1263: Since the method is not visible outside of the package, I think it makes sense to remove the checks in check429. I'll update it now

bcda/api/requests.go
```diff
@@ -386,17 +387,20 @@ func check429(jobs []models.Job, types []string, version string) ([]string, erro
 				continue
 			}
 
+			// If the job has timed-out we will allow another job to be in progress
+			if time.Now().After(job.CreatedAt.Add(GetJobTimeout())) {
```
mtrang1263, line 391: This is the exact opposite logic that used to exist.

We used to say "If the job HAS NOT timed out, then we'll do stuff". The new logic is "If the job HAS timed out, then we will skip (and allow the job to be worked)"

msnook (approved):
LGTM!

PatKelsh (approved):

PR #580, opened 2020-11-06
BCDA-3847 - Add blacklisted column to ACOs table

mtrang1263:
### Fixes [BCDA-3847](https://jira.cms.gov/browse/BCDA-3847)

We need to support the ability to blacklist an ACO. When an ACO is blacklisted, they will not be able to access the API. More details can be found [here](https://confluence.cms.gov/pages/viewpage.action?spaceKey=BCDA&title=How+To+Blacklist+an+ACO+From+Generating+Credentials).

This PR established the field that we will use to determine if an ACO is blacklisted. There are two other tickets that will be implemented to [write to the column](https://jira.cms.gov/browse/BCDA-3848) and [read from the column](https://jira.cms.gov/browse/BCDA-3849).

### Change Details

1. Add blacklisted column to ACOs table. The default value will be false to ensure that all existing ACOs are not blacklisted.
2. Update ACOs model to include blacklisted field.
3. Added tests to verify migration works as expected.

### Security Implications
- [x] new data stored or transmitted
We will store if an ACO is blacklisted or not.

### Acceptance Validation
CI passes.


msnook (approved):
LGTM.

db/migrations/bcda/000003_aco_blacklist.down.sql
```diff
@@ -0,0 +1,3 @@
+BEGIN;
```
PatKelsh, line 1: Could maybe use a comment explaining what is being dropped. Not really needed here since the migration is so small but would be a good practice for some of the bigger ones

PatKelsh (dismissed):
lgtm

PatKelsh (approved):
lgtm

PR #583, opened 2020-11-11
BCDA-3840 Added depulication MBIs in cclf import

PatKelsh:
### Fixes [BCDA-3840](https://jira.cms.gov/browse/BCDA-3840)

Added in filter for duplicate MBI entries in CCLF files to prevent duplicate data from appearing in API responses

### Proposed Changes

* Added duplication filter to CCLF file handler
* Modified importer and file handler to work with filtering method
* Updated tests

### Security Implications


bcda/cclf/cclf.go
```diff
@@ -266,6 +266,8 @@ func importCCLF(ctx context.Context, fileMetadata *cclfFileMetadata, importer im
 		}
 	}()
 
+	var importedMBI = make(map[string]models.CCLFBeneficiary)
```
mtrang1263, line 269: Since we only need to track the existence of the MBI, we should use a `map[string]struct{}`. This effectively allows us to use a map like a set.

bcda/cclf/cclf.go
```diff
@@ -275,7 +277,24 @@ func importCCLF(ctx context.Context, fileMetadata *cclfFileMetadata, importer im
...
+		}
+		// Filtering for duplicate benes in CCLF file
+		if _, ok := importedMBI[cclfBeneficiary.MBI]; ok {
+			continue
+		}
+
+		importedMBI[cclfBeneficiary.MBI] = cclfBeneficiary
+
+		err = importer.do(ctx, txn, cclfBeneficiary)
+		// err = importer.do(ctx, txn, cclfFile.ID, b)
```
mtrang1263, line 297: Remove?

bcda/cclf/cclf.go
```diff
@@ -275,7 +277,24 @@ func importCCLF(ctx context.Context, fileMetadata *cclfFileMetadata, importer im
 			continue
 		}
 
-		err = importer.do(ctx, txn, cclfFile.ID, b)
+		const (
+			mbiStart, mbiEnd   = 0, 11
+			hicnStart, hicnEnd = 11, 22
+		)
+		cclfBeneficiary := models.CCLFBeneficiary{
```
mtrang1263, line 284: If we're parsing the data with assumption that it's a CCLF8 file, we should remove `importCCLF` and push all of this code into `importCCLF8`.

bcda/cclf/cclf.go
```diff
@@ -275,7 +277,24 @@ func importCCLF(ctx context.Context, fileMetadata *cclfFileMetadata, importer im
...
+			FileID: cclfFile.ID,
+			MBI:    string(bytes.TrimSpace(b[mbiStart:mbiEnd])),
+			HICN:   string(bytes.TrimSpace(b[hicnStart:hicnEnd])),
+		}
+		// Filtering for duplicate benes in CCLF file
+		if _, ok := importedMBI[cclfBeneficiary.MBI]; ok {
+			continue
+		}
+
+		importedMBI[cclfBeneficiary.MBI] = cclfBeneficiary
```
mtrang1263, line 294: When the map changes, this would change to `importedMBI[cclfBeneficiary.MBI] = struct{}{}`

bcda/cclf/cclf_test.go
```diff
@@ -427,10 +427,14 @@ func createTemporaryZipFile(t *testing.T) (fileName, cclfName string) {
 
 type noopImporter struct{}
 
-func (i noopImporter) do(ctx context.Context, tx *sql.Tx, fileID uint, b []byte) error {
+func (i noopImporter) do(ctx context.Context, tx *sql.Tx, cclfBeneficiary models.CCLFBeneficiary) error {
 	return nil
 }
 
+// func (i noopImporter) do(ctx context.Context, tx *sql.Tx, fileID uint, b []byte) error {
```
mtrang1263, line 434: Remove?

bcda/cclf/importer_test.go
```diff
@@ -82,10 +82,16 @@ func (s *ImporterTestSuite) TestCCLF8ImporterHappyPath() {
...
+				cclfBeneficiary := models.CCLFBeneficiary{
+					FileID: fileID,
+					MBI:    mbi,
+					HICN:   hicn,
+				}
 				prepare.ExpectExec().WithArgs(bene.FileID, hicn, mbi).WillReturnResult(sqlmock.NewResult(1, 1))
 
-				data := append([]byte(mbi), []byte(hicn)...)
-				err := importer.do(context.Background(), s.tx, fileID, data)
+				// data := append([]byte(mbi), []byte(hicn)...)
```
mtrang1263, line 92: Remove?

bcda/cclf/importer_test.go
```diff
@@ -82,10 +82,16 @@ func (s *ImporterTestSuite) TestCCLF8ImporterHappyPath() {
...
+				err := importer.do(context.Background(), s.tx, cclfBeneficiary)
+				// err := importer.do(context.Background(), s.tx, fileID, data)
```
mtrang1263, line 94: Remove?

bcda/cclf/importer.go
```diff
@@ -14,7 +13,7 @@ import (
 )
 
 type importer interface {
-	do(ctx context.Context, tx *sql.Tx, fileID uint, b []byte) error
+	do(ctx context.Context, tx *sql.Tx, cclfBeneficiary models.CCLFBeneficiary) error
```
mtrang1263, line 16: To keep this generic can we change the argument to an interface{}? Then you can use a type assertion to verify that we passed a CCLFBeneficiary.

It would look something like this:
```
func (cclfImporter *cclf8Importer) do(ctx context.Context, tx *sql.Tx, data interface{}) {
     bene, ok := data.(models.CCLFBeneficiary)
     if !ok {
          return fmt.Errrof("invalid type sent, expected models.CCLFBeneficiary)
     }
     ....
}
```

bcda/cclf/importer_test.go
```diff
@@ -145,7 +151,12 @@ func (s *ImporterTestSuite) TestCCLF8ImporterErrorPaths() {
 			prepare := s.mock.ExpectPrepare(regexp.QuoteMeta(`COPY "cclf_beneficiaries" ("file_id", "hicn", "mbi")`))
 			hicn := string([]byte(bene.HICN)[0:11])
 			mbi := string([]byte(bene.MBI)[0:11])
-			data := append([]byte(mbi), []byte(hicn)...)
+			cclfBeneficiary := models.CCLFBeneficiary{
+				FileID: fileID,
+				MBI:    mbi,
+				HICN:   hicn,
+			}
+			// data := append([]byte(mbi), []byte(hicn)...)
```
mtrang1263, line 159: Remove?

mtrang1263 (changes requested):

mtrang1263 (commented):
Doesn't look like we have any update test cases to verify the dedupe behavior. Can you look at `TestImportCCLF8` and see if we can leverage that to verify that we ignore duplicates?

bcda/cclf/importer.go
```diff
@@ -33,9 +33,13 @@ type cclf8Importer struct {
 }
 
 // validates that cclf8Importer implements the interface
-var _ importer = &cclf8Importer{}
+// var _ importer = &cclf8Importer{}
```
mtrang1263, line 36: Can you uncomment this out?

bcda/cclf/cclf_test.go
```diff
@@ -197,7 +197,7 @@ func (s *CCLFTestSuite) TestImportCCLF8() {
 
 	beneficiaries := []models.CCLFBeneficiary{}
 	db.Find(&beneficiaries, "file_id = ?", file.ID)
-	assert.Equal(6, len(beneficiaries))
+	assert.Equal(5, len(beneficiaries))
```
mtrang1263, line 200: How come the number of beneficiaries decreased? Instead of removing `203031406M`, can add a duplicate?

bcda/cclf/cclf.go
mtrang1263: Bump on this, I don't think we need the `importCCLF` function. `importCCLF` used to be generic, but now that parse the file and assume it's a CCLF8 file, we should remove `importCCLF` and move all of the logic into `importCCLF8`.

bcda/cclf/cclf_test.go
mtrang1263: Bump on this.

mtrang1263 (changes requested):

bcda/cclf/cclf.go
```diff
@@ -153,17 +153,6 @@ func importCCLF8(ctx context.Context, fileMetadata *cclfFileMetadata) error {
...
-
-	if err != nil {
-		updateImportStatus(fileMetadata, constants.ImportFail)
-		return err
-	}
-	updateImportStatus(fileMetadata, constants.ImportComplete)
-	return nil
-}
-
-func importCCLF(ctx context.Context, fileMetadata *cclfFileMetadata, importer importer) (err error) {
```
mtrang1263, line 166: We should undo the changes that removed the `return err`. If we do not do an early return when we encounter an error, we may run into nil references/other issues. Ex: if `zip.NewReader` returns an error, the reader it returns is nil. [Ref.](https://golang.org/src/archive/zip/reader.go?s=904:953#L39)

I think I understand why the early returns were removed (so we set the import status to failed), but we can achieve it by using `defer`.

If we switch the signature to use a named return value (`func importCCLF8(ctx context.Context, fileMetadata *cclfFileMetadata) (err error)`) and switch any instances where we reallocate an error value (i.e. `err := ...`), we can do the following.

```
defer func() {
		if err != nil {
			updateImportStatus(fileMetadata, constants.ImportFail)
		}
	}()
```

We should place this defer statement after we've guaranteed that `fileMetadata` is not nil.

bcda/cclf/cclf.go
```diff
@@ -179,15 +168,13 @@ func importCCLF(ctx context.Context, fileMetadata *cclfFileMetadata, importer im
 		fmt.Printf("Could not read CCLF%d archive %s.\n", fileMetadata.cclfNum, fileMetadata.filePath)
 		err := errors.Wrapf(err, "could not read CCLF%d archive %s", fileMetadata.cclfNum, fileMetadata.filePath)
 		log.Error(err)
-		return err
 	}
 	defer r.Close()
 
 	if len(r.File) < 1 {
 		fmt.Printf("No files found in CCLF%d archive %s.\n", fileMetadata.cclfNum, fileMetadata.filePath)
 		err := fmt.Errorf("no files found in CCLF%d archive %s", fileMetadata.cclfNum, fileMetadata.filePath)
```
mtrang1263, line 176: This would be an instance where we change it to `err = fmt.Errorf("no files found in CCLF%d archive %s", fileMetadata.cclfNum, fileMetadata.filePath)`

mtrang1263 (changes requested):

bcda/cclf/importer.go
```diff
@@ -33,9 +24,13 @@ type cclf8Importer struct {
...
 
-func (cclfImporter *cclf8Importer) do(ctx context.Context, tx *sql.Tx, fileID uint, b []byte) error {
+func (cclfImporter *cclf8Importer) do(ctx context.Context, tx *sql.Tx, data interface{}) error {
```
mtrang1263, line 29: Sorry, I didn't realize the interface goes away once we removed the `importCCLF()` method. Because there's no importer interface, your old way of supplying the models.CCLFBeneficiary makes more sense.

bcda/cclf/importer.go
mtrang1263: scratch that, didn't notice that the interface was removed. We can delete the line now.

bcda/cclf/cclf.go
```diff
@@ -179,15 +168,13 @@ func importCCLF(ctx context.Context, fileMetadata *cclfFileMetadata, importer im
 		fmt.Printf("Could not read CCLF%d archive %s.\n", fileMetadata.cclfNum, fileMetadata.filePath)
 		err := errors.Wrapf(err, "could not read CCLF%d archive %s", fileMetadata.cclfNum, fileMetadata.filePath)
```
mtrang1263, line 169: This would be another instance where we want to switch to `err = errors.Wrapf(err, "could not read CCLF%d archive %s", fileMetadata.cclfNum, fileMetadata.filePath)`

bcda/cclf/cclf.go
```diff
@@ -275,7 +276,23 @@ func importCCLF(ctx context.Context, fileMetadata *cclfFileMetadata, importer im
 			continue
 		}
 
-		err = importer.do(ctx, txn, cclfFile.ID, b)
+		const (
+			mbiStart, mbiEnd   = 0, 11
+			hicnStart, hicnEnd = 11, 22
```
msnook, line 281: Since we don't use HICN anymore, maybe a future ticket is needed to cease ingestion of this data?

bcda/cclf/cclf.go
```diff
@@ -275,7 +276,23 @@ func importCCLF(ctx context.Context, fileMetadata *cclfFileMetadata, importer im
...
+		)
+		cclfBeneficiary := models.CCLFBeneficiary{
+			FileID: cclfFile.ID,
+			MBI:    string(bytes.TrimSpace(b[mbiStart:mbiEnd])),
+			HICN:   string(bytes.TrimSpace(b[hicnStart:hicnEnd])),
+		}
+		// Filtering for duplicate benes in CCLF file
+		if _, ok := importedMBI[cclfBeneficiary.MBI]; ok {
```
msnook, line 289: What kind of performance hit are we taking here?
I'm wondering if just ingesting the duplicate MBIs is simpler/faster than checking every single MBI to see if we've already ingested it.  Remember, we are protecting against sending claims data for duplicate beneficiaries using the work in [this PR](https://github.com/CMSgov/bcda-app/pull/506) and [this PR](https://github.com/CMSgov/bcda-app/pull/521).
Bottom line - is the work and potential performance hit from the work in this PR worth it?

mtrang1263: So the filtering that we're doing here is going to be in memory, so I'd wager that we won't see any performance difference. I don't think that'll we'll see any appreciable gains on the ingest performance, but don't think we'll come out slower.

I think you hit on the point that we're looking to improve. If we can have a guarantee that we do not have any dupes in the CCLF file, we can leverage simpler queries that can better use the index ( we can use the file_id index). As it is now, we're firing off two queries to get around the dupes, and our final query is leveraging CCLFBeneficiary#id which may not be as efficient since it's aunique column.

msnook: > So the filtering that we're doing here is going to be in memory, so I'd wager that we won't see any performance difference. I don't think that'll we'll see any appreciable gains on the ingest performance, but don't think we'll come out slower.
> 
> I think you hit on the point that we're looking to improve. If we can have a guarantee that we do not have any dupes in the CCLF file, we can leverage simpler queries that can better use the index ( we can use the file_id index). As it is now, we're firing off two queries to get around the dupes, and our final query is leveraging CCLFBeneficiary#id which may not be as efficient since it's aunique column.

Works for me!

msnook (approved):
This looks good to me.

mtrang1263 (approved):

PR #585, opened 2020-11-13
BCDA-3870 - Account for multiple file types in CCLF ingest process

mtrang1263:
### Discovered during [BCDA-3870](https://jira.cms.gov/browse/BCDA-3870)

Support the ability to running CCLF ingestion with multiple file types (default, runout).

### Proposed Changes

Similar to how we handle different performance years, we need the ability to handle situations where we have runout CCLF files that are mixed in with regular (default) CCLF files.

### Change Details
1. Update our fileprocessor to support multiple file types (in addition to multiple performance years).
    Resulting change was an update to the inner map from int (perf year only) to a struct containing perf year + file type.
2. Added new test case to confirm we can handle the situation.

### Security Implications

### Acceptance Validation
1. CI Passes.
2. Run smoke tests that attempt to ingest runout + regular CCLF files at once.


mtrang1263 (commented):
Confirmed with this [smoke test run](https://bcda-ci.adhocteam.us/job/BCDA%20-%20Test%20-%20Integration%20Smoke/2581/) we were able to ingest both runout and regular CCLF files in the same import-cclf-directory run.

```
Successfully imported 2500 records from CCLF8 file T.BCD.A9990.ZC8R20.D201113.T0000010.
Successfully imported 2500 records from CCLF8 file T.BCD.A9990.ZC8Y20.D201113.T0000010.

```

SMLuthi (approved):
LGTM!

msnook (approved):
👍

PR #590, opened 2020-11-18
BCDA-3752 - Update swagger docs to reference groupID runout

mtrang1263:
### Part of [BCDA-3752](https://jira.cms.gov/browse/BCDA-3752)

Updating swagger docs to reference the groupID runout.

### Change Details
1. Change groupID swagger type to enum. This allows us to force users to use all/runout.
2. Update group export description to reference runouts.

### Security Implications

### Acceptance Validation

Swagger docs updated as expected.

Screen Shot 2020-11-18 at 10 24 54 AM


bcda/models/SwaggerStructs.go
```diff
@@ -179,6 +179,7 @@ type GroupIDParam struct {
 	// ID of group export
 	// in: path
 	// required: true
+	// enum: all,runout
```
msnook, line 182: What happens if this is requested in `prod` now?  Won't it fail, since there are zero runout files ingested for the ACOs as of today?

mtrang1263: It'll fail for all of the non-synthetic ACOs. I think that'll be OK though since we want to expose this functionality in sandbox

msnook (approved):
👍

SMLuthi (approved):
LGTM!

PR #592, opened 2020-11-20
BCDA-3968 Feature: CLI for creating CCLF runout files from CCLF Zips

SMLuthi:
### Fixes [BCDA-3968](https://jira.cms.gov/browse/BCDA-3968)

Due to CCLF files not being dropped until halfway through a month there will be an interim period in January were we would be missing Runout CCLF files. To prevent the API from consuming incorrect data, we need to generate these runout files before the end of December using the December CCLF files.

### Proposed Changes

Create a CLI option in `bcdacli` that will take a directory containing CCLF files and clone these files to produce CCLF Runout files (using the proper nomenclature). A jenkins job can then be created and hook into this CLI option (done via a separate ticket).

### Change Details

- Create CLI option `generate-cclf-runout-files` that takes a `--directory` argument
- Create helper functions to Clone the zip files by reading all the CCLF files and generating a new zip file with the proper runout name.
- Read in the files within the original zip files, verify the file names within, and write them to the new runout CCLF file with the proper name.
- Make sure we dont clone files that dont match our naming schema.

### Security Implications

We are logging the name of the files that are being generated but I am assuming this is NOT PII.

### Acceptance Validation

This is being tested via unittests.

It has also been run manually using the following command:

`docker-compose run --rm api go run main.go generate-cclf-runout-files --directory test`

This is being passed a `test` directory containing a test CCLF zip file containing one CCLF file.
These are the results of the cloning (and unzipping the CCLF zip files):

image

### Feedback Requested

As always, any feedback would be great. I think I covered all the bases but let me know if I wasn't understanding something of the problem correctly!

### DIFF
--- a/bcda/bcdacli/cli.go
+++ b/bcda/bcdacli/cli.go
@@ -358,10 +358,12 @@ func setUpApp() *cli.App {
 				},
 			},
 			Action: func(c *cli.Context) error {
-				if err := cloneCCLFZips(filePath); err != nil {
+				rc, err := cloneCCLFZips(filePath)
+				if err != nil {
 					fmt.Fprintf(app.Writer, "%s\n", err)
 					return err
 				}
+				fmt.Fprintf(app.Writer, "Completed CCLF runout file generation. Generated %d zip files.", rc)
 				return nil
 			},
 		},
@@ -690,12 +692,13 @@ func renameCCLF(name string) string {
 	return cclfregex.ReplaceAllString(name, "${1}R${2}")
 }
 
-func cloneCCLFZips(path string) error {
+func cloneCCLFZips(path string) (int, error) {
 	files, err := ioutil.ReadDir(path)
 	if err != nil {
-		return err
+		return 0, err
 	}
 
+	rcount := 0 // Track the number of runout files that are created
 	// Iterate through all cclf zip files in provided directory
 	for _, f := range files {
 		// Make sure to not clone non CCLF files in case the wrong directory is given
@@ -706,11 +709,12 @@ func cloneCCLFZips(path string) error {
 		fn := renameCCLF(f.Name())
 		err := cloneCCLFZip(filepath.Join(path, f.Name()), filepath.Join(path, fn))
 		if err != nil {
-			return err
+			return rcount, err
 		}
+		rcount++
 		log.Infof("Created runout file: %s", fn)
 	}
-	return nil
+	return rcount, nil
 }
 
 func cloneCCLFZip(src, dst string) error {
--- a/bcda/bcdacli/cli.go
+++ b/bcda/bcdacli/cli.go
@@ -726,7 +726,11 @@ func CloneCCLFZip(path, name string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	defer newf.Close()
+	defer func() {
+		if err = newf.Close(); err != nil {
+			log.Fatal(err)
+		}
+	}()
 
 	zw := zip.NewWriter(newf)
 	defer zw.Close()
@@ -743,7 +747,7 @@ func CloneCCLFZip(path, name string) (string, error) {
 		if err != nil {
 			return filename, err
 		}
-		_, err = io.Copy(w, r)
+		_, err = io.Copy(w, r) // #nosec G110
 		if err != nil {
 			return filename, err
 		}
--- a/bcda/bcdacli/cli.go
+++ b/bcda/bcdacli/cli.go
@@ -1,9 +1,12 @@
 package bcdacli
 
 import (
+	"archive/zip"
 	"bufio"
 	"encoding/json"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net/http"
 	"os"
 	"path/filepath"
@@ -344,6 +347,33 @@ func setUpApp() *cli.App {
 			},
 		},
 		{
+			Name:     "generate-cclf-runout-files",
+			Category: "Data import",
+			Usage:    "Clone CCLF files and rename them as runout files",
+			Flags: []cli.Flag{
+				cli.StringFlag{
+					Name:        "directory",
+					Usage:       "Directory where CCLF files are located",
+					Destination: &filePath,
+				},
+			},
+			Action: func(c *cli.Context) error {
+				files, err := ioutil.ReadDir(filePath)
+				if err != nil {
+					fmt.Fprintf(app.Writer, "%s\n", err)
+					return err
+				}
+
+				for _, f := range files {
+					if err := cloneZipFile(filePath, f.Name()); err != nil {
+						fmt.Fprintf(app.Writer, "%s\n", err)
+						return err
+					}
+				}
+				return nil
+			},
+		},
+		{
 			Name:     "import-suppression-directory",
 			Category: "Data import",
 			Usage:    "Import all 1-800-MEDICARE suppression data files from the specified directory",
@@ -657,4 +687,48 @@ func setBlacklistState(cmsID string, blacklistState bool) error {
 	defer db.Close()
 
 	return db.Model(&aco).Update("blacklisted", blacklistState).Error
+}
+
+// CCLF file name pattern and regular expression
+const cclfPattern = `((?:T|P)\.BCD\..*\.ZC\d*)Y(\d{2}\.D\d{6}\.T\d{7})`
+
+var cclfRe = regexp.MustCompile(cclfPattern)
+
+func cloneZipFile(path, name string) error {
+	zr, err := zip.OpenReader(filepath.Join(path, name))
+	if err != nil {
+		return err
+	}
+	defer zr.Close()
+
+	zfname := cclfRe.ReplaceAllString(name, "${1}R${2}")
+	zfile, err := os.Create(filepath.Join(path, zfname))
+	if err != nil {
+		return err
+	}
+	defer zfile.Close()
+
+	zw := zip.NewWriter(zfile)
+
+	for _, f := range zr.File {
+		if err = cloneCCLFFile(f, zw); err != nil {
+			return err
+		}
+	}
+
+	return zw.Close()
+}
+
+func cloneCCLFFile(f *zip.File, zw *zip.Writer) error {
+	r, err := f.Open()
+	if err != nil {
+		return err
+	}
+	rfname := cclfRe.ReplaceAllString(f.Name, "${1}R${2}")
+	w, err := zw.Create(rfname)
+	if err != nil {
+		return err
+	}
+	io.Copy(w, r)
+	return r.Close()
 }
--- a/bcda/bcdacli/cli.go
+++ b/bcda/bcdacli/cli.go
@@ -358,7 +358,7 @@ func setUpApp() *cli.App {
 				},
 			},
 			Action: func(c *cli.Context) error {
-				if err := CloneCCLFZips(filePath); err != nil {
+				if err := cloneCCLFZips(filePath); err != nil {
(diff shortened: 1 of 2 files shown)


bcda/bcdacli/cli.go
```diff
@@ -658,3 +680,71 @@ func setBlacklistState(cmsID string, blacklistState bool) error {
 
 	return db.Model(&aco).Update("blacklisted", blacklistState).Error
 }
+
+// CCLF file name pattern and regex
+const cclfPattern = `((?:T|P)\.BCD\..*\.ZC\d*)Y(\d{2}\.D\d{6}\.T\d{7})`
+
+var cclfregex = regexp.MustCompile(cclfPattern)
+
+func RenameCCLF(name string) string {
```
mtrang1263, line 689: Do you see value in having this available outside of the cli package? If not, I think we should not export it and make it `renameCCLF`.

bcda/bcdacli/cli.go
```diff
@@ -658,3 +680,71 @@ func setBlacklistState(cmsID string, blacklistState bool) error {
 
 	return db.Model(&aco).Update("blacklisted", blacklistState).Error
 }
+
+// CCLF file name pattern and regex
+const cclfPattern = `((?:T|P)\.BCD\..*\.ZC\d*)Y(\d{2}\.D\d{6}\.T\d{7})`
```
mtrang1263, line 685: Great pattern! Able to match zip and sub files!

bcda/bcdacli/cli.go
```diff
@@ -658,3 +680,71 @@ func setBlacklistState(cmsID string, blacklistState bool) error {
...
+		fn, err := CloneCCLFZip(path, f.Name())
+		if err != nil {
+			return err
+		}
+		log.Infof("Created runout file: %s", fn)
+	}
+	return nil
+}
+
+func CloneCCLFZip(path, name string) (string, error) {
```
mtrang1263, line 715: nit - would function make more sense with the `(src, dest string)` args? It would make this look more like io.Copy(). So it could be called with something like this: `paths.Join(path, f.Name()), paths.Join(path, RenameCCLF(f.Name())`

mtrang1263: Do you see value in having this available outside of the cli package? If not, I think we should not export it and make it cloneCCLFZip.

bcda/bcdacli/cli.go
```diff
@@ -658,3 +680,74 @@ func setBlacklistState(cmsID string, blacklistState bool) error {
...
+	zw := zip.NewWriter(newf)
+	defer zw.Close()
+
+	// For each CCLF file in src zip, rename and write to dst zip
+	for _, f := range zr.File {
+		r, err := f.Open()
+		if err != nil {
+			return filename, err
+		}
+		defer r.Close()
```
mtrang1263, line 740: Something to note in the future with using `defer...`, if we have unbounded number of files in the zip (we don't for CCLF), we may end up in a bad state since do not close the file until the very end of the function. So we may end up running out of file descriptors (or something else (OOM?))

If there that is a concern and you still want to leverage defer, you can always wrap the entire code block in an anonymous function.

```
r, err := func() (string, error) {
                f.Open()
		if err != nil {
			return filename, err
		}
		defer r.Close()
		w, err := zw.Create(RenameCCLF(f.Name))
		if err != nil {
			return filename, err
		}
		_, err = io.Copy(w, r)
		if err != nil {
			return filename, err
		}()
```

It **does not apply to CCLF** since we know it should have at most 10 files.

bcda/bcdacli/cli_test.go
```diff
@@ -836,3 +834,110 @@ func getRandomPort(t *testing.T) int {
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return w.Close()
+}
+
+func getFileCount(path string) int {
+	f, _ := ioutil.ReadDir(path)
```
mtrang1263, line 867: We shouldn't ignore this error. To keep the return value clean, you can pass in a test object and use the `assert.NoError(t, err)` here. That way you can still return the file count while not ignoring the error.

bcda/bcdacli/cli_test.go
```diff
@@ -836,3 +834,110 @@ func getRandomPort(t *testing.T) int {
 
 	return listener.Addr().(*net.TCPAddr).Port
 }
+
+func CreateTestZipFile(file string) error {
```
mtrang1263, line 838: Do you see value in having this available outside of the cli package? If not, I think we should not export it and make it createTestZipFile.

mtrang1263: Nit - should we move these helper functions beneath the two test functions?

bcda/bcdacli/cli.go
```diff
@@ -343,6 +346,25 @@ func setUpApp() *cli.App {
...
+			Flags: []cli.Flag{
+				cli.StringFlag{
+					Name:        "directory",
+					Usage:       "Directory where CCLF files are located",
+					Destination: &filePath,
+				},
+			},
+			Action: func(c *cli.Context) error {
+				if err := CloneCCLFZips(filePath); err != nil {
+					fmt.Fprintf(app.Writer, "%s\n", err)
```
mtrang1263, line 362: Good to know about reporting these errors!

bcda/bcdacli/cli_test.go
```diff
@@ -836,3 +834,110 @@ func getRandomPort(t *testing.T) int {
...
+	zf, err := os.Create(file)
+	if err != nil {
+		return err
+	}
+	defer zf.Close()
+
+	w := zip.NewWriter(zf)
+	var files = []string{
```
mtrang1263, line 846: To keep isolation, should the caller of this function supply the files to create? So we change the `createTestZipFile` to `createTestZipFile(parent string, children ...string)` Then the caller of the function has an expectation on what file should exist in the zip.

bcda/bcdacli/cli_test.go
```diff
@@ -836,3 +834,110 @@ func getRandomPort(t *testing.T) int {
...
+
+	return w.Close()
+}
+
+func getFileCount(path string) int {
+	f, _ := ioutil.ReadDir(path)
+	return len(f)
+}
+
+func TestCloneCCLFZips(t *testing.T) {
```
mtrang1263, line 871: Nit - I think we can consolidate `TestCloneCCLFZips` and `TestCloneCCLFZip`

If we test it by invoking using testApp.run (ex: https://github.com/CMSgov/bcda-app/blob/master/bcda/bcdacli/cli_test.go#L115), we can combine the assertions that you're making in both of these functions. 1: Confirming number of zip files, 2. Confirming number of subfiles within each zip.

bcda/bcdacli/cli_test.go
```diff
@@ -836,3 +834,110 @@ func getRandomPort(t *testing.T) int {
...
+		"T.BCD.A0002.ZCY18.D181120.T9999992",
+		"not_a_cclf_file",
+	}
+
+	for _, srcZip := range srcZips {
+		err := CreateTestZipFile(srcZip)
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer func(f string) { os.Remove(f) }(srcZip)
```
mtrang1263, line 884: May want to consider creating a temporary directory (https://golang.org/pkg/io/ioutil/#TempDir) and putting all of these files underneath the temp dir.

So at the beginning you can do something like this.
```
dir, err := ioutil.TempDir("", "clone_cclf")
	if err != nil {
		log.Fatal(err)
	}
defer os.RemoveAll(dir)
```

So worse comes to worse, we fail to clean up the temporary directory. If you run these tests locally, it'll create these files in bcda-app, which would have to be manually cleaned up.
```

**Review by mtrang1263 — COMMENTED** [review]
Mostly nits, but I think we can improve on what is exposed outside of the CLI package.

Impl looks good to me!

File: bcda/bcdacli/cli.go [hunk 11]
```diff
@@ -658,3 +680,74 @@ func setBlacklistState(cmsID string, blacklistState bool) error {
...
+		if err != nil {
+			return filename, err
+		}
+		defer r.Close()
+
+		w, err := zw.Create(RenameCCLF(f.Name))
+		if err != nil {
+			return filename, err
+		}
+		_, err = io.Copy(w, r)
```
mtrang1263, line 746: To avoid the gosec you want to add the following:
```
/* #nosec G110 - not concerned with OOM since we're using this in a controlled setting */ 
```

Reference: https://github.com/CMSgov/bcda-app/blob/master/test/smoke_test/bcda_client.go#L161

bcda/bcdacli/cli.go
```diff
@@ -658,3 +680,74 @@ func setBlacklistState(cmsID string, blacklistState bool) error {
...
+	}
+	defer zr.Close()
+
+	// Create destination runout zip file with proper nomenclature
+	filename := RenameCCLF(name)
+	newf, err := os.Create(filepath.Join(path, filename))
+	if err != nil {
+		return "", err
+	}
+	defer newf.Close()
```
mtrang1263: Didn't realize it was unsafe, but the gosec error seems valid and should be addressed. One approach would be to use named returns. So something like this:
```
func CloneCCLFZip(path, name string) (newFile string, err error)
```

and then this is updated to something like this
```
defer func() {
    if err1 := newf.Close(); err1 != nil && err == nil {
            err = err1
    }
}()
```

So we won't ignore the main error, but if we have an error that was caused by closing the file, we'll return it out.

mtrang1263: Looking more into this, there's a helper function that you can use `utils.CloseFileAndLogError`

bcda/bcdacli/cli_test.go
SMLuthi: I consolidated this into one giant test. I'm not 100% sold on this being the cleanest way to implement a test (though I see the value in using `s.testApp.Run(args)`). But, regardless, it tests all aspects of the implementation.

bcda/bcdacli/cli_test.go
SMLuthi: This was a great idea. Made clean up much simpler and much safer now that I only have to manage cleaning up one folder and not a bunch of files.

bcda/bcdacli/cli.go
```diff
@@ -686,51 +688,50 @@ const cclfPattern = `((?:T|P)\.BCD\..*\.ZC\d*)Y(\d{2}\.D\d{6}\.T\d{7})`
...
 	if err != nil {
-		return "", err
+		return err
 	}
 	defer zr.Close()
 
 	// Create destination runout zip file with proper nomenclature
-	filename := RenameCCLF(name)
-	newf, err := os.Create(filepath.Join(path, filename))
+	// filename := renameCCLF(name)
```
mtrang1263, line 729: Remove?

bcda/bcdacli/cli.go
```diff
@@ -658,3 +682,77 @@ func setBlacklistState(cmsID string, blacklistState bool) error {
...
+// CCLF file name pattern and regex
+const cclfPattern = `((?:T|P)\.BCD\..*\.ZC\d*)Y(\d{2}\.D\d{6}\.T\d{7})`
+
+var cclfregex = regexp.MustCompile(cclfPattern)
+
+func renameCCLF(name string) string {
+	return cclfregex.ReplaceAllString(name, "${1}R${2}")
+}
+
+func cloneCCLFZips(path string) (int, error) {
```
mtrang1263, line 695: Nice change!

mtrang1263 (approved):
LGTM! Just one comment that should be removed before this is merged.

mtrang1263 (approved):

PR #593, opened 2020-11-30
BCDA-3936 - Fix smoke tests to report errors

mtrang1263:
### Fixes [BCDA-3936](https://jira.cms.gov/browse/BCDA-3936)

We discovered an issue with how we were triggering our smoke tests. We were not reporting failures because calling `wait` without a PID always returns a 0 exit code (success).

With this change, we encountered several errors when running smoke tests via Github actions (timeouts, etc.). Now that they are being reported, we need to update our client to handle these transient errors.

### Change Details
1. Call `wait PID` to ensure that errors are correctly reported.
2. Refactored bcda_client.go to handle transient errors (timeouts, etc.).

### Security Implications

### Acceptance Validation

1. Verified that errors in smoke_test.sh are correctly reported.


test/smoke_test/bcda_client.go
```diff
@@ -44,46 +40,51 @@ func init() {
 	flag.StringVar(&apiVersion, "apiVersion", "v1", "resourceType to test")
 	flag.IntVar(&timeout, "timeout", 300, "amount of time to wait for file to be ready and downloaded.")
 	flag.StringVar(&endpoint, "endpoint", "", "base type of request endpoint in the format of Patient or Group/all or Group/new")
+	flag.IntVar(&httpRetry, "httpRetry", 3, "amount of times to retry an http request")
```
PatKelsh, line 43: could this potentially be an environmental config?

test/smoke_test/bcda_client.go
```diff
@@ -99,70 +100,132 @@ func startJob(endpoint, resourceType string) *http.Response {
 
 	req.Header.Add("Prefer", "respond-async")
 	req.Header.Add("Accept", "application/fhir+json")
-	req.Header.Add("Authorization", fmt.Sprintf("Bearer %s", accessToken))
 
-	resp, err := client.Do(req)
+	resp, err := c.Do(req)
 	if err != nil {
-		panic(err)
+		return "", err
```
PatKelsh, line 106: This looks like a much better way to handle errors

test/smoke_test/bcda_client.go
mtrang1263: It can be, but since bcda_client.go is an executable (with the presence of the func main), we're able to leverage the [flags package](https://golang.org/pkg/flag/) to supply these config options instead of using environment variables.

When we invoke the bcda_client via `go run`, we can supply the `httpRetry` flag whose value is controlled by an external environment variable. There's some examples of this in the [bulk_data_requests.sh](https://github.com/CMSgov/bcda-app/blob/master/test/smoke_test/bulk_data_requests.sh#L7).

test/smoke_test/bcda_client.go
```diff
@@ -99,70 +100,132 @@ func startJob(endpoint, resourceType string) *http.Response {
...
 	}
+	defer resp.Body.Close()
 
-	return resp
+	switch resp.StatusCode {
```
SMLuthi, line 110: Since this switch is only handling two cases (accepted and default) is it worth keeping it like this or could it simply be an if else.

SMLuthi (approved):
This lgtm, just a small question

test/smoke_test/bcda_client.go
PatKelsh: ok, makes sense

PatKelsh (approved):

SMLuthi (approved):
LGTM!

PR #598, opened 2020-12-09
bcda-3838 Feature: Update errors to use MBI id

SMLuthi:
### Fixes [BCDA-3838](https://jira.cms.gov/browse/BCDA-3838)

Errors that are written to the `errors.ndjson` either use the BB id or the beneficiary id. Neither of these are useful to us and the error should be updated to use another id that can be helpful.

### Proposed Changes

- Update error message to use the MBI identifier instead.

### Change Details

- Error message was updated to use MBI id.
- `handleBBError` func was moved entirely as it wasn't necessary for the couple times it was called. The logic was moved into the top level function where the `errorCount` lives (and we dont have to to pass a pointer around anymore).
- `beneBBID` was renamed to `getBeneficiary` since it now returns the entire beneficiary instead of just the BB id.
- Top level code was updated to use a closure method that passes back the err and error message that will be written to the file.

### Security Implications

### Acceptance Validation

Unittests have been updated to check for the correct error messages.

### Feedback Requested

Since we are now check for the MBI I changed the `TestWriteEOBDataToFile_BlueButtonIDNotFound` in `main_test.go` to search for the cclf bene when making assertions. I haven't done a lot of db searches so I just wanted to make sure this a solid choice. Or would it be smarter to keep a slice of `cclfBeneficiarys` from the beginning (when they are created) and just iterate over those later when making assertions...

### DIFF
--- a/bcdaworker/main.go
+++ b/bcdaworker/main.go
@@ -218,20 +218,25 @@ func writeBBDataToFile(ctx context.Context, bb client.APIClient, db *gorm.DB, cm
 	failed := false
 
 	for _, beneID := range jobArgs.BeneficiaryIDs {
-		blueButtonID, err := beneBBID(beneID, bb, db)
-
-		if err != nil {
-			handleBBError(ctx, err, &errorCount, fileUUID, fmt.Sprintf("Error retrieving BlueButton ID for cclfBeneficiary %s", beneID), jobArgs.ID)
-		} else {
-			b, err := bundleFunc(blueButtonID)
+		bundle := func() (string, error) {
+			bene, err := getBeneficiary(beneID, bb, db)
 			if err != nil {
-				handleBBError(ctx, err, &errorCount, fileUUID,
-					fmt.Sprintf("Error retrieving %s for beneficiary %s in ACO %s", jobArgs.ResourceType, blueButtonID, jobArgs.ACOID),
-					jobArgs.ID)
-			} else {
-				fhirBundleToResourceNDJSON(ctx, w, b, jobArgs.ResourceType, beneID, cmsID, fileUUID, jobArgs.ID)
+				return fmt.Sprintf("Error retrieving BlueButton ID for cclfBeneficiary MBI %s", bene.MBI), err
 			}
+			b, err := bundleFunc(bene.BlueButtonID)
+			if err != nil {
+				return fmt.Sprintf("Error retrieving %s for beneficiary MBI %s in ACO %s", jobArgs.ResourceType, bene.MBI, jobArgs.ACOID), err
+			}
+			fhirBundleToResourceNDJSON(ctx, w, b, jobArgs.ResourceType, beneID, cmsID, fileUUID, jobArgs.ID)
+			return "", nil
 		}
+
+		if errMsg, err := bundle(); err != nil {
+			log.Error(err)
+			errorCount++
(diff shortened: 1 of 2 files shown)


bcdaworker/main.go
```diff
@@ -218,20 +218,25 @@ func writeBBDataToFile(ctx context.Context, bb client.APIClient, db *gorm.DB, cm
 	failed := false
 
 	for _, beneID := range jobArgs.BeneficiaryIDs {
-		blueButtonID, err := beneBBID(beneID, bb, db)
-
-		if err != nil {
-			handleBBError(ctx, err, &errorCount, fileUUID, fmt.Sprintf("Error retrieving BlueButton ID for cclfBeneficiary %s", beneID), jobArgs.ID)
-		} else {
-			b, err := bundleFunc(blueButtonID)
+		bundle := func() (string, error) {
```
mtrang1263, line 221: Do you need this to be of type `func()`? If you just need the result, you can call the anonymous function directly and get the result.

```
errMsg, err := func() (string, err) {
...
}()
```
The last () will invoke the  function

bcdaworker/main_test.go
```diff
@@ -254,8 +254,8 @@ func (s *MainTestSuite) TestWriteEOBDataToFileWithErrorsAboveFailureThreshold()
...
-{"resourceType":"OperationOutcome","issue":[{"severity":"error","code":"exception","details":{"coding":[{"system":"http://hl7.org/fhir/ValueSet/operation-outcome","code":"Blue Button Error","display":"Error retrieving ExplanationOfBenefit for beneficiary a1000065301 in ACO %s"}],"text":"Error retrieving ExplanationOfBenefit for beneficiary a1000065301 in ACO %s"}}]}`, s.testACO.UUID, s.testACO.UUID, s.testACO.UUID, s.testACO.UUID)
+	ooResp := fmt.Sprintf(`{"resourceType":"OperationOutcome","issue":[{"severity":"error","code":"exception","details":{"coding":[{"system":"http://hl7.org/fhir/ValueSet/operation-outcome","code":"Blue Button Error","display":"Error retrieving ExplanationOfBenefit for beneficiary MBI a1000089833 in ACO %s"}],"text":"Error retrieving ExplanationOfBenefit for beneficiary MBI a1000089833 in ACO %s"}}]}
```
mtrang1263, line 257: My guess is that these IDs here need to be updated.

mtrang1263 (approved):
LGTM - just a couple comments, but nothing blocking.

bcdaworker/main_test.go
```diff
@@ -303,6 +303,8 @@ func (s *MainTestSuite) TestWriteEOBDataToFile_BlueButtonIDNotFound() {
 
 	errorFileScanner := bufio.NewScanner(bytes.NewReader(d))
 	for _, cclfBeneID := range cclfBeneficiaryIDs {
+		var cclfBeneficiary models.CCLFBeneficiary
```
mtrang1263, line 306: :+1: This approach looks good to me

bcdaworker/main_test.go
SMLuthi: I thought so too but... apparently they dont. I assume whatever is creating these test entries is just using the same id for MBI as it is for the bene id.

SMLuthi: Ah yes, on lines `232` and `234` you can see it setting the MBI to the test beneficiary ID's

mtrang1263 (approved):

PR #601, opened 2020-12-11
bcda-3942 Feature: Upgrade to gorm v2

SMLuthi:
### Fixes [BCDA-3942](https://jira.cms.gov/browse/BCDA-3942)

Gorm is out of date. We've been using v1 for awhile and we need to upgrade to v2.

### Proposed Changes

Upgrade the package and fix any breaking changes.

### Change Details

Pull the new gorm v2 package from `"gorm.io/gorm"`.

The following fixes had to be applied:

- The gorm DB struct no longer supports the `Close()` method. We now pull the sql DB struct from the gorm db and close it directly. There is a `Close` method in our database utility file. This has been appropriated where needed.
- When doing a gorm `db.DB()` call it now requires you to handle an error return.
- The `RecordNotFound` error isnt a method call anymore on the db query return. You have to call `Error` and check it directly.
- The `Count` method on a query requires the integer to be an `int64`... not just an `int`.
- `gorm.Open(...)` has changed to take a dialect and config rather than a string with the type of sql database and a database object. To adjust for this the sqlmock new call had to be changed to a `NewWithDSN`. This allows us to give a dsn directly to the mock AND be able to provide it for the gorm open call as well.
- The explicit `Ping()` check in `database.go` connection setup has been removed. Gorm now does this automatically when setting up a new connection and will throw an error accordingly.

### Security Implications

### Acceptance Validation

This is being tested via all the unittests and smoke tests.

### Feedback Requested

I think I hit all the spots and didn't create any weird oddities with the database. For future reference, I dont think we are doing a lot of heavy lifting with gorm. Most of the queries are basic and a lot of them are actually just raw sql. In the future, we could probably move ahead and remove gorm entirely and replace it with something a lot more light weight that allows for more versatility (like this query builder: https://github.com/Masterminds/squirrel which can run on pgx). The main lift would be updating our models with functions that describe how to unpack the data into the models themselves.

As always, let me know if theres a better way to do what Im doing in Go, feedback is always appreciated!


bcda/api/requests.go
```diff
@@ -173,21 +179,24 @@ func (h *Handler) bulkRequest(resourceTypes []string, w http.ResponseWriter, r *
...
-				log.Error(err)
-				oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, responseutils.Processing, "")
-				responseutils.WriteError(oo, w, http.StatusInternalServerError)
+				resourceTypes = types
 			}
-
-			return
-		} else {
-			resourceTypes = types
 		}
```
mtrang1263, line 200: I can't comment on this unfortunately, but in the JobStatus function, we are using db.Find(), I believe we're relying on the ErrRecordNotFound, which I think requires us to use db.First()

bcda/auth/models_test.go
```diff
@@ -7,10 +7,10 @@ import (
...
 	"github.com/CMSgov/bcda-app/bcda/database"
 	"github.com/CMSgov/bcda-app/bcda/testUtils"
-	"github.com/jinzhu/gorm"
 	"github.com/pborman/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/suite"
+	"gorm.io/gorm"
 )
 
 type ModelsTestSuite struct {
```
mtrang1263, line 16: In `TestCMSID` I think we're misusing, the db.Find(), I think that Find + Pluck doesn't really play well. I think we should Model + Where + Pluck

bcda/api/v1/api.go
```diff
@@ -113,7 +113,7 @@ func JobStatus(w http.ResponseWriter, r *http.Request) {
 	defer database.Close(db)
 
 	var job models.Job
-	err := db.Find(&job, "id = ?", jobID).Error
+	err := db.First(&job, "id = ?", jobID).Error
```
mtrang1263, line 116: Nit - Since you're looking by the primary key, you don't need to specify the `id = ?` argument. 

Ref: https://gorm.io/docs/query.html#Retrieving-objects-with-primary-key

bcda/database/utils.go
```diff
@@ -3,12 +3,16 @@ package database
...
-	"github.com/jinzhu/gorm"
 	log "github.com/sirupsen/logrus"
+	"gorm.io/gorm"
 )
 
 func Close(db *gorm.DB) {
-	if err := db.Close(); err != nil {
+	dbc, err := db.DB()
+	if err != nil {
+		log.Infof("failed to retrieve db connection: %v", err)
```
mtrang1263, line 13: We should do an early return if we fail to get the connection. Otherwise, `dbc` may be nil and we'll panic.

bcda/auth/okta_test.go
```diff
@@ -25,14 +27,17 @@ type OktaAuthPluginTestSuite struct {
 
 func (s *OktaAuthPluginTestSuite) SetupSuite() {
 	db := database.GetGORMDbConnection()
+
 	defer func() {
-		if err := db.Close(); err != nil {
+		dbc, _ := db.DB()
```
mtrang1263, line 32: Should we be using `database.Close()` instead?

mtrang1263 (commented):
Since this is affecting quite a bit of the codebase, might be good to deploy to dev to verify everything is working as expected

bcda/auth/okta_test.go
```diff
@@ -33,6 +33,7 @@ func (s *OktaAuthPluginTestSuite) SetupSuite() {
 		if err := dbc.Close(); err != nil {
 			assert.Failf(s.T(), err.Error(), "okta plugin test")
 		}
```
mtrang1263: Now that you're calling database.Close(), we don't need this code, right?

mtrang1263: Including line 32

mtrang1263 (approved):
LGTM!

PR #602, opened 2020-12-18
Relax regex requirements to handle CEC/NGACOs

mtrang1263:
We encountered an issue when attempting to copy the CCLF files for NGACOs. The regex for the NGACO file names did not match the expectations.

To resolve this, we relaxed the regex requirements.

### Change Details
1. Relax the regex requirements for renaming the CCLF files.
2. Adding tests with NGACO/CEC files to verify the rename works as expected.

### Security Implications

### Acceptance Validation
1. Added unit tests to verify CEC/NGACO files would be renamed.
2. Verified fix by running the updated CLI on our NGACO files. They were renamed correctly.

### DIFF
--- a/bcda/bcdacli/cli.go
+++ b/bcda/bcdacli/cli.go
@@ -684,7 +684,7 @@ func setBlacklistState(cmsID string, blacklistState bool) error {
 }
 
 // CCLF file name pattern and regex
-const cclfPattern = `((?:T|P)\.BCD\..*\.ZC\d*)Y(\d{2}\.D\d{6}\.T\d{7})`
+const cclfPattern = `((?:T|P).*\.ZC[A-B0-9]*)Y(\d{2}\.D\d{6}\.T\d{7})`
 
 var cclfregex = regexp.MustCompile(cclfPattern)
 
--- a/bcda/bcdacli/cli_test.go
+++ b/bcda/bcdacli/cli_test.go
@@ -862,6 +862,8 @@ func (s *CLITestSuite) TestCloneCCLFZips() {
 	// cclf file names that are contained within the cclf zip files
 	cclfFiles := []string{
 		"T.BCD.A0001.ZC48Y18.D181120.T1000001",
+		"T.BCD.A0001.ZCAY18.D181120.T1000001",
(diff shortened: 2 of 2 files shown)


bcda/bcdacli/cli.go
```diff
@@ -684,7 +684,7 @@ func setBlacklistState(cmsID string, blacklistState bool) error {
 }
 
 // CCLF file name pattern and regex
-const cclfPattern = `((?:T|P)\.BCD\..*\.ZC\d*)Y(\d{2}\.D\d{6}\.T\d{7})`
+const cclfPattern = `((?:T|P).*\.ZC[A-B0-9]*)Y(\d{2}\.D\d{6}\.T\d{7})`
```
SMLuthi, line 687: Im assuming the check for `[A-B...` will always directly follow `ZC` maybe something like this will be more specific.
`((?:T|P).*\.ZC[A-B]?[0-9]*)Y(\d{2}\.D\d{6}\.T\d{7})`

bcda/bcdacli/cli_test.go
```diff
@@ -855,13 +855,17 @@ func (s *CLITestSuite) TestCloneCCLFZips() {
...
+		"P.BCD.V002.ZCY20.D201002.T0811490",
 		"not_a_cclf_file",
 	}
 	// cclf file names that are contained within the cclf zip files
 	cclfFiles := []string{
 		"T.BCD.A0001.ZC48Y18.D181120.T1000001",
 		"T.BCD.A0001.ZC48Y18.D181120.T1000002",
 		"T.BCD.A0001.ZC48Y18.D181120.T1000003",
+		"P.V001.ACO.ZC8Y20.D201002.T0806400",
+		"P.CEC.ZC8Y20.D201108.T0958300",
```
SMLuthi, line 868: Maybe include a test file to verify for the A/B's?

SMLuthi (changes requested):

bcda/bcdacli/cli.go
mtrang1263: Since we can have digits OR A/B, we'll leave the regex as is

SMLuthi (approved):
LGTM!

PR #606, opened 2020-12-31
BCDA-3776 - Update swagger to generate v1 and v2 API definitions

mtrang1263:
### Fixes [BCDA-3776](https://jira.cms.gov/browse/BCDA-3776)

We need to update our documentation process to publish swagger definitions for our V2 endpoints. We will publish a dedicate set of V2 swagger docs the live alongside the v1 docs.

### Change Details
1. Remove existing swagger-ui code from `./swaggerui`. Since the UI references a single swagger.json file, we'll download the UI from [swagger-ui repository](https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/installation.md#plain-old-htmlcssjs-standalone). By downloading the UI, we do not have to duplicate the UI resources for the v2 swagger definitions.
2. Update Makefile to account for race condition between documentation and openapi builds. The openapi requires that documentation completes. The `depends_on` semantic in docker-compose does not guarantee completion. 
3. Add new ResourceType definition for v2 endpoints. This parameter is required since V2 does not support EOB resources.
4. Remove duplicate Optional qualifier for Swagger fields. Since the UI displays what fields are required, we do not need to specify what fields are optional.
5. Adding Postman test to verify swagger endpoints. We had tested this in a unit test, but required some special path resolution to pass. By adding this to the Postman test suite, it will validate the we are generating the swagger page as expected.
### Security Implications

We are downloading the latest version of the swagger-ui. We were previously using a version from a couple years ago (https://github.com/CMSgov/bcda-app/pull/32)

### Acceptance Validation
1. Confirmed that we are now publishing v1/v2 swagger definitions.
2. Confirmed /api/v1/swagger does not contain v2 endpoints
3. Confirmed that /api/v2/swagger contains all of the methods needed to request v2 data. To request v2 data, we need to include the auth endpoints as well as the job endpoints.
3. Confirmed that our generated OpenAPI file includes v1 and v2 definitions.

Screen Shot 2021-01-19 at 5 10 34 PM

Screen Shot 2021-01-19 at 5 10 23 PM

### Feedback Requested
1. Run `docker-compose run documentation` and verify that we are including the correct endpoints for v2 API.


bcda/web/router_test.go
```diff
@@ -76,20 +75,6 @@ func (s *RouterTestSuite) TestDataRoute() {
 	assert.Equal(s.T(), http.StatusUnauthorized, res.StatusCode)
 }
 
-func (s *RouterTestSuite) TestFileServerRoute() {
```
mtrang1263, line 79: Test replaced by: https://github.com/CMSgov/bcda-app/pull/606/files#diff-842f2e0111020a6f5af8ab0ad66d148de552df214ba184469f9587155f9d5337R175

Makefile
```diff
@@ -1,7 +1,8 @@
 package:
 	# This target should be executed by passing in an argument representing the version of the artifacts we are packaging
 	# For example: make package version=r1
-	docker-compose up --build documentation openapi
```
mtrang1263, line 4: Since openapi relies on documentation completing, we'll run them separately. Even though `openapi` depends on `documentation`, there's no guarantee that documentation finishes. See: https://docs.docker.com/compose/startup-order/

Makefile
```diff
@@ -140,6 +141,7 @@ docker-build:
 
 docker-bootstrap:
 	$(MAKE) docker-build
+	docker-compose up --build documentation
```
mtrang1263, line 144: Since openapi relies on documentation completing, we'll run them separately. Even though openapi depends on documentation, there's no guarantee that documentation finishes. See: https://docs.docker.com/compose/startup-order/

mtrang1263:
Going to rework this a bit to have the single swagger page w/ all API calls rendered.

bcda/web/router_test.go
adamayd: I can't get this link to work for me.  It takes me back to the top of this pull request.

adamayd: Thanks for the help guiding to the right spot.  LGTM.

adamayd (dismissed):
LGTM

katyasoup (approved):
LGTM!

SMLuthi (approved):
LGTM!

PR #609, opened 2021-01-05
BCDA-4082 - Remove fresh dependency

mtrang1263:
### Part of [BCDA-4082](https://jira.cms.gov/browse/BCDA-4082)

With our migration from `dep` to `go modules`, we need to remove the deprecated `fresh` dependency from bcda-app

### Change Details
1. Remove fresh dependency from our `./vendor` folder.
2. Update Dockerfiles to no longer rely on fresh for building and running the application. Instead of fresh, we'll run `go install` and launch the binary explicitly.
3. Update Makefile to reference new command location. Since we're running `go install`, the binaries are located in our PATH and can be referenced directly. When using fresh, it would place the binary in `tmp/`.

### Security Implications

We are removing our dependency on fresh.

### Acceptance Validation
CI Passes.

Since we only use fresh for testing, this PR does not affect how we deploy to our remote environments.


adamayd (approved):
LGTM!

katyasoup (approved):

PR #610, opened 2021-01-05
Introduce job status constants

mtrang1263:
### Needed for [BCDA-3879](https://jira.cms.gov/browse/BCDA-3879)

With [BCDA-3879](https://jira.cms.gov/browse/BCDA-3879), we need to introduce a new status to indicate that the failed data has been cleaned up. Instead of introducing another String to track this, we decided to introduce a new type that will allow us to easily add new statuses in the future.

### Change Details
1. Introduce a new data type JobStatus, which contains all of our currently supported states.
2. Update code base to replace the string representation of the status with the JobStatus equivalent.

### Security Implications

### Acceptance Validation
CI passes.

Since this is swap strings with constants, it does not need to be deployed to our remote environments.

### Feedback Requested

Please double check that I have not missed any other places.


SMLuthi (commented):
This is a test to see if I can leave a review

SMLuthi (approved):
LGTM!

PR #612, opened 2021-01-07
bcda-3792 Feature: filter empty ndjson files in jobstatus

SMLuthi:
### Fixes [BCDA-3792](https://jira.cms.gov/browse/BCDA-3792)

Currently, the jobstatus endpoint in the bcda api will pull in all files under the job and read them in. Some of these files could possibly not contain any data (due to query restrictions for example).

### Proposed Changes

Filter out these empty files before we read them in so they dont get processed.

### Change Details

1. The bcda worker will check file sizes after processing. If the file size is equal to 0 (ie empty) it will replace the filename with a default name of `blank.ndjson`
2. The bcda api, when serving the JobStatus endpoint will filter on the default blank name and ignore any of those files.

### Security Implications

### Acceptance Validation

API no longer returns empty NDJSON files when reporting job status. 

### Feedback Requested

I think I'd like to introduce some logging when we do encounter a blank file in the worker and log that information. Also, I'd like to figure out a way to test for when the worker processes a job that has a file without any contents to make sure the size check works as intended.


test/postman_test/BCDA_Tests_Sequential.postman_collection.json
```diff
@@ -1633,48 +1632,6 @@
...
-					"listen": "test",
-					"script": {
-						"id": "18400bb7-7974-49fd-a96e-4df89748d3c9",
-						"exec": [
-							"pm.test(\"Status code is 200\", function() {",
-							"    pm.response.to.have.status(200);",
-							"});",
-							"",
-							"pm.test(\"Body does not contain data\", function() {",
-							"    pm.expect(pm.response.length === 0)",
```
adamayd, line 1649: Ah, I get what you were talking about now.

adamayd (approved):
LGTM.  Sorry it took me a bit to figure it out.  I was trying to follow it, but didn't understand completely until I looked at the ticket and the test.

mtrang1263 (dismissed):
LGTM!

mtrang1263 (approved):

PR #613, opened 2021-01-07
BCDA-3880 Fix: Update archive job to handle transient error

adamayd:
### Fixes [BCDA-3880](https://jira.cms.gov/browse/BCDA-3880)

During the job archive process, we move the data files to the archive directory. After, we mark the job as archived. If this update fails, the job will be stuck in the Completed state because the data move will fail (since it does not exist in the data directory)

### Proposed Changes

* ~~Add checks for existing archive files on a job marked completed.~~
* Add checks for missing payload files on job marked completed.
* Mark job as archived if ~~archive~~ payload files are found.
* ~~Update~~ Added test for missing payload condition.
* Added threshold flag to command for use from CLI.

### Change Details

Refactored logic to include file checks moving variable creation outside of the threshold conditional.  If the ~~archive files are found~~ payload files are missing, it marks the job as archived and moves on allowing clean-up to remove the files as designed.  Added ~~a section in the `TestArchiveExpiring`~~ `TestArchiveExpiringWithoutPayloadDir` in `cli_test.go` to test for this ~~"orphaned" archive~~ missing files condition.  Also added the threshold optional flag to the  `archive-job-files` command which retains a default value of 24 and looks for the environment variable.  This allows updating the `TestArchiveExpiringWithThreshold` test to use the CLI instead of calling the `archiveExpiring` function directly.  

### Security Implications

### Acceptance Validation

Ran individual tests and full unit tests suite locally and all passed.

### Feedback Requested

Feedback requested/commented while in Draft PR: would like discussion about combining or DRYing out the unit tests between `TestArchiveExpiring` and `TestArchiveExpiringWithThreshold`.  Looking at the added flag and making sure that smells good and the logic for the missing files is okay (should there be an else condition if both directories somehow exist and how should that be handled).  Also I left in comments about moving the update status to a separate function, but not sure if that could get messy passing db connections around in functions.


bcda/bcdacli/cli.go
```diff
@@ -598,19 +610,35 @@ func archiveExpiring(hrThreshold int) error {
...
+			log.Info("Archived job files found, setting status to archived...")
+			// TODO: Separate Status update function - might get messy passing db connection.
+			j.Status = models.JobStatusArchived
+			err = db.Save(j).Error
 			if err != nil {
 				log.Error(err)
 				lastJobError = err
-				continue
+			}
+		} else if int(elapsed) >= hrThreshold {
```
mtrang1263, line 631: Instead of filtering out the jobs after we get them back from the DB, we should add an additional clause to our db lookup above. We're doing a similar approach in the `cleanupArchive` method here: https://github.com/CMSgov/bcda-app/blob/b14e1435ee87640bbed303149bdadc43ebbc0b7b/bcda/bcdacli/cli.go#L633

I think if we do it this way, the TODO above is addressed since all jobs satisfy the expiration criteria.

And since the database is handling the filtering, we do not have to check that ` int(elapsed) >= hrThreshold`

bcda/bcdacli/cli.go
```diff
@@ -598,19 +610,35 @@ func archiveExpiring(hrThreshold int) error {
...
-			expDir := fmt.Sprintf("%s/%d", os.Getenv("FHIR_ARCHIVE_DIR"), id)
-
-			err = os.Rename(jobDir, expDir)
+		id := int(j.ID)
+		jobDir := fmt.Sprintf("%s/%d", os.Getenv("FHIR_PAYLOAD_DIR"), id)
+		_, err = os.Stat(jobDir)
+		jobDirExist := err == nil
+		expDir := fmt.Sprintf("%s/%d", os.Getenv("FHIR_ARCHIVE_DIR"), id)
+		_, err = os.Stat(expDir)
+		expDirExist := err == nil
```
mtrang1263, line 619: I think we can simplify this logic a little bit. If a job exists in the `FHIR_PAYLOAD_DIR`, we can make an assumption that it does not exist in the `FHIR_ARCHIVE_DIR`, and vise-versa.

With that assumption, I think we reduce the logic below into a single conditional. If the payloadDir exists, then move it to the archiveDir, then update the job status. That way we handle the error scenario where the payloadDir was moved to the archiveDir, but the job failed to update in the database.

snippet (w/o error handling)
```
if payloadDirExist {
    os.Rename(...)
    ...
}
db.Model(&j).Update("status",  models.JobStatusArchive)
```

bcda/bcdacli/cli.go
```diff
@@ -598,19 +610,35 @@ func archiveExpiring(hrThreshold int) error {
...
 		elapsed := time.Since(t).Hours()
-		if int(elapsed) >= hrThreshold {
-
-			id := int(j.ID)
-			jobDir := fmt.Sprintf("%s/%d", os.Getenv("FHIR_PAYLOAD_DIR"), id)
-			expDir := fmt.Sprintf("%s/%d", os.Getenv("FHIR_ARCHIVE_DIR"), id)
-
-			err = os.Rename(jobDir, expDir)
+		id := int(j.ID)
+		jobDir := fmt.Sprintf("%s/%d", os.Getenv("FHIR_PAYLOAD_DIR"), id)
```
mtrang1263, line 614: Can we rename this to payloadDir?

bcda/bcdacli/cli.go
```diff
@@ -598,19 +610,35 @@ func archiveExpiring(hrThreshold int) error {
...
 		t := j.UpdatedAt
 		elapsed := time.Since(t).Hours()
-		if int(elapsed) >= hrThreshold {
-
-			id := int(j.ID)
-			jobDir := fmt.Sprintf("%s/%d", os.Getenv("FHIR_PAYLOAD_DIR"), id)
-			expDir := fmt.Sprintf("%s/%d", os.Getenv("FHIR_ARCHIVE_DIR"), id)
-
-			err = os.Rename(jobDir, expDir)
+		id := int(j.ID)
```
mtrang1263, line 613: I don't think we need to cast the j.ID (uint)to an int. Sprintf will handle all numeric types.

bcda/bcdacli/cli.go
adamayd: how about jobPayloadDir and jobArchiveDir so that it matches naming with cleanupArchive.

bcda/bcdacli/cli.go
adamayd: I agree on simplifying the logic to a single, but do you think we should look at the archive dir instead though?  If the original condition of this issue was that archive dirs were being orphaned then causing the rename to fail, should we check for those first then move on instead of assuming that if the payload dir exists try to copy.  Wouldn't that put us in the same situation where the job is marked completed, the payload file is created then the rename fails because the orphaned archive dir exists?

adamayd: From offline conversation I was looking for the wrong failing condition.  Updated the code to follow this pattern and look for the FHIR_PAYLOAD_DIR existing.  Was able to reduce the conditional and variable creation.

bcda/bcdacli/cli.go
adamayd: Matched logic in `cleanupArchive` function and removed the additional conditional in both functions that was always true due to the DB call change.  Tests passing on both and full suite.

mtrang1263 (commented):
Let's add a test to verify that a job that is marked as completed that does not have a payloadDir is successfully marked as archived.

We can leverage `TestArchiveExpiring` and create an additional job (w/ no data in the payloadDir).

mtrang1263 (approved):
LGTM!

PR #614, opened 2021-01-08
BCDA-4119 - Fix GetCCLFBeneficiaries query

mtrang1263:
### Fixes [BCDA-4119](https://jira.cms.gov/browse/BCDA-4119)

When we upgrade for gorm V1 to V2 ( #601) , there was a behavior change in the `db.Not()` operator.

For [v1](https://v1.gorm.io/docs/query.html#Not):
```
// Not In
db.Not("name", []string{"jinzhu", "jinzhu 2"}).Find(&users)
//// SELECT * FROM users WHERE name NOT IN ("jinzhu", "jinzhu 2");
```

For [v2](https://gorm.io/docs/query.html#Not-Conditions):
```
// Not In
db.Not(map[string]interface{}{"name": []string{"jinzhu", "jinzhu 2"}}).Find(&users)
// SELECT * FROM users WHERE name NOT IN ("jinzhu", "jinzhu 2");
```

Our gorm call was `query.Not(mbi, suppressedMBIs)`
We were expecting v1 behavior (`AND MBI NOT IN ($2, $3)`). Instead, we had `AND mbi <> ($2, $3)`.

Example in gormV2 docs:
```
// Plain SQL
db.Not("name = ?", "jinzhu").First(&user)
//// SELECT * FROM users WHERE NOT(name = "jinzhu") ORDER BY id LIMIT 1
```

This explains why we had used a not equals to operator (`<>`) instead of using `NOT IN`

We **did not** encounter any test failures via unit tests or smoke tests because none of our data had suppression data that satisfied the time window requirements. See postgres#GetSuppressedMBIs for more information on the query used to retrieve suppressed MBIs.

We only had one instance of using the `db.not` in bcda-app. No changes in bcda-ssas-app, it does not use `db.not`.
 ### Proposed Changes

1. Update our unit test data and smoke test data to reproduce the error. This will ensure that we can verify the fix before testing it out in the remote environments.
2. Fix query in `GetCCLFBeneficiaries` to use the correct arguments.

### Change Details
1. Update our test db snapshot to include suppression datas. This will allow us to test the behavior in our api/v1 and api/v2 tests.
2. Update our Makefile to refresh the suppression data dates to ensure our smoke tests utilize the suppression data.
3. Fix the query by switching to `Where("mbi NOT IN (?), suppressedMBIs)`.

### Security Implications

### Acceptance Validation
1. CI Passes. With our new changes we have guarantees that the bug is tested in the CI process.
2. Ran two [negative](https://bcda-ci.adhocteam.us/job/BCDA%20-%20Test%20-%20Integration%20Smoke/2798/) [tests](https://bcda-ci.adhocteam.us/job/BCDA%20-%20Test%20-%20Integration%20Smoke/2799/) to verify that the issue is reproduced in the dev environment.

These were verified using an [updated smoke test job](https://github.com/CMSgov/bcda-ops/pull/637) which updates the suppression dates.


adamayd (approved):
LGTM!!

PR #615, opened 2021-01-13
BCDA-4080: Migrate BCDA-APP to Go Modules

katyasoup:
### Fixes [BCDA-4080](https://jira.cms.gov/browse/BCDA-4080)

Go Modules is the preferred dependency management approach for Go projects as of version 1.11; our current dependency manager `dep` has been deprecated.

### Proposed Changes
- Migrate current dependencies to Go Modules
- Remove `dep`

### Change Details
- Migrate current dependencies:
    - `go mod init`and `go mod tidy` to generate and populate a `go.mod` file with the appropriate dependencies from the vendor files
    - Audit dependency versions between `Gopkg.lock` and the new `go.mod` file
    - Manually adjust versions as needed to achieve parity with current dependencies 
- Remove `dep`:
    - Remove `dep` installation and `dep ensure` commands from Dockerfiles
    - Remove `vendor` directory and `Gopkg.*` files from project root

### Security Implications

### Acceptance Validation
- [x] `vendor` directory and `Gopkg.*` files removed
- [x] references to dep removed
- [x] `go.mod` dependency versions match those from `Gopkg.lock`
- [x] documentation updated
- [x] built and tested locally
- [x] built and deployed to dev
dev-deploy

- [x] post-deploy smoke tests passed

smoke-tests

### Feedback Requested
An extra set of eyes on the `testify`, `go-agent`, `vegeta` and `crypto` dependencies in particular would be appreciated to confirm correct versions and usage


.vscode/settings.json
```diff
@@ -9,7 +9,4 @@
     "go.testEnvFile": "${workspaceFolder}/shared_files/decrypted/local.env",
     "go.testTimeout": "2m",
     "go.buildOnSave": "package",
-    "go.toolsEnvVars": {
```
mtrang1263, line 12: :heart:

bcda/client/bluebutton.go
```diff
@@ -16,7 +16,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/cenkalti/backoff"
+	"github.com/cenkalti/backoff/v4"
```
mtrang1263, line 19: Good catch! For other reviewers, we were always using V4, just the package path changes as a result of moving to go modules.

go.mod
```diff
@@ -0,0 +1,47 @@
+module github.com/CMSgov/bcda-app
+
+go 1.15
+
+require (
```
mtrang1263, line 5: Wanting to confirm, we've double checked all of the direct dependencies here and all of the versions match up with what was existing in the Gopkg.lock file?

.vscode/settings.json
mtrang1263: vscode users just have to restart the IDE for these changes to take affect, right?

mtrang1263 (commented):
LGTM! I left a couple of clarifying questions, but all the code changes look good to me.

Double checked all references to the `dep` tool and other files were removed.

go.mod
katyasoup: The ones called out in the feedback section above are the ones I'm not sure of - `go-agent` seems to use an import path both with and without `/v3` (see usage in `metrics_test.go`; all other instances us the `/v3` version); `testify` and `crypto` I only caught this morning that there are different versions between the two files so double-checking those

.vscode/settings.json
katyasoup: I believe so!

go.mod
mtrang1263: Let me check those packages you listed and I'll respond back here. Thanks for listing those out!

katyasoup: From offline conversation, confirmed correct usage of `go-agent` and `vegeta`; `crypto` and `testify` have been updated to match the revision/version from the most recent `Gopkg.lock` file. Made a note to discuss future work around upgrading some of these dependencies after we've moved over to Go Modules.

README.md
```diff
@@ -166,8 +168,9 @@ apiKey := os.Getenv("OKTA_CLIENT_TOKEN")
 
 Optionally, you can edit your `~/.zshrc` or `~/.bashrc` file to eliminate the need to source the file for each shell start by appending this line
 ```
-source $GOPATH/src/github.com/CMSgov/bcda-app/.env.sh
+source [src-path]/bcda-app/.env.sh
```
mtrang1263, line 171: Thanks for getting this all cleanedup!

go.mod
```diff
@@ -44,4 +44,30 @@ require (
 	gorm.io/gorm v1.20.8
 )
 
-replace github.com/opencensus-integrations/gomongowrapper => github.com/eug48/gomongowrapper v0.0.3
+replace (
```
mtrang1263, line 47: Thanks for going through all of the gopkg.lock entries and matching revisions!

mtrang1263 (approved):
LGTM! Great working on wrangling all of these dependencies!

Dockerfiles/Dockerfile.ssas
```diff
@@ -6,7 +6,6 @@ RUN apk add git openssl
 RUN openssl genrsa -out /var/local/private.pem 2048
 RUN openssl rsa -in /var/local/private.pem -outform PEM -pubout -out /var/local/public.pem
 
-RUN go get -u github.com/golang/dep/cmd/dep
```
adamayd, line 9: Interesting that there was no `dep ensure` in this Dockerfile.

adamayd (approved):

PR #619, opened 2021-01-21
Feature/config

CosmicSyntax:
### Fixes [BCDA-4086](https://jira.cms.gov/browse/BCDA-4086)

Currently, we configure our application using environment variables and init() functions. We would prefer to move to a configuration file based approach. By doing this, we remove hundreds of expensive OS calls and further secure the API by making it immutable.
### Proposed Changes
1. Create a new internal package that can load in a configuration file once into memory, and expose a function to the rest of the application to get values for keys.
2. Replace all environment variable calls with said solution in 1.
3. Implement the first stage of a three stage transition to a single configuration file setup.

- First: Implement a configuration package that will load a configuration package in. Those variables read in are immutable and cannot change. Configuration file may not have all the variables needed, so search the environment variables once if key value pairs not found in the package. If found in environment variables, bring it into package and make it immutable. This will only affect local dev environment. TEST/PROD not affect.
- Second: Lock down on all variables for local env. Updated configuration package so it no longer looks at environment variables, making the application essentially immutable. TEST/PROD not affect.
- Third: Updated package and update AWS architecture to ave TEST/PROD match local, in that environment variables no longer used.

### Change Details
1. Add a new package called "conf" at the base directory of bcda-app.
2. Fine and replace all os.Setenv, os.Getenv, os.Lookupenv. 
3. Make sure all test passes.

### Security Implications

Viper for reading in configuration file: https://github.com/spf13/viper

### Acceptance Validation
1. CI passes. 
2. Deploy to lower environment dev/test to confirm functionality.


config/config.go
```diff
@@ -0,0 +1,229 @@
...
+const testmode bool = false
+
+// Implementing a state machine tracking how things are going in this package
+const (
+	configgood    uint8 = 0
+	configbad     uint8 = 1
+	noconfigfound uint8 = 2
+)
+// State is a public variable for diagnostic purposes.
+var State uint8 = configgood // if config found and loaded, value not changed
```
mtrang1263, line 32: I'm not seeing any usages of State outside of this package, do you think we can make it private?

config/config.go
```diff
@@ -0,0 +1,229 @@
...
+
+		// If the file exists
+		if success := fileexistattempt(3, v, i); success {
+			envVars = *setup(v)
+			break
+		}
+
+		// Exhausted the for loop, config file not found :/
+		if i == len(locationSlice) - 1 {
+			State = noconfigfound
```
mtrang1263, line 108: You can move this assignment outside of the loop, that way by default you're setting the noconfigfound and don't have to check if you've exhausted all of your options.

mtrang1263 (approved):
Testing approval to see if blocks merging due to Snyk errors

mtrang1263 (approved):

mtrang1263 (changes requested):
Reverting approval - was just approving to get verify Snyk was not blocking.

Makefile
```diff
@@ -16,7 +16,7 @@ package:
 LINT_TIMEOUT ?= 3m
 lint:
 	docker-compose -f docker-compose.test.yml build tests
-	docker-compose -f docker-compose.test.yml run --rm tests golangci-lint run --deadline=$(LINT_TIMEOUT) --verbose
+	docker-compose -f docker-compose.test.yml run --rm tests golangci-lint run --exclude="(conf\.(Un)?[S,s]etEnv)" --deadline=$(LINT_TIMEOUT) --verbose
```
mtrang1263, line 19: Can we create a ticket to fix these moving forward?

Also, can we update the settings.json so our local VS code setup can ignore these errors for now?

conf/config.go
```diff
@@ -0,0 +1,193 @@
...
+// Private global variable:
+
+// An instance of the viper struct containing the config information. Only made
+// accessible through public functions: GetEnv, SetEnv, etc.
+var envVars viper.Viper
+
+// Implementing a state machine that tracks the status of the config ingestion.
+// This state machine should go away when in stage 3.
+const (
+	configgood    uint8 = 0
```
mtrang1263, line 37: does it make sense to create a new type to encapsulate these options?
```
type configState uint8

const (
configGood = iota
configBad
noConfigFound
)

Comment (mtrang1263) [inline conf/config.go:37]: can we also used mixedCaps here? so configGood versus configgood?

File: conf/config.go [hunk 5]
```diff
@@ -0,0 +1,193 @@
...
+   init:
+   First thing to run when this package is loaded by the binary.
+   Even if multiple packages import conf, this will be called and ran ONLY once.
+*/
+func init() {
+
+	// Possible config file locations: local and PROD/DEV/TEST respectfully.
+	var locationSlice = [2]string{
+		"/go/src/github.com/CMSgov/bcda-app/shared_files/decrypted",
+		"/go/src/github.com/CMSgov/DoesNotExistYet", // This is a placeholder for now
```
mtrang1263, line 78: Do we need to specify an extra path that will fail? Maybe if we have a comment instead to describe that we could possibly add new paths for the various environments that would be just as effective?

conf/config.go
```diff
@@ -0,0 +1,193 @@
...
+/*
+   findEnv is a helper function that will determine what environment the application
+   is running in: local or PROD/TEST/DEV. Each environment should have a distinct
+   path where the configuration file is located. First it check the local path,
+   then the PROD/DEV/TEST. If both not found, defaults to just using environment vars.
+*/
+func findEnv(location []string) (bool, string) {
+
+	// Check if the configuration file exists
+	if _, err := os.Stat(location[0] + "/local.env"); err == nil {
```
mtrang1263, line 100: Would it be clearer for us to do: 
```
for _, loc := range location {
if _, err := os.Stat(location[0] + "/local.env"); err == nil {
return true, loc
}
}

return false, ""
```

mtrang1263: That way you do not have your exit condition explicitly check for the end of the list. It happens for free with the range.

mtrang1263 (approved):
LGTM!

PR #622, opened 2021-01-25
bcda-3663 Feature: Job status endpoint handles cancelled jobs

SMLuthi:
### Fixes [BCDA-3663](https://jira.cms.gov/browse/BCDA-3663)

We are enabling the capabilities for a consumer to cancel a group export request. Since this request is sequestered off to a job queue we mark the job as being in a cancelled state. We need to convey this to the consumer if a status is requested. The return should be a `404` and have an operational outcome json payload as per the fhir spec.

### Proposed Changes

Add a cancelled state check to the case statement in the `api.go` job status endpoint method.

### Change Details

- Updated the models.go job constant to include a `Cancelled` state
- Updated the `api.go` case statement to handle the job cancelled state and return a `404` with an operational outcome. 

### Security Implications

### Acceptance Validation

ACO-API now returns a correct response code for cancelled jobs.

### Feedback Requested

Couple things:

- I ended up not having to update the swagger since theres already a `404` response coming from the job status endpoint (potentially). This is already present in the documentation and it looks like its accurate.
- Thanks to the table driven tests, I only had to add the new job cancelled status to the tests and it worked! But let me know if I should introduce any other tests.

-

### DIFF
--- a/bcda/api/v1/api.go
+++ b/bcda/api/v1/api.go
@@ -203,6 +203,9 @@ func JobStatus(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Expires", job.UpdatedAt.Add(api.GetJobTimeout()).String())
 		oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Exception, responseutils.Deleted, "")
 		responseutils.WriteError(oo, w, http.StatusGone)
+	case models.JobStatusCancelled:
+		oo := responseutils.CreateOpOutcome(responseutils.Error, responseutils.Not_found, responseutils.Deleted, "Job has been cancelled.")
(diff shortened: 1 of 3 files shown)


mtrang1263 (approved):
LGTM!

adamayd (dismissed):
LGTM with one question.  Would it make sense to use the 410 GoneResponse instead of the 404 if the job was cancelled?  The description for GoneResponse is "The requested resource is no longer available."  Or is that response intended solely for the data file itself?

katyasoup (dismissed):
LGTM! The 404 seems appropriate, but open to discussing further

SMLuthi:
@adamayd I dont disagree! Its a 404 due to the FHIR spec.

mtrang1263 (approved):
🍕

PR #625, opened 2021-01-27
Daniel/bcda 4126 fix duplicate filename

CosmicSyntax:
### Fixes [BCDA-4126](https://jira.cms.gov/browse/BCDA-4126)
During the most recent CCLF Import run, we encountered a duplicate file name error. We confirmed that the CCLF zip file for this ACO WAS NOT repeated.

The issue was due to mutiple ACOs being return in a string slice when prioritizing ACOs. See changes for more detail.

### Change Details
1. Added a function called "Dedup()" that takes a slice of a string and removes duplicates in the quickest way possible using hash maps. This has been added to the "utilts' package.
2. Take the string slice from getPriorityACOs() as input for Dedup, and use that output instead which should not have any duplicates AOCs.
3. Test has been edited to check for duplicates.

### Security Implications

### Acceptance Validation
1. CI Passes
2. Make sure the solution looks sound and watch out for future CCLF ingestion issues.


bcda/cclf/cclf_test.go
```diff
@@ -243,6 +243,18 @@ func (s *CCLFTestSuite) TestOrderACOs() {
 	assert.Equal(s.T(), "A8765", acoOrder[1])
 	assert.Regexp(s.T(), "A1111|A0246", acoOrder[2])
 	assert.Regexp(s.T(), "A1111|A0246", acoOrder[3])
+
+    // Check for no duplicates
+    var dupcheck = make(map[string]bool, 4)
+    for _, v := range acoOrder {
+        // Loop through the acoOrder and place then into mwp with a value of true
```
CosmicSyntax, line 250: Oops typo... :/

bcda/utils/common.go
```diff
@@ -79,3 +79,25 @@ func CloseFileAndLogError(f *os.File) {
 		logrus.Error(ferr)
 	}
 }
+
+// This is function that takes a string slice and removes any duplicates
```
mtrang1263, line 83: Can we update this so we start with the declared name? So: `Dedupe takes a string...`

Reference: https://golang.org/doc/effective_go.html#commentary

bcda/utils/common.go
```diff
@@ -79,3 +79,25 @@ func CloseFileAndLogError(f *os.File) {
...
+func Dedup(slice []string) []string {
```
mtrang1263, line 84: Can we add an explicit test for this function?

bcda/utils/common.go
```diff
@@ -79,3 +79,25 @@ func CloseFileAndLogError(f *os.File) {
...
+	// Get the length of the slice
+	var n = len(slice)
+
+	// Use the length to allocate memory once for new slice and map
+	var newslice = make([]string, 0, n)
+	var dupcheck = make(map[string]bool, n)
```
mtrang1263, line 90: If you'd want to save the most space, you can use a `map[string]struct{}`, since you're just looking for existence. The empty struct doesn't occupy any memory.

Usage would be:
```
if _, found := dupcheck[v]; !found {
    dupcheck[v] = struct{}{}
}
```

bcda/utils/common.go
```diff
@@ -79,3 +79,25 @@ func CloseFileAndLogError(f *os.File) {
...
 	}
 }
+
+// This is function that takes a string slice and removes any duplicates
+func Dedup(slice []string) []string {
+	// Get the length of the slice
+	var n = len(slice)
+
+	// Use the length to allocate memory once for new slice and map
+	var newslice = make([]string, 0, n)
```
mtrang1263, line 89: Can we update these variables so that we use MixedCaps? Reference: https://golang.org/doc/effective_go.html#mixed-caps

bcda/cclf/cclf_test.go
```diff
@@ -243,6 +243,18 @@ func (s *CCLFTestSuite) TestOrderACOs() {
 	assert.Equal(s.T(), "A8765", acoOrder[1])
 	assert.Regexp(s.T(), "A1111|A0246", acoOrder[2])
 	assert.Regexp(s.T(), "A1111|A0246", acoOrder[3])
+
+    // Check for no duplicates
+    var dupcheck = make(map[string]bool, 4)
+    for _, v := range acoOrder {
```
mtrang1263, line 249: This test will pass as-is, right? Meaning if you didn't have the utils.Dedup() [here](https://github.com/CMSgov/bcda-app/pull/625/files#diff-59e95aaede55a09c483036234c8a21d5210e6fa6a0948a26a8c0b90fda49ca8fR403), this test would be pass with this additional check?

Can we update the unit-test-db data to include a duplicate ACO entry? To do that, I think we have to create another entry in the systems table that references the same group_id.


Example:
```
bcda_test=# INSERT INTO systems (group_id, deleted_at) VALUES ('dbbd1ce1-ae24-435c-807d-ed45953077ff', NULL) RETURNING id;
 id
----
  9
(1 row)

INSERT 0 1
bcda_test=# INSERT INTO secrets (system_id) VALUES (9);
INSERT 0 1
```
See this README entry for details on updating the unit-test-db: https://github.com/CMSgov/bcda-app/blob/master/README.md#updating-seed-data-for-unit-tests

mtrang1263 (changes requested):
I believe the implementation is correct but think we need to update some of our test data to confirm the fix

bcda/cclf/cclf_test.go
CosmicSyntax: Yep

bcda/utils/common.go
SMLuthi: If you change it to using an empty struct I would document that heavily. Using a bool is much more explicit to checking existence!

CosmicSyntax: I think I'll keep it as a bool to be explicit. Bool takes up 1 byte, and this is not taking into consideration any compiler optimizations.

mtrang1263: Makes sense! I shouldn't have referenced the size savings. More of awareness if you see this in other spots of the code/in other projects. I think that it's Go's way of implementing sets w/o introducing a new container (or generics 😄 )

bcda/cclf/cclf_test.go
CosmicSyntax: A duplicate "A4321" added to group table.  When the Dedup function is commented out... the following test fails.
Screen Shot 2021-01-27 at 6 31 43 PM

bcda/cclf/cclf_test.go
```diff
@@ -230,19 +230,38 @@ func (s *CCLFTestSuite) TestImportCCLF8_Invalid() {
 func (s *CCLFTestSuite) TestOrderACOs() {
 	var cclfMap = map[string]map[metadataKey][]*cclfFileMetadata{
 		"A1111": map[metadataKey][]*cclfFileMetadata{},
-		"A8765": map[metadataKey][]*cclfFileMetadata{},
+		"A4321": map[metadataKey][]*cclfFileMetadata{},
```
CosmicSyntax, line 233: Changed from A8765 to A4321, because duplicate A4321 was added to the test group db.

bcda/cclf/cclf_test.go
```diff
@@ -230,19 +230,38 @@ func (s *CCLFTestSuite) TestImportCCLF8_Invalid() {
...
+
 	// A3456 and A8765 have been added to the database == prioritized over the other two
-	assert.Len(s.T(), acoOrder, 4)
 	assert.Equal(s.T(), "A3456", acoOrder[0])
-	assert.Equal(s.T(), "A8765", acoOrder[1])
-	assert.Regexp(s.T(), "A1111|A0246", acoOrder[2])
-	assert.Regexp(s.T(), "A1111|A0246", acoOrder[3])
+	assert.Equal(s.T(), "A4321", acoOrder[1])
+	assert.Regexp(s.T(), "A1111|A0246", acoOrder[n - 1]) // second to last in index
+	assert.Regexp(s.T(), "A1111|A0246", acoOrder[n]) // last in index
```
CosmicSyntax, line 249: Introducing duplicates mess / clash with the order.  The order can be fine, but a duplicate as seen in the test db, will trigger 248 and 249. To prevent this clash, removed the hard coded index, and use "n" instead.

bcda/cclf/cclf_test.go
```diff
@@ -230,19 +230,38 @@ func (s *CCLFTestSuite) TestImportCCLF8_Invalid() {
...
+        // If there are no duplicates, dupcheck[v] should not return true
+        if !dupcheck[v] {
+            dupcheck[v] = true
+        } else {
+            assert.Fail(s.T(), "There was a duplicate found in the acoOrder.")
+        }
+    }
+
+    // Check for increase in number of ACOs not attributed to duplicates
+	assert.Len(s.T(), acoOrder, 4)
```
CosmicSyntax, line 264: A duplicate will always trigger than length check.  To check increase in ACOs is not directly related to duplicates, moved this check last.

bcda/utils/common.go
```diff
@@ -79,3 +79,34 @@ func CloseFileAndLogError(f *os.File) {
...
+		// If false, we have not encountered the string before
+		// If it is true, do nothing
+        if !dupcheck[v] {
+			// Not a duplicate
+			dupcheck[v] = true
+			newSlice = append(newSlice, v)
+		}
+	}
+
+	return newSlice
```
CosmicSyntax, line 111: Mixed Cases :)

bcda/cclf/cclf.go
```diff
@@ -399,6 +399,8 @@ func orderACOs(cclfMap map[string]map[metadataKey][]*cclfFileMetadata) []string
 	defer db.Close()
 
 	priorityACOs := getPriorityACOs(db)
+	// Ensure there are no duplicate ACOs by wrapping it in Dedup()
+    priorityACOs = utils.Dedup(priorityACOs)
```
CosmicSyntax, line 403: This was commented out to make sure the testing works.  Now commented back in.

bcda/cclf/cclf.go
```diff
@@ -399,6 +399,8 @@ func orderACOs(cclfMap map[string]map[metadataKey][]*cclfFileMetadata) []string
 	defer db.Close()
 
 	priorityACOs := getPriorityACOs(db)
+	// Ensure there are no duplicate ACOs by wrapping it in Dedup()
+	priorityACOs = utils.Dedup(priorityACOs)
```
CosmicSyntax, line 403: Indenting issue in last commit. Sorry. Has bee fixed.

bcda/cclf/cclf.go
```diff
@@ -417,11 +419,11 @@ func orderACOs(cclfMap map[string]map[metadataKey][]*cclfFileMetadata) []string
 
 func getPriorityACOs(db *sql.DB) []string {
 	const query = `
-	SELECT trim(both '["]' from g.x_data::json->>'cms_ids') "aco_id" 
-	FROM systems s JOIN groups g ON s.group_id=g.group_id 
-	WHERE s.deleted_at IS NULL AND g.group_id IN (SELECT group_id FROM groups WHERE x_data LIKE '%A%' and x_data NOT LIKE '%A999%') AND
-	s.id IN (SELECT system_id FROM secrets WHERE deleted_at IS NULL);
-	`
+    SELECT trim(both '["]' from g.x_data::json->>'cms_ids') "aco_id"
```
CosmicSyntax, line 422: Not sure why this is happening. Used gofmt.  And it doesn't look like this in vscode.

mtrang1263: Looks good! I believe we're entering the tabs vs. spaces debate.

This was tabs, now it's spaces. Since it's in the multi-line string, `gofmt` doesn't seem to care. So I think we're good!

bcda/cclf/cclf_test.go
```diff
@@ -230,31 +230,38 @@ func (s *CCLFTestSuite) TestImportCCLF8_Invalid() {
...
 		"A3456": map[metadataKey][]*cclfFileMetadata{},
 		"A0246": map[metadataKey][]*cclfFileMetadata{},
 	}
 
 	acoOrder := orderACOs(cclfMap)
 
+	// A4321 duplicate as been added to test table groups as of 27 JAN 2021 to
+	// bolster testing for duplicates. Since a failure in removal of duplicate A4321
+	// pushed the index by one, check for A1111 and A0246 no longer hardcoded.
+	n := len(acoOrder) - 1
```
mtrang1263, line 243: Thanks for putting the explanation

bcda/utils/common.go
```diff
@@ -80,13 +80,22 @@ func CloseFileAndLogError(f *os.File) {
...
+	   2. Don't use map... maybe use merge sort deletion
+	*/
+
 	// Get the length of the slice
 	var n = len(slice)
 
 	// Use the length to allocate memory once for new slice and map
-	var newslice = make([]string, 0, n)
+	var newSlice = make([]string, 0, n)
+	// Empty structs do no occupty space
```
mtrang1263, line 98: Comments not valid now, right?

bcda/utils/common_test.go
```diff
@@ -0,0 +1,36 @@
...
+// Testing the Dedup() function that removes duplicates in string slice
+func (s *CommonTestSuite) TestDedup() {
+
+	// Sample data to be used for testing
+	var sampleSlice = []string{"one", "two", "two", "three"}
+	// Results should not have 2 "two"s
+	var result = Dedup(sampleSlice)
+
+	// Iterate through the results, and only encounter "two" once
+	// If counter is 1 after the for loopp, good... anything else, not good
```
mtrang1263, line 22: typo: `loop` instead of `loopp`

bcda/utils/common_test.go
```diff
@@ -0,0 +1,36 @@
...
+	var counter = 0
+
+	for _, v := range result {
+		if v == "two" {
+			counter++
+		}
+	}
+
+	assert.Equal(s.T(), counter, 1)
```
mtrang1263, line 31: Can we add a couple more checks here? Verifying that `"one"` and `"three"` exist? Just to make sure our dedupe returns all results?

Testify has `assert.Contains` that you can use here. If you use that, then you can use `assert.Len` and not have to do the loop since you know that there would be three items and all of them are accounted for/ unique.

mtrang1263 (commented):
thanks for adding the tests/data. Just put a couple last comments

bcda/utils/common.go
CosmicSyntax: Oh yes. Let me remove this!

bcda/cclf/cclf_test.go
CosmicSyntax: Also will change as -> has.

mtrang1263 (approved):
LGTM!

SMLuthi (approved):
This LGTM! The only thing I would maybe add is some logging in case we DO encounter a duplicate. Though I'm not sure if that is worth logging in the first place. @mtrang1263 ?

mtrang1263:
> This LGTM! The only thing I would maybe add is some logging in case we DO encounter a duplicate. Though I'm not sure if that is worth logging in the first place. @mtrang1263 ?

I think it's OK for us to not log it. Don't think it's a big concern of the CCLF side when we encounter a dupe because it's not really in charge of that data source.

I think you bring up a good point about logging these data discrepancies though. I think there's some assumptions in other components that are not being satisfied which results in the problem we're fixing here.

PR #633, opened 2021-02-10
bcda-4135 Feature: remove deleted at columns from db

SMLuthi:
### Fixes [BCDA-4135](https://jira.cms.gov/browse/BCDA-4135)

With our move from GORM to sql builder we lost access to some automatic GORM functionality that created and updated bookkeeping columns in our tables. `deleted_at` is one of those columns.

### Proposed Changes

Create a migration that removes the `deleted_at` column from all the tables that contain it since we do not use this data for anything.

### Change Details

- Migration up removes `deleted_at` column.
- Migration down adds `deleted_at` column.
- Unit tests verify migrations

### Security Implications

### Acceptance Validation

Migration removes `deleted_at` column.

The migration was applied manually to the unit test database and the unit tests run to assure the migration wont affect the unit tests in the future.

image

[jenkins smoke test](https://bcda-ci.adhocteam.us/job/BCDA%20-%20Test%20-%20Integration%20Smoke/2908/) ran successfully after applying the [migration](https://bcda-ci.adhocteam.us/job/BCDA%20-%20SQL%20Migrate/114/) to dev.

### Feedback Requested

I did not update our unit test db snapshot since, according to @mtrang1263 there is an upcoming change that will apply our migrations automatically to our unit test database making the snapshot change unnecessary. Let me know if that bothers anybody to the point of where I need to go in and apply the changes manually.

### DIFF
--- a/db/migrations/migrations_test.go
+++ b/db/migrations/migrations_test.go
@@ -86,16 +86,6 @@ var migration5Tables = []interface{}{
 	"suppression_files",
 }
 
-var migration7Tables = []string{
-	"acos",
-	"cclf_beneficiaries",
-	"cclf_files",
-	"job_keys",
-	"jobs",
-	"suppressions",
-	"suppression_files",
-}
-
 func (s *MigrationTestSuite) TestBCDAMigration() {
 	migrator := migrator{
 		migrationPath: "./bcda/",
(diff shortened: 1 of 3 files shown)


db/migrations/migrations_test.go
```diff
@@ -86,6 +86,16 @@ var migration5Tables = []interface{}{
 	"suppression_files",
 }
 
+var migration7Tables = []string{
```
mtrang1263, line 89: Since we're only using this in the TestBCDAMigration() function, should we define this in that function?

mtrang1263 (commented):
Can we run this migration on the dev environment and see how it works with existing data?

db/migrations/migrations_test.go
SMLuthi: We can, it would make sense to put it there as well!

SMLuthi:
@mtrang1263 Sounds good to me. I ended up applying the changes manually to our unit test database and now running the unit tests to make sure this wont break anything in the future when migrate it automatically. But I can also apply the migration to dev to make sure nothing breaks there with "live" data.

mtrang1263 (approved):
LGTM!

CosmicSyntax (approved):
LGTM!

PR #636, opened 2021-02-11
BCDA-3661: Support Job Cancellation

katyasoup:
### Fixes [BCDA-3661](https://jira.cms.gov/browse/BCDA-3661)
We want to support the ability for a caller to stop in-progress bulk data requests. [BCDA-3660](https://github.com/CMSgov/bcda-app/pull/629/files) introduces a new endpoint for cancelling jobs, and this ticket introduces the logic needed to support that cancellation.

### Proposed Changes
When stopping an in-progress request, our system will remove any jobs associated with the bulk data request. All pending (non-running) jobs will not be executed. In-progress jobs should be stopped.

### Change Details
- When the queue manager receives a queuejob, it will start a goroutine that periodically checks if the parent job has been cancelled. If the parent job has been cancelled, the current in-process queuejob will be cancelled via context.cancel()
- While the worker process a job (worker#writeBBDataToFile), if a queuejob's context is cancelled, it will fail the job. The job status should not update to Failed, but remain as Cancelled.

### Security Implications

### Acceptance Validation
- unit tests passing
- manually tested (local) to confirm parent job cancellation, stopping of in-progress queue job, and that no remaining queue jobs were started

### Feedback Requested
- General feedback on the approach; making sure I've leveraged context correctly; no gaping holes, etc.
- Any extra cleanup needed for pending queue jobs?
- Ideas on how to improve test coverage/mock a job that gets cancelled part-way through


bcdaworker/queueing/manager/que.go
```diff
@@ -107,6 +110,24 @@ func (q *queue) processJob(job *que.Job) error {
...
 
+	// start a goroutine that will periodically check the status of the parent job
+	go func() {
+		for {
+			select {
+			case <-time.After(15 * time.Second):
+				parentCancelled, _ := q.isParentJobCancelled(jobArgs.ID)
+				// parent job is cancelled, cancel this queuejob
+				if parentCancelled {
+					cancel()
```
CosmicSyntax, line 121: Will there be additional code added here to stop the job?

bcdaworker/queueing/manager/que.go
```diff
@@ -107,6 +110,24 @@ func (q *queue) processJob(job *que.Job) error {
 		return nil
 	}
 
+	// start a goroutine that will periodically check the status of the parent job
+	go func() {
+		for {
+			select {
+			case <-time.After(15 * time.Second):
+				parentCancelled, _ := q.isParentJobCancelled(jobArgs.ID)
```
mtrang1263, line 118: Can we log this error? Maybe at a warn level to see if this could start causing us grief if we see too many errors.

bcdaworker/worker/worker.go
```diff
@@ -100,7 +101,8 @@ func (w *worker) ProcessJob(ctx context.Context, job models.Job, jobArgs models.
 
 	// This is only run AFTER completion of all the collection
 	if err != nil {
-		err = w.r.UpdateJobStatus(ctx, job.ID, models.JobStatusFailed)
+		// only inProgress jobs should move to a failed status (i.e. don't move a cancelled job to failed)
+		err = w.r.UpdateJobStatusCheckStatus(ctx, job.ID, models.JobStatusInProgress, models.JobStatusFailed)
```
mtrang1263, line 105: Can we also update the call where we move from the "In Progress" to the "Completed" state to ensure that we do not move from cancelled to Completed?

bcdaworker/worker/worker.go
```diff
@@ -172,6 +174,12 @@ func writeBBDataToFile(ctx context.Context, r repository.Repository, bb client.A
 	failed := false
 
 	for _, beneID := range jobArgs.BeneficiaryIDs {
+		// if the parent job was cancelled, stop processing beneIDs and fail the job
```
mtrang1263, line 177: Should we check for ctx.Err == context.Canceled here like you are doing on line 220?

bcdaworker/worker/worker_test.go
```diff
@@ -466,6 +466,37 @@ func (s *WorkerTestSuite) TestProcessJob_NoBBClient() {
...
+	j := models.Job{
+		ACOID:      uuid.Parse("DBBD1CE1-AE24-435C-807D-ED45953077D3"),
+		RequestURL: "/api/v1/Patient/$export",
+		Status:     models.JobStatusCancelled,
+		JobCount:   1,
+	}
+	postgrestest.CreateJobs(s.T(), s.db, &j)
+
+	complete, err := checkJobCompleteAndCleanup(ctx, s.r, j.ID)
+	// no jobs should be completed since status has always been cancelled
```
mtrang1263, line 480: I don't think we need this, right?

mtrang1263 (commented):
For 

1. New worker_test.go test where you create the job with status Pending and we have a jobEqnueArgs with the BeneficiaryIDs set to a few hundred. This should guarantee that the job does not complete.
Create a context.WithCancel() and supply that to worker.ProcessJob()
Wait a little bit (few hundred milliseconds mabye) and then call cancel().

This should move the job to a failed state.

2. Add a que.go test that does something similar, but instead of cancelling the context directly, it'll update the job status to cancelled. I think to help with testing, allow us to supply the amount of time to poll for the cancelled status. Maybe 10ms or something quite small.

Let's set up sometime and maybe we can discuss more.

bcdaworker/worker/worker.go
mtrang1263: Can we also check for ErrJobNotUpdated? I think it shouldn't be an error scenario if the job status does not match up

if goerrors.Is(err, repository.ErrJobNotUpdated)

bcdaworker/queueing/manager/que.go
```diff
@@ -107,6 +110,24 @@ func (q *queue) processJob(job *que.Job) error {
 		return nil
 	}
 
+	// start a goroutine that will periodically check the status of the parent job
+	go func() {
+		for {
+			select {
+			case <-time.After(15 * time.Second):
```
CosmicSyntax, line 117: How long does a typical job take? Is there a chance that a cancel is requested, but the job is completed anyway because it took less than 15 seconds?

I think this is fine for now, since we're aiming to cancel more time consuming jobs. Maybe we can put in a ticket to propagate the cancel at an earlier time -- like from the web package -- and get rid of the timer?

bcdaworker/queueing/manager/que.go
```diff
@@ -107,6 +110,24 @@ func (q *queue) processJob(job *que.Job) error {
...
+				parentCancelled, _ := q.isParentJobCancelled(jobArgs.ID)
+				// parent job is cancelled, cancel this queuejob
+				if parentCancelled {
+					cancel()
+					return
```
CosmicSyntax, line 122: More code here in the future to stop the job?

bcda/models/service.go
```diff
@@ -153,6 +156,27 @@ func (s *service) GetJobAndKeys(ctx context.Context, jobID uint) (*Job, []*JobKe
...
+	job, err := s.repository.GetJobByID(ctx, jobID)
+	if err != nil {
+		return 0, err
+	}
+
+	// Check if the job is pending or in progress.
+	if job.Status == JobStatusPending || job.Status == JobStatusInProgress {
+		job.Status = JobStatusCancelled
+		err = s.repository.UpdateJob(ctx, *job)
+		if err != nil {
```
SMLuthi, line 170: Nitpick but you could cut down on the returns this way by checking for err being nil
```
...
    if err == nil {
      return jobID, nil
    }
  }
  return 0, ErrJobNotCancelled
}
```

adamayd: This was from my code that Katie based her branch on, but this is a trick on the eyes.  There are 3 different returns here, the difference is`ErrJobNotCancelled` and `ErrJobNotCancellable`  This is not the first time getting them confused though.  Should I have a better name for one to better distinguish between them when reading through the code?

SMLuthi: I shouldn't look at PRs too late >_<

bcdaworker/queueing/manager/que.go
katyasoup: so cancelling here should mean it gets picked up by `worker.go#writeBBDataToFile` where it will stop and exit out of the loop, and return an error indicating that the parent job was cancelled. I'll add in some more comments to make those steps clearer!

bcdaworker/worker/worker_test.go
katyasoup: ah, nope!

bcdaworker/worker/worker.go
katyasoup: tweaked this a bit - UpdateJobStatusCheckStatus was causing test failures because we currently allow for Failed to jobs to be marked as Completed, in addition to Pending jobs. So, I kept it as UpdateJobStatus but added a check for job.Status != Cancelled first, with a log indicating the Cancelled job was not updated to Completed

mtrang1263: Can we go ahead and remove/clean up those failing tests? I think the failed -> completed would also be an invalid case.

Or, since it's behavior that existed before, can we go ahead and make a ticket to fix this behavior?

bcdaworker/worker/worker.go
```diff
@@ -342,14 +357,21 @@ func checkJobCompleteAndCleanup(ctx context.Context, r repository.Repository, jo
 			return false, err
 		}
 
-		if err := r.UpdateJobStatus(ctx, j.ID, models.JobStatusCompleted); err != nil {
-			return false, err
-		}
+		// make sure only inProgress jobs get moved to completed (don't update Cancelled jobs)
+		if j.Status != models.JobStatusCancelled {
```
mtrang1263, line 361: Can we move this conditional check outside of this for loop? So we can make the check after we've made the Completed check on line 329.

 I think logging and returning true is a good response.

bcdaworker/worker/worker.go
katyasoup: Sure, I'll write up a new ticket for it!

mtrang1263 (approved):
LGTM!

PR #638, opened 2021-02-17
BCDA-4267 - Conf Experimental Feature

CosmicSyntax:
### Proposed Solution

For the conf package on the master branch, we essentially swapped out OS environment calls with functions that look for information within the application memory space made available by a configuration file. While this is a step in the right direction, the conf package is too ubiquitous; holds too many points of failure.  This PR takes a jab at fixing this.

So what is the tentative solution?

We want to purge GetEnv, LookupEnv, SetEnv, and UnsetEnv calls scattered all over the application. Instead of using the said functions to retrieve information, various packages in our application will instead pass their structs to the conf package instead. The conf package will walk the struct and fill in the information needed.

Example:
```
// Here is a struct that may be used in worker.go
type worker struct {
	r repository.Repository
        FHIR_STAGING_DIR string
}
// Instance of worker struct
x := worker{}
// Pass it to conf as a reference
conf.Checkout(&x)
// FHIR_STAGING_DIR should now no longer be a empty string, but with the information you need
fmt.Println(x)
```
Currently, this experimental feature checks all "exportable" fields and tries to fill it with information. "Exportable" is just a fancy word for a field that starts with a capital letter.

A more specific targeting can be done by using tags:
```
// Note the custom tag for second field
type worker struct {
	r repository.Repository
        FHIR_DIR string `conf:"FHIR_STAGING_DIR"`
}
```
There is code in place to utilize tags; but for now, nothing is done with them.

### Change Details

There is NO CHANGE to how the application uses the conf package in this PR. Changes are all internal to conf package, and these changes are:

1. Added an experimental function that may potentialy replace GetEnv, LookupEnv, SetEnv, and UnsetEnv.
2. Using GOPATH to determine the location of local.env file. Previously, this was hardcoded to what runs in the docker container.

### Security Implications

### Acceptance Validation
Make sure this PR does nothing to the application.

### Feedback Requested
This is a draft and will probably stay as one for a bit.  Feel free to explore and provide any other solutions or changes that come to mind!


conf/config.go
```diff
@@ -190,3 +216,102 @@ func UnsetEnv(protect *testing.T, key string) error {
...
+			if el.Kind() == reflect.Struct {
+
+				// Get ready to walk the fields of the structs
+				num := el.NumField() // Number of fields in struct
+				elType := el.Type()  // Used for tags
+
+				for i := 0; i < num; i++ {
+					field := el.Field(i)
+					// Ensure the field is "exportable" and is a type string
+					if field.CanInterface() && field.IsValid() && field.Type().Name() == "string" {
```
mtrang1263, line 273: Might be something later down the road - but should we add support for other data types? 

Here's a reference on how the json.Unmarshal is achieving the similar functionality with reflection: https://golang.org/src/encoding/json/decode.go?s=4081:4129#L784

conf/config.go
```diff
@@ -190,3 +216,102 @@ func UnsetEnv(protect *testing.T, key string) error {
...
+						// Get the field name
+						name := elType.Field(i).Name
+						// Tags are supported, but currently they do nothing
+						_ = elType.Field(i).Tag.Get("conf")
```
mtrang1263, line 277: 👍  - Can't wait to see this go live!

conf/config.go
```diff
@@ -190,3 +216,102 @@ func UnsetEnv(protect *testing.T, key string) error {
...
+
+			// Dereference the pointer
+			el := check.Elem()
+
+			// Is the data type a struct?
+			if el.Kind() == reflect.Struct {
+
+				// Get ready to walk the fields of the structs
+				num := el.NumField() // Number of fields in struct
+				elType := el.Type()  // Used for tags
```
mtrang1263, line 268: If we know it's a struct - can we call the `walk` method directly instead of having the parsing logic here?

conf/config.go
```diff
@@ -190,3 +216,102 @@ func UnsetEnv(protect *testing.T, key string) error {
...
+/***********************************************************************************************
+
+    CODE BELOW THIS BLOCK IS EXPERIMENTAL AND ONLY BEING USED INTERNALLY BY CONF PACKAGE
+
+***********************************************************************************************/
+
+// Gopath function exposes the gopath of the application while keeping it immutable. Golang does
+// not allow const to be strings, and a var would make it mutable if made public. This could be
+// something the config / viper struct keeps track. For now it's separate.
+func Gopath() string {
```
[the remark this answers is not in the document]
CosmicSyntax: Other then being used internally, not that I know of.

I'll go ahead and either remove this or comment it out.

conf/config.go
```diff
@@ -190,3 +216,102 @@ func UnsetEnv(protect *testing.T, key string) error {
...
+	return envVars.gopath
+}
+
+/*
+Checkout function takes a reference to a struct or a slice of string. It will traverse both
+data structures and look up key value pairs in the conf / viper struct by the name of the field
+for structs and string values for the slice. The function works with pointers so no value is
+returned. An error is returned if the wrong data structure is provided.
+*/
+func Checkout(list interface{}) error {
```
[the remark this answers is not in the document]
CosmicSyntax: If we were to rank team members by who can come up with the worst variable name, I probably would top that list.

"v" sounds good.

conf/config.go
```diff
@@ -190,3 +216,102 @@ func UnsetEnv(protect *testing.T, key string) error {
...
+
+	// Check if the data type provided is supported
+	switch list := list.(type) {
+	// If it's a slice of strings
+	case []string:
+
+		for n, str := range list {
+			if val, exists := LookupEnv(str); exists {
```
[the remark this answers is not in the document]
CosmicSyntax: I left this alone just in case the user may have a "default" value in there if conf package is not tracking. But if you think "default" value should not a be a thing, it makes sense to replace it with an empty string.

conf/config.go
CosmicSyntax: I think so!

Perhaps we start with the two we have now, and expand if needed during ALR implementation?

conf/config.go
CosmicSyntax: Yes!

conf/config.go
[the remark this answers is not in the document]
mtrang1263: I might be missing the implementation that you have in mind.

if I have the following config values that I want to retrieve:
["DB_URL", "TIMEOUT", "NUM_CONNECTIONS"] and none of them exist in the config, I would be left with ["DB_URL", "TIMEOUT", "NUM_CONNECTIONS"] right?

I'm thinking it would be clearer to return ["", "", ""] to indicate that none of them are set.

conf/config.go
mtrang1263: Yup that works for me. My thinking with expanded types is to catch the situation where we say that we expect a numeric value, but it turns out that the value is set to some non-numeric input. I'm thinking we would want to catch this as soon as possible, similar to how JSON unmarshal would fail in the similar situation.

conf/config.go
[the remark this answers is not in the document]
CosmicSyntax: Sorry. I thought it was aimed at the struct. Yes, this makes sense!

conf/config_test.go
```diff
@@ -187,12 +188,12 @@ func TestLookupEnv(t *testing.T) {
 		want  string
 		want1 bool
 	}{
-		{
-			"Query a variable that exists in local.env but does not have value",
-			args{"TEST_EMPTY"},
-			"",
-			true,
-		},
+		//{
```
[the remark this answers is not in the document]
CosmicSyntax: Yes.

Currently, if the value is blank, it defaults to os environment variables. So this doesn't really test what it should right now, and it always fails on local machine.

Will bring it back when stage 2 is done.

conf/config.go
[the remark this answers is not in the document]
CosmicSyntax: Actually. Holding off on the error when the usage of tags are implemented.

mtrang1263 (dismissed):
LGTM!

conf/config.go
```diff
@@ -290,10 +290,14 @@ func walk(field reflect.Value) {
 		innerField := field.Field(i)
 		// Ensure the field is "exportable" and is a type string
 		if innerField.CanInterface() && innerField.IsValid() && innerField.Type().Name() == "string" {
-			// Get the field name
-			name := fieldType.Field(i).Name
-			// Tags are supported, but currently they do nothing
-			_ = fieldType.Field(i).Tag.Get("conf")
+
+			// Tags are supported
+			name := fieldType.Field(i).Tag.Get("conf")
```
mtrang1263, line 295: Nice!!

conf/config.go
```diff
@@ -290,10 +290,14 @@ func walk(field reflect.Value) {
...
+			// If the Tag is not there, then default to the name of the struct
+			if name == "" {
```
mtrang1263, line 297: If we are adding support for tags, should we go ahead and support ignored fields? 

Reference: https://golang.org/src/encoding/json/encode.go?s=6458:6501#L1259

mtrang1263 (approved):
LGTM!

PR #639, opened 2021-02-18
Use wait-for-it script to ensure unit test db is running

mtrang1263:
### Fixes issue introduced by [PR](https://github.com/CMSgov/bcda-app/pull/635)

In this [PR](https://github.com/CMSgov/bcda-app/pull/635), we rebuild the database container and run the migration script to ensure that the database schema matches our expectations. To ensure that we start with a clean state, we teardown and rebuild the database container. We expect that this process does not take longer 5s.

However, our build were failing on the CI pipeline since the sleep time was not adequate. 

### Change Details

Use the strategy listed [here](https://github.com/vishnubob/wait-for-it/issues/57#issuecomment-497661237) to ensure that the unit-test-db container is started as expected

### Security Implications

We are using this [project](https://github.com/vishnubob/wait-for-it) as recommended by [Docker team](https://docs.docker.com/compose/startup-order/) to ensure that our database instance works as expected.
 
### Acceptance Validation
1. Successful CI run [here](https://bcda-ci.adhocteam.us/job/BCDA%20-%20App%20Build%20and%20Package/587/)
     * Looking at the logs, it was taking 19s to finish. We had set a sleep timer of 5s.
     ```
     wait-for-it: waiting 30 seconds for db-unit-test:5432
     wait-for-it: db-unit-test:5432 is available after 19 seconds
     ```
2. Successful CI build run with concurrent SSAS build [here](https://bcda-ci.adhocteam.us/job/BCDA%20-%20App%20Build%20and%20Package/589/)
     * Looking at the logs, it was taking 26s to finish. We had set a sleep timer of 5s.
     ```
     wait-for-it: waiting 30 seconds for db-unit-test:5432
     wait-for-it: db-unit-test:5432 is available after 26 seconds
     ```
     
Since the run with concurrent SSAS build took 26s, we chose a 60s timeout. This should give us plenty of head room in case the CI build machine is under heavy load.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ unit-test-db:
 	docker-compose -f docker-compose.test.yml up -d db-unit-test
 	
 	# Wait for the database to be ready
-	docker-compose -f docker-compose.wait-for-it.yml run --rm wait wait-for-it -h db-unit-test -p 5432 -t 30
+	docker-compose -f docker-compose.wait-for-it.yml run --rm wait wait-for-it -h db-unit-test -p 5432 -t 60
 	
 	# Perform migrations to ensure matching schemas
 	docker-compose -f docker-compose.migrate.yml run --rm migrate  -database "postgres://postgres:toor@db-unit-test:5432/bcda_test?sslmode=disable&x-migrations-table=schema_migrations_bcda" -path /go/src/github.com/CMSgov/bcda-app/db/migrations/bcda up
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ unit-test-db:
 	docker-compose -f docker-compose.test.yml up -d db-unit-test
 	
 	# Wait for the database to be ready
-	sleep 5
+	docker-compose -f docker-compose.wait-for-it.yml run --rm wait wait-for-it -h db-unit-test -p 5432 -t 30
 	
 	# Perform migrations to ensure matching schemas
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -77,7 +77,7 @@ unit-test-db:
 	docker-compose -f docker-compose.test.yml up -d db-unit-test
 	
 	# Wait for the database to be ready
-	sleep 5
+	docker-compose -f docker-compose.wait-for-it.yml run --rm wait wait-for-it -h db-unit-test -p 5432 -t 60
```
CosmicSyntax, line 80: Could we forgo docker-compose and use netcat instead?

Somewhere in the Makefile (maybe at the end of the file), you can have:
```
define waitforit
	@num=0; while ! nc -z localhost $(1); do \
		sleep 1; \
		echo $$num; \
		((num = num + 1)); \
		if [ $$num -eq $(2) ]; then\
		echo "Timed out..."; \
		break; \
		fi \
	done
endef
```
And then here you can have
```
$(call waitforit, 5432, 60)
```

mtrang1263: Makes sense to me! I was thinking about using docker-compose to be able to use the service name (db-unit-test), but since we expose a local port on all of the containers, that step is not necssary.

mtrang1263: After trying it - it looks like using the external port is successful but our attempt to run the migrate command fails due to connection refused. I'm think we should continue using this strategy.

SMLuthi (approved):
LGTM! I've had _some_ experience using dockerize so that would have been my other recommendation.

katyasoup (approved):
LGTM!

CosmicSyntax (approved):
LGTM!

PR #640, opened 2021-02-18
BCDA 4202: Fix response codes from SSAS

adamayd:
### Fixes [BCDA-4202](https://jira.cms.gov/browse/BCDA-4202)

When the API makes calls to SSAS to verify/authenticate the caller, we always return a 401 Unauthorized response when we encounter errors from SSAS. This includes issues like timeouts, preventing us from effectively retrying our requests.

### Proposed Changes

* Update our interactions with SSAS client to return the appropriate response codes.
  * Example, if we get a timeout when making a call to SSAS, we should return a 500 error to indicate that the caller can retry that action.
  * If SSAS returns an HTTP response code, we should return that same code back tot he caller. Ex: If SSAS returns a 400, we return 400 back to the caller.

### Change Details

* `auth/errors/provider.go` - Added file that provides the provider error to `bcda/auth` and `bcda/auth/clients` while avoiding cyclical imports with how `bcda/auth` is currently structured.
* `middleware.go#RequireTokenAuth` - Updated the `RequireTokenAuth` function by way of `AuthorizeAccess` -> `VerifyToken` -> `bcda/auth/client/ssas.go#VerifyPublicToken` method to return clear error types to have us choose the correct status code to return.
* `api.go#GetAuthToken` - Updated `MakeAccessToken` fucntion by way of `bcda/auth/client/ssas.go#GetToken` to return clearer error types so we can choose the correct status code to return.

### Security Implications

### Acceptance Validation

Build and all tests passed locally

### Feedback Requested

Tear it up.


mtrang1263:
Spoke with @DeirdreHolub offline about this PR. We were intentionally obfuscating any information that would be valuable to a potential attacker. If we forwarded the response codes/error message that we received from SSAS, it may give the attacker clues on if the request was malformed, invalid, etc.

PR #644, opened 2021-02-24
BCDA-4217: Update beneficiary lookup process to account for termination date

katyasoup:
### Fixes [BCDA-4217](https://jira.cms.gov/browse/BCDA-4217)
We currently support retrieving CCLF files with optional upper and lower time bounds. We should leverage this to retrieve CCLF files based on termination details. 

### Proposed Changes
Adds logic to `service.go` to determine which CCLF file to retrieve based on upper and lower time bounds:
- For determining a lower bound (cutoffTime): if the caller is terminated and has a non-Zero attributionDate, we should not use a cutoffTime; ACO should use OLD attribution
- For determining an upper bound (_since): if the caller is terminated and has a non-Zero attributionDate, and makes a request with a _since parameter that is *after* the attributionDate, we should treat all beneficiaries as "OLD" i.e. do not return full claims data (see diagram [here](https://confluence.cms.gov/pages/viewpage.action?pageId=429721042)).

### Change Details
- Update `service.go` to look up the latest CCLF file based on the terminated ACO's attribution date.
  - Based on termination type, we may use termination date as a cap on the attribution data.
    - Set the upper bound to termination date - this would cap the attribution data to when the caller was terminated
    - No upper bound set - use latest attribution data.
  - If we do not have termination information available, we can assume caller is not terminated and will use latest attribution data.

### Security Implications

### Acceptance Validation
- unit tests updated and passing locally

### Feedback Requested
I'm not wild about the if/else logic layout in `GetQueJobs`, so any improvements there are welcome.
General suggestions/test improvements, etc.


bcda/models/service.go
```diff
@@ -106,52 +106,57 @@ type runoutParameters struct {
...
 func (s *service) GetQueJobs(ctx context.Context, conditions RequestConditions) (queJobs []*que.Job, err error) {
-
 	if err := s.setTimeConstraints(ctx, conditions.ACOID, &conditions); err != nil {
 		return nil, fmt.Errorf("failed to set time constraints for caller: %w", err)
 	}
 
-	conditions.fileType = FileTypeDefault
-	switch conditions.ReqType {
-	case Runout:
+	var (
```
mtrang1263, line 113: Nice cleanup!

bcda/models/service.go
```diff
@@ -261,23 +266,37 @@ func (s *service) getNewAndExistingBeneficiaries(ctx context.Context, conditions
...
 		cutoffTime time.Time
+		upperBound time.Time
 	)
 
-	if s.stdCutoffDuration > 0 {
+	// if the _since parameter is after the attribution date,
+	// we should only return "old" benes; set the upper bound
+	// to the attribution date to avoid returning benes
+	// that were assigned after the attribution date
+	if !conditions.attributionDate.IsZero() &&
```
mtrang1263, line 276: Since we make the check [here](https://github.com/CMSgov/bcda-app/pull/644/files#diff-58fa56dc44c74157c85532e295a887dd487681523c03de7b7161263f856312e4R128), we don't need this in here right? We should assume the since can be satisfied and we can query for new/old benes?

bcda/models/service_test.go
```diff
@@ -445,24 +445,36 @@ func (s *ServiceTestSuite) TestGetQueJobs() {
 	}
 
 	since := time.Now()
+	terminationDetails := &Termination{
+		AttributionStrategy: AttributionHistorical,
+		TerminationDate:     time.Now().Add(-30 * 24 * time.Hour).Round(time.Millisecond).UTC(),
+	}
+
+	sinceAfterTermination := terminationDetails.TerminationDate.Add(10 * 24 * time.Hour).Round(time.Millisecond).UTC()
```
mtrang1263, line 453: Nit - we shouldn't need the Round/UTC call here since we've already performed that on the TerminationDate.

bcda/models/service.go
katyasoup: the `hasAttributionDate` check isn't run in the `getNewAndExisitingBeneficiaries` block (since that could be a valid path if the ACO isn't terminated - and we set the attr details to time.Time{} both if an ACO uses latest attribution OR if an ACO is not terminated)

bcda/models/service_test.go
katyasoup: ah, good point!

bcda/models/service.go
katyasoup: cleared up per offline conversation, we can remove this whole first if condition :)

mtrang1263 (approved):
LGTM!

PR #645, opened 2021-02-25
BCDA-4266 - Create utilities to create and deserialize ALR CSV data

mtrang1263:
#### ALR generation
Using the 2021 ALRDataDictionary found [here](https://www.cms.gov/files/zip/py-2020-shared-savings-program-report-templates.zip). introduced a package that can generate a synthetic ALR file.

The strategy defines columns and associated value generators. The generated values follow the valid values listed in the data dictionary. All values are schematically correct. 

Two shortcomings exist with this approach. GEO_SSA_CNTY_CD_NAME is using full state names since the library we are using does not have county name generators. Since we do not have a valid county names, we cannot generate a valid STATE_COUNTY_CD.

Generators have been defined for ALR Table 1-1. This approach can be extended to the other tables as needed.

Random values are generated using this [package](github.com/Pallinder/go-randomdata). 

#### ALR deserialization
Using [gota](github.com/go-gota/gota), we deserialize the ALR CSV file into a dataframe. By using dataframes, we can leverage JOINs to combine multiple ALR files together.

We currently require the following fields to be present for the ALR file to be valid:

- BENE_MBI_ID
- BENE_HIC_NUM
- BENE_1ST_NAME
- BENE_LAST_NAME
- BENE_SEX_CD
- BENE_BRTH_DT
- BENE_DEATH_DT

Once the merged dataframe is generated, we walk through the records and generate an Alr struct. We map all of the fields above to specific fields in the Alr. All other fields end up in the KeyValue map.

To handle [Byte Order Mark](https://en.wikipedia.org/wiki/Byte_order_mark), we leverage this [package](github.com/dimchansky/utfbom). See this [issue](https://github.com/golang/go/issues/33887) for more information.

### Security Implications

Three new dependencies added:

1. github.com/dimchansky/utfbom
2. github.com/go-gota/gota
3. github.com/Pallinder/go-randomdata

This PR will enable us to generate random ALR data which will be used to populate our system with synthetic data. This PR **does not** add any new data to any system.

When used, we plan on leveraging synthetic MBIs, so no PHI/PII will be used to generate synthetic ALR data.

### Acceptance Validation

CI passes - This code is not in the main code path so unit tests should be comprehensive enough to validate behavior.

### Next Steps
1. Adding new generators for the other ALR files.
2. CLI tool(s) to generate synthetic ALR data for our set of synthetic beneficiaries.


bcda/alr/csv/alr.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"time"
+
+	"github.com/CMSgov/bcda-app/bcda/models"
+	"github.com/sirupsen/logrus"
+)
+
+// Wrap models.Alr to allow us to create setter that will allow us to incrementally build
+// an ALR from a CSV file
+type alr struct {
```
CosmicSyntax, line 12: 👍

bcda/alr/csv/csv.go
```diff
@@ -0,0 +1,139 @@
...
+	sex,
+	birth, death,
+}
+
+// Fields that will be used to join multiple dataframes together
+var joinFields = requiredFields
+
+// ToALR reads in a CSV file(s) and unmarshals the data into an ALR model.
+// CSV files are joined based on a predetermined list of fields
+func ToALR(names ...string) ([]models.Alr, error) {
```
CosmicSyntax, line 35: Maybe change name to csvPaths or alrCsvPaths?

bcda/alr/csv/csv.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+		if err := validate(df); err != nil {
+			return nil, fmt.Errorf("dataframe from %s is not valid: %w",
+				name, err)
+		}
+
+		if len(mergedDF.Names()) == 0 {
+			mergedDF = df
+		} else {
+			mergedDF = mergedDF.InnerJoin(df, joinFields...)
```
CosmicSyntax, line 52: We may have to react to cases like:
BENE_MBI_ID | BENE_HIC_NUM | BENE_1ST_NAME | BENE_LAST_NAME | BENE_SEX_CD | BENE_BRTH_DT | BENE_DEATH_DT
1-1
1234, 1234, John, Doe, Male, 1-15-1950, 
1-2
1234, <Missing>, John, Doe, Male, 1-15-1950, 

Or even
1-1
1234, 1234, John, Doe, Male, 1-15-1950, 
1-2
1234, 1234, John, Doe, M, 1-15-1950, 

Hopefully not.

bcda/alr/csv/csv.go
```diff
@@ -0,0 +1,144 @@
...
+	// See: https://github.com/golang/go/issues/33887
+	reader := utfbom.SkipOnly(f)
+
+	df := dataframe.ReadCSV(reader, dataframe.HasHeader(true), dataframe.DetectTypes(false))
+	// Any error from this read operation is written to the Err field
+
+	return df, df.Err
+}
+
+func validate(df dataframe.DataFrame) error {
```
CosmicSyntax, line 83: Awesome!

bcda/alr/csv/csv.go
```diff
@@ -0,0 +1,144 @@
...
+// ToALR reads in a CSV file(s) and unmarshals the data into an ALR model.
+// CSV files are joined based on a predetermined list of fields
+func ToALR(names ...string) ([]models.Alr, error) {
+	var mergedDF dataframe.DataFrame
+	for _, name := range names {
+		df, err := toDataFrame(name)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create dataframe: %w", err)
+		}
+		if err := validate(df); err != nil {
```
CosmicSyntax, line 44: Should we talk to Bill or Nicole about what to do when a particular ALR table is missing the required fields. Do we just not process the whole ALR submission or just skip that particular table.

Though if one table is sparse, probably a good chance other tables are sparse too. I lean towards former, and not process the whole submission like you have here.

bcda/alr/csv/csv.go
mtrang1263: I wouldn't be surprised if we end up having to relax the joined columns so we only have MBI present. I'll put a comment in the code to describe the usage of the joinFields. 

For the Male -> M, I'm thinking the data dictionary should save us here. It actually calls out the expected values that we should have for each column. If we end up getting some unknown value, I think it makes sense to convey that when we encode it to the particular FHIR resource as UNKNOWN, not populate, etc.

bcda/alr/csv/csv.go
mtrang1263: Thinking the former as well. I'm thinking if an ALR violates some data assumptions, we should skip this conversion to ensure that we have some guarantees on the data that we're going to serve. We'd be able to say that we will have Table 1-1, 1-2, etc. Instead of saying that we may have situations where we have 1-1, 1-3, 1-4. 1-2 had some bad data.

We should keep in mind a "best-effort" approach too, but I think that should be weighted w/ Nicole, Laureen, and H-Team if we decide to pivot to that approach.

bcda/alr/csv/csv.go
```diff
@@ -0,0 +1,143 @@
...
+		a := &alr{}
+		a.KeyValue = make(map[string]string, len(row))
+		for idx, val := range row {
+			setter := setters[idx]
+			// No specific field set, we'll add this to the generic K:V
+			if setter == nil {
+				a.KeyValue[headers[idx]] = val
+			} else {
+				setter(a, val)
+			}
```
CosmicSyntax, line 111: 👍

bcda/alr/csv/csv.go
```diff
@@ -0,0 +1,143 @@
...
+func validate(df dataframe.DataFrame) error {
+	fields := df.Names()
+	m := make(map[string]struct{}, len(fields))
+	for _, field := range fields {
+		m[field] = struct{}{}
+	}
+
+	for _, required := range requiredFields {
+		if _, ok := m[required]; !ok {
+			return fmt.Errorf("required filed '%s' not found", required)
```
CosmicSyntax, line 91: field?

bcda/alr/gen/synthetic.go
```diff
@@ -0,0 +1,152 @@
...
+	regexp.MustCompile("BRTH_DT"):   func() string { return randomDate(minBirthDate, maxBirthDate) },
+	regexp.MustCompile("DEATH_DT"): func() string {
+		return randomEmpty(less,
+			func() string { return randomDate(minDeathDate, time.Now()) })
+	},
+	regexp.MustCompile("GEO_SSA_CNTY_CD_NAME"): func() string { return randomdata.City() }, // No county data source
+	regexp.MustCompile("GEO_SSA_STATE_NAME"):   func() string { return randomdata.State(randomdata.Large) },
+	// NOTE: This CNTY_CODE will not match up with the provided CNTY/STATE tuple.
+	// Need to integrate valid counties + state codes using the FIPS data set to align
+	// See: https://en.wikipedia.org/wiki/FIPS_county_code
```
CosmicSyntax, line 44: Should we create a ticket for this?

mtrang1263: https://jira.cms.gov/browse/BCDA-4308

CosmicSyntax (approved):
LGTM!

PR #646, opened 2021-02-25
BCDA-4216: Accommodate termination details when retrieving suppressions data

katyasoup:
### Fixes [BCDA-4216](https://jira.cms.gov/browse/BCDA-4216)
Right now, we receive the most recent opt-out beneficiaries by their effective date. There's no way for us to retrieve opt-out beneficiaries from a certain time.

### Proposed Changes
Update the query to accept an upper time bound when retrieving suppressed MBIs. This will return the most recent opt-out beneficiaries as of the upper bound.

All requests already have time constraints set in `GetQueJobs` to either a zero value or a date calculated based on the ACO's termination details, so no additional logic is needed to supply a value for the upper bound.

### Change Details
- Update `GetSuppressedMBIs` to accept an upper bound, and append an additional Where clause to the query if applicable
- Update `service.go` to supply the upper bound by passing RequestConditions to `getBenesByFileID`

### Security Implications

### Acceptance Validation
- Unit tests updated and passing locally

### Feedback Requested
- General improvements or suggestions


bcda/models/service.go
```diff
@@ -34,7 +34,7 @@ type RequestConditions struct {
 	fileType CCLFFileType
 
 	attributionDate time.Time
-	optOptDate      time.Time
```
mtrang1263, line 37: 🤦

bcda/models/postgres/repository.go
```diff
@@ -275,6 +275,10 @@ func (r *Repository) GetSuppressedMBIs(ctx context.Context, lookbackDays int) ([
 	sb.From(sb.BuilderAs(subSB, "h")).Join("suppressions s", "s.mbi = h.mbi", "s.effective_date = h.max_date")
 	sb.Where(sb.Equal("preference_indicator", "N"))
 
+	if !upperBound.IsZero() {
+		sb.Where(sb.LessEqualThan("effective_date", upperBound))
```
mtrang1263, line 279: If we want to get the "frozen in time" view of suppressions data, I think we need to adjust the statement above:
```
subSB.Between("effective_date", sqlbuilder.Raw(fmt.Sprintf("NOW() - interval '%d days'", lookbackDays)), sqlbuilder.Raw("NOW()"))
```

As is and interval of 10, we would still be looking at effective dates between NOW() and NOW() - 10 days AND effective_date <= upperBound.

So if we have 2021-07-01 as the upper bound, but make the call in 2021-08-01, nothing would be returned (date has to be between 2021-07-21 and 2021-08-01 while being less than 2021-07-01).

mtrang1263: Thinking out loud here, by using NOW() we are using the system time on the posgres instance. I don't think it's necessary here.

So instead of using the BETWEEN verb, we can use the sb.Where(sb.GreaterEqualThan(...), sb.LessEqualThan(...))

mtrang1263: Reference to usage of time.Now() in our queries: https://github.com/CMSgov/bcda-app/blob/master/bcda/models/service.go#L278

katyasoup: Ah, yep, totally overlooked that that would block a lot of valid dates. So just to clarify, we're looking at two scenarios: 1, there is an upper bound set, so we want less than or equal to the upper bound (with the lookback included, too?), vs. 2, no upper bound set, so we want greater than or equal to a date in the past specified by the lookback period

mtrang1263: I'm thinking the upper bound can be treated the same in this query. So in the `GetSuppressedMBIs` method, caller supplies an upperBound. We'll say that the upperBound - N days <= effective_date <= upperBound.

The caller of `GetSuppressedMBIs` can either supply. optOutDate (if set) or time.Now() for the upperBound. That way the logic is pretty straightforward in the query (upperBound always set) and we can make the decision of what to supply to the query in the service.

Let me know if this makes sense. Might be something we can chat about if you'd like.

mtrang1263 (approved):
LGTM!

PR #647, opened 2021-02-26
BCDA 4237 Adding ALR Models

CosmicSyntax:
### Completes [BCDA-4237](https://jira.cms.gov/browse/BCDA-4237)
We are getting ready to start consuming and serving up Assignment List Report (ALR) data to our customers. To do this, the initial step is to create a model and then make the necessary migrations to prep the database for ingestion.
### Proposed Changes
1. Create ALR Model. We need to keep in mind that a typical ALR submission may have more rows than a typical CCLF submission. In addition, we are currently assuming that seven fields (out of possible hundreds) should be present in each ALR submssions.  The rest of the fields are possibly violate and we cannot guarantee they will be present in each submission.
2. Make the appropriate migration.
### Change Details
1. The ALR model will have two tables. One will house the meta data and will be shorter in terms of rows compared to the second. We will query the first table by ACOs and timestamp to obtain an ID, which will serve as a foreign key to the second table.  The second table will house the actual ALR data submission, and we will filter by the foreign key value from the previous query.
2. Add migration # 10 to include ALR models.
### Security Implications
### Acceptance Validation
Migration looks sound and passes all unit test.

### DIFF
--- a/bcda/models/alr.go
+++ b/bcda/models/alr.go
@@ -6,7 +6,7 @@ import (
 
 type Alr struct {
 	ID            uint
-	ACOID         uint // Link to AlrMetaData
+	MetaKey       uint // Foreign Key
 	BeneMBI       string
 	BeneHIC       string
 	BeneFirstName string
@@ -15,20 +15,11 @@ type Alr struct {
 	BeneDOB       time.Time
 	BeneDOD       time.Time
 	Timestamp     time.Time
-	KeyValue      string // This stores all the other "violate" fields
-	CreatedAt     time.Time
-	UpdatedAt     time.Time
+	KeyValue      map[string]string // All "violate" fields
 }
 
 type AlrMetaData struct {
-	MetaDataID uint // Foreign key
-	ACO        string
-	Timestamp  time.Time
-	CreatedAt  time.Time
-	UpdatedAt  time.Time
+	ID        uint // Primary Key
+	ACO       string
+	Timestamp time.Time
 }
-
-// Thought in progress...
-/* type CSVContents struct {
-	Dataframe dataframe.DataFrame
-} */
--- a/bcda/models/alr.go
+++ b/bcda/models/alr.go
@@ -14,7 +14,6 @@ type Alr struct {
 	BeneSex       string
 	BeneDOB       time.Time
 	BeneDOD       time.Time
-	Timestamp     time.Time
 	KeyValue      map[string]string // All "violate" fields
 }
 
--- a/bcda/models/alr.go
+++ b/bcda/models/alr.go
@@ -23,6 +23,7 @@ type Alr struct {
 type AlrMetaData struct {
 	MetaDataID uint // Foreign key
 	ACO        string
+	Timestamp  time.Time
 	CreatedAt  time.Time
 	UpdatedAt  time.Time
 }
--- a/bcda/models/alr.go
+++ b/bcda/models/alr.go
@@ -1,0 +1,33 @@
+package models
+
+import (
+	"time"
+)
+
+type Alr struct {
+	ID            uint
+	ACOID         uint // Link to AlrMetaData
+	BeneMBI       string
+	BeneHIC       string
+	BeneFirstName string
+	BeneLastName  string
+	BeneSex       string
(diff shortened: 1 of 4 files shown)


db/migrations/bcda/000010_alr.up.sql
```diff
@@ -0,0 +1,38 @@
...
+
+ALTER TABLE ONLY public.alr_meta
+    ADD CONSTRAINT primary_key_alr_meta PRIMARY KEY (id);
+ALTER TABLE ONLY public.alr
+    ADD CONSTRAINT foreign_key_alr FOREIGN KEY (metakey) REFERENCES public.alr_meta(id) ON UPDATE RESTRICT ON DELETE RESTRICT;
+
+CREATE SEQUENCE IF NOT EXISTS public.metaid_seq START WITH 1 INCREMENT BY 1 CACHE 1 OWNED BY public.alr_meta.id;
+ALTER TABLE ONLY public.alr_meta ALTER COLUMN id SET DEFAULT nextval('public.metaid_seq');
+
+COMMIT;
```
mtrang1263, line 38: We should add two update triggers that keep the updated_at refreshed.

Example: https://github.com/CMSgov/bcda-app/blob/master/db/migrations/bcda/000006_set_updated_at_now_on_update.up.sql#L83

bcda/models/alr.go
```diff
@@ -0,0 +1,25 @@
+package models
+
+import (
+	"time"
+)
+
+type Alr struct {
```
mtrang1263, line 7: Should we add a unique on MetaKey + BeneMBI?

CosmicSyntax: I don't think so?

If you look at table 1-2 (as well as 1-3 and etc.) in the data dictionary:

"Each record represents a unique MBI-participant TIN combination."  Which sounds like there can be more than one row per MBI.

This just reminded me though. Are we in the business of remove duplicate rows?

db/migrations/bcda/000010_alr.up.sql
```diff
@@ -0,0 +1,65 @@
...
+ALTER TABLE ONLY public.alr_meta
+    ADD CONSTRAINT primary_key_alr_meta PRIMARY KEY (id);
+ALTER TABLE ONLY public.alr
+    ADD CONSTRAINT foreign_key_alr FOREIGN KEY (metakey) REFERENCES public.alr_meta(id) ON UPDATE RESTRICT ON DELETE RESTRICT;
+
+-- Set Incrementing to both IDs
+CREATE SEQUENCE IF NOT EXISTS public.alr_meta_id_seq START WITH 1 INCREMENT BY 1 CACHE 1 OWNED BY public.alr_meta.id;
+ALTER TABLE ONLY public.alr_meta ALTER COLUMN id SET DEFAULT nextval('public.alr_meta_id_seq');
+CREATE SEQUENCE IF NOT EXISTS public.alr_id_seq START WITH 1 INCREMENT BY 1 CACHE 1 OWNED BY public.alr.id;
+ALTER TABLE ONLY public.alr ALTER COLUMN id SET DEFAULT nextval('public.alr_id_seq');
```
CosmicSyntax, line 43: I forgot to put the incrementing for the alr table in the last commit.

db/migrations/bcda/000010_alr.up.sql
```diff
@@ -0,0 +1,65 @@
...
+-- Trigger trigger_set_timestamp function on updated_at column for alr & alr_meta
+CREATE TRIGGER set_timestamp
+BEFORE UPDATE ON public.alr
+FOR EACH ROW
+EXECUTE PROCEDURE trigger_set_timestamp();
+
+CREATE TRIGGER set_timestamp
+BEFORE UPDATE ON public.alr_meta
+FOR EACH ROW
+EXECUTE PROCEDURE trigger_set_timestamp();
```
CosmicSyntax, line 63: Triggers add.

bcda/models/alr.go
mtrang1263: Ah good find! I think we should discuss further about cleanup in general.

mtrang1263 (approved):
LGTM!

PR #660, opened 2021-03-15
Dan/bcda 4238 alr workers

CosmicSyntax:
### Completes [BCDA-4238](https://jira.cms.gov/browse/BCDA-4238)
Currently, we have a worker instance that is responsible for querying BFD for Patient, Coverage, and EOB resource types. We want to create modify the existing worker instance to take on the additional responsibility of retrieving ALR data.

### Change Details
1. Add new job related definitions in models.go
2. Update bcdaworker to listen and work on ALr jobs
3. Deserialize the job to retrieve the by ACO_ID, MBIs, and time range.
4. Using those arguments, the worker should query the database to find the ALR entries the satisfy the arguments.
5. Return the latest set of ALR data that satisfy the arguments.
6. Marshall data into FHIR compliant ndjson and write to disk

### Security Implications

### Acceptance Validation
Code is sound and passes test

### Feedback Requested
Any to make code better


bcdaworker/worker/alr_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+// Test ProcessAlrJob
+func (s *AlrWorkerTestSuite) TestProcessAlrJob() {
+	ctx := context.Background()
+	err := s.alrWorker.ProcessAlrJob(ctx, s.jobArgs)
+	// Check Job is processed with no errors
+	assert.NoError(s.T(), err)
+	// Make sure the filename is available for retrieval
+	assert.NotEmpty(s.T(), s.alrWorker.ndjsonFilename)
+	// Check the contents of the ndjson
```
CosmicSyntax, line 106: Not checked in this ticket.

bcdaworker/worker/alr.go
```diff
@@ -0,0 +1,215 @@
...
+		c <- data{patient, observations}
+	}
+
+	// Wait on the go routine and return back results
+	err = <-result
+
+	return err
+}
+
+func (a *alrWorker) UpdateJobAlrStatus(ctx context.Context, jobID uint,
```
CosmicSyntax, line 169: Not tested in this ticket.

bcdaworker/worker/alr.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	if affected == 0 {
+		return repository.ErrJobNotUpdated
+	}
+
+	return nil
+}
+
+func (a *alrWorker) GetAlrJobStatus(ctx context.Context, jobID uint) (*models.JobStatus,
```
CosmicSyntax, line 195: Not tested in this ticket.

bcdaworker/queueing/manager/alr.go
```diff
@@ -0,0 +1,163 @@
...
+		// This means the job did not finish for various reason
+		q.alrLog.Warnf("Failed to update job to complete for '%s' %s", job.Args, err)
+		// Re-enqueue the job
+		return nil
+	}
+
+	return nil
+}
+
+// func StartAlrQue(log *logrus.Logger, queueDatabaseURL string, numWorkers int) *alrQueue {
```
CosmicSyntax, line 110: If ALR would use its own pool.  Will be deleted before merge.

bcdaworker/worker/alr.go
```diff
@@ -0,0 +1,215 @@
...
+		id, fileUUID))
+	if err != nil {
+		logrus.Error(err)
+		return err
+	}
+	w := bufio.NewWriter(f)
+	defer utils.CloseFileAndLogError(f)
+
+	// Serialize data into JSON
+	marshaller, err := jsonformat.NewPrettyMarshaller(jsonformat.STU3)
```
mtrang1263, line 90: I think we should use the std marshaller. Since we are returning NDJSON files, we want every ALR record to be contained in a single line.

bcdaworker/worker/alr.go
```diff
@@ -0,0 +1,215 @@
...
+	result := make(chan error)
+
+	// Go rountine + closure that will write to file
+	// Two channels involved with this closure. One receiving "data", and the
+	// other will send an error is there was any error. The worker thread will
+	// will also use the result channel to wait for the IO to finish.
+	go func(c chan data, w *bufio.Writer, result chan error) {
+		// Receive data through channel
+
+		for i := range c {
```
mtrang1263, line 114: 🔥  - Just to confirm, the loop exits when the caller does close(c)?

bcdaworker/worker/alr.go
```diff
@@ -0,0 +1,215 @@
...
+	// data is what will be sent over the channel to the go routine
+	type data struct {
+		patient      *resources_go_proto.Patient
+		observations []*resources_go_proto.Observation
+	}
+
+	// Creating channels for go routine
+	// c is buffered b/c IO operation is slower than unmarshalling
+	c := make(chan data, 1000) // 1000 rows before blocking
+	result := make(chan error)
```
mtrang1263, line 105: Should we call:

```
defer close(c)
```

bcdaworker/worker/alr.go
```diff
@@ -0,0 +1,215 @@
...
+
+		// Everything went a-ok... send nil
+		result <- nil
+
+	}(c, w, result)
+
+	// Marshall into JSON and send it over the channel
+	for i := range alrModels {
+		patient, observations := alr.ToFHIR(&alrModels[i], alrModels[i].Timestamp)
+		c <- data{patient, observations}
```
mtrang1263, line 160: I'm wondering if we need to use the above goroutine and channel here. I'm thinking if we had a process that incrementally reuturn the postgres results (like a paging mechanism), then we would be able to see a lot of impacts by keeping the db IO and file IO separate, but by the time you get here, you've finished all of the db IO, and are now doing all of the file IO operations.

bcdaworker/worker/alr.go
mtrang1263: I'm thinking we should update these and place the query logic in the bcdaworker/repository.go (and postgres/repository.go). Then the worker can use the interface instead of using the db directly.

bcdaworker/worker/alr.go
CosmicSyntax: Actually, I need to explicitly close this after the for loop through alrModels, but before listening to result.

bcdaworker/worker/alr.go
CosmicSyntax: Yes.

bcda/models/alr.go
```diff
@@ -23,3 +24,17 @@ type AlrMetaData struct {
 	ACO       string
 	Timestamp time.Time
 }
+
+// Data Structure for Jobs
+type AlrJobs struct {
```
mtrang1263, line 29: We should have a new table for this right?

bcda/models/alr.go
```diff
@@ -23,3 +24,17 @@ type AlrMetaData struct {
...
+	ID     uint
+	ACO    string
```
mtrang1263, line 31: What do you think about matching this with our current pattern of CMSID? We would change this struct and the AlrMetaData one as well to match up.

bcda/models/alr.go
CosmicSyntax: Yes. Should the migration be included in this ticket?

mtrang1263: Yeah, let's put it in this ticket

bcda/models/fhir/alr/hcc.go
```diff
@@ -31,7 +31,10 @@ func init() {
 	)
 	// Scan through local path then the location of the file from the RPM
 	// See: ./ops/build_and_package.sh#49
-	paths := []string{"./hcc_crosswalk.tsv", "/etc/sv/api/hcc_crosswalk.tsv"}
+	paths := []string{"./hcc_crosswalk.tsv",
+		"/etc/sv/api/hcc_crosswalk.tsv",
+		"/go/src/github.com/CMSgov/bcda-app/bcda/models/fhir/alr/hcc_crosswalk.tsv",
+		os.Getenv("GOPATH") + "/src/github.com/CMSgov/bcda-app/bcda/models/fhir/alr/hcc_crosswalk.tsv"}
```
CosmicSyntax, line 37: The last two are pretty much the same thing; I'll remove the third entry.

bcda/models/postgres/alr.go
```diff
@@ -179,7 +179,13 @@ func (r *AlrRepository) AddAlr(ctx context.Context, aco string, timestamp time.T
...
 }
 
-func (r *AlrRepository) GetAlr(ctx context.Context, ACO string, lowerBound time.Time, upperBound time.Time) ([]models.Alr, error) {
+func (r *AlrRepository) GetAlr(ctx context.Context, ACO string, MBIs []string, lowerBound time.Time, upperBound time.Time) ([]models.Alr, error) {
+
+	// Convert []string{} to []interface{}
+	mbis := make([]interface{}, len(MBIs))
+	for i, v := range MBIs {
+		mbis[i] = v
+	}
```
CosmicSyntax, line 188: Need this conversion to build meta.In(), as seen below.

bcda/models/postgres/repository_test.go
```diff
@@ -876,6 +876,7 @@ func (r *RepositoryTestSuite) TestAlr() {
 	exMap["test1"] = "Test 01!"
 	exMap["test2"] = "Test 02!"
 	aco := "A1234"
+	MBIs := []string{"abd123abd01", "abd123abd02"}
```
CosmicSyntax, line 879: Test updated since GetAlr now takes slice of MBIs.

bcdaworker/queueing/manager/alr.go
```diff
@@ -0,0 +1,83 @@
...
+	err = q.alrWorker.ProcessAlrJob(ctx, jobArgs)
+	if err != nil {
+		// This means the job did not finish for various reason
+		q.alrLog.Warnf("Failed to complete job.Args '%s' %s", job.Args, err)
+		// Re-enqueue the job
+		return err
+	}
+
+	// Update DB that work is done / success
+	err = q.repository.UpdateJobStatus(ctx, jobArgs.ID, models.JobStatusCompleted)
```
CosmicSyntax, line 74: Method is shared with BFD.

bcda/models/alr.go
```diff
@@ -21,15 +23,18 @@ type Alr struct {
...
 type AlrMetaData struct {
 	ID        uint // Primary Key
-	ACO       string
+	CMSID     string
 	Timestamp time.Time
 }
 
-// Data Structure for Jobs
+// Data Structure for Jobs... this uses the same postgreSQL table as Job
 type AlrJobs struct {
```
mtrang1263, line 31: I'm thinking we can remove this, right?

CosmicSyntax: Yes, we should be able to do that.

mtrang1263 (dismissed):
LGTM!

mtrang1263 (approved):
LGTM!

PR #665, opened 2021-03-17
BCDA-4338 - Use singleton db reference for main and queue databases

mtrang1263:
### Fixes [BCDA-4338](https://jira.cms.gov/browse/BCDA-4338)

When working on [big int migration](https://jira.cms.gov/browse/BCDA-4042), we encountered issues with the underlying db clients (e.g. connection timed out errors). This can lead to client facing issues like the one reported in our [Google Group](https://groups.google.com/g/bc-api/c/Gq6aS_OVN5g).

To solve this, we need to use a health checker to constantly validate the underlying connection pool. See this [issue](https://github.com/jackc/pgx/issues/494) for more details.

### Proposed Changes

Utilize a singleton reference for both the main db (*sql.DB) and the queue db (*pgx.Pool). This will allow us to guarantee that each of these connections have the underlying health checker goroutine running.

We had this health checker in a couple of spots (in requests.go and in the bcdaworker), but it wasn't consistent across all usages of the database.GetDbConnection() method.

### Change Details

1. Adding configuration struct that allows us to define the connection defaults.
2. Create to singletons that connect with the main database (bcda) and the queue database (bcda_queue). All users should use the database.Connection and database.QueueConnection to retrieve the connection. They **should not** be creating a new set of connections.
3. Update all our packages to reference the new connections.
4. Remove usages of the db.Close(). Since these are shared for the lifecycle of the app, we do not need to close the database.

### Security Implications

### Acceptance Validation
1. Verified the fix locally by running the following scenario:
```
1. Execute an export request for data
2. Make request to job endpoint.
3. Restart the databases by running docker-compose restart db queue
4. Make request to job endpoint. This would have failed without this PR since it would be using a bad set of connections. After this fix, the application recovers and can satisfy the request
```
2. Verified the fix on the dev environment by running the following scenario:
```
1. Execute an export request for data
2. Make request to job endpoint.
3. Restart the databases by using the RDS console.
4. Make request to job endpoint. This would have failed without this PR since it would be using a bad set of connections. After this fix, the application recovers and can satisfy the request
```


CosmicSyntax (approved):
LGTM!

This really cleans things up! 👍

katyasoup (approved):
LGTM!

bcda/models/postgres/alr.go
```diff
@@ -135,6 +136,7 @@ func (a *alrCopyFromSource) Values() ([]interface{}, error) {
 func (r *AlrRepository) AddAlr(ctx context.Context, aco string, timestamp time.Time, alrs []models.Alr) error {
 	// Grab pgx.Conn
 	conn := getPgxConn(r.DB)
+	defer utils.CloseAndLog(logrus.WarnLevel, func() error { return stdlib.ReleaseConn(r.DB, conn) })
```
CosmicSyntax, line 139: Totally forgot about this. TY!

PR #669, opened 2021-03-18
BCDA-4244 - Push environment variables from docker-compose to local.env 

mtrang1263:
### Fixes [BCDA-4244](https://jira.cms.gov/browse/BCDA-4244)

This PR implements the Phase 2 of our configuration refactor. It moves all of the environment variables that used to be defined in the docker-compose.yml file into the local.env file.

### Change Details
1. Consolidate all config parameters into the local.env that used to exist in the docker-compose file associated with api and worker.

NOTE: We did not remove the `env` section for ssas since ssas relies on environment variables.
2. Update our docker-compose.test.yml file to include the overrides needed for the main db and queue db. This is needed since unit tests are using a different database.
3. Small tweak in the bluebutton_test.go to ensure that paging is disabled. This was needed since we've consolidated the env vars from the docker-compose file into the local.env.

### Security Implications

- [] new software dependencies

### Acceptance Validation
1. CI Passes
2. Deploy to dev and ensure it passes. This step should not be needed since it should only affect the local setup, but since we have modified the conf package, we should be sure that the remote deployments are not affected.


conf/config.go
```diff
@@ -180,12 +180,6 @@ func LookupEnv(key string) (string, bool) {
 	if state == configGood {
```
CosmicSyntax, line 180: We may also need to remove lines 157 to 166.

CosmicSyntax (approved):
LGTM!

PR #672, opened 2021-03-23
BCDA-4263: Remove blacklisted column from the ACOs table

adamayd:
### Fixes [BCDA-4263](https://jira.cms.gov/browse/BCDA-4263)

Currently the ACO blacklisted value is stored in the DB as a field.  To make use of the addition of Termination details we will provide default values to the Termination details of ACOs that are blacklisted, and calculate if the ACO is terminated base on the details provided.

### Proposed Changes

Remove the `Blacklisted` property from the `ACO` Struct.  Create a helper function to calculated if the ACO is blacklisted when called. Add the default values to the `termination_details` column in the DB where `blacklisted` is currently true.  Remove the `blacklisted` column.

### Change Details

1. Update blacklist logic to use TerminationDetails. Right now we are using ACO#Blacklisted column.
  1. If the ACO#TerminationDetails#BlacklistType is set to Involuntary or Voluntary, the ACO can be considered blacklisted.
  2. Remove the Blacklisted field in the ACO struct
  3. Create `Blacklisted()` helper function on the ACO type that encapsulates this logic.
2. Add migration script that will do the following:
  For every ACO whose blacklisted column == true, then we should write the following JSON data to the ACO#termination_details column
  `{
    "TerminationDate": "2020-12-31T23:59:59Z",
    "CutoffDate": "2020-12-31T23:59:59Z",
    "BlacklistType": 0
  }`
3. Create migration script that will drop the blacklisted column
4. Remove Blacklisted field from the ACO struct
5. Updated tests and fixtures to test for new implementation.

### Security Implications

### Acceptance Validation

1. `blacklisted` column is no longer present in the database.
2. Local tests including Postman tests checking for blacklisting have all passed.

Screenshot from 2021-03-23 11-30-35

Screenshot from 2021-03-23 11-31-02

Screenshot from 2021-03-23 11-31-29

### Feedback Requested

Better way to handle this, check my go and sql.


bcda/bcdacli/cli_test.go
```diff
@@ -812,8 +818,11 @@ func (s *CLITestSuite) TestBlacklistACO() {
 	s.Error(s.testApp.Run([]string{"bcda", "unblacklist-aco", "--cms-id", notFoundCMSID}))
 	s.Error(s.testApp.Run([]string{"bcda", "blacklist-aco", "--cms-id", notFoundCMSID}))
 
-	s.False(postgrestest.GetACOByUUID(s.T(), s.db, blacklistedACO.UUID).Blacklisted)
-	s.True(postgrestest.GetACOByUUID(s.T(), s.db, notBlacklistedACO.UUID).Blacklisted)
+	dbBlacklistedACO := postgrestest.GetACOByUUID(s.T(), s.db, blacklistedACO.UUID)
```
mtrang1263, line 821: Now that we have a name for it, should this be something like `newlyUnblacklistedACO`/`newlyBlacklistedACO`?

db/migrations/migrations_test.go
```diff
@@ -256,6 +255,109 @@ func (s *MigrationTestSuite) TestBCDAMigration() {
...
+				}
+				createTestRow(t, db, "acos", aco)
+
+				defer postgrestest.DeleteACO(t, db, acoUUID)
+
+				migrator.runMigration(t, "11")
+				assertColumnExists(t, true, db, "acos", "termination_details")
+
+				sb := sqlFlavor.NewSelectBuilder()
+				sb.Select("blacklisted", "termination_details").
```
mtrang1263, line 279: I'm thinking that we don't need to read out the blacklisted column since it's part of the WHERE clause?

db/migrations/migrations_test.go
```diff
@@ -256,6 +255,109 @@ func (s *MigrationTestSuite) TestBCDAMigration() {
...
+					Where(
+						sb.Equal("blacklisted", true),
+					)
+
+				query, args := sb.Build()
+				rows, err := db.Query(query, args...)
+				assert.NoError(t, err)
+				defer rows.Close()
+
+				for rows.Next() {
```
mtrang1263, line 290: I wonder if we should add a check to ensure that we actually read out a row here. This test could pass because the query returns no results.

So something like:
```
var rowsFound bool
for rows.Next() {
rowsFound = true
...
}
assert.True(t, rowsFound)

File: db/migrations/migrations_test.go [hunk 4]
```diff
@@ -256,6 +255,109 @@ func (s *MigrationTestSuite) TestBCDAMigration() {
...
+					var blacklisted bool
+					var termination_details string
```
mtrang1263, line 292: I'm thinking that we would want to use a `sql.NullString` here. And for the check, you can say `termination_details.IsValid()`.

Without this, we would fail at the rows.Scan method.

db/migrations/migrations_test.go
```diff
@@ -256,6 +255,109 @@ func (s *MigrationTestSuite) TestBCDAMigration() {
...
+
+				defer postgrestest.DeleteACO(t, db, aco.UUID)
+
+				postgrestest.CreateACO(t, db, aco)
+				assertColumnExists(t, false, db, "acos", "blacklisted")
+				migrator.runMigration(t, "11")
+				assertColumnExists(t, true, db, "acos", "blacklisted")
+
+				sb := sqlFlavor.NewSelectBuilder()
+				sb.Select("termination_details", "blacklisted").
```
mtrang1263, line 332: Thinking that we do not need to read out the termination_details here, right?

db/migrations/migrations_test.go
mtrang1263: And small nit pick (nit) on naming, can we used mixedCaps here?

db/migrations/migrations_test.go
```diff
@@ -256,6 +255,109 @@ func (s *MigrationTestSuite) TestBCDAMigration() {
...
+					Where(
+						sb.IsNotNull("termination_details"),
+					)
+
+				query, args := sb.Build()
+				rows, err := db.Query(query, args...)
+				assert.NoError(t, err)
+				defer rows.Close()
+
+				for rows.Next() {
```
mtrang1263, line 343: Same comment as above, can we ensure that we actually read out rows?

mtrang1263 (commented):
LGTM! Some test suggestions in the migration side, but impl looks solid

msnook (commented):
I saw this PR fly by in the Slack notifications, and was wondering if any of our ops scripts/config need to be updated with this change?
For example: https://github.com/CMSgov/bcda-ops/search?q=blacklisted

bcda/bcdacli/cli_test.go
adamayd: Fixed

db/migrations/migrations_test.go
adamayd: Fixed

db/migrations/migrations_test.go
adamayd: Fixed

db/migrations/migrations_test.go
adamayd: Fixed

db/migrations/migrations_test.go
adamayd: Fixed

adamayd:
> I saw this PR fly by in the Slack notifications, and was wondering if any of our ops scripts/config need to be updated with this change?
> For example: https://github.com/CMSgov/bcda-ops/search?q=blacklisted

Going through Ops for these changes now.  PR to follow.

db/migrations/migrations_test.go
```diff
@@ -33,45 +36,41 @@ const (
...
+	if _, err := db.Exec("CREATE DATABASE " + bcdaQueueDB); err != nil {
 		assert.FailNowf(s.T(), "Could not create bcda_queue db", err.Error())
 	}
-}
 
-func (s *MigrationTestSuite) TearDownSuite() {
-	if _, err := s.db.Exec("DROP DATABASE " + s.bcdaDB); err != nil {
-		assert.FailNowf(s.T(), "Could not drop bcda db", err.Error())
-	}
+	s.T().Cleanup(
```
CosmicSyntax, line 64: 👍

bcda/models/models_test.go
```diff
@@ -25,3 +28,33 @@ func (s *ModelsTestSuite) TestJobStatusMessage() {
 	j = Job{Status: JobStatusCompleted, JobCount: 25, CompletedJobCount: 25}
 	assert.Equal(s.T(), string(JobStatusCompleted), j.StatusMessage())
 }
+
+func (s *ModelsTestSuite) TestACOBlacklist() {
+	blackListValues := Termination{
+		TerminationDate: time.Date(2020, time.December, 31, 23, 59, 59, 0, time.Local),
+		CutoffDate:      time.Date(2020, time.December, 31, 23, 59, 59, 0, time.Local),
+		BlacklistType:   Involuntary,
+	}
```
CosmicSyntax, line 37: Should we also add a case for Voluntary?

bcda/models/models.go
```diff
@@ -64,10 +64,22 @@ type ACO struct {
...
 	TerminationDetails *Termination `json:"termination"`
 }
 
+// Blacklisted returns bool based on TerminationDetails.
+func (aco *ACO) Blacklisted() bool {
+	if aco.TerminationDetails != nil {
+		if aco.TerminationDetails.BlacklistType == Involuntary || aco.TerminationDetails.BlacklistType == Voluntary {
+			return true
+		} else {
+			return false
```
SMLuthi, line 76: Clarifying question: How common is it to have a blacklist type of `limited`? And is that worth logging... and is it worth logging if blacklisted ACO's are attempting to make requests?

SMLuthi (approved):
Left a question but otherwise LGTM (hinging on the ops PR of course :) )

db/migrations/migrations_test.go
adamayd: Missed resolving this one.

bcda/models/models_test.go
adamayd: Added tests for Voluntary and Limited.

bcda/models/models.go
adamayd: Was this answered in the insights meeting?

adamayd:
@SMLuthi per @msnook the Ops changes are only for insights and won't need to hold this PR up and only become an issue at release.

mtrang1263 (approved):
LGTM!

PR #675, opened 2021-03-26
BCDA-4238 ALR worker more fault-tolerant

CosmicSyntax:
### Fixes [BCDA-4238](https://jira.cms.gov/browse/BCDA-4238)
Currently the worker for BFD, as well as ALR since it was modeled off BFD, has several points of failure where the worker can end up serving incomplete data to the end user. We should come up with a solution that is less fault tolerant.
### Proposed Changes
1. Since we are on some roadmap to remove que-go and replace it with something better, while good amount of thought has been put in, changes are not grand. Confluence document will follow to catalog what was researched.
2. Compared to BFD, implement a more robust way of keeping track of worker failing and re-queuing in line.
3. The entire job is marked as failed if any of the smaller job fail 3x.
4. If a small job fails, it is re-queue and any files created related to that worker is deleted.
### Security Implications

### Acceptance Validation
Passes CI test
### Feedback Requested
Any improvements. Currently, I may be making it worse.


bcdaworker/queueing/manager/alr.go
```diff
@@ -19,64 +27,199 @@ type alrQueue struct {
...
-			case <-ctx.Done():
-				return
+	// Keep track of how many times a small job has failed.
+	// If the threshold is reached, fail the job
+	alrJobTracker.Lock()
+	if _, exists := alrJobTracker.tracker[jobArgs.ID]; !exists {
+
+		alrJobTracker.tracker[jobArgs.ID] = alrJobAttempt{
+			status:         models.JobStatusInProgress,
+			attemptTracker: make(map[uint]uint),
```
CosmicSyntax, line 127: May be a good idea to pre-allocate, since small jobs may be in the 100s or 1000s.

bcdaworker/queueing/manager/alr.go
```diff
@@ -19,64 +27,199 @@ type alrQueue struct {
...
 	err := json.Unmarshal(job.Args, &jobArgs)
+	// If we cannot unmarshall this, it would be very problematic.
+	if err != nil {
+		// TODO: perhaps just fail the job?
+		q.alrLog.Fatalf("Failed to unmarhall job.Args '%s' %s.",
+			job.Args, err)
+	}
+
+	// Check if this is already a failed job, and don't continue if it is
+	alrJobs, err := q.repository.GetJobByID(ctx, jobArgs.ID)
```
mtrang1263, line 108: Since we commit the transaction that contains the new model.Job, we may encounter a situation where we have the queue jobs but not the parent job yet.

We handle it like this in the que.go manager now. https://github.com/CMSgov/bcda-app/blob/master/bcdaworker/queueing/manager/que.go#L134 can we follow a similar approach?

bcdaworker/queueing/manager/alr.go
```diff
@@ -19,64 +27,199 @@ type alrQueue struct {
...
-					// cancelled context will get picked up by worker.go#writeBBDataToFile
-					cancel()
-					return
-				}
-			case <-ctx.Done():
-				return
+	// Keep track of how many times a small job has failed.
+	// If the threshold is reached, fail the job
+	alrJobTracker.Lock()
+	if _, exists := alrJobTracker.tracker[jobArgs.ID]; !exists {
```
mtrang1263, line 123: On the que.Job, we actually encode the failure count. See: https://github.com/bgentry/que-go/blob/master/que.go#L39.

It'll get incremented everytime we have to fail the job. Can we leverage that instead?

bcdaworker/queueing/manager/que_test.go
```diff
@@ -165,3 +172,136 @@ func TestProcessJobFailedValidation(t *testing.T) {
...
+	// Set up data based on testfixtures
+	db := database.Connection
+	alrWorker := worker.NewAlrWorker(db)
+
+	// Create synthetic Data
+	// TODO: Replace this with Martin's testing strategy from #4239
+	// To do this, this requires bringing in worker que migrations
+
+	// Create the key value pair
+	exMap := make(map[string]string)
```
mtrang1263, line 185: Can we start using the strategy for the ALR data which lives in the bcda database?

I was playing around with the fixtures and did a data dump for the keyvalue field. Here's a snippet:
```
- created_at: 2021-04-08T10:03:51.843824-06:00
  updated_at: 2021-04-08T10:03:51.843824-06:00
  id: 54
  metakey: 1
  mbi: 4SN0A00AA00
  hic: lagP5i2Akdd9
  firstname: Sofia
  lastname: Miller
  sex: "1"
  dob: 1903-05-19T00:00:00Z
  dod: 0001-01-01T00:00:00Z
  keyvalue: 0x0eff81040102ff8200010c010c000004ff820000
```

The keyvalue encoded here is actually an empty map.

bcdaworker/queueing/manager/alr.go
```diff
@@ -21,52 +31,139 @@ type alrQueue struct {
 	alrWorker worker.AlrWorker
 }
 
+// Max number of retries either set by ENV or default value of 3
+var maxRetry = int32(utils.GetEnvInt("BCDA_WORKER_MAX_JOB_NOT_FOUND_RETRIES", 3))
```
mtrang1263, line 35: You think it's good time to introduce a conf file/package for the worker? Similar to the [service config](https://github.com/CMSgov/bcda-app/blob/master/bcda/service/config.go)?

bcdaworker/queueing/manager/alr.go
```diff
@@ -21,52 +31,139 @@ type alrQueue struct {
...
+		q.alrLog.Warnf("ALR big job has been failed, worker will not be tasked for %s",
+			job.Args)
+		return nil
+	}
+	// End of validation
+
+	// Keep track of how many times a small job has failed.
+	// If the threshold is reached, fail the parent job
+	// This is additional logic that does not exist for BFD worker
+	if _, exists := alrJobTracker[alrJobs.ID]; !exists {
```
mtrang1263, line 132: Right now the job tracker is used to see if we need to update the job status to in progress. What if the job is marked as in progress, fails and gets picked up by another work instance? We would update that status to InProgress as well, right?

Does it  make sense to always do an update and use a conditional update like we do in the BFD worker? [It does a check then set](https://github.com/CMSgov/bcda-app/blob/041a24f75c650ffe7dd07597c6b9723c22180bd3/bcdaworker/repository/postgres/repository.go#L118).

bcdaworker/worker/alr.go
```diff
@@ -115,6 +115,7 @@ func goWriter(c chan data, w *bufio.Writer,
 func (a *AlrWorker) ProcessAlrJob(
 	ctx context.Context,
 	jobArgs models.JobAlrEnqueueArgs,
+	ndjsonFilename uuid.UUID,
```
mtrang1263, line 118: This is not used, is it planned to be updated at a later time?

CosmicSyntax: I think it was a thought, but I don't think so. I'll remove this.

mtrang1263 (approved):
LGTM!

mtrang1263 (approved):

PR #676, opened 2021-03-29
Update bcda_client to avoid recursive retry loop

SMLuthi:
### Fixes [BCDA-xxx](https://jira.cms.gov/browse/BCDA-xxx)

It was possible to enter a recursive `update token` retry loop with the new retry logic for the bcda client for the smoke tests.

### Change Details

Instead of using the `retryhttp` client, `updateAccessToken` will be run on the default http client that avoids all that retry logic. This will prevent it from falling into a recursive retry loop.

### Security Implications

### Acceptance Validation

Im keeping this a draft because Im trying to figure out the best way to test this. I know Martin ran into this issue originally but I dont remember how to recreate it.

### Feedback Requested


adamayd (dismissed):
LGTM!!

katyasoup (approved):
LGTM!

test/smoke_test/bcda_client.go
```diff
@@ -307,7 +309,10 @@ func (c *client) updateAccessToken() error {
 	req.SetBasicAuth(clientID, clientSecret)
 	req.Header.Add("Accept", "application/json")
 
-	resp, err := c.httpClient.Do(req)
+	// The retry logic may try to update the access token, to avoid a recursive
+	// retry loop we dont want to use the retry logic on updating the access token
+	http.DefaultClient.Timeout = httpTimeout
```
mtrang1263, line 314: Does it make sense to put this on line 52?

SMLuthi: That would be fine. I kept it where it is since its the only place we currently reference the http client... but if we move it up it'll only set it once.

mtrang1263 (dismissed):
LGTM!

mtrang1263 (approved):
LGTM!

PR #682, opened 2021-04-12
Add error check to ssas parsetoken middleware

SMLuthi:
### Fixes [BCDA-4398](https://jira.cms.gov/browse/BCDA-4398)

If the `Issuer` is set to `ssas` we have not been checking the error in the `ParseToken` middleware for our requests. Up until now any error thrown from `adFromClaims` which extracts available values from SSAS token claims has been ignored.

### Change Details

Check for an error, log, and server.

This has been updated so that `ParseToken` will now serve an error if the token its trying to parse is incorrect or the ACO cannot be found. However, since not all endpoints require a token, we will continue serving requests if no auth has been provided and we let `RequireToken` handle the errors.

### Security Implications

### Acceptance Validation

Check and handle  the error instead of ignoring it.

### Feedback Requested

I've been trying to figure out if there was a convenient way to test this but we may just have to wait until we mock out the ssas client in the future.


bcda/auth/middleware.go
```diff
@@ -71,7 +71,13 @@ func ParseToken(next http.Handler) http.Handler {
 			// okta token
 			switch claims.Issuer {
 			case "ssas":
-				ad, _ = adFromClaims(repository, claims)
+				ad, err = adFromClaims(repository, claims)
+				if err != nil {
+					log.Error(err)
+					next.ServeHTTP(w, r)
```
mtrang1263, line 77: Can we update this to return the 401 status code here instead of moving on to the next handler? Like we are doing [here](https://github.com/CMSgov/bcda-app/blob/master/bcda/auth/middleware.go#L120).

I think this'll be a good opportunity for us to update the other error scenarios in ParseToken

SMLuthi: Great idea!

mtrang1263:
👍 LGTM! Thanks for consolidating the logic above.

mtrang1263 (approved):

katyasoup (approved):
LGTM!

PR #679, opened 2021-04-06
Dan/bcda 4354 remove okta alpha

CosmicSyntax:
### Fixes [BCDA-4354](https://jira.cms.gov/browse/BCDA-4354)
We are using SSAS auth in our deployments & our smoke tests. We no longer need to include the other auth providers.
### Proposed Changes
1. Update the auth package to start and always use SSAS as the authentication provider
2. Update unit tests and ensure they pass.
3. Remove Alpha and Okta from code-base.
4. Again ensure tests pass.
5. Ensure unit-test, postman, and smoke-test are not affected.
6. Make edits to Makefile and bash scripts to stop application from restarting and resetting credentials.
### Change Details
No change from proposed changes above.
Some tests were commented out since SSAS should be testing them.

Another ticket may be required to thoroughly sweep through all code base to remove any remnants of alpha and okta are removed and additional unit-tests are added.
### Security Implications
- [ ] new software dependencies
- [ ] security controls or supporting software altered
- [ ] new data stored or transmitted
- [ ] security checklist is completed for this change
- [x] requires more information or team discussion to evaluate security implications
- [x] no PHI/PII is affected by this change
### Acceptance Validation
CI Passes and no security concerns.
### Feedback Requested
Encouraged

PR #684, opened 2021-04-13
BCDA-4372 - Use mock auth provider  in unit tests

mtrang1263:
### Part of [BCDA-4372](https://jira.cms.gov/browse/BCDA-4372)

With the removal of Okta and Alpha auth providers in this [PR](https://github.com/CMSgov/bcda-app/pull/679), we needed to have all of our unit tests using SSAS. This is not needed for most of our unit tests since we are not testing the SSAS package.

Instead of requiring SSAS to be running in the unit tests, we added a mock_provider.go which implements the auth.Provider interface. In tests that need an auth provider but do not need it to be SSAS, we spin up the mock provider.

### Change Details
1. No longer require real SSAS for unit tests.
2. Introduce mock Provider and function to set the provider. By using a separate function, we can ensure that we only set the mock provider in test cases. The developer will not be able to set the mock auth provider outside of tests.

### Security Implications

### Acceptance Validation
1. CI passes - only affects our unit tests.


SMLuthi (approved):
LGTM! <3

katyasoup (approved):
LGTM! 🎉 Can you add the link to the referenced PR in the description?

PR #688, opened 2021-04-15
BCDA-4372 - Switch from fatal to error logging

mtrang1263:
### Fixes error raised by [BCDA-4372](https://jira.cms.gov/browse/BCDA-4372)

With this [PR](https://github.com/CMSgov/bcda-app/pull/684), we exited early if we could not set up the auth provider.

This caused issues with packages that import the auth package, but do not use it. It was causing errors in our NFS instances.

### Change Details
1. Switch init() block to use Errorf instead of Fatalf

### Security Implications

### Acceptance Validation
1. Running deploy on dev environment and verify smoke tests pass

### DIFF
--- a/bcda/auth/provider.go
+++ b/bcda/auth/provider.go
@@ -28,7 +28,7 @@ func init() {
 
 	c, err := client.NewSSASClient()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


SMLuthi (approved):
LGTM! Annoying little side effect >_<

PR #690, opened 2021-04-19
bcda-4400 Feature: Allow cclf ingest termination

SMLuthi:
### Fixes [BCDA-4400](https://jira.cms.gov/browse/BCDA-4400)

We ran into an issue when we deployed in the middle of the cclf ingestion process which ended up causing it to shutdown unexpectedly. This resulted in duplicate files and resulted in us having to manually remove these files from the database before starting the import process over again.

### Proposed Changes

We need to make the import process more robust. This includes first checking whether the cclf file has already been imported by checking the database (and skipping the import if it does). Secondly, we need to wrap the entire file import into a database transaction so that we can be certain if the file already exists in the database (as mentioned before) we can guarantee the import was completely successful.

### Change Details

1. Implement the interfaces for pgx/sql database so the trx can be interchangeable.
2. Update `repository.go` to allow a pgx transaction. Create a new method for retrieving a cclf file by `name`.
3. Update `cclf.go` to check for the existence of a cclf file before importing and skip if true. Wrap the import process in a transaction and allow rollback if the import process encounters an error.
4. Delete ACO prioritizing code as this is no longer necessary since our import process has reduced from tens of minutes to seconds.

### Security Implications

### Acceptance Validation

We do not need to perform any manual changes to the DB/CCLF files if the import fails mid run.

### Feedback Requested

The usual stuff. Better Go, etc...


bcda/cclf/cclf.go
```diff
@@ -152,26 +149,55 @@ func importCCLF0(ctx context.Context, fileMetadata *cclfFileMetadata) (map[strin
 }
 
 func importCCLF8(ctx context.Context, fileMetadata *cclfFileMetadata) (err error) {
-	db := database.Connection
-
-	repository := postgres.NewRepository(db)
-
 	if fileMetadata == nil {
```
mtrang1263, line 152: Can we remove this logic here? I think we can guarantee that fileMetadata is not nil at this point.

bcda/cclf/cclf.go
```diff
@@ -152,26 +149,55 @@ func importCCLF0(ctx context.Context, fileMetadata *cclfFileMetadata) (map[strin
...
 		log.Error(err)
 		return err
 	}
 
+	db := database.Connection
+	repository := postgres.NewRepository(db)
+
+	e, err := repository.GetCCLFFileExistsByName(ctx, fileMetadata.name)
+	if err != nil {
+		fmt.Printf("failed to check existence of CCLF%d file.\n", fileMetadata.cclfNum)
```
mtrang1263, line 164: Something we can consider in the future is a [MultiWriter](https://golang.org/pkg/io/#MultiWriter). We can build our logger so it writes to a file as well as write to stdout. That way we don't have to remember to double log all of these entries.

bcda/cclf/cclf.go
```diff
@@ -152,26 +149,55 @@ func importCCLF0(ctx context.Context, fileMetadata *cclfFileMetadata) (map[strin
...
 		err = errors.New("CCLF file not found")
 		fmt.Println(err.Error())
 		log.Error(err)
 		return err
 	}
 
+	db := database.Connection
+	repository := postgres.NewRepository(db)
+
+	e, err := repository.GetCCLFFileExistsByName(ctx, fileMetadata.name)
```
mtrang1263, line 162: nit - can we use `exists` instead of `e`?

mtrang1263 (approved):
LGTM!

FYI @CosmicSyntax - this PR adds the interfaces needed to allow us to use the pgx.Tx in the respository code.

bcda/cclf/cclf.go
CosmicSyntax: https://github.com/CMSgov/bcda-app/blob/1dee98bd451adfeb0287f16b8ef454a7e0039e9d/bcda/cclf/cclf.go#L335

bcda/cclf/cclf.go
CosmicSyntax: Great idea.

bcda/cclf/cclf.go
```diff
@@ -152,26 +149,55 @@ func importCCLF0(ctx context.Context, fileMetadata *cclfFileMetadata) (map[strin
...
+		fmt.Printf("failed to check existence of CCLF%d file.\n", fileMetadata.cclfNum)
+		err = errors.Wrapf(err, "failed to check existence of CCLF%d file", fileMetadata.cclfNum)
+		log.Error(err)
+		return err
+	}
+
+	if e {
+		fmt.Printf("CCL%d file %s already exists in database, skipping import...", fileMetadata.cclfNum, fileMetadata)
+		log.Infof("CCL%d file %s already exists in database, skipping import...", fileMetadata.cclfNum, fileMetadata)
+		return nil
```
CosmicSyntax, line 173: Could we change the return type here (preferably a custom error) so that the caller of importCCLF8 knows if the file was already imported, other than through logs?

I don't think this needs to be done now, but maybe in a future ticket.

bcda/database/database.go
```diff
@@ -0,0 +1,76 @@
+package database
```
CosmicSyntax, line 1: 👍 Sweet

CosmicSyntax (dismissed):
LGTM

mtrang1263 (approved):
LGTM!

PR #695, opened 2021-04-21
BCDA 4337: Docker Multi-Stage Builds

adamayd:
### Fixes [BCDA-4337](https://jira.cms.gov/browse/BCDA-4337)

We want to clean up our Dockerfiles to leverage multistage builds. By using multistage builds, we should be able to leverage the stage cache better (faster build times) as well as slim down our image size.

### Proposed Changes

* Create separate development images
* Implement Entrypoint in production images
* Implement Multi-Stage builds
* Add layer caching (handled in previous PR)
* Implement BuildKit

### Change Details

* Remove the CMD argument from the various Dockerfiles and replace them with the entry point.
Using this approach, we are able to treat our Docker images as containers that have all of the binaries built but do not run any commands.
* Use multistage builds to separate the downloading of dependencies (go get, go mod download, go install) from the copying of the source.
* Added `Dockerfile.bcda_prod` and `Dockerfile.bcdaworker_prod` for future progression to containerized application while keeping development images using the `watch` script.  Development images use `dev` as the stage target name.  Other images that do not have separate development builds use `prod` as the final stage target name.
* Added BuildKit to make better use of layer caching.
  * This change will require the following environment variables to be added locally.
    * `DOCKER_BUILDKIT=1`
    * `COMPOSE_DOCKER_CLI_BUILD=1`
  * These environment variables were added to Github Actions through the `ci-workflow.yml`

### Security Implications

### Acceptance Validation

1. We have multistage builds resulting in smaller images for bcda-app, ssas-app and worker
  * The sizes of all the images have been greatly reduced.  `api` alone went from 3.09GB to 438MB
2. We see significant improvements in build times because of layer caching.
  * The end result is rebuilt times in the minutes instead of seconds.  `api` will rebuild in 600ms without change, 7.5s with code change, and 36s from a fresh build.
3. All tests have passed locally and in Github Actions.

### Feedback Requested

Bring the pain!!


Makefile
```diff
@@ -164,6 +164,12 @@ debug-worker:
 		docker-compose -f docker-compose.yml -f docker-compose.debug.yml run --no-deps -T --rm -v $(shell pwd):/go/src/github.com/CMSgov/bcda-app worker dlv debug"
 
 bdt:
+	# Set up valid client credentials
```
mtrang1263, line 167: Nice improvement!

mtrang1263 (approved):
LGTM! Great job on improving the build times!

CosmicSyntax (dismissed):
LGTM!

mtrang1263 (approved):
🔥 🔥 🔥 🔥

SMLuthi (approved):
LGTM! solid work.

PR #699, opened 2021-04-27
BCDA 4452 Fix: Pin Postgres image version to 10

lostpurplecow:
### Fixes [BCDA-4452](https://jira.cms.gov/browse/BCDA-4452)

The current deployed major version of Postgres in AWS for ACO API is Postgres 10.  However, in the docker-compose stacks for bcda-app we are pulling from postgres:latest (see here for example).

We need to ensure all references to pulling Postgres images from Dockerhub in bcda-app align with the version of Postgres that we are using in AWS.  This should be `postgres:10` image in Dockerhub.

### Proposed Changes

- Pin Postgres image version to 10

### Security Implications

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -17,7 +17,7 @@ services:
       - ${HOME}/.cache/go-build:/root/.cache/go-build
       - ${GOPATH}/pkg/mod:/go/pkg/mod
   db-unit-test:
-    image: postgres:10
+    image: postgres:13
     environment:
       - POSTGRES_PASSWORD=toor
       - POSTGRES_DB=bcda_test
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -17,7 +17,7 @@ services:
       - ${HOME}/.cache/go-build:/root/.cache/go-build
       - ${GOPATH}/pkg/mod:/go/pkg/mod
   db-unit-test:
-    image: postgres
+    image: postgres:10
     environment:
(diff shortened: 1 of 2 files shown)


CosmicSyntax (approved):
LGTM

CosmicSyntax:
Looks like the test is failing with unit-test-db. I am wondering if we are using features not available in version 10...

lostpurplecow:
> Looks like the test is failing with unit-test-db. I am wondering if we are using features not available in version 10...

yes, seems like it, will investigate

msnook:
Could it be caused by the `pgdump` which initializes the DB?  https://github.com/CMSgov/bcda-app/tree/master/db/testing/docker-entrypoint-initdb.d

docker-compose.test.yml
```diff
@@ -17,7 +17,7 @@ services:
       - ${HOME}/.cache/go-build:/root/.cache/go-build
       - ${GOPATH}/pkg/mod:/go/pkg/mod
   db-unit-test:
-    image: postgres
+    image: postgres:13
```
msnook, line 20: Just wanted to comment for anyone looking back at this PR.  The unit test DB relies on a `pgdump` from Postgres 13 DB that makes it very difficult to use that same dump in Postges 10.  Because it is simply the unit test DB, we determined it acceptable to keep this database at 13 while aligning the databases for the API with AWS.

msnook (approved):
👍

PR #700, opened 2021-04-27
BCDA 4454: CI Speed Improvements

adamayd:
### Fixes [BCDA-4454](https://jira.cms.gov/browse/BCDA-4454)

Our current Github Actions solution is not making use of the `actions/cache` action.

### Proposed Changes

Enabled Github Actions `actions/cache` and point it to the Docker cached layers.

### Change Details

Removed the vestigial Github Actions cache steps and the Download Dependencies steps.  Tested to make sure the cache was not providing speed performance due to some other style of caching over the individual runner temp cache.  Since we are using BuildKit we have a few more hoops to jump through because BuildKit expects images to be a part of a registry.  *~This creates a file-system base registry and attaches that to the Github Actions cache.~*  The `api` and `worker` images were split to use a hybrid builder/multi-stage setup since you can use an external image as a stage in multi-stage builds.  *~This allows us to build the builder images locally, then push them to the file-system based cache in Github Actions.  This means any dependency updates will require also rebuilding the `api_builder` and `worker_builder` images and pushing them to the cache.  This is handled in a new `Makefile` section called `builders`.~*

### Security Implications

### Acceptance Validation

Github Actions cache is being used for the "builder" images as a file-system based Docker registry.

### Feedback Requested

This is probably a temporary solutions and [Github Packages](https://github.com/features/packages) would be a better, more stable long-term solution once BuildKit is stable in Docker Community Edition and Github Packages has been vetted as a viable option.   Thoughts on this would be great.


docker-compose.yml
```diff
@@ -33,9 +45,22 @@ services:
       - "3000:3000"
       - "3001:3001"
     depends_on:
```
katyasoup, line 47: should we add this for the worker_builder too?

katyasoup (approved):
One small Q but otherwise LGTM!

docker-compose.yml
adamayd: Those ports are only exposed to reach the API, they aren't needed for the worker.

SMLuthi (approved):
LGTM!

docker-compose.yml
katyasoup: Ah gotcha 👍

PR #703, opened 2021-05-03
bcda-4427 Feature: improve error loggin in err ndjson

SMLuthi:
### Fixes [BCDA-4427](https://jira.cms.gov/browse/BCDA-4427)

Our `*-error.ndjson` files are currently not leveraging the `issue.code`. We want to expand on this and return a `NOT_FOUND` to give better information the user so they can opt to not retry if the request is not possible due to an mbi not being in the system. Our logging also does not store the cms id / job id when an error is encountered in the job processor. We want to leverage this as well so that we can query splunk easier.

### Change Details

- Add fhir code `NOT_FOUND` to error ndjson logs if applicable.
- Add cms id and job id to error log when encountering an error when processing a job. 

### Security Implications

### Acceptance Validation

- Logging in place to associate errors with a particular CMS ID + jobID
- New Operational Outcome errors sent out to the community. This should describe what type of errors they should see and if it's retriable or not.

### Feedback Requested

Should we add the cms id/job id anywhere else?


CosmicSyntax (approved):
LGTM!

CosmicSyntax:
I can't think of any other place to put cmsID + jobID other than the worker log.  I think this suffice for now.

PR #705, opened 2021-05-12
bcda-4520 Feature: allow aco disabling via configs

SMLuthi:
### Fixes [BCDA-4520](https://jira.cms.gov/browse/BCDA-4520)

Currently, if an ACO has valid credentials there is no way programmatically for us to stop them making requests.

### Proposed Changes

Add a flag that allows us to mark them as `disabled` and prevent them from making export requests.

### Change Details

- Add a `Disabled` flag to the `aco.yml` file that will be added to the aco configs (default to `false`)
- Mark `DC` Model as disabled.
- Update the service config with a new method to verify is an ACO is enabled using the new flag.
- Add new middleware to the export job routes that verifies an ACO is not disabled and allowed to make the call.

### Security Implications

### Acceptance Validation

The ACO, if marked as disabled, should no longer be able to make an export job call without getting an error.

### Feedback Requested

This was my proposed solution... there may be a better way to do this.


bcda/web/middleware/middleware.go
```diff
@@ -24,3 +30,26 @@ func SecurityHeader(next http.Handler) http.Handler {
 		next.ServeHTTP(w, r)
 	})
 }
+
+func IsACOEnabled(next http.Handler) http.Handler {
```
SMLuthi, line 34: Gonna rename this. I dont like having both `IsACOEnabled` and `IsACODisabled`

msnook:
Topic for discussion... 
We already have a `blacklisted` mechanism, and we can control data access to an ACO by way of their entry in the ACO table.  Would adding this additional `isAcoDisabled` function be an additional redundant layer?  What if we enable/disable disable ACO access to requesting/accessing data per their entry in the ACO table?  [This ParseToken middleware function](https://github.com/CMSgov/bcda-app/blob/81e56f4d5f6f49541c1b63e438b86e9f0ef7b48a/bcda/auth/ssas.go#L150-L153) fails if the ACO ID in the token does not exist in our database, which means we'd be protected on all auth-required endpoints.  Could we simply address this control via DB migration scripts (i.e., load/unload the ACO in the ACO table)?  Or maybe an `is_active` field for the ACO in the ACO table?
@CosmicSyntax @SMLuthi

bcda/service/config_test.go
```diff
@@ -32,6 +33,24 @@ func TestLoadConfig(t *testing.T) {
...
+func TestIsACODisabled(t *testing.T) {
+	tests := []struct {
+		name     string
+		cmsID    string
+		expected bool
+		cfg      *Config
+	}{
+		{"ACOIsDisabled", "TEST1234", true, &Config{ACOConfigs: []ACOConfig{{patternExp: compileRegex(t, "TEST\\d{4}"), Disabled: true}}}},
+		{"ACOIsEnabled", "TEST1234", false, &Config{ACOConfigs: []ACOConfig{{patternExp: compileRegex(t, "TEST\\d{4}"), Disabled: false}}}},
+		{"ACOIsDoesNotExist", "DNE1234", true, &Config{ACOConfigs: []ACOConfig{{patternExp: compileRegex(t, "TEST\\d{4}"), Disabled: false}}}},
```
katyasoup, line 45: can you also test that an ACOConfig with no `Disabled` field will be treated with the appropriate default?

SMLuthi: That is not really possible... since the conf package front loads this. The defaults are only used in that regard with the viper package. This test kind of circumvents all that.

katyasoup: Gotcha 👍

katyasoup (approved):
LGTM!

CosmicSyntax (approved):
LGTM! Great work!

PR #711, opened 2021-06-10
bcda-4565 Feature: Signal handling for cli tools

SMLuthi:
### Fixes [BCDA-4565](https://jira.cms.gov/browse/BCDA-4565)

Currently, if our cclf and 1800 medicare ingestions process gets a termination signal from AWS it will stop immediately. Since we attempt to ingest on an hourly basis we would like these ingestions to get a chance to finish.

### Proposed Changes

Add signal handling to both ingestion cli process commands that ignores the termination signals.

Note that AWS will send a power down request to its ec2 instances when trying to shut them down. This results in the instance giving a `SIGTERM` to its running processes to kill them gracefully. If the instance takes to long shutting down, AWS will then send a `SIGKILL` signal instead to force it to shutdown immediately. That wait time depends on how our ops team has set that up... though by default the wait time should be more than long enough for our ingestions to finish.

### Change Details

- add `ignoreSignals` method that consumes the `SIGINT` and `SIGTERM` signals, logs that they have been received, and ignores them. 
- add test to verify `ignoreSignals` ignores the signals correctly.

### Security Implications

### Acceptance Validation

CCLF ingestion and 1800 Medicare ingestion now ignores `SIGINT` and `SIGTERM` signals to allows running ingestions to finish.

### Feedback Requested

This was my proposed solution... there may be a better way to do this.

### DIFF
--- a/bcda/bcdacli/cli.go
+++ b/bcda/bcdacli/cli.go
@@ -10,10 +10,12 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"os/signal"
 	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/CMSgov/bcda-app/bcda/alr/csv"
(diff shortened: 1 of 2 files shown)


CosmicSyntax (approved):
LGTM!

According to this:
https://aws.amazon.com/blogs/containers/graceful-shutdowns-with-ecs/
We have 30 seconds from SIGTERM before SIGKILL is sent.  Since both suppression and CCLF is way under this, we should be fine.

CosmicSyntax (approved):
LGTM

katyasoup (approved):
LGTM!

PR #713, opened 2021-06-16
bcda-4516 Feature: implement testing for v2

SMLuthi:
### Fixes [BCDA-4516](https://jira.cms.gov/browse/BCDA-4516)

Since we are almost ready to roll out v2 endpoints in production we want to make sure our test coverage is substantial enough for those endpoints in all our testing infrastructures.

### Proposed Changes

- Update postman tests to include more v2 testing
- Update smoke tests to include more v2 testing
- Update unit tests to include  more v2 testing

### Change Details

- Added more v2 export calls to postman including tests for all endpoints that fail due to not having provided a token and creating and deleting a job. The test order has also been slightly restructured so that the postman tests runner will kick off ALL export jobs (v1 and v2) before starting to consume them by checking the job statuses. Im not sure if this will affect performance at all but theoretically the postman runner wont have to query job statuses as long if they are all queued at once.
- Smoke tests have been updated to include the same amount of tests for v2 as we currently have for v1 endpoints.
- Unit tests coverage for the v2 api endpoints is consistent with what we have currently for v1.

### Security Implications

### Acceptance Validation

We have increased the v2 testing to a level that is acceptable for allowing it to go to production.

### Feedback Requested

I want to make a note here and point out that we have A LOT of postman tests for v1. The testing goes pretty in depth by using the `since` for some tests, testing backlisted acos, and repeating resources in the requests (as some examples). We _could_ double up on these tests for v2 as well... though I think this may be over kill since a lot of this code is used by v1 and v2.

But if theres a consensus for us to want these tests they can be added.

### DIFF
--- a/test/postman_test/BCDA_Tests_Sequential.postman_collection.json
+++ b/test/postman_test/BCDA_Tests_Sequential.postman_collection.json
@@ -3431,6 +3431,253 @@
 			"response": []
 		},
 		{
+			"name": "Start Group/all export v2 without Since",
+			"event": [
+				{
+					"listen": "test",
+					"script": {
+						"exec": [
+							"pm.test(\"Status code is 202\", function() {",
+							"    pm.response.to.have.status(202);",
+							"});",
+							"",
+							"pm.test(\"Has Content-Location header\", function() {",
+							"    pm.response.to.have.header(\"Content-Location\");",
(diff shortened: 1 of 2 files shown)


CosmicSyntax (approved):
🎉

PR #716, opened 2021-06-22
bcda-4649 Feature: API v2 response r4 compatible

SMLuthi:
### Fixes [BCDA-4649](https://jira.cms.gov/browse/BCDA-4649)

Currently, the API uses fhir `stu3` as its response type for both `v1` and `v2` of our API. However, since BFD has moved its response for their `v2` to fhir `R4` we want to do the same for when we create Error Responses.

### Proposed Changes

Identify when requests are being made via `v1` or v2` and use the correct response writer. Create a new writer for `v2` that integrates with the `R4` fhir spec and builds the responses appropriately.

### Change Details

- Create `v2` writer.go in `responseutils` that creates responses using the correct fhir library.
- Move the creating of error responses to both `v1` and `v2` writer files and, therefore, removing fhir library access away from the `requests.go` file.
- Create and interface in requests.go that will satisfy the responsewriter methods we are using.
- Update and add testing where needed.
- Update middleware package to properly deduce API version and load the appropriate responseutility
Note: The auth middleware methods will default to the `v1` response writer because the auth middleware is also used in routes (like the data router) that do not contain version information. This should be changed and updated in the future.
- Update `.../_version` to not respond with a fhir body

### Security Implications

### Acceptance Validation

`v2` calls now build responses using the R4 fhir library spec. Note that there are some differences in there besides the severity and type codes. Things have moved a little bit and certain fields have been removed (i.e. `AcceptUnknown`).
 
### Feedback Requested

This is the solution I came up with. Since the errors we create using the severity and type code are just binary at this point the `ResponseWriter` is still small. However, if these error types ever get larger we will have to add more ResponseWriter error build methods. I dont see this happening but I think its important to note.

### DIFF
--- a/bcda/api/requests.go
+++ b/bcda/api/requests.go
@@ -93,7 +93,7 @@ func newHandler(resources []string, basePath string, apiVersion string, db *sql.
 	case "v2":
 		h.RespWriter = responseutilsv2.NewResponseWriter()
 	default:
-		log.API.Fatalf("Unexpected API version: %s", h.apiVersion)
+		log.API.Fatalf("unexpected API version: %s", h.apiVersion)
 	}
 
 	return h
--- a/bcda/api/requests.go
+++ b/bcda/api/requests.go
@@ -16,9 +16,6 @@ import (
 	"net/http"
 	"time"
 
-	fhircodes "github.com/google/fhir/go/proto/google/fhir/proto/stu3/codes_go_proto"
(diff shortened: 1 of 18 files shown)


CosmicSyntax (commented):
Looks good, but need updates done for the middleware package.

We may need to discuss how to approach bcdaworker and api/v1 (e.g. GetVersion).

CosmicSyntax (approved):
🎉

PR #717, opened 2021-06-28
BCDA-4571 Add support for R4 for ALR

CosmicSyntax:
### Fixes [BCDA-4571](https://jira.cms.gov/browse/BCDA-4571)
When ALR package was originally coded, it solely used Standard for Trial Use 3 (STU3). V2 endpoint are required to be in R4 format. Since we will be serving ALR data through V1 and V2 endpoints, we need to add support for R4.
### Proposed Changes
We need to refactor the directory structure and provide some sort of abstraction where the caller doesn't have to worry about the implementation details. You call a function, and you get JSON either in R4 or STU3.
### Change Details
The original STU3 all have been moved into a new V1 directory. V2 directory contains V1 counterparts for R4.
The AlrFhirBulk struct, which originally represented a row of ALR data, now represents either a row for V1 or V2. When it represents a row for V1, V2 is nil, and vice versa.
The JSON FHIR marshaller is no longer required to be implemented by the caller. It is now part of methods for V1 and V2.
### Security Implications
### Acceptance Validation
Github actions passes and code looks sound.
### Feedback Requested
Any feedback is good feedback!


bcda/models/fhir/alr/v1/coverage.go
```diff
@@ -1,14 +1,17 @@
...
+	"github.com/CMSgov/bcda-app/bcda/models/fhir/alr/utils"
 	fhirdatatypes "github.com/google/fhir/go/proto/google/fhir/proto/stu3/datatypes_go_proto"
 	fhirmodels "github.com/google/fhir/go/proto/google/fhir/proto/stu3/resources_go_proto"
 )
 
 // This part of the package houses the logical to create coverage resource type data
 
 // coverage takes a beneficiary and their respective K:V enrollment and returns FHIR
-func coverage(mbi string, keyValue []kvPair) *fhirmodels.Coverage {
+func coverage(mbi string, keyValue []utils.KvPair, lastUpdated time.Time) *fhirmodels.Coverage {
```
katyasoup, line 14: Do we want to add in any error handling here (and the other resources)? Thinking especially for resources with non-mandatory columns - if we fail to populate certain fields, will that read as the value not being present in the original file? (Maybe not an issue here where everything is a string, but for resources that rely on regex matches, e.g., how do we handle a k:v that has a non-nil value that doesn't match any patterns)?

bcda/models/fhir/alr/v2/observation.go
```diff
@@ -0,0 +1,78 @@
...
+
+	for _, kv := range keyValue {
+
+		// Get information from HCC Crosswalk
+		hccinfo := utils.HccData(version, kv.Key)
+
+		if hccinfo == nil {
+			// If a nil is returned, we could not give the field in the crosswalk...
+			// for now will skip
+			log.API.Warnf("We would not find %s in the crosswalk for %s. with value %s", kv.Key, version, kv.Value)
```
katyasoup, line 53: re: earlier comment about error handling, maybe more checks like this?

bcda/models/fhir/alr/v1/coverage.go
CosmicSyntax: I think for now, we don't include any K:v not mapped to a pattern. But logging it would be a great idea.

katyasoup (dismissed):
LGTM!

CosmicSyntax:
The PR may be borked. Opening a new one.

PR #726, opened 2021-07-14
DCGEO-137: Adding data access control to the ACO Config file

ericbuckley:
### Proposed Changes

- adding the `data` key to the [ACO Config file to control which datasets model participants can access](https://confluence.cms.gov/display/BCDA/Add+data+access+rules+to+ACO+Config)
- adding a new testing ACO Model, `TEST`, so we can experiment with pre-adjudicated data before releasing to end users

### Change Details

- updated the ACO Config file with a new model and the `data` key on every model block

### Security Implications

  - adding a new key to the ACO Config will **eventually** affect which users can access adjudicated (CCW) and pre-adjudicated (RDA) data, however within this PR, it will have no affect as it's not being read by the application

### Acceptance Validation

- verify that the new `data` key meets expectations from the [proposal](https://confluence.cms.gov/display/BCDA/Add+data+access+rules+to+ACO+Config)
- verify that the new `TEST` ACO model will not conflict with existing ACO models


KyRobbins (approved):

bcda/service/config.go
```diff
@@ -53,10 +53,11 @@ type RunoutConfig struct {
...
 	Model              string
-	Pattern            string `conf:"name_pattern"`
-	PerfYearTransition string `conf:"performance_year_transition"`
-	LookbackYears      int    `conf:"lookback_period"`
-	Disabled           bool   `conf:"disabled" conf_default:"false"`
+	Pattern            string   `conf:"name_pattern"`
+	PerfYearTransition string   `conf:"performance_year_transition"`
+	LookbackYears      int      `conf:"lookback_period"`
+	Disabled           bool     `conf:"disabled" conf_default:"false"`
+	Data               []string `conf:"data"`
```
SMLuthi, line 60: Since it looks like every current model has the `"adjudicated"` trait, would it make sense to set this as the default in our config so we have to less to set in the config file(s) itself.

SMLuthi (approved):
LGTM! Just one comment for review.

bcda/service/config.go
ericbuckley: yeah that makes sense, is the following the correct format for `conf_default`?
```
	Data               []string `conf:"data" conf_default:"[1]string{\"adjudicated\"}"`
```

SMLuthi: Im actually not sure myself when it comes to a list, ill double check as well.

ericbuckley: @SMLuthi I took a deep look into this and I don't see an easy path for supporting default values for complex (slices, maps, etc) fields in field tags.

If you want to pair on this some more and try to figure out what changes we can make to `conf/config.go` to support this, I'm open.

SMLuthi: @ericbuckley Like we talked about yesterday. I think this is outside the scope of your change. We can note this down as a ticket for future improvement but I'd rather you not get bogged down trying to make a change to the configuration package!

katyasoup (approved):
LGTM!

PR #727, opened 2021-07-15
BCDA-4663/4696 Feature: Python script to flatten NDJSON -> CSV

jphechter:
### Proposed Changes
- Create a directory for scripts relevant to the application, but not part of primary tooling
- Add a script to convert NDJSON structured data to a single flat CSV
- Update .gitignore
- add README to accompany this script

### Change Details

The script works by flattening nested JSON data, and then builds a CSV.

### Security Implications

- [X] no PHI/PII is affected by this change

### Acceptance Validation
There are no tests for this work currently. This is a standalone script that is meant to be more of a stopgap solution than a fully fledged feature. 

- Review logic of the script
- Recommend following the included README to build Python venv, install package dependencies, and test script
- Review output for all 3 resource types
  - I used data from the sandbox environment to build this. Recommend pulling the same data for testing purposes. I can also save you the time and upload that somewhere.
- Need to manually generate ALR data by running test with output flag:
```go test -v github.com/CMSgov/bcda-app/bcda/models/fhir/alr -run TestGenerateAlr -output -version 1``` 

### Feedback Requested
Open to all forms of feedback and flying accusations. If there's a better way to approach this, I'd love to hear it.

### CODE REVIEW

File: scripts/ndjson_to_csv/ndjson_to_csv.py [hunk 1]
```diff
@@ -0,0 +1,74 @@
...
+        else:
+            items.append((new_key, v))
+    return dict(items)
+
+
+if __name__=='__main__':
+    print("This tool can convert FHIR data stored in an NDJSON file to a CSV format.")
+    while True:
+        # Ingest the report results
+        filename = input("Enter the name of the NDJSON file you wish to convert: ")
```
SMLuthi, line 44: I would recommend using a CLI interface to grab the file name at run time. Added benefit would also be able to include a `-h` helper blurb for the script itself.

scripts/ndjson_to_csv/ndjson_to_csv.py
```diff
@@ -0,0 +1,74 @@
...
+        
+        output_filename = '{}_{}.csv'.format(
+            filename,
+            datetime.datetime.today().date()
+        )
+        print("Creating CSV")
+        output.to_csv(output_filename, index=False)
+
+        print("File converted successfully. Output file '{}'".format(output_filename))
+        if input("Would you like to convert another file? (y/n) ") == 'n':
```
SMLuthi, line 72: As stated above, CLI interface would probably be my preferred approach instead of requesting input from the user during runtime. If flattening more than one file at a time is desired you could have the CLI handle a list of file names at run time.

SMLuthi (commented):
Overall this is looking good! I made a few comments but whether you want to adopt them is up to you.

scripts/ndjson_to_csv/ndjson_to_csv.py
```diff
@@ -0,0 +1,74 @@
...
+
+if __name__=='__main__':
+    print("This tool can convert FHIR data stored in an NDJSON file to a CSV format.")
+    while True:
+        # Ingest the report results
+        filename = input("Enter the name of the NDJSON file you wish to convert: ")
+        f = open(filename, 'r')
+        
+        # Format data
+        string_json = f.read()
```
ericbuckley, line 48: what size NDJSON files do you expect users to be able to use this script with?  I've heard that some of the files ACO's produce are GBs in size, reading this in all at once might cause memory errors

jphechter: That's a really good point that I hadn't taken into account.

scripts/ndjson_to_csv/ndjson_to_csv.py
jphechter: Included `-h`, `--help`, and help message. The script takes 2 positional arguments (source + output file name), verifies their existence or non-existence, and then proceeds.

scripts/ndjson_to_csv/ndjson_to_csv.py
jphechter: @ericbuckley I've completely refactored the way that this works.

Instead of reading the source into memory, I'm iterating over the file line by line. This _should not_ keep too many things in memory, however, I am using a tempfile that will be roughly the same size as the source while the script is running. Let me know if you want to discuss how that may or may not impact folks.

scripts/ndjson_to_csv/ndjson_to_csv.py
```diff
@@ -0,0 +1,122 @@
...
+
+    Usage: ndjson_to_csv.py <inputfile> <output_filename>
+    
+    <inputfile>        Source file to be converted (.json, .ndjson, .txt)
+    <output_filename>  Optional desired output filename
+    """
+
+    argv = sys.argv[1:]
+    
+    if '-h' in argv or '--help' in argv:
```
ericbuckley, line 68: take a look at [argparse](https://docs.python.org/3/library/argparse.html) it can make some of this easier

scripts/ndjson_to_csv/ndjson_to_csv.py
```diff
@@ -0,0 +1,122 @@
...
+
+    # Create Tempfile
+    tf = tempfile.NamedTemporaryFile(prefix=f'{outputfile}_temp_', mode="w+t", delete=True)
+    counter = 0
+
+    # Flatten data structures
+    print(f'Flattening "{inputfile}"...')
+    with open(inputfile, 'r') as source_file:
+        for line in source_file:
+            if line[-1:] == '\n':
```
ericbuckley, line 97: is there typically an extra `\n` per line in NDJSON files?

scripts/ndjson_to_csv/ndjson_to_csv.py
```diff
@@ -0,0 +1,122 @@
...
+    # Error handling
+    if not Path(inputfile).is_file():
+        print(f'Source file "{inputfile}" does not exist.')
+        exit()
+    if Path(outputfile).is_file():
+        print(f'A file with the name "{outputfile}" already exists in this directory.')
+        exit()
+
+    # Create Tempfile
+    tf = tempfile.NamedTemporaryFile(prefix=f'{outputfile}_temp_', mode="w+t", delete=True)
```
ericbuckley, line 90: This may be fine, but you also may need to benchmark this against some example NDJSON files to test performance....

First, great thought on having to iterate through the file twice to accurately get a list of the CSV headers.  I see that you're writing the data to a temp file to avoid having to call flatten twice per line, however there is a cost there in you'll need double the amount of file space to render the result (1 output file for the temp file and another for the csvwriter).  The first should get deleted when the program exits, but that won't be to the very end.

It be interesting to see what this looks like if the first pass only adds to the `col_header` set, and the second pass writes the data to file.  It may be slower computationally, but I believe it would require fewer resources to compute.

Just a thought.

scripts/ndjson_to_csv/ndjson_to_csv.py
```diff
@@ -0,0 +1,122 @@
...
+                            items.extend(flatten(item, new_key, sep).items())
+                else:
+                    for item in v:
+                        items.extend(flatten(item, new_key, sep).items())
+            else:
+                list_strings = [str(item) for item in v]
+                col_headers.add(new_key)
+                items.append((new_key, ' '.join(list_strings)))
+        else:
+            col_headers.add(new_key)
```
ericbuckley, line 51: I think the program might read better if you initialize `col_headers` in the main, then after `flatten` is invoked just add the headers.
```
data = flatten(json.loads(line))
col_headers.update(data.keys())
tf.write(f'{json.dumps(data)}\n')
```

scripts/ndjson_to_csv/ndjson_to_csv.py
jphechter: Not totally sure, but while I was working with test data I think I did encounter some trailing new lines.

scripts/ndjson_to_csv/ndjson_to_csv.py
jphechter: > It be interesting to see what this looks like if the first pass only adds to the col_header set, and the second pass writes the data to file. It may be slower computationally, but I believe it would require fewer resources to compute.

I'm pretty sure the most time consuming part of this is building the CSV. In the interest of saving disk space, this might be worth it.

scripts/ndjson_to_csv/ndjson_to_csv.py
jphechter: Good thought

jphechter: included.

scripts/ndjson_to_csv/ndjson_to_csv.py
jphechter: TL;DR - I'm pushing the changes you suggested and we're flattening the file x2

OK, so I created a larger test file ~340 MB to try and get a better idea for how performant these 2 options are.

**Option 1 (Tempfile)** - Completed in ~47 seconds
**Option 2 (Flatten 2x)** - Completed in ~55 seconds

So by using the tempfile approach we save about 8 seconds and for a brief period triple the disk space used. That extrapolates to 23.5 sec / 1 GB. While that will add up quickly for people who have larger amounts of data, I'm sure they'll appreciate not creating unnecessarily large files on their systems.

If you're curious, it looks as though each pass through `flatten()` was taking ~12 seconds.

scripts/ndjson_to_csv/ndjson_to_csv.py
jphechter: included

carlpartridge:
Closing due to inactivity.

PR #729, opened 2021-07-28
DCGEO-138: ACO authorization changes based on "Data" attribute

KyRobbins:
### Fixes [DCGEO-138](https://jira.cms.gov/browse/DCGEO-138)

Updating the `requests` module to access control on the `type` of FHIR resource requested based on the `Data` permissions defined per ACO Model in the configuration file.  If a model has `adjudicated` access, they will obtain permissions to request `Patient, Coverage and EOB` resources.  If the model has `pre-adjudicated` access, they will obtain permissions to request `Claim and ClaimResponse` resources.  Additionally, the default FHIR resource types have been updated to match the above logic.

### Proposed Changes

Update to allow data-source specification of resources and make ACO level configurations for authorized data types

### Change Details

- Add additional valid Resource types
- Make Resource types indicate valid data source types
- Update validation checks to include authorization check for data type

### Security Implications

ACOs should only have access to their authorized claim data and FHIR Resource types


bcda/api/requests.go
```diff
@@ -34,6 +34,11 @@ import (
 	"github.com/CMSgov/bcda-app/log"
 )
 
+type ResourceType struct {
```
ericbuckley, line 37: can we add a comment block here describing the purpose of this struct.

bcda/api/requests.go
```diff
@@ -280,13 +282,9 @@ func (h *Handler) bulkRequest(w http.ResponseWriter, r *http.Request, reqType se
...
 
-	resourceTypes := rp.ResourceTypes
-	// If caller does not supply resource types, we default to all supported resource types
-	if len(resourceTypes) == 0 {
-		resourceTypes = []string{"Patient", "ExplanationOfBenefit", "Coverage"}
-	}
+	resourceTypes := h.getResourceTypes(rp, ad.CMSID)
 
-	if err = h.validateRequest(resourceTypes); err != nil {
+	if err = h.validateRequest(resourceTypes, ad.CMSID); err != nil { //TODO: Validate type of claim & resource data for user
```
ericbuckley, line 287: is the TODO still needed?

bcda/api/requests.go
```diff
@@ -424,17 +422,61 @@ func (h *Handler) bulkRequest(w http.ResponseWriter, r *http.Request, reqType se
...
+func (h *Handler) authorizedResourceAccess(resource ResourceType, cmsId string) bool {
+	if cfg, ok := h.Svc.GetACOConfigForId(cmsId); ok {
+		return (resource.Adjudicated && contains(cfg.Data, "adjudicated")) ||
+			(resource.PreAdjudicated && contains(cfg.Data, "preadjudicated"))
+	}
+
+	return false
+}
+
+func contains(array []string, str string) bool {
```
ericbuckley, line 470: can we use `ContainsString` instead?  https://github.com/CMSgov/bcda-app/blob/master/bcda/utils/common.go#L63

bcda/api/v1/api.go
```diff
@@ -23,7 +23,14 @@ import (
 var h *api.Handler
 
 func init() {
-	h = api.NewHandler([]string{"Patient", "Coverage", "ExplanationOfBenefit", "Observation"}, "/v1/fhir", "v1")
+	resources := make(map[string]api.ResourceType, 6)
+	resources["Patient"] = api.ResourceType{Adjudicated: true}
+	resources["Coverage"] = api.ResourceType{Adjudicated: true}
+	resources["ExplanationOfBenefit"] = api.ResourceType{Adjudicated: true}
+	resources["Observation"] = api.ResourceType{Adjudicated: true}
+	resources["Claim"] = api.ResourceType{Adjudicated: false, PreAdjudicated: true}
```
ericbuckley, line 31: Claim and ClaimResponse shouldn't be options for v1, BFD will only support pre-adj data in v2

bcda/service/service.go
```diff
@@ -456,6 +458,16 @@ func (s *service) GetJobPriority(acoID string, resourceType string, sinceParam b
 	return priority
 }
 
+func (s *service) GetACOConfigForId(cmsId string) (*ACOConfig, bool) {
```
ericbuckley, line 461: thanks for this, I think other parts of the codebase can reuse this function.  mind adding a docstring?

ericbuckley (commented):
@KyRobbins just a few comments.  Logic seems pretty sound, I think we just need to clean up a few things

ericbuckley (approved):
LGTM

brianburton (dismissed):
LGTM

ericbuckley (dismissed):

ericbuckley (approved):
LGTM

brianburton (approved):
LGTM

bcda/api/requests.go
```diff
@@ -514,17 +515,55 @@ func (h *Handler) bulkRequest(w http.ResponseWriter, r *http.Request, reqType se
...
+			}
+		}
+	}
+
+	return resourceTypes
+}
+
+func (h *Handler) validateRequest(resourceTypes []string, cmsID string) error {
 
 	for _, resourceType := range resourceTypes {
```
katyasoup, line 539: `resourceType` is a bit confusing here, as it's referring to a string, rather than the ResourceType struct. Could we maybe use something like `resourceString` instead (and perhaps `resourceType` in place of `resource`)? Or on the flip side, maybe `ResourceType` could be `ResourceAccess`/`ResourceData` (or similar)?

bcda/api/requests.go
```diff
@@ -514,17 +515,55 @@ func (h *Handler) bulkRequest(w http.ResponseWriter, r *http.Request, reqType se
...
-		if _, ok := h.supportedResources[resourceType]; !ok {
-			return fmt.Errorf("invalid resource type %s. Supported types %s", resourceType, h.supportedResources)
+		resource, ok := h.supportedResources[resourceType]
+
+		if !ok {
+			resources := make([]string, 0, len(h.supportedResources))
+			for k := range h.supportedResources {
```
katyasoup, line 544: Could we move this out (maybe to the `init()`) and avoid having to run this extra loop for every resource param? Having the slice of resource strings available might be handy elsewhere, too.

bcda/api/requests.go
```diff
@@ -514,17 +515,55 @@ func (h *Handler) bulkRequest(w http.ResponseWriter, r *http.Request, reqType se
 	}
 }
 
-func (h *Handler) validateRequest(resourceTypes []string) error {
+func (h *Handler) getResourceTypes(parameters middleware.RequestParameters, cmsID string) []string {
+	resourceTypes := parameters.ResourceTypes
```
katyasoup, line 519: see below re: resourceType vs. ResourceType confusion

katyasoup (commented):
Looking good! A few clarifying Qs in `requests.go`

bcda/api/requests.go
ericbuckley: @katyasoup yeah I agree, the `resource type` seems to have a couple meanings and can be confusing depending on the context.  I was thinking about changing up the labels here to the following, what do you think?

resource type = FHIR resource type (eg Patient, EOB, Claim, Coverage, etc)
data type = upstream source data (eg adjudicated vs pre-adjudicated)

katyasoup: That sounds great! Then we'd have a map of `[resourceType]dataType` which I think is much easier to follow

katyasoup (dismissed):
LGTM! Nice work!

brianburton (approved):

ericbuckley (approved):

bcda/api/requests_test.go
```diff
@@ -219,7 +232,13 @@ func (s *RequestsTestSuite) TestRunoutDisabled() {
...
+	resourceMap := map[string]DataType{
+		"Patient":              {Adjudicated: true},
+		"Coverage":             {Adjudicated: true},
+		"ExplanationOfBenefit": {Adjudicated: true},
+	}
+
+	h := newHandler(resourceMap, "/v1/fhir", "v1", s.db)
 
 	// Use a mock to ensure that this test does not generate artifacts in the queue for other tests
 	enqueuer := &queueing.MockEnqueuer{}
```
CosmicSyntax, line 244: Should we add a test for validating adjudicated? Please let me know if this is already done somewhere and I missed it.

CosmicSyntax (commented):
This looks great. Just one comment on testing in request_test.go.

bcda/api/requests_test.go
ericbuckley: Do you mean a test for validating pre-adjudicated?  From the looks of it, the `TestRequests` function is validating v1 adjudicated requests.

CosmicSyntax: Is adjudicated always set to true, no matter the ACO model?

ericbuckley: No, we're having it be explicitly set, thus the need for having the ACO config models updated across the environments before this is merged.  If an ACO model does not have adjudicated listed in the configuration, they will get a permissions error when making a request for any of the CCW BFD data.

CosmicSyntax: So I believe all request tests are done using this dummy ACO: "ba21d24d-cd96-4d7d-a691-b0e8c88e67a5." Which has a cms_id of "A0002." Should we add another dummy model and test for a negative where adjudicated is not set?

ericbuckley: Ah ha, I see what you're saying now.  Yes I agree, we should ideally have tests for ACOs testing all 4 possibly states (or at least the first 3).
```
1. [adjudicated, pre-adjudicated]
2. [adjudicated]
3. [pre-adjudicated]
4. []
```

bcda/api/requests_test.go
```diff
@@ -216,10 +231,116 @@ func (s *RequestsTestSuite) TestRunoutDisabled() {
 	s.Contains(string(body), "Invalid group ID")
 }
 
+func (s *RequestsTestSuite) TestDataTypeAuthorization() {
```
ericbuckley, line 234: thanks for adding this test case @KyRobbins!

ericbuckley (approved):
🎉

bcda/api/requests_test.go
```diff
@@ -229,6 +231,107 @@ func (s *RequestsTestSuite) TestRunoutDisabled() {
...
+		{"Auth Pre-Adj, Request Adj", "C0000", []string{"Patient"}, http.StatusBadRequest},
+		{"Auth Pre-Adj, Request Pre-Adj", "C0000", []string{"Claim"}, http.StatusAccepted},
+		{"Auth None, Request Adj", "D0000", []string{"Patient"}, http.StatusBadRequest},
+		{"Auth None, Request Pre-Adj", "D0000", []string{"Claim"}, http.StatusBadRequest},
+	}
+
+	for _, test := range tests {
+		s.T().Run(test.name, func(t *testing.T) {
+			w := httptest.NewRecorder()
+			r, _ := http.NewRequest("GET", "http://bcda.ms.gov/api/v2/Group/$export", bytes.NewReader(jsonBytes))
```
CosmicSyntax, line 315: If we are sending an empty body, we can leave this as nil. But not a big deal to change.

bcda/api/requests_test.go
```diff
@@ -322,3 +425,47 @@ func (s *RequestsTestSuite) genASRequest() *http.Request {
 
 	return req.WithContext(ctx)
 }
+
+type MockService struct {
```
CosmicSyntax, line 429: We usually auto-generate mocks using mockery. But we can create a separate ticket to do this later.

CosmicSyntax (approved):
🎉

ericbuckley (approved):

katyasoup (approved):
🎉 LGTM!

bcda/api/requests_test.go
SMLuthi: Actually, I think we should switch to using the mock service thats already in place to avoid stacking up tech debt. Look for usage of `mockSvc` in this file.

SMLuthi (changes requested):

SMLuthi (approved):
Since Kyle is addressing the mock service stuff in his next PR ill just approve this!

https://github.com/CMSgov/bcda-app/pull/733/files#diff-7fc3cc46a29844127337e64bc5dbd5d50fbfa043073ae134ea665a8c07cedd9bR305

PR #732, opened 2021-08-04
Removed printing config into log for service and db

CosmicSyntax:
### Fixes Configuration to Remove Unnecessary entries into log
When we added a new feature to better track and manage our environment variables (EV) into a single location, we had logs be created when the EV were successfully load. We are currently logging data that is better kept off.
### Proposed Changes
This is short PR to remove code that is responsible for logging configuration data.
Two files are affect:
1. config.go in database
2. config.go in service
### Change Details
Codes mentioned above have been removed.
### Security Implications
### Acceptance Validation
All unnecessary data logged previous no longer occurs, and CI passes.

### DIFF
--- a/bcda/database/config.go
+++ b/bcda/database/config.go
@@ -31,6 +31,5 @@ func LoadConfig() (cfg *Config, err error) {
 		return nil, errors.New("invalid config, QueueDatabaseURL must be set")
 	}
 
-	logrus.Infof("Successfully loaded config %+v.", cfg)
 	return cfg, nil
 }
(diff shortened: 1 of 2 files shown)


bcda/database/config.go
```diff
@@ -31,6 +31,5 @@ func LoadConfig() (cfg *Config, err error) {
 		return nil, errors.New("invalid config, QueueDatabaseURL must be set")
 	}
 
-	logrus.Infof("Successfully loaded config %+v.", cfg)
```
msnook, line 34: Does it make sense to just say `Successfully loaded config`?

msnook (approved):
👍

PR #733, opened 2021-08-12
DCGEO-149: Add job queing for pre-adjudicated data types

KyRobbins:
### Fixes [DCGEO-149](https://jira.cms.gov/browse/DCGEO-149)

Need to update BCDA to integrate with BFD for new fhir resources and allow new ACO configuration models

### Proposed Changes

- Add job processing for new fhir resources
- Add/update testing to account for new ACO configuration models

### Change Details

- Jobs are now added on a per DataType basis, i.e. adjudcated data jobs are run separate from pre-adjudicated jobs
- The current assumption (based on current needs) is that Claim/ClaimResponse are always pre-adjudicated (bcdaworker/worker/worker.go), this may change in the future, requiring rework of worker code.
- New bcda/service/resources.go added for DataType logic

### Security Implications

- Only ACOs with access to pre-adjudicated data should be able to access the pre-adjudicated resources
  - (Likewise for adjudicated resources)

  - request verification altered to account for ACO data type allowances

### Acceptance Validation

- ACO configurations should allow for existing features to continue working as they were
- New resources should be served if requested and requesting ACO has rights to pre-adj data
- Any existing jobs in the queue should not cause errors with these changes if deploy happens mid processing.

### Feedback Requested

- Do the supplied changes follow existing coding practices?
- Are the new features being adequately tested?
- Is there any anticipation the changes will cause any issues with existing unfinished jobs on deploy?


bcda/client/bluebutton.go
```diff
@@ -185,7 +185,7 @@ func (bbc *BlueButtonClient) GetClaim(mbi, jobID, cmsID, since string, transacti
 	mbiHash := HashIdentifier(mbi)
 	params := GetDefaultParams()
 	params.Set("mbi", mbiHash)
-	params.Set("excludeSAMHSA", "true")
+	//params.Set("excludeSAMHSA", "true")
```
brianburton, line 188: Why leave this in commented out?  It would be cleaner to just remove it.  Ditto for the other cases.

bcda/service/resources.go
```diff
@@ -0,0 +1,68 @@
...
+
+// init creates a map of ResourceType => DataType configurations to be used by later functions
+func init() {
+	dataTypeMap = map[string]DataType{
+		"Patient":              {Adjudicated: true},
+		"Coverage":             {Adjudicated: true},
+		"ExplanationOfBenefit": {Adjudicated: true},
+		"Observation":          {Adjudicated: true},
+		"Claim":                {Adjudicated: false, PreAdjudicated: true},
+		"ClaimResponse":        {Adjudicated: false, PreAdjudicated: true},
```
brianburton, line 33: Assuming values default to false could we just remove the Adjudicated value for Claim and ClaimResponse?

brianburton (approved):
LGTM!

brianburton (dismissed):

bcdaworker/worker/worker.go
```diff
@@ -169,6 +170,8 @@ func writeBBDataToFile(ctx context.Context, r repository.Repository, bb client.A
 		bundleFunc = func(bene models.CCLFBeneficiary) (*fhirmodels.Bundle, error) {
 			return bb.GetPatient(bene.BlueButtonID, strconv.Itoa(jobArgs.ID), cmsID, jobArgs.Since, jobArgs.TransactionTime)
 		}
+		//TODO: The assumption is Claim/ClaimResponse is always pre-adjudicated, future work may require checking what
```
ericbuckley, line 173:
```suggestion
		//NOTE: The assumption is Claim/ClaimResponse is always pre-adjudicated, future work may require checking what
```

ericbuckley (dismissed):
looks good @KyRobbins thanks for adding the createQueueJobs tests!

bcda/service/resources_test.go
```diff
@@ -0,0 +1,142 @@
...
+			DataType{Adjudicated: true, PreAdjudicated: true},
+			constants.Adjudicated,
+			true,
+		},
+		{
+			"Valid Pre-Adjudicated Type",
+			DataType{Adjudicated: true, PreAdjudicated: true},
+			constants.PreAdjudicated,
+			true,
+		},
```
brianburton, line 37: Since both permissions are true in these will the test really ensure the appropriate one was being checked?  I would suggest making the one not being tested false.

brianburton (dismissed):
Looks good!

brianburton (approved):
Looks good!

ericbuckley (approved):
LGTM

bcda/service/resources.go
```diff
@@ -0,0 +1,68 @@
...
+// GetDataTypes creates a map of the given resourceNames with their associated DataType objects
+// It returns the resource map and a status flag, with true meaning all resources were found and mapped.
+func GetDataTypes(resourceNames ...string) (map[string]DataType, bool) {
+	foundAll := true
+
+	returnMap := make(map[string]DataType, len(resourceNames))
+
+	if len(resourceNames) == 0 {
+		// If no resource specified, return copy of full map
+		for name, entry := range dataTypeMap {
```
CosmicSyntax, line 53: Could we replace this block of code with:
returnMap = dataTypeMap

ericbuckley: Sorry to be late to the conversation, but won't a `returnMap = dataTypeMap` assignment return the original private variable when called?  Meaning, if a downstream method altered the map it would change the value defined in the `init` method?  I'm new to go, but wouldn't it be safer to do the copy?

ericbuckley: @CosmicSyntax is the above correct or will `returnMap` always be a copy?

CosmicSyntax: It is a copy, but a copy of a "smart pointer."  Which means yes, any change down stream would change the globally defined variable. 

This map isn't mutated anywhere, and it's on a hot path. But; on second thought now, I think you're right if you're alluding to future memory safety is worth a little bit of the overhead.

CosmicSyntax: In other words: great point Eric and sorry Kyle; let's change this back.

ericbuckley (approved):
👍

CosmicSyntax (approved):
🍺

katyasoup (approved):

PR #739, opened 2021-08-20
consolidate dockerfiles and docker services

ericbuckley:
Reducing the number of Dockerfiles and docker compose services in use to make it slightly easier to understand when running locally.  This does have some potential performance implications as well, since few services will need to run in the compose network (albeit the services removed are using little to no resources, so the gain is probably neglagible).

* Some of these changes (notabely removing the `_builder` services) makes it easier to reuse the Dockerfiles in custom docker-compose environments created outside this project (the DC Geo team has been doing this to test pre-adj claims integration between BFD & BCDA)

In general I've used 3 strategies to reduce Dockerfiles and compose services
1) [Docker multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) allows us to use intermediate Dockerfiles for build operations, using this feature allows us to consolidate the `_builder` Dockerfiles into their intended service
2) docker compse allows us to override both the `entrypoint` and the `command` instructions so we can use non-prod commands locally
3) docker run commands, with mounted volumes and networks, let us run image specific operations on existing services without the need for building our own images and/or compose services 

### Acceptance Validation
- developers should be able to run `make docker-bootstrap` and have their local docker environments spin up without any errors

### Feedback Requested

- Are these changes *actually* easier to read, or does having additional Dockerfiles and compose services make it easier?


SMLuthi (approved):
This LGTM! I've been wanting to consolidate our Dockerfiles using multistage builds for awhile so Im excited for this! Plus pushing out the `wait-for-it` and `migrate` sections to pre-existing services is great for clean up. Thanks for doing this!

Dockerfiles/Dockerfile.bcda_prod
```diff
@@ -1,7 +1,32 @@
...
+WORKDIR /go/src/github.com/CMSgov/bcda-app/bcda
+RUN --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    go install -v 
+
 # syntax = docker/dockerfile:1-experimental
 FROM golang:1.15-alpine3.12 
+ARG ENVIRONMENT
+# only add dev packages if the environment argument was set to development
+RUN [ "$ENVIRONMENT" = "development" ] && apk add git openssl entr bash
```
brianburton, line 27: Shouldn't this really be in the start script instead of the Dockerfile?  We should be using the same images locally as we would deploy to production and apk could be run once locally if we need to add some special tooling there.

Dockerfiles/Dockerfile.bcdaworker_prod
```diff
@@ -1,6 +1,31 @@
...
+WORKDIR /go/src/github.com/CMSgov/bcda-app/bcdaworker
+RUN --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    go install -v
+
 # syntax = docker/dockerfile:1-experimental
 FROM golang:1.15-alpine3.12 
+ARG ENVIRONMENT
+# only add dev packages if the environment argument was set to development
+RUN [ "$ENVIRONMENT" = "development" ] && apk add git openssl entr bash
```
brianburton, line 27: Same question as above.

Dockerfiles/Dockerfile.bcda_prod
ericbuckley: Are you suggesting we run the `apk add...` command at the start of the watch script?  If so one problem with that is the watch script is written in bash, so we need to have bask installed before we can execute it

CosmicSyntax (dismissed):
🎉

ericbuckley:
@CosmicSyntax are you ok if I merge these changes or would you like to see more reviews first?

jphechter (dismissed):
LGTM 👍

CosmicSyntax (approved):
🎉

PR #740, opened 2021-08-20
faster loading of synthetic cclf data

ericbuckley:
As is, the `make load-synthetic-cclf-data` target iterates over the ACO size list and issues a docker run command to load the data for each size.  By running the docker container once, and looping inside the container execution, we can remove the cost of container setup and teardown with each iteration of the ACO size.

Locally, I observed ~3x speed increase when running the `make load-synthetic-cclf-data` target

### Proposed Changes

- adding a new `scripts/` directory to manage top level shell scripts
- creating a new `bulk_import_synthetic_cclf_package.sh` script to execute the import command multiple times (once for each size given in the arguments list)

### Acceptance Validation

- verify the target still load all the testing CCLF data
- verify the new version of the target execute faster locally

### Feedback Requested

- thoughts on using a `scripts/` directory for managing shell scripts?


CosmicSyntax (approved):
🎉

PR #764, opened 2022-02-10
PACA-429: Add logging details to ServeData

KyRobbins:
### Fixes [PACA-429](https://jira.cms.gov/browse/PACA-429)

Add logging details about ServeData requests including ACO_ID, CMS_ID, and Resource type of the file being downloaded.

### Security Implications


bcda/api/v1/api.go
```diff
@@ -285,6 +293,36 @@ func ServeData(w http.ResponseWriter, r *http.Request) {
...
+		var jobKeys, err = repository.GetJobKeys(context.Background(), uint(jobIdInt))
+
+		if err == nil {
+			var resourceType string
+
+			for _, j := range jobKeys {
+				if strings.TrimSpace(j.FileName) == strings.TrimSpace(fileName) {
+					resourceType = j.ResourceType
+				}
+			}
```
brianburton, line 310: This is calling `trimSpace` on `fileName` multiple times.

Is it harmless if `resourceType` isn't set?

KyRobbins: ah, in the loop, misunderstood, fixed this

brianburton (approved):
LGTM

bcda/api/v1/api.go
```diff
@@ -285,6 +293,38 @@ func ServeData(w http.ResponseWriter, r *http.Request) {
 	} else {
 		http.ServeFile(w, r, fmt.Sprintf("%s/%s/%s", dataDir, jobID, fileName))
 	}
+
+	// Logging request for auditing
+	var jobIdInt, err = strconv.ParseUint(jobID, 10, 64)
+
+	if err == nil {
+		var jobKeys, err = repository.GetJobKeys(context.Background(), uint(jobIdInt))
```
ericbuckley, line 301: what do you think about adding a new function to repository with the following signature to encapsulate some of this logic?  `GetJobKey(ctx context.Context, jobID uint, fileName string) (*models.JobKey, error)`

ericbuckley: I think it will be easier to write a test case for that if it stands alone as its own func

bcda/models/postgres/repository.go
```diff
@@ -509,6 +509,22 @@ func (r *Repository) GetJobKeys(ctx context.Context, jobID uint) ([]*models.JobK
...
+func (r *Repository) GetJobKey(ctx context.Context, jobID uint, fileName string) (*models.JobKey, error) {
+	sb := sqlFlavor.NewSelectBuilder().Select("id", "file_name", "resource_type").From("job_keys")
+	sb.Where(sb.And(sb.Equal("job_id", jobID), sb.Equal("file_name", fileName)))
+
+	query, args := sb.Build()
+	row := r.QueryRowContext(ctx, query, args...)
+
+	jk := models.JobKey{JobID: jobID}
+
+	if err := row.Scan(&jk.ID, &jk.FileName, &jk.ResourceType); err != nil {
```
ericbuckley, line 521: If the select statement returns 0 rows, will Scan return an error or will it just fill null values into the struct?

bcda/api/v1/api.go
```diff
@@ -285,6 +293,28 @@ func ServeData(w http.ResponseWriter, r *http.Request) {
...
+	var jobIdInt, err = strconv.ParseUint(jobID, 10, 64)
+
+	if err == nil {
+		var jobKey, err = repository.GetJobKey(context.Background(), uint(jobIdInt), strings.TrimSpace(fileName))
+
+		if err == nil {
+			var authData, ok = r.Context().Value(auth.AuthDataContextKey).(auth.AuthData)
+
+			if ok {
+				log.API.Infof("ACO_ID: '%s', CMS_ID: '%s', Resource: '%s'", authData.ACOID, authData.CMSID, jobKey.ResourceType)
```
ericbuckley, line 307: I think we only need this logging statement; the 3 else blocks below indicate other issues with the request and or the job in question that would result in a non 200 response from ServeData.

bcda/models/postgres/repository.go
KyRobbins: Using a filename that wasn't in the DB, it seemed to generate an error.

ericbuckley (approved):

brianburton (dismissed):

bcda/api/v1/api.go
```diff
@@ -285,6 +293,22 @@ func ServeData(w http.ResponseWriter, r *http.Request) {
 	} else {
 		http.ServeFile(w, r, fmt.Sprintf("%s/%s/%s", dataDir, jobID, fileName))
 	}
+
+	// Logging request for auditing
+	var jobIdInt, err = strconv.ParseUint(jobID, 10, 64)
+
+	if err == nil {
+		var jobKey, err = repository.GetJobKey(context.Background(), uint(jobIdInt), strings.TrimSpace(fileName))
```
whytheplatypus, line 301: Should probably use the request context here.

bcda/models/postgres/repository.go
```diff
@@ -509,6 +509,22 @@ func (r *Repository) GetJobKeys(ctx context.Context, jobID uint) ([]*models.JobK
...
 }
 
+func (r *Repository) GetJobKey(ctx context.Context, jobID uint, fileName string) (*models.JobKey, error) {
+	sb := sqlFlavor.NewSelectBuilder().Select("id", "file_name", "resource_type").From("job_keys")
+	sb.Where(sb.And(sb.Equal("job_id", jobID), sb.Equal("file_name", fileName)))
+
+	query, args := sb.Build()
+	row := r.QueryRowContext(ctx, query, args...)
+
+	jk := models.JobKey{JobID: jobID}
```
whytheplatypus, line 519: Make this a pointer rather than reference it `&` below.

bcda/models/postgres/repository.go
```diff
@@ -509,6 +509,22 @@ func (r *Repository) GetJobKeys(ctx context.Context, jobID uint) ([]*models.JobK
 	return keys, nil
 }
 
+func (r *Repository) GetJobKey(ctx context.Context, jobID uint, fileName string) (*models.JobKey, error) {
+	sb := sqlFlavor.NewSelectBuilder().Select("id", "file_name", "resource_type").From("job_keys")
+	sb.Where(sb.And(sb.Equal("job_id", jobID), sb.Equal("file_name", fileName)))
+
+	query, args := sb.Build()
+	row := r.QueryRowContext(ctx, query, args...)
```
whytheplatypus, line 517: Make sure to `close` the row

bcda/api/v1/api.go
```diff
@@ -22,8 +27,11 @@ import (
 )
 
 var h *api.Handler
+var repository *postgres.Repository
 
 func init() {
+	repository = postgres.NewRepository(database.Connection)
```
whytheplatypus, line 33: It looks like this is already available through the `h.Svc` variable. You may have to add to it's set of functions.

bcda/api/v1/api.go
```diff
@@ -285,6 +293,22 @@ func ServeData(w http.ResponseWriter, r *http.Request) {
...
+	if err == nil {
+		var jobKey, err = repository.GetJobKey(context.Background(), uint(jobIdInt), strings.TrimSpace(fileName))
+
+		if err == nil {
+			var authData, ok = r.Context().Value(auth.AuthDataContextKey).(auth.AuthData)
+
+			if ok {
+				log.API.Infof("ACO_ID: '%s', CMS_ID: '%s', Resource: '%s'", authData.ACOID, authData.CMSID, jobKey.ResourceType)
+			}
+		}
```
whytheplatypus, line 309: Move this into a function to make it possible to simplify the error handling.

whytheplatypus (changes requested):

bcda/models/postgres/repository.go
KyRobbins: I'm not super familiar with go, I was looking to try to information on closing a single row.  I found this in the documentation which seems to infer that the wrapped `Rows` object is automatically closed after Scan()?

https://github.com/jackc/pgx/blob/v3.6.2/query.go#L23

bcda/api/v1/api.go
KyRobbins: Would you mind clarifying the intent here?  Should lines 298 to 310 into a separate function and simply invoke it from here?  We weren't currently doing anything with the err because appropriate responses were being returned by later logic in the request chain

bcda/models/postgres/repository.go
whytheplatypus: Great find! I was hunting for this in their library and didn't see it, ty ty!

bcda/api/v1/api.go
whytheplatypus: yeah sorry for the lack of clarity. If you move this into a separate function that returns the data you need and an error you can avoid the nested `if`s. This is a minor readability request more than anything.

whytheplatypus: right so it would have things like:
```go
var jobIdInt, err = strconv.ParseUint(jobID, 10, 64)
if err != nil {
    return error, nil // or whatever the signature would be.
}
```

The signature could either return an error and the data you need, or it could simply perform the log. It will be easier to test if it returns the error and data.

whytheplatypus (approved):

ericbuckley (approved):

ericbuckley (approved):
looks great, thanks for updating the mocks

brianburton (approved):
Looks good.

whytheplatypus (approved):

PR #766, opened 2022-02-24
PACA-465: enabling SAMHSA filtering on all Claim/ClaimResponse calls

ericbuckley:
### Fixes [PACA-465](https://jira.cms.gov/browse/PACA-465)

We need to exclude all claims that have SAMHSA diagnosis and/or procedure codes in it.  The filtering was added to the BFD API in the Fall of 2021, but never enabled here.

### Proposed Changes

On all Claim or ClaimResponse API calls, we should be sending the query parameter `excludeSAMHSA=true` to filter out claims with SAMHSA codes.

### Security Implications


brianburton (approved):

KyRobbins (approved):

whytheplatypus (approved):
lgtm

PR #767, opened 2022-04-26
Log the resource type associated with a download job

whytheplatypus:
- starts a pattern for adding fields to the contextual log (the log associated with the request).
- using middleware to decorate functions with additional request level logging.

### Proposed Changes

- [x] Moving to structured logging
- [x] Upping test coverage
- [x] Experimenting with CI pipeline checks

### Change Details

Using the structured contextual logging provided by the chi middleware

### Security Implications
None: adding only one non-pii/phi field to logging.

None

### DIFF
--- a/bcda/logging/middleware.go
+++ b/bcda/logging/middleware.go
@@ -13,9 +13,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/CMSgov/bcda-app/bcda/auth"
-	"github.com/CMSgov/bcda-app/bcda/database"
 	"github.com/CMSgov/bcda-app/bcda/models"
-	"github.com/CMSgov/bcda-app/bcda/models/postgres"
 	"github.com/CMSgov/bcda-app/bcda/servicemux"
 	"github.com/CMSgov/bcda-app/log"
 )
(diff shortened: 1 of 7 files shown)


4ell0 (approved):
Looks good

bcda/logging/middleware.go
```diff
@@ -83,6 +86,48 @@ func (l *StructuredLoggerEntry) Panic(v interface{}, stack []byte) {
 	})
 }
 
+type ResourceTypeLogger struct {
+	Repository models.JobKeyRepository
+}
+
+func (rl *ResourceTypeLogger) LogJobResourceType(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		next.ServeHTTP(w, r)
```
ericbuckley, line 95: Am I reading this correctly that only a 200 response from `v1.ServeData` will result in the resource type being logged?  Not necessarily a bad thing, just wanted to make sure I was interpreting this middleware function correctly.

ericbuckley (approved):
looks great, thanks for moving this into middleware

bcda/logging/middleware.go
whytheplatypus: This should run regardless, although if ServeData isn't returning a 200 it's likely that the same thing that caused that will mean there's no ResourceType to log (e.g. the job id was wrong, or something)

ian-a-sawyer (approved):
Looks great!

KyRobbins (approved):

PR #770, opened 2022-05-11
Updates to allow `make unit-test` to pass

whytheplatypus:
Before this running `make unit-test` on it's own would fail due to a missing external dependency. This is used by Jenkins.

### Change Details

Include local `ssas` as a dependency of the `unit-test` make command.


4ell0 (approved):
Ran locally and make unit-test passed.

PR #771, opened 2022-05-12
Correct test coverage for all packages

whytheplatypus:
Before this test coverage would not be counted for called functions from
dependencies (sub directories)

### Change Details

Updated test call to check for how tests cover other packages when running

Locally this upped coverage from 70% to 78%


4ell0 (approved):
LGTM

PR #774, opened 2022-06-22
BCDA-5795 Increase timeout to 5 seconds for SSAS http client

magile32:
### Fixes [BCDA-5795](https://jira.cms.gov/browse/BCDA-5795)

Increasing the timeout on the client side for SSAS api requests


magile32:
Update env variables instead. Closing PR

PR #795, opened 2022-09-16
BCDA-6022: Provide expires_in value in Response Body for /auth/token

ian-a-sawyer:
### Proposed Changes
- Add new key "expires_in" to /auth/token endpoint response to follow https://tools.ietf.org/html/rfc6749#section-5.1 recommendations
- Example response below in postman
Example ExpiresIn

### Security Implications

### Acceptance Validation
- Tests pass
- Verify response in postman
- (`make credentials` command allows you to create credentials for local env)

### DIFF
--- a/bcda/auth/api_test.go
+++ b/bcda/auth/api_test.go
@@ -27,6 +27,7 @@ import (
 
 type TokenResponse struct {
 	AccessToken string `json:"access_token"`
+	ExpiresIn   string `json:"expires_in,omitempty"`
 	TokenType   string `json:"token_type"`
 }
 
@@ -50,9 +51,9 @@ func (s *AuthAPITestSuite) TestAuthToken() {
 	clientID, clientSecret, accessToken := uuid.New(), uuid.New(), uuid.New()
 	mock := &auth.MockProvider{}
 	mock.On("MakeAccessToken", auth.Credentials{ClientID: clientID, ClientSecret: clientSecret}).
-		Return(accessToken, nil)
+		Return(accessToken, "", nil)
 	mock.On("MakeAccessToken", auth.Credentials{ClientID: "not_a_client", ClientSecret: "not_a_secret"}).
-		Return("", errors.New("some auth error"))
+		Return("", "", errors.New("some auth error"))
 	auth.SetMockProvider(s.T(), mock)
 
 	// Missing authorization header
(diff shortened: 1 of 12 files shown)


bcda/auth/api_test.go
```diff
@@ -50,9 +51,9 @@ func (s *AuthAPITestSuite) TestAuthToken() {
 	clientID, clientSecret, accessToken := uuid.New(), uuid.New(), uuid.New()
 	mock := &auth.MockProvider{}
 	mock.On("MakeAccessToken", auth.Credentials{ClientID: clientID, ClientSecret: clientSecret}).
-		Return(accessToken, nil)
+		Return(accessToken, "", nil)
```
krobertson3314, line 54: this is the happy path scenario, right? if so can we mock the return as a value (ex. 1200) instead of empty string, and then assert it returned the correct value (vs asserting ExpiresIn returned empty)?

bcda/auth/client/ssas_test.go
```diff
@@ -273,12 +273,13 @@ func (s *SSASClientTestSuite) TestGetToken() {
 		s.FailNow(constants.CreateSsasErr, err.Error())
 	}
 
-	respKey, err := client.GetToken(authclient.Credentials{ClientID: "happy", ClientSecret: "client"})
+	respKey, respExp, err := client.GetToken(authclient.Credentials{ClientID: "happy", ClientSecret: "client"})
```
krobertson3314, line 276: can we have this happy path test case return respExp with a value, that we can then assert as populated? (would need to update line 260 to have w.Write include the new expires_in with sample value used for testing)

bcda/auth/ssas_test.go
```diff
@@ -212,34 +212,40 @@ func (s *SSASPluginTestSuite) TestMakeAccessToken() {
 	}
 	s.p = SSASPlugin{client: c, repository: s.r}
 
-	ts, err := s.p.MakeAccessToken(Credentials{ClientID: "mock-client", ClientSecret: "mock-secret"})
+	ts, ei, err := s.p.MakeAccessToken(Credentials{ClientID: "mock-client", ClientSecret: "mock-secret"})
 	assert.Nil(s.T(), err)
 	assert.NotEmpty(s.T(), ts)
+	assert.Equal(s.T(), ei, "")
```
ian-a-sawyer, line 218: I will update this test as well.

bcda/auth/ssas.go
```diff
@@ -96,14 +96,14 @@ func (s SSASPlugin) RevokeSystemCredentials(ssasID string) error {
 }
 
 // MakeAccessToken mints an access token for the given credentials.
-func (s SSASPlugin) MakeAccessToken(credentials Credentials) (string, error) {
-	ts, err := s.client.GetToken(client.Credentials{ClientID: credentials.ClientID, ClientSecret: credentials.ClientSecret})
+func (s SSASPlugin) MakeAccessToken(credentials Credentials) (string, string, error) {
+	ts, ei, err := s.client.GetToken(client.Credentials{ClientID: credentials.ClientID, ClientSecret: credentials.ClientSecret})
```
krobertson3314, line 100: let's clean up the variable naming here. I know they had 'ts' existing in codebase but to make it cleaner code let's change it tokenString and expiresIn instead of 'ts' and 'ei'.

krobertson3314 (approved):

PR #799, opened 2022-10-11
PACA-794 Feature: Support ACO entries for individual entities

brianburton:
Retains original order of ACOConfigs rather than hashing them.

### Fixes [PACA-794](https://jira.cms.gov/browse/PACA-794)

To support a pilot release of PACA data in production, we may want to limit the number of entities that have access to the data within the GPDC model.  To do so, we need to modify the ACO config parsing logic to support ACOIDs that potentially match more than one pattern.

### Proposed Changes

Search ACO patterns in original order from config file with first matching config being used.

### Change Details

- Replaces `acoConfig` map in `service` struct with `acoConfigs` which is simply the original array of `ACOConfigs` objects.
- Changes `setClaimsDate` and `GetACOConfigForID` to iterate through the array instead of through keys from the map.

### Security Implications


bcda/service/service.go
```diff
@@ -496,8 +489,8 @@ func (s *service) setClaimsDate(args *models.JobEnqueueArgs, conditions RequestC
 		args.ClaimsWindow.UpperBound = conditions.claimsDate
 	}
 
-	for pattern, cfg := range s.acoConfig {
-		if pattern.MatchString(conditions.CMSID) {
+	for _, cfg := range s.acoConfigs {
+		if cfg.patternExp.MatchString(conditions.CMSID) {
 			args.ClaimsWindow.LowerBound = cfg.LookbackTime()
 			break
 		}
```
KyRobbins, line 496: It would seem this is config order dependent?  i.e. if two models overlap (such as with a pilot) it will just use whichever is found first.  Do we need to make a special note of that somewhere?

ericbuckley: Yeah that's correct, I'll look to see if there is a confluence spec somewhere on the ACO config file, but maybe we should also a) document it in the yaml and b) document it in the ACOConfig struct.   Thoughts?

brianburton: I have added comments in the struct and in each place the configs are searched.

ericbuckley (approved):

KyRobbins (approved):

krobertson3314 (approved):

PR #812, opened 2022-12-05
BCDA-6109: Remove jobId/clientId headers for non-bulk requests

ian-a-sawyer:
### Proposed Changes
- Split addRequestHeaders function into multiple functions pertaining to non-bulk/bulk requests

### Security Implications

### Acceptance Validation
- Tests pass

### Feedback Requested
- Verify code looks correct


bcda/client/bluebutton_test.go
```diff
@@ -516,6 +534,7 @@ func (s *BBRequestTestSuite) TestValidateRequest() {
 				noExcludeSAMHSAChecker,
 				noIncludeAddressFieldsChecker,
 				noIncludeTaxNumbersChecker,
+				hasDefaultRequestHeaders,
```
krobertson3314, line 537: For this scenario, can you also add an assertion method that confirms the bulk request headers are not present? (instead of leaving the assertion off for hasBulkRequestHeaders) Since this call makes an underlying call to the getRawData.

ian-a-sawyer: Updated

ian-a-sawyer: https://github.com/CMSgov/bcda-app/pull/812/files#diff-3642e6e64fc5fe565b5224862f2f624cba8e76e0ebc297dd0c778d98b0fa2d89R922

PR #816, opened 2022-12-22
PACA-875: Manual migration to partially adjudicated runout file records.

brianburton:
### Fixes [PACA-875](https://jira.cms.gov/browse/PACA-875)

Runout queries are not working for PACA test accounts in sandbox environment.  The problem appears to be caused by the type and timestamp parameter values in the runout file records.  This PR includes a manual migration that would need to be executed in the sandbox database to change these values.

### Proposed Changes

Adds manual migration file to update `type` and `timestamp` in PACA runout `cclf_files` records.

### Change Details

Adds manual migration file to update `type` and `timestamp` in PACA runout `cclf_files` records.

### Security Implications

### Feedback Requested

Is this change correct for its intended purpose?

### DIFF
--- a/db/migrations/manual/20221222-partially-adj-synthea-rel1-runout-update.sql
+++ b/db/migrations/manual/20221222-partially-adj-synthea-rel1-runout-update.sql
@@ -1,0 +1,14 @@
+-- Updating the sandbox runout files for partially adjudicated TEST99x accounts
+-- to allow runout testing.
+
+BEGIN;
+UPDATE cclf_files
+    set type=1,
+        updated_at=now(),
+        timestamp='2022-09-29',
+        performance_year=22
+    where
+        cclf_num=8
+        and name in ('T.BCD.TEST990.ZC8R21.D210929.T1834180','T.BCD.TEST991.ZC8R21.D210929.T1834180','T.BCD.TEST992.ZC8R21.D210929.T1834180')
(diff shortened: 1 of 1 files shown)


KyRobbins (approved):

db/migrations/manual/20221222-partially-adj-synthea-rel1-runout-update.sql
```diff
@@ -0,0 +1,14 @@
+-- Updating the sandbox runout files for partially adjudicated TEST99x accounts
+-- to allow runout testing.
+
+BEGIN;
+UPDATE cclf_files
+    set type=1,
+        updated_at=now(),
+        timestamp='2022-09-29',
```
KyRobbins, line 8: Not sure if timestamp year should be updated, or if it needs to match the year in the name set, question for BCDA

daryll: i suggest we update one variable at a time; to see if this works as expected

daryll: i wonder about the performance_year
will this stop working on 1/1/2023 ?

daryll (approved):

cbrunefearless (approved):
lgtm

PR #820, opened 2023-01-07
BCDA-6660: synthetic runout file script & test data

krobertson3314:
### Security Implications

### Acceptance Validation

See the script to generate test data that was used to test out the synthetic runout file script in lower environment (DEV) for different scenarios

### DIFF
--- a/db/migrations/manual/20230106-synthetic-runout-file-creation.sql
+++ b/db/migrations/manual/20230106-synthetic-runout-file-creation.sql
@@ -51,7 +51,7 @@ END LOOP;
 
 --verify for each entity that they only have 1 December CCLF file that was ingested for it.
 IF (SELECT count(dec_file_count) FROM temp_affected_acos WHERE dec_file_count > 1) > 0 THEN
-RAISE EXCEPTION 'Aborting: Found more than 1 cclf file record per aco_cms_id in December!';
+RAISE EXCEPTION 'Aborting: Found more than 1 cclf file record for at least one aco_cms_id in December!';
 END IF;
 
 --Step 2: create synthetic runout file records in cclf_files & obtain the new generated file id for each entity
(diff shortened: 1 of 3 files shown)


db/migrations/manual/20230106-synthetic-runout-file-creation.sql
```diff
@@ -0,0 +1,133 @@
...
+december_range_end_date cclf_files.created_at%TYPE := '2022-12-31';
+file_performance_year cclf_files.performance_year%TYPE := 22;
+synthetic_file_name_convention cclf_files.name%TYPE := CONCAT('.ACO.ZC8R','22','.D','221231','.SYNTHETICRUNOUTFILE');
+
+items RECORD;
+affected_aco_count INTEGER;
+total_synthetic_file_count INTEGER;
+total_mbi_count INTEGER;
+
+BEGIN
```
magile32, line 26: Is this needed when there is a BEGIN on line 11?

magile32 (approved):

db/migrations/manual/20230106-synthetic-runout-file-creation.sql
krobertson3314: I typed a lengthy answer for the PR and accidentally deleted the tab instead of clicking back to it :facepalm:

In a nutshell, from what I could find it seems both END and COMMIT for statements that begin with BEGIN do the same thing (but END being a postgres sql extension for COMMIT) in Postgres.
Some years back I worked with SQL Server for a bit and there I believe was a diff between a BEGIN ... END and a BEGIN .. COMMIT. The BEGIN / END was a begin and end to a code block, and BEGIN / COMMIT was used more for beginning and ending transactions.
Esp since doing a DECLARE first with variables used in the script, I then wanted to proceed with a BEGIN ... END to mark the boundaries of the actual execution.

PR #841, opened 2023-04-28
BCDA-4456: Remove duplicate calls to SSAS /introspect endpoint

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-4456

## 🛠 Changes

- Rely on ParseToken for token parsing and validation, if it is provided
- Remove validation logic from `RequireTokenAuth` (only check if token has been pre-validated and stored in the request context)
- Instead of deleting the `AuthorizeAccess` method...
  - Merge validation logic from `ParseToken` to `AuthorizeAccess` and clean up unreachable code

i.e. flow before and after for authorized endpoints:

```
# Before
--> ParseToken (validation logic) 
--> RequireTokenAuth (confirm token exists) 
    --> calls AuthorizeAccess (validation logic)

# After
--> ParseToken
    --> calls AuthorizeAccess (validation logic)
--> RequireTokenAuth (confirm token exists)
```

### Current Logic

---

Currently, token middleware is handled through two functions (ParseToken and RequireTokenAuth).

- Unauthenticated endpoints use ParseToken
- Authenticated endpoints use both ParseToken and RequireTokenAuth

→ **ParseToken** method checks if a token was provided in the headers:

- If a token is provided, parse + validate it and store it in the request context.
- If no token is provided, do not return an error (continue with the request.) 

→ **RequireTokenAuth** method checks that a token was stored in the request context.

- If a token is stored, validate it (again).
- If no token is stored, return a 401 Unauthorized error.

Since ParseToken is _guaranteed_ to be called before RequireTokenAuth, this means that every request to an endpoint requiring authorization will make duplicate calls to:

- the SSAS /introspect endpoint
- the `getAuthDataFromClaims` function, which makes database calls

### New Logic

---

This PR removes the validation from RequireTokenAuth. If the token is stored in the request context, _it has already been validated by ParseToken_ and does not need to be re-validated.

Instead of removing AuthorizeAccess entirely, I have merged the existing validation logic in `ParseToken` with the logic in `AuthorizeAccess`. They essentially do the same thing with minor differences.

## ✅ Acceptance Validation

Existing unit and integration tests (with a fake/live SSAS server) verify the expected authorization behavior of the BCDA API. I have only updated the mocks to reflect that RequireTokenAuth no longer makes any calls to the validation methods.

Note that the auth module logs do not reach Splunk, so it is not possible to fully verify whether the error conditions in RequireTokenAuth have ever been reached in production. However, since validation already happens in ParseToken, it is expected that the token validation in RequireTokenAuth _always succeeds_.

## 🔒 Security Implications

This change is not expected to impact the functionality of the authorization layer. Reviewers should confirm this.

- [ ] New data is stored or transmitted.
  
- [ ] This change introduces new software dependencies.
  
- [ ] This change impacts security controls or alters supporting software.
  
- [ ] A [security checklist](https://confluence.cms.gov/display/BCDA/Security+Checklists) is completed for this change.
  
- [ ] More team discussion required to evaluate security implications.

### DIFF
--- a/bcda/auth/middleware.go
+++ b/bcda/auth/middleware.go
@@ -86,8 +86,8 @@ func AuthorizeAccess(tokenString string) (*jwt.Token, AuthData, error) {
 	}
 
 	claims, ok := token.Claims.(*CommonClaims)
-	if !ok {
-		// This is already validated by VerifyToken so in theory it's unreachable code.
+	if !ok || !token.Valid {
+		// These should already trigger an error within VerifyToken, so in theory it's unreachable code.
 		return nil, ad, errors.New("invalid ssas claims")
 	}
 
--- a/bcda/auth/middleware.go
+++ b/bcda/auth/middleware.go
@@ -85,11 +85,9 @@ func AuthorizeAccess(tokenString string) (*jwt.Token, AuthData, error) {
 		return nil, ad, err
 	}
 
-	// Maybe split this back out to ensure that we don't start failing requests that used to succeed...
-	// except it's only in this specific scenario that it fails. otherwise it continues to getAuthDataFromClaims
-	// and returns an error anyways, so we're probably good to continue with this approach.
 	claims, ok := token.Claims.(*CommonClaims)
-	if !ok || !token.Valid {
+	if !ok {
+		// This is already validated by VerifyToken so in theory it's unreachable code.
 		return nil, ad, errors.New("invalid ssas claims")
 	}
 
--- a/bcda/auth/middleware.go
+++ b/bcda/auth/middleware.go
@@ -91,19 +91,10 @@ func AuthorizeAccess(tokenString string) (*jwt.Token, AuthData, error) {
 		return nil, ad, errors.New("invalid ssas claims")
 	}
 
-	switch claims.Issuer {
-	case "ssas":
-		ad, err = GetProvider().getAuthDataFromClaims(claims)
-		if err != nil {
-			tknEvent.help = fmt.Sprintf("failed getting AuthData; %s", err.Error())
-			operationFailed(tknEvent)
-			return nil, ad, err
-		}
-	default:
-		tknEvent.help = fmt.Sprintf("Unsupported claims issuer; %s", claims.Issuer)
+	ad, err = GetProvider().getAuthDataFromClaims(claims)
+	if err != nil {
+		tknEvent.help = fmt.Sprintf("failed getting AuthData; %s", err.Error())
 		operationFailed(tknEvent)
-		msg := fmt.Sprintf("Claim issuer '%s' is not supported", claims.Issuer)
-		err := &customErrors.UnsupportedClaimsIssuerError{Err: errors.New(msg), Msg: msg}
 		return nil, ad, err
 	}
 
@@ -120,8 +111,6 @@ func handleTokenVerificationError(w http.ResponseWriter, rw fhirResponseWriter, 
 			rw.Exception(w, http.StatusUnauthorized, responseutils.ExpiredErr, "")
 		case *customErrors.EntityNotFoundError:
 			rw.Exception(w, http.StatusForbidden, responseutils.UnauthorizedErr, responseutils.UnknownEntityErr)
-		case *customErrors.UnsupportedClaimsIssuerError:
-			rw.Exception(w, http.StatusNotFound, responseutils.TokenErr, "")
 		case *customErrors.RequestorDataError:
 			rw.Exception(w, http.StatusBadRequest, responseutils.InternalErr, "")
 		case *customErrors.RequestTimeoutError:
--- a/bcda/auth/middleware.go
+++ b/bcda/auth/middleware.go
@@ -58,48 +58,29 @@ func ParseToken(next http.Handler) http.Handler {
 
 		tokenString := authSubmatches[1]
 
-		token, err := GetProvider().VerifyToken(tokenString)
+		token, ad, err := GetProvider().AuthorizeAccess(tokenString)
 		if err != nil {
-			log.Auth.Errorf("Unable to verify token; %s", err)
 			handleTokenVerificationError(w, rw, err)
 			return
 		}
 
-		var ad AuthData
-		if claims, ok := token.Claims.(*CommonClaims); ok && token.Valid {
-			switch claims.Issuer {
-			case "ssas":
-				ad, err = GetProvider().getAuthDataFromClaims(claims)
-				if err != nil {
-					handleSsasAuthDataError(w, rw, err)
-					return
-				}
-			default:
-				log.Auth.Errorf("Unsupported claims issuer %s", claims.Issuer)
-				rw.Exception(w, http.StatusNotFound, responseutils.TokenErr, "")
-				return
-			}
-		}
 		ctx := context.WithValue(r.Context(), TokenContextKey, token)
 		ctx = context.WithValue(ctx, AuthDataContextKey, ad)
 		next.ServeHTTP(w, r.WithContext(ctx))
 	})
 }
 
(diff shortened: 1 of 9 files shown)


bcda/auth/middleware.go
```diff
@@ -58,48 +59,71 @@ func ParseToken(next http.Handler) http.Handler {
...
 }
 
-func handleSsasAuthDataError(w http.ResponseWriter, rw fhirResponseWriter, err error) {
-	log.Auth.Error(err)
-	if _, ok := err.(*customErrors.EntityNotFoundError); ok {
-		rw.Exception(w, http.StatusForbidden, responseutils.UnauthorizedErr, responseutils.UnknownEntityErr)
-	} else {
-		rw.Exception(w, http.StatusUnauthorized, responseutils.TokenErr, "")
+// AuthorizeAccess asserts that a base64 encoded token string is valid for accessing the BCDA API.
+func AuthorizeAccess(tokenString string) (*jwt.Token, AuthData, error) {
```
kyeah, line 75: AuthorizeAccess was basically doing the same validation as ParseToken with some additional logging, so I merged the two to create this helper function that is now used by ParseToken.

bcda/auth/middleware.go
```diff
@@ -58,48 +59,69 @@ func ParseToken(next http.Handler) http.Handler {
...
+	var ad AuthData
+
+	if err != nil {
+		tknEvent.help = fmt.Sprintf("VerifyToken failed in AuthorizeAccess; %s", err.Error())
+		operationFailed(tknEvent)
+		return nil, ad, err
+	}
+
+	claims, ok := token.Claims.(*CommonClaims)
+	if !ok || !token.Valid {
```
kyeah, line 89: The `token.Valid` check is new, pulled from `parseToken`. Note however that VerifyToken always sets token.Valid to true, so this check should always pass.

bcda/auth/middleware.go
```diff
@@ -58,48 +59,69 @@ func ParseToken(next http.Handler) http.Handler {
...
+
+	switch claims.Issuer {
+	case "ssas":
+		ad, err = GetProvider().getAuthDataFromClaims(claims)
+		if err != nil {
+			tknEvent.help = fmt.Sprintf("failed getting AuthData; %s", err.Error())
+			operationFailed(tknEvent)
+			return nil, ad, err
+		}
+	default:
```
kyeah, line 102: The claims.Issuer check is also brought in from `parseToken`.

bcda/auth/middleware.go
kyeah: Note also that this introduces one small functional change to unauthenticated endpoints. Previously if this check failed, `ParseToken` would not fail the unauthenticated request. Now the request will fail.

I checked in the code to confirm that this is expected to be unreachable code, because...

- [VerifyToken already checks for token.Claims](https://github.com/CMSgov/bcda-app/blob/dd356ecda6aa85fb9befddc02b09a9e8f7e7907c/bcda/auth/ssas.go#L238-L242)
- [VerifyToken already sets token.Valid to true](https://github.com/CMSgov/bcda-app/blob/dd356ecda6aa85fb9befddc02b09a9e8f7e7907c/bcda/auth/ssas.go#L208)

bcda/auth/middleware.go
```diff
@@ -58,48 +59,69 @@ func ParseToken(next http.Handler) http.Handler {
...
-					handleSsasAuthDataError(w, rw, err)
-					return
-				}
-			default:
-				log.Auth.Errorf("Unsupported claims issuer %s", claims.Issuer)
-				rw.Exception(w, http.StatusNotFound, responseutils.TokenErr, "")
-				return
-			}
-		}
 		ctx := context.WithValue(r.Context(), TokenContextKey, token)
```
kyeah, line 68: I confirmed that `ParseToken` is the only place in the codebase where the token is set in the request context. This ensures that if the request context has a token set, it is a valid, active token.

bcda/auth/middleware_test.go
```diff
@@ -114,36 +113,6 @@ func (s *MiddlewareTestSuite) TestReturn401WhenExpiredToken() {
 	assert.Nil(s.T(), err)
 }
 
-// unit test
-func (s *MiddlewareTestSuite) TestRequireTokenAuthReturn401WhenInvalidToken() {
```
kyeah, line 118: This test is no longer needed because RequireTokenAuth does not do token validation anymore. The user scenario here (invalid token) is still covered by the following method:

- TestReturn400WhenInvalidTokenAuthWithInvalidSignature

Technically this 401 error was unreachable because the user should run into the 400 error from ParseToken before RequireTokenAuth is ever called.

bcda/auth/middleware.go
```diff
@@ -58,48 +59,58 @@ func ParseToken(next http.Handler) http.Handler {
...
+		token, ad, err := AuthorizeAccess(tokenString)
 		if err != nil {
-			log.Auth.Errorf("Unable to verify token; %s", err)
 			handleTokenVerificationError(w, rw, err)
 			return
 		}
 
-		var ad AuthData
-		if claims, ok := token.Claims.(*CommonClaims); ok && token.Valid {
-			switch claims.Issuer {
```
kyeah, line 70: This switch case is unreachable because VerifyToken already does the issuer check here: 

https://github.com/CMSgov/bcda-app/blob/dd356ecda6aa85fb9befddc02b09a9e8f7e7907c/bcda/auth/ssas.go#L246

laurenkrugen-navapbc: LGTM! leaving a note:

VerifyToken() -> confirmTokenStringLegitimacy() -> confirmRequestorTokenPayload is path to issue check in linked comment

bcda/auth/middleware.go
laurenkrugen-navapbc: What would happen previously from the consumer side? I see in the old code that it has a `return error` statement.

bcda/auth/provider.go
```diff
@@ -84,10 +84,6 @@ type Provider interface {
 	// RevokeAccessToken a specific access token identified in a base64 encoded token string
 	RevokeAccessToken(tokenString string) error
 
-	// TODO refactor input to be AuthData
-	// AuthorizeAccess asserts that a base64 encoded token string is valid for accessing the BCDA API
-	AuthorizeAccess(tokenString string) error
```
laurenkrugen-navapbc, line 89: this is removed, per the merge/refactor here, correct?

https://github.com/CMSgov/bcda-app/pull/841/files#r1180735589

bcda/auth/ssas_test.go
```diff
@@ -274,7 +274,7 @@ func (s *SSASPluginTestSuite) TestAuthorizeAccessErrIsNilWhenHappyPath() {
 	c, err := client.NewSSASClient()
 	require.NotNil(s.T(), c, sSasClientErrorMsg, err)
 	s.p = SSASPlugin{client: c, repository: s.r}
-	err = s.p.AuthorizeAccess(tokenString)
+	_, _, err = AuthorizeAccess(tokenString)
```
karinamzalez, line 277: Just looked into blank identifiers -- makes sense! Just curious re: why it's preferred instead of the previous version here.

karinamzalez (approved):

bcda/auth/provider.go
kyeah: Yes that's correct. This is removed from the SSAS/auth provider and is now just a helper function that lives directly in the middleware.

bcda/auth/ssas_test.go
kyeah: Previously this methid only returned an error (or nil). I updated the method to return three objects, but for the test we are only interested in checking if there's an error returned 🙂

laurenkrugen-navapbc (approved):

PR #843, opened 2023-05-03
BCDA-6997: PR Templates - Security Section

gfreeman-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-6997

## 🛠 Changes

Adds standard security section to pull request template.


.github/pull_request_template.md
```diff
@@ -14,20 +14,13 @@ https://jira.cms.gov/browse/BCDA-xxx
 
 (How were the changes verified? Did you fully test the acceptance criteria in the ticket? Provide reproducible testing instructions and screenshots if applicable.)
 
-## 🔒 Security Implications
+### What Security Implications Does This PR Have?
```
kyeah, line 17: Thanks @gfreeman-navapbc! Small request — could we keep the same section formatting structure as the rest of the template for consistency? I don't mind having the actual checklist copy-pasted from Jason but if we could keep this at an h2 (`##` instead of `###`) and keep the emoji that would be appreciated 🙂

gfreeman-navapbc: oops, sorry, tried to do that where I could, will go ahead and fix this on the rest of the PRs as well

kyeah (approved):

PR #845, opened 2023-05-09
Add new relic spans to db transactions in repository.go

karinamzalez:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-6916

## 🛠 Changes

- adds nrpgx library for DB tracing 
- update DM queries in requests.go and middleware.go so that they use request context instead of background context. This is needed for new relic db tracing.

## ℹ️ Context for reviewers

(Background context, more in-depth details of the implementation, and anything else you'd like to call out or ask reviewers.)

- nrpgx allows us to trace db transactions without having to explicitly start and stop new relic spans

## ✅ Acceptance Validation

- validate that local DB transactions are traced in new relic when calls are made via postman

## 🔒 Security Implications

- [ ] New data is stored or transmitted.
  

- [ ] This change introduces new software dependencies.
  

- [ ] This change impacts security controls or alters supporting software.
  

- [ ] A [security checklist](https://confluence.cms.gov/display/BCDA/Security+Checklists) is completed for this change.
  

- [ ] More team discussion required to evaluate security implications.


bcda/monitoring/monitoring.go
```diff
@@ -26,6 +27,27 @@ func (a apm) Start(msg string, w http.ResponseWriter, r *http.Request) *newrelic
 	return nil
 }
 
+func (a apm) NewTransaction(name string, ctx context.Context) (*newrelic.Transaction, context.Context) {
+	if a.App != nil {
+		txn := a.App.StartTransaction(name)          // transaction trace
+		newRelicCtx := newrelic.NewContext(ctx, txn) // parent context
+		return txn, newRelicCtx
```
laurenkrugen-navapbc, line 34: do we need to return the `txn` variable? it looks like in the [example for the integration](https://github.com/newrelic/go-agent/blob/b704e94bdd0cdd982464fa7ae55e2cb5e8e51f0a/v3/integrations/nrpgx/example/sql_compat/main.go#L34), we can just return the context and use that in the query to the database. 
```
txn := app.StartTransaction("postgresQuery")
ctx := newrelic.NewContext(context.Background(), txn)
row := db.QueryRowContext(ctx, "SELECT count(*) FROM pg_catalog.pg_tables")
```

laurenkrugen-navapbc: [these docs](https://pkg.go.dev/github.com/newrelic/go-agent/v3/integrations/nrpgx) also tell us that the above would require us to add the integration in during the database setup, as well: 
```db, err := sql.Open("nrpgx", "host=localhost port=5432 user=postgres dbname=postgres password=docker sslmode=disable")```

File: bcda/monitoring/monitoring.go [hunk 2]
```diff
@@ -26,6 +27,27 @@ func (a apm) Start(msg string, w http.ResponseWriter, r *http.Request) *newrelic
...
+	}
+	return nil, nil
+}
+
+func NewSpan(parentCtx context.Context, name string) (close func()) {
+	txn := newrelic.FromContext(parentCtx)
+	if txn == nil {
+		log.API.Warn("No transaction found. Cannot create child.")
+	}
+	segment := txn.StartSegment(name)
```
laurenkrugen-navapbc, line 44: If use an integration package (like this one: https://pkg.go.dev/github.com/newrelic/go-agent/v3/integrations/nrpgx), then we don't have to create a [segment manually each time](https://github.com/newrelic/go-agent/blob/master/GUIDE.md#datastore-segments)e, which is a cool functionality!

bcda/monitoring/monitoring.go
laurenkrugen-navapbc: I would also be curious if a new relic transaction is meant to hold a specific segment (like a database call) or if we should be creating a transaction earlier when a request comes in.

laurenkrugen-navapbc: that makes sense to me! there is already a helper function that starts a transaction for [web requests here](https://github.com/CMSgov/bcda-app/blob/2ab83d60a3bb007661a7fd2fad23c547096a9e3c/bcda/monitoring/monitoring.go#L19).

laurenkrugen-navapbc:
Looks good! We'll just need to figure out why the Build and Test workflow is failing.

laurenkrugen-navapbc (dismissed):

laurenkrugen-navapbc (approved):

PR #861, opened 2023-07-17
BCDA-todo: Split opt out utilities into a separate module

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-todo

## 🛠 Changes

Create a new Golang module for optout code that can be imported separate from BCDA code. This includes:
  - All related models (SuppressionFile, Suppression, SuppressionFileMetadata)
  - Utility function for parsing metadata from filenames
  - Utility function for parsing a single line from an opt out file

## ℹ️ Context for reviewers

AB2D and DPC are currently in the process of automating data-sharing opt-out requests. It is expected that the file formats will be consistent with the existing files that BCDA receives, so a lot of the existing Go code is reusable.

DPC is likely going to follow the pre-established pattern of building lambdas in Go, so we expect that BCDA's code could be directly used as part of the lambda. This PR pulls out generic utilities that are not related to BCDA's implementation in any way (e.g. reading from the local filesystem or saving to the BCDA database.)

## ✅ Acceptance Validation

- Unit tests continue to pass
- Was able to create a new local go program and import these utilities using `go get github.com/CMSgov/bcda-app/output@7603537d5966b1d4734e1956b5d47b9c54a3f39b`.

## 🔒 Security Implications

### DIFF
--- a/bcda/models/models.go
+++ b/bcda/models/models.go
@@ -111,27 +111,6 @@ type CCLFBeneficiary struct {
 	BlueButtonID string
 }
 
-type SuppressionFile struct {
-	ID           uint
-	Name         string
-	Timestamp    time.Time
-	ImportStatus string
-}
-
-type Suppression struct {
-	ID                  uint
-	FileID              uint
-	MBI                 string
-	SourceCode          string
-	EffectiveDt         time.Time
-	PrefIndicator       string
-	SAMHSASourceCode    string
-	SAMHSAEffectiveDt   time.Time
-	SAMHSAPrefIndicator string
-	ACOCMSID            string
-	BeneficiaryLinkKey  int
-}
-
 type JobEnqueueArgs struct {
 	ID              int
 	ACOID           string
--- a/bcda/suppression/suppression.go
+++ b/bcda/suppression/suppression.go
@@ -7,13 +7,12 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strconv"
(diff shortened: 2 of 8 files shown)


go.work
laurenkrugen-navapbc, line 1: this is neat. do we have to make changes in both modules or does one pack up the changes and apply it to the other?

scripts/unit_test.sh
```diff
@@ -13,7 +13,7 @@ mkdir -p test_results/latest
 echo "Running unit tests and placing results/coverage in test_results/${timestamp} on host..."
 
 # Avoid the db/migrations package since it only contains test code
-PACKAGES_TO_COVER=$(go list ./... | egrep -v 'test|mock|db/migrations' | tr "\n" "," | sed -e 's/,$//g')
+PACKAGES_TO_COVER=$(go list ./... ./optout | egrep -v 'test|mock|db/migrations' | tr "\n" "," | sed -e 's/,$//g')
```
laurenkrugen-navapbc, line 16: do we need to also update the `lint` command in the makefile to include this package?

laurenkrugen-navapbc (approved):

go.work
kyeah: I think the workspace file basically just tells each module about the other, so this makes the optout module visible to the main bcda module.

When bcda is packaged, it'll figure out that the optout package is housed together so it'll be able to package them together for the API 😄

karinamzalez (approved):

PR #864, opened 2023-09-06
BCDA-7135: bcdaworker logging improvements

laurenkrugen-navapbc:
## 🎫 Ticket

## 🛠 Changes

- The bcda worker.go and que.go will utilize context to create and pass around log fields. 
- fields `cmsID` and `jobID` have been renamed to `cms_id` and `job_id` to for consistency and querying
- log strings have been modified so that field values are in their own individual fields instead of the free text `msg` field

## ℹ️ Context for reviewers

The BCDA worker does not use the initial request context when processing jobs. It does use the Job ID, which is a common identifier that we can use to query events on. When we use the `transaction_id` in splunk we can view related events for the job and use the job ID to query events further.

The que.go file has been modified so that when we create our context for the job processing, it uses it to pass around log fields to other function calls. This has eliminated the need to declare the fields repeatedly when we use `WithFields({...})` and removed redundant or unnecessary log calls.

## ✅ Acceptance Validation

all tests passing.

## 🔒 Security Implications


bcdaworker/log/log.go
```diff
@@ -0,0 +1,24 @@
+package workerlog
```
laurenkrugen-navapbc, line 1: this doesn't need to be in its own package; I created these types so that they could be added to the context within bcdaworker. We could probably add this to one of the other log packages.

bcdaworker/queueing/manager/que.go
```diff
@@ -113,37 +114,44 @@ func (q *queue) processJob(job *que.Job) error {
 		return nil
 	}
 
+	ctxFields := logrus.Fields{"job_id": jobArgs.ID, "aco_id": jobArgs.ACOID}
```
laurenkrugen-navapbc, line 117: When we call logrus's `WithFields` function, it returns an entry and takes a map of fields of the type `logrus.Fields`. This is where the log entry starts to "snowball" fields and pick up relevant data as the worker processes the job.

laurenkrugen-navapbc: we later use the `cms_id`, so `aco_id` might not be something that we necessarily need or want here.

bcdaworker/queueing/manager/que.go
```diff
@@ -113,37 +114,44 @@ func (q *queue) processJob(job *que.Job) error {
...
 		if job.ErrorCount >= maxNotFoundRetries {
-			q.log.Errorf("No job found for ID: %d acoID: %s. Retries exhausted. Removing job from queue.", jobArgs.ID,
-				jobArgs.ACOID)
+			q.log.WithFields(ctxFields).Errorf("No job found. Retries exhausted. Removing job from queue.")
 			// By returning a nil error response, we're singaling to que-go to remove this job from the jobqueue.
 			return nil
 		}
 
-		q.log.Warnf("No job found for ID: %d acoID: %s. Will retry.", jobArgs.ID, jobArgs.ACOID)
+		q.log.WithFields(ctxFields).Warnf("No job found, retrying")
```
laurenkrugen-navapbc, line 140: like the log statements above and many of the following, I opted to remove values from the message that are now in their own field in the log message.

bcdaworker/queueing/manager/que_test.go
```diff
@@ -133,10 +133,10 @@ func TestProcessJobFailedValidation(t *testing.T) {
 		expectedErr error
 		expLogMsg   string
 	}{
-		{"ParentJobCancelled", worker.ErrParentJobCancelled, nil, `^queJob \d+ associated with a cancelled parent Job`},
-		{"NoBasePath", worker.ErrNoBasePathSet, nil, `^Job \d+ does not contain valid base path`},
-		{"NoParentJob", worker.ErrParentJobNotFound, repository.ErrJobNotFound, `^No job found for ID: \d+ acoID.*Will retry`},
-		{"NoParentJobRetriesExceeded", worker.ErrParentJobNotFound, nil, `No job found for ID: \d+ acoID.*Retries exhausted`},
+		{"ParentJobCancelled", worker.ErrParentJobCancelled, nil, `^Removing queuejob from que; parent job \d+ cancelled.`},
```
laurenkrugen-navapbc, line 136: Simplified log messages; no longer using the interpolation

bcdaworker/worker/worker.go
```diff
@@ -62,6 +62,8 @@ func (w *worker) ValidateJob(ctx context.Context, jobArgs models.JobEnqueueArgs)
 
 func (w *worker) ProcessJob(ctx context.Context, job models.Job, jobArgs models.JobEnqueueArgs) error {
 
+	logFields := workerlog.GetLogFields(ctx)
```
laurenkrugen-navapbc, line 65: We have two log fields from `que.processJob()`'s context. Here we are using are function defined in log.go to retrieve those values so we can continue to add new fields as we encounter them.

bcdaworker/worker/worker.go
```diff
@@ -70,28 +72,28 @@ func (w *worker) ProcessJob(ctx context.Context, job models.Job, jobArgs models.
...
+		log.Worker.WithFields(logFields).Error(err)
+		return err
 	}
 
-	log.Worker.WithFields(logrus.Fields{
-		"jobID": jobArgs.ID,
-		"cmsID": *aco.CMSID,
-	})
+	logFields["cms_id"] = aco.CMSID
+	ctx = workerlog.WithLogFields(ctx, logFields)
```
laurenkrugen-navapbc, line 81: adding new fields to the context.

bcdaworker/worker/worker.go
```diff
@@ -381,21 +383,22 @@ func fhirBundleToResourceNDJSON(ctx context.Context, w *bufio.Writer, b *fhirmod
...
-			log.Worker.Error(err)
+			log.Worker.WithFields(logFields).Error(err)
 			appendErrorToFile(ctx, fileUUID, fhircodes.IssueTypeCode_EXCEPTION,
 				responseutils.InternalErr, fmt.Sprintf("Error writing %s to file for beneficiary %s in ACO %s", jsonType, beneficiaryID, acoID), jobID)
 		}
 	}
 }
 
 func checkJobCompleteAndCleanup(ctx context.Context, r repository.Repository, jobID uint) (jobCompleted bool, err error) {
+	logFields := workerlog.GetLogFields(ctx)
```
laurenkrugen-navapbc, line 401: calling `GetLogFields` here so we can log the warn statement with the appropriate data.

bcdaworker/worker/worker_test.go
```diff
@@ -164,7 +165,10 @@ func (s *WorkerTestSuite) TestWriteResourceToFile() {
 		s.T().Run(tt.resource, func(t *testing.T) {
 			jobArgs := models.JobEnqueueArgs{ID: s.jobID, ResourceType: tt.resource, BeneficiaryIDs: cclfBeneficiaryIDs,
 				Since: since, TransactionTime: transactionTime, ClaimsWindow: claimsWindow}
-			uuid, size, err := writeBBDataToFile(context.Background(), s.r, &bbc, *s.testACO.CMSID, jobArgs)
+			ctxFields := logrus.Fields{"job_id": s.jobID, "aco_id": s.testACO.ID}
```
laurenkrugen-navapbc, line 168: ProcessJob test fails when the logrus.Fields map is nil/empty.

bcdaworker/worker/worker_test.go
```diff
@@ -282,11 +292,6 @@ func (s *WorkerTestSuite) TestWriteEOBDataToFileWithErrorsBelowFailureThreshold(
 	fData = fData[:len(fData)-1]
 	assertEqualErrorFiles(s.T(), ooResp, string(fData))
 
-	// Assert cmsID and jobID fields are being added to the logs
-	assert.Equal(s.T(), 2, len(logHook.Entries))
```
laurenkrugen-navapbc, line 286: these were removed because we no longer add `cmsID` and `jobID` in the `writeBBDataToFile` function

bcdaworker/queueing/manager/que.go
karinamzalez: Ah! I see, so this is also answers my other question regarding the interpolation that was removed.

bcdaworker/worker/worker.go
karinamzalez: So will each worker have its own personalized set of log fields depending on the contents of ctx? Would it be possible to document somewhere all of the different types of log fields that could be returned?

bcdaworker/queueing/manager/que_test.go
laurenkrugen-navapbc: I tried to move any fields that are an identifier out of the `msg` string field and into their own fields; it seemed redundant to list it again in the message! 😄

bcdaworker/worker/worker.go
laurenkrugen-navapbc: the log fields won't be specific to the job but their values will be. The `logrus.Fields{}` will contain a new map of log values for each context that is created. Every time the worker processes a job, it creates a new context and when it is complete, the context is no longer used and the cycle starts all over again. 

The log fields are not documented anywhere but the code base and in a future PR i'd like to at least make the log field **names** be constants so we can avoid typos and that sort of thing.

bcdaworker/worker/worker.go
kyeah: Ah interesting. Is it a performance optimization to avoid having an `AddLogFields(ctx, ...)` function that automatically creates a new context with the new fields, instead of having to:

1. call GetLogFields
2. update logFields
3. call WithLogFields

? I wonder if it's possible to have a nice-looking `AddLogFields` function that can accept a variable number of key-value pairs.

```
ctx = AddLogFields(ctx, {
  "file_id": fileId,
  "file_size": fileSize
})
```

bcdaworker/worker/worker.go
```diff
@@ -70,28 +72,28 @@ func (w *worker) ProcessJob(ctx context.Context, job models.Job, jobArgs models.
...
-		"cmsID": *aco.CMSID,
-	})
+	logFields["cms_id"] = aco.CMSID
+	ctx = workerlog.WithLogFields(ctx, logFields)
 
 	err = w.r.UpdateJobStatusCheckStatus(ctx, job.ID, models.JobStatusPending, models.JobStatusInProgress)
 	if goerrors.Is(err, repository.ErrJobNotUpdated) {
 		// could also occur if job was marked as cancelled
-		log.Worker.Warnf("Failed to update job. Assume job already updated. Continuing. %s", err.Error())
+		log.Worker.WithFields(logFields).Warnf("Failed to update job. Assume job already updated. Continuing. %s", err.Error())
```
kyeah, line 86: I think this is where it starts to be a little unfortunate to not have the logger housed directly in the context and updated with the fields as needed 😅 It's very easy to forget to insert the logFields when logging a message.

kyeah (commented):
Thanks! I just did a first pass/skim at some of the main bits. Added some notes 👋

bcdaworker/queueing/manager/que.go
laurenkrugen-navapbc: I would prefer to remove the `aco_id` to be honest. I can have a look through our reporting to see if that ID is used anywhere and if it's not, am I OK to remove?

bcdaworker/queueing/manager/que_test.go
laurenkrugen-navapbc: that's a good point -- let me do some looking through splunk to make sure no reports break with these changes.

bcdaworker/worker/worker.go
laurenkrugen-navapbc: We could set it up similar to here, where we store an Entry. As new fields are added, we would need to add those to the Entry in the context, using the `WithField(s)` function. We could adjust this logic so that when a new field is added to the log Entry, instead of including the WithFields(logFields) line each time, we could update the Entry in the context and return it immediately for use again. 

If we add, say, two new fields with a function call, we would still need to immediately update the context but that would eliminate the need to include the new fields in each log call.

 What do you think?

laurenkrugen-navapbc: A potential downside is that the `Entry` being returned will no longer be referred to as `log.API...`, but whatever new name we decide to give to it. It could still use the `log.API` logger as the [Entrys Logger](https://pkg.go.dev/github.com/sirupsen/logrus#Entry).

bcda/cclf/metrics/metrics.go
```diff
@@ -146,7 +147,7 @@ type noopTimer struct {
 }
 
 func (t *noopTimer) new(parentCtx context.Context, name string) (ctx context.Context, close func()) {
-	return context.Background(), noop
```
laurenkrugen-navapbc, line 149: I changed this to use the parentCtx because it creates a new nil context when there's not a newrelic timer; this overwrites the logger passed around in the context. this only occurs when the license key is not set and should affect local development only.

bcda/logging/middleware.go
laurenkrugen-navapbc, line 1: Moved `StructuredLogger` and the two functions that it implements (`Panic` and `Write`) in the top level `log/logger.go` folder because both the bcda (api) and bcdaworker (worker) utilize those. Made more sense to have common or shared components in the root level

bcda/logging/middleware_test.go
laurenkrugen-navapbc, line 1: package renaming here. Looks like some of my changes got [pulled in from the last PR](https://github.com/CMSgov/bcda-app/pull/865/files#diff-aee68ab10d854948370950275cc3159dbf7d39d09dc64c1da1ead0874055ae23) during rebase.

PR #865, opened 2023-09-08
BCDA-7135: bcda api logging improvements

laurenkrugen-navapbc:
## 🎫 Ticket

## 🛠 Changes

- added new middleware in logging.go that adds request ID and the cms ID to the context for next handler
- updated requests.go logging statements to utilize logger that is passed via context

## ℹ️ Context for reviewers

Updating the request processing to add a logger to the context via middleware. Request ID and CMS ID can be used in future logging calls and new fields can be added to the requests context logger as needed.

Looked into database logging to possibility be included in this PR, but database is using the default `logrus.logger`. Creating a separate ticket to address that. 

## ✅ Acceptance Validation

Current tests pass, new test added, and ran locally to check logs & context values.

## 🔒 Security Implications

### DIFF
--- a/bcda/api/requests.go
+++ b/bcda/api/requests.go
@@ -530,8 +530,7 @@ func (h *Handler) bulkRequest(w http.ResponseWriter, r *http.Request, reqType se
 	}
 
 	if newJob.ID != 0 {
-		ctx = logging.SetCtxLogger(ctx, "job_id", newJob.ID)
-		logger = logging.GetCtxLogger(ctx)
+		ctx, logger = logging.SetCtxLogger(ctx, "job_id", newJob.ID)
 	}
 
 	// request a fake patient in order to acquire the bundle's lastUpdated metadata
--- a/bcda/api/requests.go
+++ b/bcda/api/requests.go
@@ -11,7 +11,7 @@ import (
 	"strings"
 
 	"github.com/go-chi/chi/v5"
-	gcmw "github.com/go-chi/chi/v5/middleware"
+
 	"github.com/pkg/errors"
 
 	"net/http"
@@ -23,6 +23,7 @@ import (
 	"github.com/CMSgov/bcda-app/bcda/client"
 	"github.com/CMSgov/bcda-app/bcda/constants"
 	"github.com/CMSgov/bcda-app/bcda/database"
+	logging "github.com/CMSgov/bcda-app/bcda/logging"
 	"github.com/CMSgov/bcda-app/bcda/models"
 	"github.com/CMSgov/bcda-app/bcda/models/postgres"
 	responseutils "github.com/CMSgov/bcda-app/bcda/responseutils"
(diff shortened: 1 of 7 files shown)


bcda/api/requests.go
```diff
@@ -517,20 +524,20 @@ func (h *Handler) bulkRequest(w http.ResponseWriter, r *http.Request, reqType se
...
+		logger.Error(err)
 		h.RespWriter.Exception(w, http.StatusInternalServerError, responseutils.DbErr, "")
 		return
 	}
 
 	if newJob.ID != 0 {
-		requestID := gcmw.GetReqID(r.Context())
-		log.API.Infof("requestID %s jobID %d", requestID, newJob.ID)
+		logging.LogEntrySetField(ctx, "job_id", newJob.ID)
+		logger = logging.GetLogEntry(r.Context())
```
kyeah, line 534: I wonder if `LogEntrySetField` should return the logger instead of, or in addition to, the new context? seems like it would simplify lines like this.

```go
logger = logging.LogEntrySetField(...)
```

or 

```go
logger, _ = logging.LogEntrySetField(...)
```

I think the other confusing thing is that `r.Context()` here would be the _old_ context, right? Since `LogEntrySetField` returns the new context 🤔 So in theory if the request context didn't already have the log key, I think there wouldn't be logger here when you look for it from `r.Context()`.

bcda/logging/middleware.go
```diff
@@ -136,3 +137,46 @@ func Redact(uri string) string {
 	}
 	return uri
 }
+
+type CommonLogCtxKeyType string
+
+const CommonLogCtxKey CommonLogCtxKeyType = "commonLog"
+
+func NewCommonLogFields(next http.Handler) http.Handler {
```
kyeah, line 145: nit — could we add some doc comments for these functions? I know the codebase in general doesn't really do this consistently, but I think it might help with maintainability since these are functions that will likely be used generally across the code.

bcda/api/requests.go
laurenkrugen-navapbc: I left the `GetLogEntry` and `LogEntrySetField` (whose name I will be updating to `SetLogEntry`) separate since these functions will be used in the worker, too. There may be cases where we just want to retrieve the logger without updating the fields at the same time. 

Maybe we leave the `GetLogEntry` and update `LogEntrySetField`  to additionally return the logger when it gets updated?

That is also my mistake on using the old context there -- I'll fix that line right now!

laurenkrugen-navapbc: OK, I updated the names to be more clear and I also realized that we are utilizing one of Chis middleware functions that would have had some conflict (GetLogEntry). 

I didn't update the return type yet -- let me know what your thoughts are on my last comment! :)

bcda/logging/middleware.go
laurenkrugen-navapbc: comments added!

bcda/api/requests.go
kyeah: Yeah I think we can leave (what used to be) GetLogEntry alone, and update the SetEntry method to return the logger. Thanks!

laurenkrugen-navapbc: changes have been pushed!

PR #868, opened 2023-09-26
BCDA-7124 - Improve coverage for "bcda/auth/client/ssas.go"

alex-dzeda:
Update unit tests to go from approximately ~65% test coverage of auth/client/ssas.go to 93.3% coverage.

## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7124

## 🛠 Changes

Updated unit tests to cover more conditions (generally, handling inter-process communication errors, as well as incorrect config file scenarios). There will remain a subset of conditions that should never rear their head, such as json marshaling failing (it looks like this is quite difficult to do when working solely with strings). 

## ℹ️ Context for reviewers

Should be relatively straightforward; most of the uncovered conditions in the existing file were repeatable across different functions. 

## ✅ Acceptance Validation

Passes unit tests, and has an output of 93.3% coverage.
image

## 🔒 Security Implications


bcda/auth/client/ssas_test.go
```diff
@@ -105,200 +158,163 @@ func (s *SSASClientTestSuite) TestNewSSASClient_TLSFalseNoURL() {
 	assert.EqualError(s.T(), err, "SSAS client could not be created: no URL provided")
```
laurenkrugen-navapbc, line 158: separate from the work in this PR, but it might be useful to update this error message to something like:

`SSAS client could not be created: no certificate provided`

bcda/auth/client/ssas_test.go
```diff
@@ -105,200 +158,163 @@ func (s *SSASClientTestSuite) TestNewSSASClient_TLSFalseNoURL() {
...
 		if err != nil {
-			log.Fatal(err)
+			s.FailNow(constants.CreateSsasErr, err.Error())
 		}
-	})
-	server := httptest.NewServer(router)
 
-	conf.SetEnv(s.T(), "SSAS_URL", server.URL)
-	conf.SetEnv(s.T(), "SSAS_USE_TLS", "false")
+		fmt.Println(tc)
```
laurenkrugen-navapbc, line 197: Is this set in case of a test failure/visibility?

bcda/auth/client/ssas_test.go
```diff
@@ -465,3 +487,265 @@ func (s *SSASClientTestSuite) TestGetToken() {
...
+		{fnInput: []string{}, header: http.StatusOK, env: EnvVars{BCDA_SSAS_CLIENT_ID: "-1"}, errorExpected: true, message: "missing clientID or secret"},
+	}
+
+	for _, tc := range tests {
+		router := chi.NewRouter()
+		keyStr := "123456"
+		router.Get("/system/{systemID}/key", func(w http.ResponseWriter, r *http.Request) {
+			_, err := w.Write([]byte(`{ "client_id": "123456", "public_key": "` + keyStr + `" }`))
+			if err != nil {
+				log.Fatal(err)
```
laurenkrugen-navapbc, line 544: should any errors that occur here result in a test failure instead of logging and exiting?

laurenkrugen-navapbc:
really thorough and great coverage! 🎉

bcda/auth/client/ssas_test.go
alex-dzeda: Good catch, going through and deleting the print statements

bcda/auth/client/ssas_test.go
alex-dzeda: On this one, I think I have a conceptual question: In the context of a test, would we behave any differently for one vs. the other?

In this case, we're mocking the router to return a fake public key. An error here would imply that the router failed to write a string. If the response is written incorrectly/errors out (and we don't log + exit), the test that relies upon that response would have an unexpected outcome compared to what we're asserting, and the test would then fail. 
Thinking further, my personal thoughts would be that if we're setting up the testing conditions incorrectly, we shouldn't necessarily trust the validity of the test result, so indeed it may make sense to exit where the setup failed? I can see it both ways, not particularly attached to one approach over the other, and interested to get your perspective.

laurenkrugen-navapbc: That's a really good point! I think maybe we should leave that in, then. The `testing` package offers a logger that will use the same behavior but continue test execution: https://pkg.go.dev/testing#T.Fatal

Maybe that would be an option to use `t.Fail(err)`? What are your thoughts?

alex-dzeda: Sounds good, thank you! I've updated that + removed the fmt.println statements in a new commit that should go through the c/i flow shortly.

laurenkrugen-navapbc: perfect, thanks for the update!

laurenkrugen-navapbc (approved):

PR #869, opened 2023-09-26
BCDA-7135: bcdaworker logging improvements

laurenkrugen-navapbc:
…worker uses ctx to pass logger

## 🎫 Ticket

## 🛠 Changes

- The bcda worker.go and que.go will utilize context to create and pass around log fields.
- ields cmsID and jobID have been renamed to cms_id and job_id to for consistency and querying
- log strings have been modified so that field values are in their own individual fields instead of the free text msg field
- common logging components have been moved to log/logger.go and import names updated
- add a single `logger.info()` in `requests.go/bulkrequest()` so that job ID and request ID can be correlated

## ℹ️ Context for reviewers

The BCDA worker does not use the initial request context when processing jobs. It does use the Job ID, which is a common identifier that we can use to query events on. When we use the request_id in splunk we can view related events for the job and use the job ID to query events further. This requires the addition of one informational log event when the job ID is created.

The que.go file has been modified so that when we create our context for the job processing, it uses it to pass around log fields to other function calls. This has eliminated the need to declare the fields repeatedly when we use WithFields({...}) and removed redundant or unnecessary log calls.

## ✅ Acceptance Validation

Tests updated and passing, tested app running locally.

## 🔒 Security Implications

### DIFF
--- a/bcda/api/requests.go
+++ b/bcda/api/requests.go
@@ -23,7 +23,6 @@ import (
 	"github.com/CMSgov/bcda-app/bcda/client"
 	"github.com/CMSgov/bcda-app/bcda/constants"
 	"github.com/CMSgov/bcda-app/bcda/database"
-	logging "github.com/CMSgov/bcda-app/bcda/logging"
 	"github.com/CMSgov/bcda-app/bcda/models"
 	"github.com/CMSgov/bcda-app/bcda/models/postgres"
 	responseutils "github.com/CMSgov/bcda-app/bcda/responseutils"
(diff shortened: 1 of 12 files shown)


bcda/api/requests.go
laurenkrugen-navapbc, line 1: package name updates

bcda/api/requests_test.go
laurenkrugen-navapbc, line 1: package name updates

bcda/api/v1/api_test.go
laurenkrugen-navapbc, line 1: package name updates

alex-dzeda (approved):
Looks good, I think the additional granularity provided by being able to triage on the basis of individual job IDs / request IDs without having to generate new log entries will be useful.

PR #872, opened 2023-10-05
BCDA: Improve coverage for "bcda/api/requests.go"

alex-dzeda:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7123

## 🛠 Changes

Updated unit tests for BCDA-7123: bcda/api/requests.go
Scrubbed unnecessary information that was sent to a client (go overflow error)

## ℹ️ Context for reviewers

Most of these are straightforward, adding on use cases to the existing tests and covering functions more fully.

## ✅ Acceptance Validation

Unit test coverage on this file is now 85.1%; this can be increased in the future by implementing the MockRepository.
image
Awaiting the CodeCov report, as it should be closer to 90% coverage for the entire file now. 

## 🔒 Security Implications

### DIFF
--- a/bcda/api/requests.go
+++ b/bcda/api/requests.go
@@ -229,7 +229,10 @@ func (h *Handler) JobStatus(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		err = errors.Wrap(err, "cannot convert jobID to uint")
 		logger.Error(err)
-		h.RespWriter.Exception(w, http.StatusBadRequest, responseutils.RequestErr, err.Error())
+		//We don't need to return the full error to a consumer.
+		//We pass a bad request header (400) for this exception due to the inputs always being invalid for our purposes
+		h.RespWriter.Exception(w, http.StatusBadRequest, responseutils.RequestErr, "")
+
 		return
 	}
 
@@ -238,7 +241,9 @@ func (h *Handler) JobStatus(w http.ResponseWriter, r *http.Request) {
 		logger.Error(err)
 		// NOTE: This is a catch all and may not necessarily mean that the job was not found.
 		// So returning a StatusNotFound may be a misnomer
+		//In contrast to above, if the input COULD be valid, we return a not found header (404)
 		h.RespWriter.Exception(w, http.StatusNotFound, responseutils.DbErr, "")
+
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


bcda/api/requests.go
```diff
@@ -229,7 +229,10 @@ func (h *Handler) JobStatus(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		err = errors.Wrap(err, "cannot convert jobID to uint")
 		logger.Error(err)
-		h.RespWriter.Exception(w, http.StatusBadRequest, responseutils.RequestErr, err.Error())
+		//We don't need to return the full error to a consumer.
+		//We pass a bad request header (400) for this exception due to the inputs always being invalid for our purposes
+		h.RespWriter.Exception(w, http.StatusBadRequest, responseutils.RequestErr, "")
```
laurenkrugen-navapbc, line 234: good call on removing that error from the consumer side!

bcda/api/requests_test.go
```diff
@@ -506,50 +544,87 @@ func (s *RequestsTestSuite) TestDataTypeAuthorization() {
...
-		cmsId        string
-		resources    []string
-		expectedCode int
-		acoConfig    *service.ACOConfig
+		cmsId           string
+		resources       []string
+		expectedCode    int
+		acoConfig       *service.ACOConfig
+		supplyAuthData  bool
+		breakBB         bool
```
laurenkrugen-navapbc, line 552: what's the difference here between `breakBB` and `breakBBInit`?

alex-dzeda: breakBB will cause the call to get a patient to fail, since the basepath / connection isn't initialized when the BB client is returned.
breakBBInit breaks the actual client creation itself, since it induces an error condition where it's trying to pull a certificate/key that does not exist (control char).

bcda/responseutils/v2/writer.go
```diff
@@ -161,7 +161,9 @@ func CreateOpOutcome(severity fhircodes.IssueSeverityCode_Value, code fhircodes.
 }
 
 func WriteError(outcome *fhirmodelOO.OperationOutcome, w http.ResponseWriter, code int) {
-	w.Header().Set(constants.ContentType, constants.JsonContentType)
+	//Write application/fhir+json header on OperationOutcome responses
+	//https://build.fhir.org/ig/HL7/bulk-data/export.html#response---error-status-1
+	w.Header().Set(constants.ContentType, constants.FHIRJsonContentType)
```
laurenkrugen-navapbc, line 166: What are the implications of updating the header for our consumers? Just wondering if there a compatibility between `JsonContentType` and `FHIRJsonContentType` or if our consumers would need to make any code changes on their end.

alex-dzeda: They _shouldn't_ need to make any changes, unless they've adopted a practice of expecting OperationOutcomes to be accompanied by Content-Type=application/json. Since the CapabilityStatement from BCDA includes application/json and application/fhir+json, I'd imagine the risk of introducing an error in this manner is very low.

laurenkrugen-navapbc (approved):
LGTM. Thanks for following up on that content header!

PR #880, opened 2023-10-25
Surface CCLF import error messages

kyeah:
## 🎫 Ticket

N/A

## 🛠 Changes

- Surface error messages that were not being printed/logged

## ℹ️ Context for reviewers

This would have helped us when debugging CCLF import failures.

## 🔒 Security Implications

### DIFF
--- a/bcda/cclf/fileprocessor.go
+++ b/bcda/cclf/fileprocessor.go
@@ -59,7 +59,7 @@ func (p *processor) walk(path string, info os.FileInfo, err error) error {
 	zipReader, err := zip.OpenReader(filepath.Clean(path))
 	if err != nil {
 		p.skipped = p.skipped + 1
-		msg := fmt.Sprintf("Skipping %s: file is not a CCLF archive.", path)
+		msg := fmt.Sprintf("Skipping %s: file could not be opened as a CCLF archive. %s", path, err.Error())
 		fmt.Println(msg)
 		log.API.Warn(msg)
 		return nil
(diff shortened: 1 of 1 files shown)


alex-dzeda:
This is good, I think there's an additional part that might want to add in the CCLF0/CCLF8 import in cclf.go, where if we're trying to import a CCLF file, but the file name != name from the metadata, we log the names of files that are present. This would bubble up directory issues, since file.name does not pass directory info as is.

kyeah:
@alex-dzeda Thanks, that's a good callout. I did notice that all of the "unknown files" _should_ be printed earlier in the process through these logs:

```
Unknown file found: %s
```

However, those work based on regex string submatching and the check you mentioned in cclf.go require exact filename matching. It feels like those should match so we could update cclf.go to allow for substring matching?

alex-dzeda:
> @alex-dzeda Thanks, that's a good callout. I did notice that all of the "unknown files" _should_ be printed earlier in the process through these logs:
> 
> ```
> Unknown file found: %s
> ```
> 
> However, those work based on regex string submatching and the check you mentioned in cclf.go require exact filename matching. It feels like those should match so we could update cclf.go to allow for substring matching?

Perfect, and switching to matching on the substring would also likely have led to KCF being unaffected by the changed structure.

alex-dzeda:
Only other thing I can think of, too, is logging the permissions of the file (if that doesn't bubble up via your update in line 62), since that appears to be a likely culprit of these errors.

kyeah:
@alex-dzeda I'm going to break out those separate updates into a ticket since they're a little more involved. The existing logging updates in this PR I think will help with the recurring unzipping issues, so I'd like to get those in today if possible 👋

alex-dzeda (approved):

kyeah:
A hastily-made ticket, feel free to update 😄 

https://jira.cms.gov/browse/BCDA-7501

PR #882, opened 2023-11-02
BCDA-6594: Improve logging when receiving BFD 400/500 status codes

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-6594

## 🛠 Changes

Using struct `JobEnqueueArgs` in functions to replace individual arguments of `cmsID` and `jobID`, so that we can include those data points in our requests to BFD. The field of `cmsID` has been added to the model `JobEnqueueArgs`. I ended up the available struct with the additional field because many of the functions that required a changed needed some field data from JobEnqueueArgs. This reducing the number of parameters that functions expect and (hopefully) improves readability.

## ℹ️ Context for reviewers

Our BB client has two methods `GetMetaData` and `GetPatientHashIdentifier` that do not include the job ID or CMS ID in the request headers, which results in those values being empty when the request is logged. 

## ✅ Acceptance Validation

Wrote new test and modified existing; all are passing.

## 🔒 Security Implications

### DIFF
--- a/bcda/api/requests.go
+++ b/bcda/api/requests.go
@@ -538,8 +538,16 @@ func (h *Handler) bulkRequest(w http.ResponseWriter, r *http.Request, reqType se
 		logger.Info("job id created")
 	}
 
+	jobData := models.JobEnqueueArgs{
+		ID:              int(newJob.ID),
+		ACOID:           acoID.String(),
+		Since:           "",
+		TransactionTime: time.Now(),
+		CMSID:           ad.CMSID,
+	}
+
 	// request a fake patient in order to acquire the bundle's lastUpdated metadata
-	b, err := bb.GetPatient("0", strconv.FormatUint(uint64(newJob.ID), 10), acoID.String(), "", time.Now())
+	b, err := bb.GetPatient(jobData, "0")
 	if err != nil {
 		logger.Error(err)
 		h.RespWriter.Exception(w, http.StatusInternalServerError, responseutils.FormatErr, "Failure to retrieve transactionTime metadata from FHIR Data Server.")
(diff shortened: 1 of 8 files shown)


bcda/api/requests.go
```diff
@@ -538,8 +538,16 @@ func (h *Handler) bulkRequest(w http.ResponseWriter, r *http.Request, reqType se
 		logger.Info("job id created")
 	}
 
+	jobData := models.JobEnqueueArgs{
```
laurenkrugen-navapbc, line 541: since GetPatient now requires the JobEnqueueArgs type as a parameter, the function here will now create the struct for the purposes of this initial request to get the bundle's metadata.

bcdaworker/worker/worker_test.go
```diff
@@ -124,112 +125,163 @@ func TestWorkerTestSuite(t *testing.T) {
 	suite.Run(t, new(WorkerTestSuite))
 }
 
-func (s *WorkerTestSuite) TestWriteResourceToFile() {
+func (s *WorkerTestSuite) TestWriteEOBToFile() {
```
laurenkrugen-navapbc, line 128: TestWriteResourceToFile has been broken up from its table test format because the test setup + mocking calls require jobArgs to have the resource types before the tests run, so that they can be used in the mock calls. This resulted in a sort of "chicken and egg" test setup. 

The difference between the old test and the new, is that it only tests one bene and it's no longer table driven.

alex-dzeda (approved):
Thank you! I think this looks cleaner and more maintainable through the use of the struct vs. individual args, and common headers regardless of request type will help with debugging.

PR #883, opened 2023-11-03
BCDA-7480: Add alerting when CCLF-8 import job fails to open a file

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7480

## 🛠 Changes

Updated the cli command `import-cclf-directory` to return the error when one occurs. If no errors come from the `importcclfdirectory` function in cli.go but a file is skipped or fails import, then a new error is created and returned to be handled by the jenkins script.

## ℹ️ Context for reviewers

Jenkins pipeline currently always runs successfully even if cclf files are skipped or fail to import. 

## ✅ Acceptance Validation

Tests still passing. This jenkins pipeline does not have the option to run from a development branch, for both the pipeline and the app. This will need to be deployed and a mock test file will need to be created in order to test. 

## 🔒 Security Implications

### DIFF
--- a/bcda/bcdacli/cli.go
+++ b/bcda/bcdacli/cli.go
@@ -369,7 +369,7 @@ func setUpApp() *cli.App {
 					return err
 				}
 				if failure > 0 || skipped > 0 {
-					log.API.Error("Successfully imported %v files.  Failed to import %v files.  Skipped %v files.  See logs for more details. ", err)
+					log.API.Error("files failed to import or were skipped.  See logs for more details. ", err)
 					cli.NewExitError("some files failed to import or were skipped, see logs for more details", 1)
 					return err
 				}
--- a/bcda/bcdacli/cli.go
+++ b/bcda/bcdacli/cli.go
@@ -23,6 +23,7 @@ import (
 	"github.com/CMSgov/bcda-app/bcda/alr/gen"
 	"github.com/CMSgov/bcda-app/bcda/auth"
 	authclient "github.com/CMSgov/bcda-app/bcda/auth/client"
+
 	"github.com/CMSgov/bcda-app/bcda/cclf"
 	cclfUtils "github.com/CMSgov/bcda-app/bcda/cclf/utils"
 	"github.com/CMSgov/bcda-app/bcda/constants"
@@ -362,6 +363,17 @@ func setUpApp() *cli.App {
 			Action: func(c *cli.Context) error {
 				ignoreSignals()
 				success, failure, skipped, err := cclf.ImportCCLFDirectory(filePath)
+				if err != nil {
+					log.API.Error("error returned from ImportCCLFDirectory: ", err)
+					cli.NewExitError("an error occurred during the cclf-import process, see logs for more details", 1)
+					return err
+				}
+				if failure > 0 || skipped > 0 {
+					log.API.Error("Successfully imported %v files.  Failed to import %v files.  Skipped %v files.  See logs for more details. ", err)
+					cli.NewExitError("some files failed to import or were skipped, see logs for more details", 1)
+					return err
+				}
+				log.API.Infof("Completed CCLF import.  Successfully imported %v files.  Failed to import %v files.  Skipped %v files.  See logs for more details.", success, failure, skipped)
 				fmt.Fprintf(app.Writer, "Completed CCLF import.  Successfully imported %v files.  Failed to import %v files.  Skipped %v files.  See logs for more details.", success, failure, skipped)
 				return err
 			},
--- a/bcda/bcdacli/cli.go
+++ b/bcda/bcdacli/cli.go
@@ -365,13 +365,13 @@ func setUpApp() *cli.App {
 				success, failure, skipped, err := cclf.ImportCCLFDirectory(filePath)
 				if err != nil {
 					log.API.Error("error returned from ImportCCLFDirectory: ", err)
-					cli.NewExitError("an error occurred during the cclf-import process, see logs for more details", 1)
-					return err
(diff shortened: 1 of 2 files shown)


laurenkrugen-navapbc:
I did the error handling at the top level of the cli command because when the cclf file fails to open, it will increment the [failed file count](https://github.com/CMSgov/bcda-app/blob/992557513be4f44b171f9ac4c44ae28ca563e913/bcda/cclf/cclf.go#L335C7-L335C16) and that will cause the exit command to be sent to jenkins. Add this error handling here will hopefully cast a wide net to catch any additional errors that occur.

bcda/bcdacli/cli.go
```diff
@@ -362,6 +363,17 @@ func setUpApp() *cli.App {
 			Action: func(c *cli.Context) error {
 				ignoreSignals()
 				success, failure, skipped, err := cclf.ImportCCLFDirectory(filePath)
+				if err != nil {
+					log.API.Error("error returned from ImportCCLFDirectory: ", err)
+					cli.NewExitError("an error occurred during the cclf-import process, see logs for more details", 1)
+					return err
+				}
+				if failure > 0 || skipped > 0 {
+					log.API.Error("Successfully imported %v files.  Failed to import %v files.  Skipped %v files.  See logs for more details. ", err)
```
karinamzalez, line 372: @laurenkrugen-navapbc out of curiosity, how are `%v` vars populated here?

bcda/bcdacli/cli.go
```diff
@@ -362,6 +363,17 @@ func setUpApp() *cli.App {
 			Action: func(c *cli.Context) error {
 				ignoreSignals()
 				success, failure, skipped, err := cclf.ImportCCLFDirectory(filePath)
+				if err != nil {
+					log.API.Error("error returned from ImportCCLFDirectory: ", err)
+					cli.NewExitError("an error occurred during the cclf-import process, see logs for more details", 1)
+					return err
```
karinamzalez, line 369: @laurenkrugen-navapbc where does log.API.Error log to again? (same Q for cli logs)

bcda/bcdacli/cli.go
```diff
@@ -362,6 +363,17 @@ func setUpApp() *cli.App {
...
+				}
+				if failure > 0 || skipped > 0 {
+					log.API.Error("Successfully imported %v files.  Failed to import %v files.  Skipped %v files.  See logs for more details. ", err)
+					cli.NewExitError("some files failed to import or were skipped, see logs for more details", 1)
+					return err
+				}
+				log.API.Infof("Completed CCLF import.  Successfully imported %v files.  Failed to import %v files.  Skipped %v files.  See logs for more details.", success, failure, skipped)
```
karinamzalez, line 376: @laurenkrugen-navapbc last question, if the instances where "failure" and "skipped" are populated are handled in the condition above, and there's a return statement in said condition -- would that not exit out of the function before getting to this log?

bcda/bcdacli/cli.go
laurenkrugen-navapbc: log.API is one of the global loggers that gets setup [here](https://github.com/CMSgov/bcda-app/blob/master/log/logger.go#L34C6-L34C11). The logs get written to a file and ingested into splunk from there!

bcda/bcdacli/cli.go
laurenkrugen-navapbc: If there is an error or a file is skipped, our import function will increment [those counters](https://github.com/CMSgov/bcda-app/blob/master/bcda/cclf/cclf.go#L332C15-L332C26) as opposed to returning an error. The error returned doesn't really go anywhere right now; the jenkins pipeline script doesn't check the output of the import job so those won't impact the outcome of the pipeline run.

bcda/bcdacli/cli.go
laurenkrugen-navapbc: I actually reverted the logging back to what it originally was, this time including the variables in the logging statement.

alex-dzeda:
Looks good, we should make sure that a failed import job using these new definitions doesn't cause panic / someone to get paged until there's changes to account for files that are still downloading

bcda/bcdacli/cli.go
karinamzalez: @laurenkrugen-navapbc would you be open to discuss this further during eng office hours? I'm not fully clear re: those counters and am curious why the jenkins pipeline doesn't check the output of the import job.

bcda/bcdacli/cli.go
karinamzalez: @laurenkrugen-navapbc what about `cli.NewExitError` ? 👀

laurenkrugen-navapbc: https://pkg.go.dev/gopkg.in/urfave/cli.v1#NewExitError

Since we are using the cli package to have the app intercept cli commands, `cli.NewExitError` is the equivalent of the `bash exit #` command. That actually exits the cli altogether, which is not what we want so I removed it.

laurenkrugen-navapbc:
> Looks good, we should make sure that a failed import job using these new definitions doesn't cause panic / someone to get paged until there's changes to account for files that are still downloading

Agreed! Posting this comment after our eng office hours today and good discussion about how we want to handle files that are still downloading and making sure we aren't inundated with alerts. 

For folks that were not in office hours: BCDA-7512 will address the the files that are still downloading. For now, I am going to leave the pipeline to fail if files are skipped or if files fail to import; 1) I think it would be a good opportunity to cast a wide net and address the failures as they come in and tweak our alerting and logic accordingly and 2) the October incident included files that were skipped, so better to catch all it all for the time being while we make changes to the import process.

PR #885, opened 2023-11-20
BCDA-7480: Update errors returned from import-cclf

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7480

## 🛠 Changes

Updating `import-cclf` command to return an error when at least one file has failed to be imported or a file has been skipped, or if the successfully imported file count is 0.

Smoke tests are failing because the suppression files get added to the skipped count, if they exist in the directory. Since we don't want false positives, a check has been added to see if the file we are currently importing is an opt out file. If it is, we ignore it altogether and don't increment the count of the files skipped.

## ℹ️ Context for reviewers

Smoke tests in dev are failing because opt-out files are being added to the skipped count, which causes the synthetic EFT import to fail.

## ✅ Acceptance Validation

modified tests that import cclf files to also check for opt out files and not add to skipped count.

## 🔒 Security Implications


bcda/cclf/fileprocessor.go
```diff
@@ -119,6 +128,15 @@ func (p *processor) handleArchiveError(path string, info os.FileInfo, cause erro
 	return err
 }
 
+func isOptOut(filename string) (isOptOut bool) {
```
kyeah, line 131: Is this based on an existing function/regex? Just wondering if the CCLF and opt out importers can share the same utility function in case we ever need to change them.

bcda/bcdacli/cli.go
```diff
@@ -368,7 +368,7 @@ func setUpApp() *cli.App {
 					return err
 
 				}
-				if failure > 0 || skipped > 0 {
+				if failure > 0 || skipped > 0 || success == 0 {
```
kyeah, line 371: Will this cause an error / alert if no files were available for import?

bcda/cclf/fileprocessor.go
laurenkrugen-navapbc: yeah - I pulled the [regex out of this file](https://github.com/CMSgov/bcda-app/blob/e83513faa7785edd9db1eefdd7c3742f1777c043/optout/utils.go#L25C6-L25C19), but can create an exported util function that both can use! I'll update.

bcda/bcdacli/cli.go
laurenkrugen-navapbc: Yes, this will cause an alert if no files were successfully imported. There is [another piece of logic](https://github.com/CMSgov/bcda-app/blob/master/bcda/cclf/cclf.go#L298) that will catch an error if we've walked the contents of the directory and have no files that can be imported. I added this because when I was reviewing the logs and the output during the CCLF incident, I did see an instance of 0 files being imported.

bcda/cclf/fileprocessor.go
laurenkrugen-navapbc: OK - this has now been moved to the utils.go file used by `optout` to reduce repetitive code. 🎉

bcda/bcdacli/cli.go
laurenkrugen-navapbc: Per our conversation during office hours, this success statement may be a bit too aggressive; removed it!

kyeah (approved):
nice!

PR #888, opened 2023-11-21
BCDA-7134: Update error logging

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7134

## 🛠 Changes

- Add new error logs when returning errors to the API caller
- Add new error/warning logs when returning from a function in response to an error or other conditional scenarios
- Ensure errors are wrapped (`err = errors.Wrap(err, "my message"))` instead of ignored (`err = errors.New("my message")`)
- Surface error from `HashIdentifier`
- Use ctx logger when plausible
- Use `error` level when logging 500 errors

## ℹ️ Context for reviewers

In several areas of the codebase, we are running into errors and taking action on them without logging the original error. We should ensure that errors are logged when possible, and that we are otherwise logging contextual information when making conditional decisions (e.g. in the `hasDuplicates` function, we should be logging each conditional check.)

I basically looked at every instance of `err != nil` and verified whether it was being logged or returned to a downstream caller (who should also be checking for err != nil, and then logging it or returning it farther downstream.)

Note that I did not make any changes to the writer files to utilize the ctx logger since they were heavily used and would likely require more involved changes there.

## ✅ Acceptance Validation

Unit Testing.

## 🔒 Security Implications


bcda/client/bluebutton_test.go
```diff
@@ -292,6 +292,20 @@ func (s *BBRequestTestSuite) TestGetClaim() {
...
 }
 
+func (s *BBRequestTestSuite) TestGetClaim_HashIdentifierError() {
+	existingPepper := conf.GetEnv("BB_HASH_PEPPER")
+
+	defer func() {
+		conf.SetEnv(s.T(), "BB_HASH_PEPPER", existingPepper)
+	}()
+
+	conf.SetEnv(s.T(), "BB_HASH_PEPPER", "Ã«ÃÃ¬Ã¹Ã")
```
laurenkrugen-navapbc, line 302: Are these characters deliberate to break the hash?

bcdaworker/worker/bluebutton.go
```diff
@@ -12,7 +12,11 @@ import (
 // This method will ensure that a valid BlueButton ID is returned.
 // If you use cclfBeneficiary.BlueButtonID you will not be guaranteed a valid value
 func getBlueButtonID(bb client.APIClient, mbi string, jobData models.JobEnqueueArgs) (blueButtonID string, err error) {
-	hashedIdentifier := client.HashIdentifier(mbi)
+	hashedIdentifier, err := client.HashIdentifier(mbi)
```
laurenkrugen-navapbc, line 15: thanks for updating this!

laurenkrugen-navapbc (approved):
LGTM - thanks for the all of the updates!

bcda/client/bluebutton_test.go
kyeah: yes!

bcda/auth/client/ssas.go
```diff
@@ -56,7 +56,7 @@ func NewSSASClient() (*SSASClient, error) {
 
 	var timeout int
 	if timeout, err = strconv.Atoi(conf.GetEnv("SSAS_TIMEOUT_MS")); err != nil {
-		log.SSAS.Info("Could not get SSAS timeout from environment variable; using default value of 500.")
+		log.SSAS.Info(errors.Wrap(err, "Could not get SSAS timeout from environment variable; using default value of 500."))
```
karinamzalez, line 59: Out of curiosity, why aren't we able to do something like `log.SSAS.Error` here? 

Also -- I've noticed `.Error` and `Errorf` -- what's the difference?

bcda/logging/middleware.go
```diff
@@ -77,23 +77,15 @@ type ResourceTypeLogger struct {
...
 		jobKey, err := rl.extractJobKey(r)
 		if err != nil {
-			log.API.Error(err)
-			return
-		}
-		// Note: could split this out into a function for adding to the context log
-		entry, ok := middleware.GetLogEntry(r).(*log.StructuredLoggerEntry)
-		if !ok {
-			log.API.Error("Incorrect type of logger used in request context")
+			logger := log.GetCtxLogger(r.Context())
```
karinamzalez, line 82: Why are some logs made using the logger while others aren't?

bcda/client/bluebutton.go
```diff
@@ -123,7 +123,7 @@ func NewBlueButtonClient(config BlueButtonConfig) (*BlueButtonClient, error) {
 	}
 	var timeout int
 	if timeout, err = strconv.Atoi(conf.GetEnv("BB_TIMEOUT_MS")); err != nil {
-		logger.Info("Could not get Blue Button timeout from environment variable; using default value of 500.")
+		logger.Info(errors.Wrap(err, "Could not get Blue Button timeout from environment variable; using default value of 500."))
```
karinamzalez, line 126: why aren't we using logger.Error here?

karinamzalez (approved):

bcda/auth/client/ssas.go
kyeah: Oh! Good question — I didn't look at what method was being used here. In general I use the following definitions for each log level:

- INFO: Informational
- WARN: Not a show-stopper, but something out-of-the-ordinary 
- ERROR: A show-stopper (unexpected error in the system, stops the rest of the function)

In this case, I'll change it to `log.SSAS.Warn` since it's not expected to happen but we have a fallback/default value.

bcda/logging/middleware.go
kyeah: We should use the context logger in most cases if we have access to a context or http.Request object. I tried to do that in this Pr except for the `writer` files which will require a more involved refactoring to get access to the request context. I'll create a ticket for that.

bcda/client/bluebutton.go
kyeah: Similar to my comment above — will change this to `Warn`.

bcda/auth/client/ssas.go
kyeah: Oh and your second question! `Error` is for basic messages and `Errorf` accepts multiple arguments for formatting.

```
log.SSAS.Error("I did something wrong")
log.SSAS.Errorf("I did something wrong for organization %s", organization.id) 
```

bcda/logging/middleware.go
kyeah: Followup ticket here: https://jira.cms.gov/browse/BCDA-7581

cc. @laurenkrugen-navapbc

PR #889, opened 2023-11-22
BCDA-7565 (1/x): Modularize opt out functionality

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7565

## 🛠 Changes

- Create a new Importer class that can take in different "file handler" (e.g. local file, S3) and "saver" (mock saver, bcda db saver) implementations
- Allow importer to take in configurations when you create the class, instead of pulling from env vars
- Move local file functionality to LocalFileHandler

## ℹ️ Context for reviewers

I made this branch a few months ago while breaking down some DPC opt out work 😅 but I think it may still be helpful for transitioning us into S3.

Opening this for initial thoughts before I either move forward with it or close it out and take a more direct approach.

## ✅ Acceptance Validation

## 🔒 Security Implications


optout/file_handler.go
```diff
@@ -0,0 +1,11 @@
+package optout
```
laurenkrugen-navapbc, line 1: should we add a comment either to the top of this file or underneath the `OptOutFileHandler` that gives a quick overview of the interface + potential for other handlers?

optout/mock_saver.go
```diff
@@ -0,0 +1,22 @@
+package optout
+
+type MockSaver struct {
+	Files         []OptOutFile
+	OptOutRecords []OptOutRecord
+}
+
+func (m *MockSaver) SaveFile(optOutFile OptOutFile) (fileID uint, err error) {
```
laurenkrugen-navapbc, line 8: testify has a tool for mocking that we use elsewhere in our tests; out of curiosity, was it not a good fit here?

kyeah: Good question! I think the difference here is to have something that can store state, whereas the testify mocking is more about mocking specific calls.

I think I could probably have used testify and done the test assertions differently (checking specific calls to "UpdateImportStatus" for example, instead of checking the internal end state) but it's a little more awkward to read and write.

This post is behind a paywall, bu the first part seems to align with my thinking about testing state vs. testing calls 🙂 I think what I'm making here is not really a "mock" saver but a "fake" in-memory saver.

https://medium.com/@june.pravin/mocking-is-not-practical-use-fakes-e30cc6eaaf4e

laurenkrugen-navapbc: ah, that makes much more sense; fake vs mock. Not required, but maybe a name update would provide more clarity? Ex: `FakeSaver` vs `MockSaver`

laurenkrugen-navapbc (approved):

PR #890, opened 2023-11-30
BCDA-7565: Support S3 file downloads for opt-out files

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7565

## 🛠 Changes

- Fix intermittent test errors due to `db.Close()` statements (switch to sqlmock)
- Add localstack (for spinning up local AWS S3 mock server)
- Add new CLI options for import-suppression-directory
- Add S3FileHandler to search and import files from S3

## ℹ️ Context for reviewers

We run a daily Jenkins job that sshes onto an NFS server and reads files from local disk. In the future, the BFD team will be serving files through their S3 bucket; this is an interim solution that switches our file handling to read from their S3 bucket when given the `--filesource s3` option.

## ✅ Acceptance Validation

I added files locally to dev S3 and then manually ran the import in dev.

[This job](https://management.bcda.cms.gov/job/BCDA%20-%201800medicare%20Import/22839/) imported files successfully.

[This job](https://management.bcda.cms.gov/job/BCDA%20-%201800medicare%20Import/22840/) correctly did nothing after seeing nothing in S3.

[This job](https://management.bcda.cms.gov/job/BCDA%20-%201800medicare%20Import/22838/) failed due to an invalid file header (which was expected.)

[This job](https://management.bcda.cms.gov/job/BCDA%20-%201800medicare%20Import/22837/) showed one skipped file due to an invalid filename (also expected.)

See [the ops PR](https://github.com/CMSgov/bcda-ops/pull/995) for more details.

## 🔒 Security Implications

- [x] This PR adds a new software dependency or dependencies.

This adds a new **local-development-only** dependency for Localstack for testing against a fake S3 server.


shared_files/encrypted/local.env
```diff
@@ -1,120 +1,122 @@
 $ANSIBLE_VAULT;1.1;AES256
```
kyeah, line 1: Updated this to include `BFD_S3_ENDPOINT=http://localhost:4566`

alex-dzeda:
@kyeah It looks like this is the first use of localstack in the bcda-app project; would this be something that needs to trigger one of the "security implications" checkboxes?

kyeah:
> @kyeah It looks like this is the first use of localstack in the bcda-app project; would this be something that needs to trigger one of the "security implications" checkboxes?

Thanks! I'll check the "new dependency" box but add a note that it's _only_ installed for local development/testing. It shouldn't need to wait for approval (cc. @StewGoin @bhodges-navapbc)

optout/s3_file_handler.go
```diff
@@ -1 +1,220 @@
...
+	}
+
+	handler.Infof("file downloaded: size=%d", numBytes)
+
+	byte_arr := buff.Bytes()
+	sc := bufio.NewScanner(bytes.NewReader(byte_arr))
+	return sc, func() {}, err
+}
+
+func (handler *S3FileHandler) CleanupOptOutFiles(suppresslist []*OptOutFilenameMetadata) error {
```
alex-dzeda, line 123: Are these opt-out files per-consumer, or shared between them?

optout/s3_file_handler.go
```diff
@@ -1 +1,220 @@
...
+	return nil
+}
+
+// Creates a new AWS S3 session. If the handler is given a custom S3 endpoint
+// and/or IAM role ARN to assume, the new session connects using those parameters.
+func (handler *S3FileHandler) createSession() (*session.Session, error) {
+	sess := session.Must(session.NewSession())
+
+	config := aws.Config{
+		Region: aws.String("us-east-1"),
```
alex-dzeda, line 181: Do we see a scenario where a file may _not_ be in us-east-1?

alex-dzeda (commented):
Looks good, comments mainly concerned with the potential for region based issues as well as cross-talk between consumers, but may be false alarms!

optout/s3_file_handler.go
kyeah: Per consumer (each API team has their own S3 folders that only they have access to.)

optout/s3_file_handler.go
kyeah: Ah thanks, I said on our call earlier that it would automatically use whatever is configured in the AWS profile, but I forgot that this is hardcoded.

The BFD S3 buckets will live in us-east-1 so I think we can leave them hardcoded for now to avoid the "too many options/parameters/environment variables" fatigue, but open to alternative thoughts! 

I did note that this region is hardcoded in several other places in the codebase (e.g. the insights pipeline and a looot of Jenkins workflows)

alex-dzeda (approved):
Looks good, thanks for addressing comments!

PR #897, opened 2023-12-08
BCDA-7602: Add Transaction ID to log events

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7602

## 🛠 Changes

Added:
- middleware that adds a transaction ID to the incoming requests context
- request headers in the BB client to include the transaction ID
- field of `TransactionID` to model JobEnqueueArgs to pass identifier to worker/job processing

Updated:
- tests to check for log messages
- updated deprecated `ioutil`

Removed:
- unused code

## ℹ️ Context for reviewers

Adding transaction ID to all log events between api, worker, ssas, and BFD so we can have better visibility into system health and alerts.

## ✅ Acceptance Validation

- added and modified tests; tests pass. 
- ran locally and verified transaction ID was added to log statements.

## 🔒 Security Implications

### DIFF
--- a/bcda/client/bluebutton.go
+++ b/bcda/client/bluebutton.go
@@ -350,7 +350,6 @@ func (bbc *BlueButtonClient) getRawData(jobData models.JobEnqueueArgs, u *url.UR
 			return err
 		}
 		addDefaultRequestHeaders(req, uuid.NewRandom(), jobData)
-		// add transaction ID to req context
 		result, err = bbc.client.DoRaw(req)
 		if err != nil {
 			logger.Error(err)
--- a/bcda/client/bluebutton.go
+++ b/bcda/client/bluebutton.go
@@ -36,8 +36,9 @@ import (
 var logger logrus.FieldLogger
 
 const (
-	clientIDHeader = "BULK-CLIENTID"
-	jobIDHeader    = "BULK-JOBID"
+	clientIDHeader      = "BULK-CLIENTID"
+	jobIDHeader         = "BULK-JOBID"
+	transactionIDHeader = "TRANSACTIONID"
 )
 
 // BlueButtonConfig holds the configuration settings needed to create a BlueButtonClient
@@ -310,7 +311,7 @@ func (bbc *BlueButtonClient) tryBundleRequest(u *url.URL, jobData models.JobEnqu
 		}
 
 		queryID := uuid.NewRandom()
-		addDefaultRequestHeaders(req, queryID, strconv.Itoa(jobData.ID), jobData.CMSID)
+		addDefaultRequestHeaders(req, queryID, jobData)
 
 		result, nextURL, err = bbc.client.DoBundleRequest(req)
 		if err != nil {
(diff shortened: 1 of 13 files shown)


bcda/client/bluebutton_test.go
```diff
@@ -871,10 +875,7 @@ func (s *BBRequestTestSuite) TestValidateRequest() {
 		s.T().Run(tt.name, func(t *testing.T) {
 
 			tsValidation := httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-				uid, err := uuid.Parse(req.Header.Get("BlueButton-OriginalQueryId"))
```
laurenkrugen-navapbc, line 874: It looked like we had two different UUID packages we used in the repo. updated to the more commonly used one here, which changes the return value of `Parse`, hence the change.

bcda/client/bluebutton.go
```diff
@@ -379,7 +380,7 @@ func (bbc *BlueButtonClient) getURL(path string, params url.Values) (*url.URL, e
 	return u, nil
 }
 
-func addDefaultRequestHeaders(req *http.Request, reqID uuid.UUID, jobID, cmsID string) {
+func addDefaultRequestHeaders(req *http.Request, reqID uuid.UUID, jobData models.JobEnqueueArgs) {
```
laurenkrugen-navapbc, line 383: since we are returning 3 values from the JobEnqueueArgs model, I removed the individual parameters and opted to pass in the whole type instead.

bcda/logging/middleware.go
```diff
@@ -60,6 +60,8 @@ func (l *StructuredLogger) NewLogEntry(r *http.Request) middleware.LogEntry {
 		logFields["cms_id"] = ad.CMSID
 	}
 
+	logFields["transaction_id"] = r.Context().Value(CtxTransactionKey).(string)
```
kyeah, line 63: Do we need to check return values here, similar to the if statement above?

```go
if tid, ok := r.Context().Value(CtxTransactionKey).(string); ok {
  ...
}
```

bcda/logging/middleware.go
```diff
@@ -122,8 +124,23 @@ func NewCtxLogger(next http.Handler) http.Handler {
...
 }
+
+// type to create context.Context key
+type CtxTransactionKeyType string
+
+// context.Context key to get the transaction ID from the request context
+const CtxTransactionKey CtxTransactionKeyType = "ctxTransaction"
+
+// Adds a transaction ID to the request context
+func TransactionHandler(next http.Handler) http.Handler {
```
kyeah, line 141: Small nit, but might be clearer to name this something like `TransactionID` to match `middleware.RequestID`  or something like `logging.NewTransactionID` to match `logging.NewStructuredLogger`.

kyeah (approved):

bcda/logging/middleware.go
laurenkrugen-navapbc: good catch, that is how we should be checking! I'll update that.

bcda/logging/middleware.go
laurenkrugen-navapbc: Yeah, I agree - will update this as well!

PR #907, opened 2024-01-25
BCDA-7581: Utilize context logger for FHIR response writers in bcda-app

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7581

## 🛠 Changes

- Passing logger context to response writers to be consistent with use of ctx logger elsewhere. 
- Metadata endpoint still using `log.API` since there is no relevant ctx to use
- auth middleware is using `log.AUTH` to create a ctx logger because auth middleware is processed before ctx logger middleware in the chi routers.
- MakeTestStructuredLoggerEntry duplication is because of circular dependencies in testUtils. Would like to work on package reorganization later to start eliminating circular dependencies here and elsewhere.

## ℹ️ Context for reviewers

We updated error logging but did not switch to using the context logger (which has standardized attributes) for the FHIR response v1 and v2 writer files (writer.go).

## ✅ Acceptance Validation

tests updated and passing.

## 🔒 Security Implications

### DIFF
--- a/Dockerfiles/Dockerfile.ssas
+++ b/Dockerfiles/Dockerfile.ssas
@@ -36,6 +36,7 @@ RUN openssl rsa -in /var/local/private.pem -outform PEM -pubout -out /var/local/
 
 # Make sure we are in the directory to ensure the config files are resolved as expected
 COPY --from=base /go/bcda-ssas-app/ /go/bcda-ssas-app/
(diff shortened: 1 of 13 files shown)


alex-dzeda (approved):
LGTM; glad that the CI build issues are fixed, too.

laurenkrugen-navapbc:
> LGTM; glad that the CI build issues are fixed, too.

Thanks for the help getting that fixed!

PR #911, opened 2024-02-16
BCDA-7811: Update logging to differentiate reasons for job not found

alex-dzeda:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7811

## 🛠 Changes

Added tests + logic to differentiate between a row not being found, and any other type of issue. 

## ℹ️ Context for reviewers

BCDA has received reports of jobs not being found that are clearly in our database, this will create different levels of logging for those that are truly not found vs. something else, so we can better address that. 

## ✅ Acceptance Validation

Added unit tests, all tests pass. 

## 🔒 Security Implications

### DIFF
--- a/bcda/api/requests.go
+++ b/bcda/api/requests.go
@@ -242,10 +242,13 @@ func (h *Handler) JobStatus(w http.ResponseWriter, r *http.Request) {
 	job, jobKeys, err := h.Svc.GetJobAndKeys(r.Context(), uint(jobID))
 	if err != nil {
 		logger.Error(err)
-		// NOTE: This is a catch all and may not necessarily mean that the job was not found.
-		// So returning a StatusNotFound may be a misnomer
-		//In contrast to above, if the input COULD be valid, we return a not found header (404)
-		h.RespWriter.Exception(r.Context(), w, http.StatusNotFound, responseutils.DbErr, "")
+		if err == sql.ErrNoRows {
+			log.API.Info("Requested job not found.")
+			h.RespWriter.Exception(r.Context(), w, http.StatusNotFound, responseutils.DbErr, "Job not found.")
+		} else {
+			log.API.Error("Error attempting to request job. Error:", err.Error())
+			h.RespWriter.Exception(r.Context(), w, http.StatusInternalServerError, responseutils.InternalErr, "Error trying to fetch job. Please try again.")
+		}
 
 		return
 	}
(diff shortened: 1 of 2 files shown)


bcda/api/requests.go
```diff
@@ -242,10 +242,13 @@ func (h *Handler) JobStatus(w http.ResponseWriter, r *http.Request) {
 	job, jobKeys, err := h.Svc.GetJobAndKeys(r.Context(), uint(jobID))
 	if err != nil {
 		logger.Error(err)
-		// NOTE: This is a catch all and may not necessarily mean that the job was not found.
-		// So returning a StatusNotFound may be a misnomer
-		//In contrast to above, if the input COULD be valid, we return a not found header (404)
-		h.RespWriter.Exception(r.Context(), w, http.StatusNotFound, responseutils.DbErr, "")
+		if err == sql.ErrNoRows {
```
kyeah, line 245: Do we need to differentiate between an error from `GetJobById` and `GetJobKeys`? It seems like GetJobById should be StatusNotFound, and GetJobKeys should be an internal error since we'd expect job keys for successfully-completed jobs.

kyeah, line 245: Also! totally non-blocking nit, but should this use `errors.Is(err, sql.ErrNoRows)`? I see this check in the bcdaworker folder, although I assume it's just another way of writing this.

kyeah (commented):
Thanks! one small question (I think this could be merged but there may be another improvement plausible here.)

I also noted similarly that `RequireTokenJobMatch` uses GetJobById and returns a StatusNotFound if there's any error. Seems like it'd be worth applying the same change there.

greg-pf:
Just a note, this should be [BCDA-7811](https://jira.cms.gov/browse/BCDA-7811) - sorry for sharing the wrong ticket # earlier @alex-dzeda

bcda/api/requests.go
alex-dzeda: Good catch, I'll update that + apply above; errors.Is will be more flexible since it can unwrap them.

alex-dzeda: Right before GetJobKeys it checks for the completed one, so that query won't end up executing if the job is completed. Since GetJobKeys uses QueryContext it _shouldn't_ return an error if there are no rows, since QueryContext returns an EOF if empty (vs QueryRowContext which throws the ErrNoRows), and for jobs that are Pending but not yet In Process, it's possible to have no job keys

kyeah: got it, thanks!

kyeah (approved):

PR #917, opened 2024-03-05
BCDA-7893: Prevent 429s on /jobs endpoint

alex-dzeda:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7893

## 🛠 Changes

Skip duplicate check on /jobs endpoint.

## ℹ️ Context for reviewers

Middleware is enabled on several BCDA routes, and an existing check would block the /jobs endpoint if the ACO had a pending or in-progress job. 

## ✅ Acceptance Validation

Added unit tests and ran locally w/ rate_limit enabled. No longer get a 429 with an in-progress/pending job when checking job list. 

## 🔒 Security Implications

### DIFF
--- a/bcda/web/middleware/ratelimit.go
+++ b/bcda/web/middleware/ratelimit.go
@@ -6,7 +6,6 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/CMSgov/bcda-app/bcda/auth"
@@ -70,9 +69,6 @@ func CheckConcurrentJobs(next http.Handler) http.Handler {
 
 func hasDuplicates(ctx context.Context, pendingAndInProgressJobs []*models.Job, types []string, version string, newRequestUrl string) bool {
 	logger := log.GetCtxLogger(ctx)
-	if strings.Contains(newRequestUrl, "/jobs") && !strings.Contains(newRequestUrl, "$export") {
-		return false
-	}
 
 	typeSet := make(map[string]struct{}, len(types))
 	for _, t := range types {
--- a/bcda/web/middleware/ratelimit.go
+++ b/bcda/web/middleware/ratelimit.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/CMSgov/bcda-app/bcda/auth"
@@ -69,6 +70,9 @@ func CheckConcurrentJobs(next http.Handler) http.Handler {
 
 func hasDuplicates(ctx context.Context, pendingAndInProgressJobs []*models.Job, types []string, version string, newRequestUrl string) bool {
 	logger := log.GetCtxLogger(ctx)
+	if strings.Contains(newRequestUrl, "/jobs") && !strings.Contains(newRequestUrl, "$export") {
+		return false
+	}
 
 	typeSet := make(map[string]struct{}, len(types))
 	for _, t := range types {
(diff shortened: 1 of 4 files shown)


laurenkrugen-navapbc:
Another option here that would not require us to check the URL would be to directly modify the handlers; we could remove `checkConcurrentJobs` from the[ list of handlers](https://github.com/CMSgov/bcda-app/blob/master/bcda/web/router.go#L62C3-L62C97) for the `/jobs` endpoint altogether. 

That would require that we also split `checkConcurrentJobs` from `[requestValidators](https://github.com/CMSgov/bcda-app/blob/master/bcda/web/router.go#L52)` so those two are separate handlers. Once those are split, we can apply the `checkConcurrentJobs` to only the `Patient` and `Group` endpoints, since those are the only two that would really require that middleware (only endpoints that create jobs).

This would not only eliminate the need for the additional code change and checking in the `checkConcurrentJobs` middleware, but would also remove that check for endpoints that don't need it, like `/jobs`. What are your thoughts?

alex-dzeda:
> Another option here that would not require us to check the URL would be to directly modify the handlers; we could remove `checkConcurrentJobs` from the[ list of handlers](https://github.com/CMSgov/bcda-app/blob/master/bcda/web/router.go#L62C3-L62C97) for the `/jobs` endpoint altogether.
> 
> That would require that we also split `checkConcurrentJobs` from `[requestValidators](https://github.com/CMSgov/bcda-app/blob/master/bcda/web/router.go#L52)` so those two are separate handlers. Once those are split, we can apply the `checkConcurrentJobs` to only the `Patient` and `Group` endpoints, since those are the only two that would really require that middleware (only endpoints that create jobs).
> 
> This would not only eliminate the need for the additional code change and checking in the `checkConcurrentJobs` middleware, but would also remove that check for endpoints that don't need it, like `/jobs`. The only remaining thing to do would be to apply this only to production. What are your thoughts?

In the context of v1, would that look something like this, you think?
`var requestValidators = []func(http.Handler) http.Handler{
		middleware.ACOEnabled(cfg), middleware.ValidateRequestURL, middleware.ValidateRequestHeaders,
	}
	nonExportRequestValidators := []func(http.Handler) http.Handler{
		middleware.ACOEnabled(cfg), middleware.ValidateRequestURL, middleware.ValidateRequestHeaders,
	}

	if conf.GetEnv("DEPLOYMENT_TARGET") != "prod" {
		r.Get("/", userGuideRedirect)
		r.Get(`/{:(user_guide|encryption|decryption_walkthrough).html}`, userGuideRedirect)
	} else {
		// Apply rate limiting on production only
		requestValidators = append(requestValidators, middleware.CheckConcurrentJobs)
	}

		r.With(append(commonAuth, requestValidators...)...).Get(m.WrapHandler("/Patient/$export", v1.BulkPatientRequest))
		if conf.GetEnv("ENABLE_ALR_ENDPOINTS") == "true" {
			r.With(append(commonAuth, requestValidators...)...).Get(m.WrapHandler("/alr/$export", v1.ALRRequest))
		}
		r.With(append(commonAuth, requestValidators...)...).Get(m.WrapHandler("/Group/{groupId}/$export", v1.BulkGroupRequest))
		r.With(append(commonAuth, auth.RequireTokenJobMatch)...).Get(m.WrapHandler(constants.JOBIDPath, v1.JobStatus))
		r.With(append(commonAuth, nonExportRequestValidators...)...).Get(m.WrapHandler("/jobs", v1.JobsStatus))
		r.With(append(commonAuth, auth.RequireTokenJobMatch)...).Delete(m.WrapHandler(constants.JOBIDPath, v1.DeleteJob))
		r.With(commonAuth...).Get(m.WrapHandler("/attribution_status", v1.AttributionStatus))
		r.Get(m.WrapHandler("/metadata", v1.Metadata))
	})`

laurenkrugen-navapbc (approved):
LGTM!

PR #921, opened 2024-03-11
BCDA-7851: Compare _since parameter in Group endpoint with CCLF file processing date

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7851

## 🛠 Changes

- Modify new beneficiaries check to compare `_since` parameter against the latest CCLF file processing date instead of attribution timestamp

## ℹ️ Context for reviewers

Currently, the since parameter is not implemented _optimally per the spec on the Bulk FHIR IG (https://build.fhir.org/ig/HL7/bulk-data/export.html#query-parameters) . 

The current logic can lead to instances (especially around EOY) where users can receive excessive historical data due to comparisons with a previous PY file for comparing beneficiary rolls. 

The issue that ACOs have run into is that today, the timestamp is being used for comparisons, but there can be a week+ delay of when a file is generated, and when it is processed.

Instead, the new beneficiaries check should be calculated from the user's point of view (using the file processing time), _not_ based on the upstream attribution/CCLF file timestamp.

## ✅ Acceptance Validation

- Unit testing

## 🔒 Security Implications

No new data implications.

### DIFF
--- a/bcda/models/postgres/repository.go
+++ b/bcda/models/postgres/repository.go
@@ -117,9 +117,9 @@ func (r *Repository) GetCCLFFileExistsByName(ctx context.Context, name string) (
 	return true, nil
 }
 
-func (r *Repository) GetLatestCCLFFile(ctx context.Context, cmsID string, cclfNum int, importStatus string, lowerBound, upperBound time.Time, fileType models.CCLFFileType) (*models.CCLFFile, error) {
+func (r *Repository) GetLatestCCLFFile(ctx context.Context, cmsID string, cclfNum int, importStatus string, lowerBound time.Time, upperBound time.Time, fileType models.CCLFFileType) (*models.CCLFFile, error) {
 	sb := sqlFlavor.NewSelectBuilder()
-	sb.Select("id", "name", "timestamp", "performance_year")
+	sb.Select("id", "name", "timestamp", "performance_year", "created_at")
 	sb.From("cclf_files")
 	sb.Where(
 		sb.Equal("aco_cms_id", cmsID),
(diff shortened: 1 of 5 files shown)


bcda/service/service.go
```diff
@@ -369,8 +386,8 @@ func (s *service) getNewAndExistingBeneficiaries(ctx context.Context, conditions
 	//
 	oldFileUpperBound := conditions.Since
 
-	// If the _since parameter is more recent than the latest CCLF file, set the upper bound
-	// for the older file to be prior to cclfFileNew.Timestamp.
+	// If the _since parameter is more recent than the latest CCLF file timestamp, set the upper bound
+	// for the older file to be prior to the newest file's timestamp.
```
kyeah, line 390: I don't think any changes to the CCLF file queries are actually necessary here, we still want to query the lower/upper bounds based on the timestamp and not created_at. But please confirm that business logic 👋

alex-dzeda: That is my understanding: We're tied to not using CCLF files (for current PY) generated over 45 days ago, regardless of when they actually arrived, due to the possibility of in-year re-alignments.

bcda/service/service.go
```diff
@@ -359,6 +359,23 @@ func (s *service) getNewAndExistingBeneficiaries(ctx context.Context, conditions
...
+		if err != nil {
+			return nil, nil, err
+		}
+
+		if len(benes) == 0 {
+			return nil, nil, fmt.Errorf("Found 0 new or existing beneficiaries from CCLF8 file for cmsID %s cclfFiledID %d",
+				conditions.CMSID, cclfFileNew.ID)
+		}
+
+		return newBeneficiaries, benes, nil
```
alex-dzeda, line 376: Style question, since newBeneficiaries in this instance would be 0 / newBeneficiaries isn't assigned, what are your thoughts on returning 0 in lieu of newBenes as a variable?

bcda/models/postgres/repository.go
```diff
@@ -117,9 +117,9 @@ func (r *Repository) GetCCLFFileExistsByName(ctx context.Context, name string) (
 	return true, nil
 }
 
-func (r *Repository) GetLatestCCLFFile(ctx context.Context, cmsID string, cclfNum int, importStatus string, lowerBound, upperBound time.Time, fileType models.CCLFFileType) (*models.CCLFFile, error) {
+func (r *Repository) GetLatestCCLFFile(ctx context.Context, cmsID string, cclfNum int, importStatus string, lowerBound time.Time, upperBound time.Time, fileType models.CCLFFileType) (*models.CCLFFile, error) {
```
laurenkrugen-navapbc, line 120: 👋 Not sure if this change is just for readability and my comment is not a request to change it: You can specify a type for a function parameter once, if there are multiple parameters of the same type. 

Ex: In this function:`func FooString(int x, foo, bar, hello, string)` foo, bar, hello would be of the type `string`.

bcda/service/service.go
alex-dzeda: Got it! Definitely don't want to seg fault, especially in a language that's not C.

alex-dzeda (approved):
Thanks for doing this + addressing comments, excited to see _since functioning as intended!

PR #930, opened 2024-03-27
BCDA-7845: Update operation outcome for error ndjson

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7845

## 🛠 Changes

- `CreateOpOutcome` conforms to [HL7 spec](https://hl7.org/fhir/R4/operationoutcome.html)
- `writeBBDataToFile` uses the fhir code of "processing" instead of "not-found" when MBIs cannot be retrieved from BlueButton.

Operation Outcome Before:
```
{
	"resourceType": "OperationOutcome",
	"issue": [
		{
			"severity": "error",
			"code": "not-found",
			"details": {
				"coding": [
					{
						"system": "http://hl7.org/fhir/ValueSet/operation-outcome",
						"code": "Blue Button Error",
						"display": "Error retrieving ExplanationOfBenefit for beneficiary MBI <> in ACO <>"
					}
				],
				"text": "Error retrieving ExplanationOfBenefit for beneficiary MBI <> in ACO <>"
			}
		}
	]
}
```

Operation Outcome After:
```
{
	"resourceType": "OperationOutcome",
	"issue": [
		{
			"severity": "error",
			"code": "processing",
			"diagnostics": "Error retrieving ExplanationOfBenefit for beneficiary MBI <> in ACO <>"
		}
	]
}
```

## ℹ️ Context for reviewers

`errType` is now an unused parameter. It was not removed because it would have greatly expanded the scope of this ticket and there is an opportunity to review and update our HTTP responses while removing that parameter in a [future ticket](https://jira.cms.gov/browse/BCDA-7962). Post man tests have also been modified to no longer check for the `errType` in the test assertions, where applicable.

HTTP error responses outside of the errors in ndjson files will also have this updated format, since they utilize `CreateOpOutcome`.

## ✅ Acceptance Validation

Tests updated and passing.

## 🔒 Security Implications


bcda/auth/middleware.go
```diff
@@ -108,17 +108,17 @@ func handleTokenVerificationError(ctx context.Context, w http.ResponseWriter, rw
...
 		switch err.(type) {
 		case *customErrors.ExpiredTokenError:
-			rw.Exception(ctx, w, http.StatusUnauthorized, responseutils.ExpiredErr, "")
+			rw.Exception(ctx, w, http.StatusUnauthorized, http.StatusText(http.StatusUnauthorized), responseutils.ExpiredErr)
 		case *customErrors.EntityNotFoundError:
-			rw.Exception(ctx, w, http.StatusForbidden, responseutils.UnauthorizedErr, responseutils.UnknownEntityErr)
+			rw.Exception(ctx, w, http.StatusForbidden, http.StatusText(http.StatusForbidden), responseutils.UnauthorizedErr)
 		case *customErrors.RequestorDataError:
-			rw.Exception(ctx, w, http.StatusBadRequest, responseutils.InternalErr, "")
+			rw.Exception(ctx, w, http.StatusBadRequest, http.StatusText(http.StatusBadRequest), responseutils.InternalErr)
```
alex-dzeda, line 115: Perhaps RequestErr instead of InternalErr?

bcdaworker/worker/worker.go
```diff
@@ -233,13 +233,13 @@ func writeBBDataToFile(ctx context.Context, r repository.Repository, bb client.A
 			bene, err := getBeneficiary(ctx, r, uint(id), bb, fetchBBId, jobArgs)
 			if err != nil {
 				//MBI is appended inside file, not printed out to system logs
-				return fmt.Sprintf("Error retrieving BlueButton ID for cclfBeneficiary MBI %s", bene.MBI), fhircodes.IssueTypeCode_NOT_FOUND, err
+				return fmt.Sprintf("Error retrieving BlueButton ID for cclfBeneficiary MBI %s", bene.MBI), fhircodes.IssueTypeCode_PROCESSING, err
```
alex-dzeda, line 236: I think this one + the one below it may work better remaining as not_found, since they come directly from another system, and the code returned from BB in these scenarios is a 404. For the ones where the OperationOutcome we return is based upon a BCDA consumer's request (eg the parent job cancelled) then processing is definitely the most appropriate.

alex-dzeda:
Good work! Added two comments in, I think that the simplified structure is much better. I think that we can pass through the not_found for actual not_found scenarios, and the other codes are relevant for their situations (eg the ones where BCDA determines the error type rather than receiving from an external source.)

bcda/auth/middleware.go
laurenkrugen-navapbc: I updated some of the mismatched HTTP statuses and the response util strings while doing a larger set of changes that I decided to move into https://jira.cms.gov/browse/BCDA-7962 and I reverted some changes unintentionally, with this being one of them - thanks for catching!

bcdaworker/worker/worker.go
laurenkrugen-navapbc: agreed, I went back and forth on changing this one. Let's leave it as `not_found`; I'll revert this code.

alex-dzeda (approved):

PR #937, opened 2024-05-06
BCDA-7895: Support CCLF file imports from S3

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7895

## 🛠 Changes

- Update CCLF CLI to support S3 flags
- Encapsulate CCLF import functionality in CclfImporter interface, which can support Local and S3 imports
- Break out shared utilities into `parser.go`
- Break down optout s3_file_handler into smaller utility functions for re-use in CCLF importer

## ℹ️ Context for reviewers

This PR contains fairly significant changes to the CCLF import code to support importing zip archive files from S3. 

In order to enable this functionality, the CCLF code utilizes S3 utilities broken down from the optout module. Since the optout module is the only shared Go module we currently have, this seemed to make the most sense; however, I recognize it feels a bit odd to reuse/wrap the optout module in this way. Ideally the reused S3 functionality would be repackaged in a separate module that both the optout and CCLF code code use.

## ✅ Acceptance Validation

- [x] Unit testing
- [x] Smoke test in dev for local files

## 🔒 Security Implications


bcda/bcdacli/cli.go
```diff
@@ -359,10 +359,49 @@ func setUpApp() *cli.App {
 					Usage:       "Directory where CCLF files are located",
 					Destination: &filePath,
 				},
+				cli.StringFlag{
+					Name:        "filesource",
+					Usage:       "Source of files. Must be one of 'local', 's3'. Defaults to 'local'",
```
alex-dzeda, line 364: This is more philosophical, since we don't run most of these commands manually, but rather through say jenkins - the below doesn't appear to enforce only local or s3 (rather anything that's not s3 is treated as local), do we want an extra level of nuance for the cli commands? Eg requiring the filesource to be specified, directory must be specified, rather than letting cli fail if not provided?

alex-dzeda: (My personal thought is yes, but if we choose to, it can be a separate ticket)

kyeah: Since this is a new parameter, keeping a default of `local` allows us to transition smoothly (the Jenkins jobs don't specify filesource right now, so it's a bit of a chicken-or-egg situation.)

But I can see it would be helpful to add the enforcement in a followup ticket for both CCLF and opt out ingestion.

alex-dzeda (approved):
Sorry for the delay; looks good! Everything passes locally for me on this branch, too. 
Approved.
I'll make a ticket to review / update CLI flags in general as a follow-up to this (but not limited to the flags introduced here).

alex-dzeda:
Created https://jira.cms.gov/browse/BCDA-8130 to address cli flags, feel free to refine!

PR #943, opened 2024-05-20
BCDA-7945: Externalize SSAS calls

austincanada:
Use NewRelic to externalize calls to SSAS from BCDA.

## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7945

## 🛠 Changes

`auth/token` seems like the only endpoint with a SSAS call from BCDA, create an external segment for New Relic.

## ✅ Acceptance Validation

https://gorgon.nr-assets.net/image/3f52053c-2f08-474c-950a-00790f7565e0?type=table

## 🔒 Security Implications


laurenkrugen-navapbc:
Could be a possible bug with new relic, but cannot see SSAS under external calls in the 'new view'. However, can see it listed in the old view and also see an additional entity listed when viewing the auth/token transactions. Thanks for getting this setup! 🎉

PR #947, opened 2024-05-31
BCDA-8104 (1/2): Add lambda entrypoint for running opt out import

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8104

## 🛠 Changes

- Translates some parameter store utilities from DPC into a `bcdaaws` package
- Adds a new `bcda/lambda` directory for housing lambda entrypoints
- Added lambda entrypoint for opt out import
- Updated database connection initializer to support loading values from Parameter Store

## ℹ️ Context for reviewers

This PR allows us to run the opt out import functionality as a lambda. A few points to note related to this:

### 1. Loading database configuration

The database connection usually requires DATABASE_URL and QUEUE_DATABASE_URL to be set as environment variables before the application is started. For lambda applications, secret values need to be pulled at runtime within the application.

Originally, I wanted to avoid using the `connection.go` functionality at all. However, it appears that the init function always runs if the `database` package is imported at any point in the import chain. Because of this, the lambda entrypoint was always looking for those env vars and then crashing.

Instead, I've added a fallback to the initializer to check parameter store if the ENV var is set (which it will be, for lambda.)

### 2. Lambda environment variables

Lambda is configured differently from our EC2 applications. In particular, we are using shared infrastructure modules and only have these environment variables available to us:

https://github.com/CMSgov/ab2d-bcda-dpc-platform/blob/f9fa64bde4bf3399d8cbfdc75e0eff0f23697b5a/terraform/services/opt-out-import/main.tf#L53-L57

When retrieving the environment, we need to use ENV instead of DEPLOYMENT_TARGET.

### 3. Command line arguments

Lambdas receive string-y events as input. With the shared opt-out import infrastructure, they will be receiving S3 event notifications through SNS for any new files dropped into the bucket directory.

Since we do not have the usual CLI arguments for the BFD assume role arn and BFD import bucket, those values are pulled from new places:

- BFD assume role ARN: Already available in parameter store, so pull it from there.
- BFD import bucket: Can be parsed from the S3 event notification when it comes in.

## ✅ Acceptance Validation

- Unit Testing

## 🔒 Security Implications


optout/s3_file_handler.go
```diff
@@ -178,7 +179,7 @@ func (handler *S3FileHandler) createSession() (*session.Session, error) {
 		config.Endpoint = &handler.Endpoint
 	}
 
-	if handler.AssumeRoleArn != "" {
+	if handler.AssumeRoleArn != "" && os.Getenv("LOCAL_STACK_ENDPOINT") == "" {
```
kyeah, line 182: Needed this hack to get unit tests to work correctly. Localstack was returning "Expired credentials" errors when attempting to assume the fake BFD role, but those parameters are required to exist in order to run the import code.

Based on documentation, this doesn't seem to be the expected behavior (we should be able run assume-role commands in Localstack without any permission errors) but this was a quick workaround to make tests work.

bcda/lambda/optout/main.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/CMSgov/bcda-app/bcda/models/postgres"
+	"github.com/CMSgov/bcda-app/bcda/suppression"
+	"github.com/CMSgov/bcda-app/bcda/utils"
+	"github.com/CMSgov/bcda-app/optout"
+
+	"github.com/CMSgov/bcda-app/conf"
+)
+
+func main() {
+	if os.Getenv("LOCAL_STACK_ENDPOINT") != "" {
```
alex-dzeda, line 25: Reference number 1

bcda/lambda/optout/main.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	logger.Info("No ObjectCreated:Put events found, skipping safely.")
+	return "", nil
+}
+
+func loadBfdS3Params() (string, error) {
+	env := conf.GetEnv("ENV")
+
+	bcdaSession, err := bcdaaws.NewSession("", os.Getenv("LOCAL_STACK_ENDPOINT"))
```
alex-dzeda, line 77: reference number 2

bcda/aws/session.go
```diff
@@ -0,0 +1,43 @@
...
+// Returns a new AWS session using the given roleArn
+func NewSession(roleArn, endpoint string) (*session.Session, error) {
+	sess := session.Must(session.NewSession())
+	var err error
+
+	config := aws.Config{
+		Region: aws.String(s3Region),
+	}
+
+	if endpoint != "" {
```
alex-dzeda, line 24: reference 3

alex-dzeda:
Looks good overall! I do have one ask, is it possible to add some descriptive comments in at least one of the 3 spots I put "reference" in order to document the use of LOCAL_STACK_ENDPOINT in the non-test files? I can tell that it's implied this environment variable will not be populated when actually deployed, but having explicit documentation of that would probably help others viewing the code.

bcda/aws/parsers.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"encoding/json"
+
+	"github.com/aws/aws-lambda-go/events"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+// TODO: Iterate over records
```
austincanada, line 11: Do we still need this todo?

austincanada (commented):
Nothing jumped out at me, but probably would like a second set of eyes due to the size of the work.

bcda/lambda/optout/main.go
kyeah: Added it here since it's the top of the file, so most likely to be seen by developers when perusing code.

alex-dzeda (approved):
Thanks for adding the comment!

PR #948, opened 2024-05-31
BCDA-8146: Make file server content-encoding aware

alex-dzeda:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8146

## 🛠 Changes

Added magic number check for files, to determine if it is gzip encoded (or not). 

## ℹ️ Context for reviewers

As part of BCDA's gzip epic, we are aiming to simultaneously decrease costs on EFS, while also increasing the overall throughput of files. When we switch over from primarily storing content uncompressed to primarily storing data gzip-encoded, there will be at least some period of time where both encodings are present. The BCDA application will now check the magic number (https://en.wikipedia.org/wiki/Magic_number_(programming)) to determine if content is encoded or not, and will handle the data appropriately (decompressing the file if an Accept-Encoding header of gzip is not present). 

There is a binary file, shared_files/gzip_feature_test/corrupt_gz_file.ndjson, that is not able to be opened with a normal text editor. It's advised to open files of this type in a hex editor for manual inspection. In this case, the file has the magic number of a gzip file, but the subsequent data is not compliant. 

## ✅ Acceptance Validation

Added unit tests to handle the various cases. 

## 🔒 Security Implications


bcda/api/v1/api.go
```diff
@@ -293,20 +301,75 @@ func ServeData(w http.ResponseWriter, r *http.Request) {
...
 		}
 	}
-
+	w.Header().Set(constants.ContentType, "application/fhir+ndjson")
 	if useGZIP {
 		w.Header().Set("Content-Encoding", "gzip")
 		gz := gzip.NewWriter(w)
 		defer gz.Close()
 
 		gzw := gzipResponseWriter{Writer: gz, ResponseWriter: w}
```
austincanada, line 310: Not sure how heavy of a process this is, did we just want to add it to the encoded-else check?

alex-dzeda: I was thinking that the scope of isGzipEncoded would be just to determine if the magic number of a file (if it can be read) implies the file is gzip-encoded, and then the actual opening / writing of the file could be handled within the main function utilizing it. 
Only one of the four paths actually opens the file again (rather than using a writer and the http.serveFile method), so I was thinking that may be an acceptable tradeoff rather than having to allocate memory for a path that will be chosen in a small minority of situations (data encoded + no gzip header, which right now is <5% of cases).

austincanada: Ah okay, thanks for the explanation!

austincanada (approved):

PR #950, opened 2024-06-05
PLT-366 Add opt-out-import workflows

gsf:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-366

## 🛠 Changes

Added workflows for the optout lambda for BCDA.

## ℹ️ Context for reviewers

These files have been copied over from the dpc-app repo. I've made some adjustments for BCDA but more will likely still need to be made to tailor to specifics in the BCDA function.

Related ticket at https://jira.cms.gov/browse/BCDA-7898.

## ✅ Acceptance Validation

This PR will need to be merged before we can test these workflows, so we should get them in pretty good shape then merge, and follow up with another PR to fix any further issues.

## 🔒 Security Implications

None.

### DIFF
--- a/.github/workflows/opt-out-import-dev-deploy.yml
+++ b/.github/workflows/opt-out-import-dev-deploy.yml
@@ -1,0 +1,38 @@
+name: opt-out-import dev deploy
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - optout/**
+      - .github/workflows/opt-out-import-dev-deploy.yml
+  workflow_dispatch:
+
+jobs:
+  test:
+    permissions:
+      contents: read
+      id-token: write
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./optout
+    environment: dev
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+      - name: Build opt-out-import zip file
+        run: |
+          go build -o bootstrap main.go parsers.go utils.go models.go db.go
+          zip function.zip bootstrap
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: ${{ vars.AWS_REGION }}
+          role-to-assume: arn:aws:iam::${{ secrets.ACCOUNT_ID }}:role/delegatedadmin/developer/bcda-dev-github-actions
+      - name: Upload and reload
+        run: |
+          aws s3 cp --no-progress function.zip \
+            s3://bcda-dev-opt-out-import-function/function-${{ github.sha }}.zip
+          aws lambda update-function-code --function-name bcda-dev-opt-out-import \
+            --s3-bucket bcda-dev-opt-out-import-function --s3-key function-${{ github.sha }}.zip
--- a/.github/workflows/opt-out-import-test-integration.yml
+++ b/.github/workflows/opt-out-import-test-integration.yml
@@ -1,0 +1,51 @@
+name: opt-out-import test integration
+
+on:
+  pull_request:
+    paths:
+      - .github/workflows/opt-out-import-test-integration.yml
+      - .github/workflows/opt-out-import-test-deploy.yml
+      - optout/**
+  workflow_dispatch:
+
+# Ensure we have only one integration test running at a time
+concurrency:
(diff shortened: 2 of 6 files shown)


.github/workflows/opt-out-import-dev-deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+    defaults:
+      run:
+        working-directory: ./optout
+    environment: dev
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+      - name: Build opt-out-import zip file
+        run: |
+          go build -o bootstrap main.go parsers.go utils.go models.go db.go
```
gsf, line 27: These filenames will need to be updated for BCDA optout.

.github/workflows/opt-out-import-unit.yml
```diff
@@ -0,0 +1,20 @@
...
+  test:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./optout
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+      - name: Run unit tests for opt-out-import lambda
+        run: go test
```
gsf, line 20: This will need to be updated to run the full unit test.

.github/workflows/opt-out-import-dev-deploy.yml
```diff
@@ -0,0 +1,38 @@
+name: opt-out-import dev deploy
```
gsf, line 1: We can rename these workflows from "opt-out-import" to just "optout" if that makes more sense for BCDA.

kyeah: This is fine, it's more explicit and will match the assumed CCLF format as well (`cclf-import`)

.github/workflows/opt-out-import-unit.yml
kyeah: The lambda is encapsulated as part of the larger application (for now), so I don't think we need this section.

.github/workflows/opt-out-import-dev-deploy.yml
kyeah: I added a make command that should build it, see here:

https://github.com/CMSgov/bcda-app/pull/947

From bcda subdirectory: `go build -o bin/opt-out-import ./lambda/optout/main.go`

.github/workflows/opt-out-import-test-integration.yml
```diff
@@ -0,0 +1,51 @@
...
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: ${{ vars.AWS_REGION }}
+          # Note that we use the BFD role with access to the bucket
+          role-to-assume: arn:aws:iam::${{ secrets.BFD_ACCOUNT_ID }}:role/bfd-test-eft-bcda-bucket-role
+          role-chaining: true
+          role-skip-session-tagging: true
+      - name: Upload test file to the BFD bucket to trigger lambda function via SNS message
+        run: |
+          aws s3 cp --no-progress synthetic_test_data/T.NGD.DPC.RSP.D240123.T1122001.IN \
```
kyeah, line 47: will need to update this path to match. I think we have two test files, one example here:

https://github.com/CMSgov/bcda-app/blob/main/shared_files/synthetic1800MedicareFiles/test/T%23EFT.ON.ACO.NGD1800.DPRF.D181120.T1000009

kyeah:
@gsf I went through just to flag areas where I know we need changes based on the codebase and #947. LMK if you'll be updating the PR or how you'd like to coordinate.

.github/workflows/opt-out-import-unit.yml
gsf: Okay so the unit test workflow can be dropped for now?

gsf:
@kyeah Thanks for the review! If there's enough to work with here for @austincanada to take over the PR I'd be fine with that.

.github/workflows/opt-out-import-unit.yml
kyeah: yep

gsf:
Continuing in #957.

PR #955, opened 2024-06-11
BCDA-7980: Implement gzip compression for files stored on EFS

alex-dzeda:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7980

## 🛠 Changes

Instead of attempting to move files, we write a gzip-encoded file to EFS and delete the temporary files after each job key

## ℹ️ Context for reviewers

This is the final ticket in the gzip epic, yay! 
Adds a new environment variable, COMPRESSION_LEVEL, that can be set between 1-9
Files are stored as gzip-encoded content, previous PRs ensure that the /data/ endpoint are content-encoding aware, and can return un-encoded content if requested, while ensuring significant gains for the 90+% of requests that do request gzip-encoded content.

## ✅ Acceptance Validation

Smoke tests + unit tests pass.
I've also performed manual validation utilizing the dev server. Throughput is as expected.

## 🔒 Security Implications

### DIFF
--- a/bcdaworker/worker/worker.go
+++ b/bcdaworker/worker/worker.go
@@ -179,11 +179,12 @@ func compressFiles(ctx context.Context, tempDir string, stagingDir string) error
 	gzipLevel, err := strconv.Atoi(os.Getenv("COMPRESSION_LEVEL"))
 	if err != nil || gzipLevel < 1 || gzipLevel > 9 { //levels 1-9 supported by BCDA.
 		gzipLevel = gzip.DefaultCompression
-		logger.Warnf("COMPRESSION_LEVEL not set to appropriate value.")
+		logger.Warnf("COMPRESSION_LEVEL not set to appropriate value; using default.")
 	}
 	for _, f := range files {
 		oldPath := fmt.Sprintf("%s/%s", tempDir, f.Name())
 		newPath := fmt.Sprintf("%s/%s", stagingDir, f.Name())
+		//Anonymous function to ensure defer statements run
 		err := func() error {
 			inputFile, err := os.Open(filepath.Clean(oldPath))
 			if err != nil {
--- a/bcdaworker/worker/worker.go
+++ b/bcdaworker/worker/worker.go
@@ -2,11 +2,13 @@ package worker
 
 import (
 	"bufio"
+	"compress/gzip"
 	"context"
 	"database/sql"
 	"encoding/json"
 	goerrors "errors"
 	"fmt"
+	"io"
 	"os"
 	"strconv"
 
@@ -144,7 +146,7 @@ func (w *worker) ProcessJob(ctx context.Context, job models.Job, jobArgs models.
 		}
 	}
 	//move the files over
-	err = moveFiles(tempJobPath, stagingPath)
+	err = compressFiles(tempJobPath, stagingPath)
 	if err != nil {
 		logger.Error(err)
 	}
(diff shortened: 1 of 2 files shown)


bcdaworker/worker/worker.go
```diff
@@ -164,20 +166,41 @@ func (w *worker) ProcessJob(ctx context.Context, job models.Job, jobArgs models.
...
+func compressFiles(tempDir string, stagingDir string) error {
 	// Open the input file
 	files, err := os.ReadDir(tempDir)
 	if err != nil {
 		err = errors.Wrap(err, "Error reading from the staging directory for files for Job")
 		return err
 	}
+	gzipLevel, err := strconv.Atoi(os.Getenv("COMPRESSION_LEVEL"))
+	if err != nil || gzipLevel < 1 || gzipLevel > 9 { //levels 1-9 supported by BCDA.
+		gzipLevel = gzip.DefaultCompression
```
kyeah, line 178: Can we log a warning here if there's an error / the COMPRESSION_LEVEL is misconfigured?

bcdaworker/worker/worker.go
```diff
@@ -164,20 +166,41 @@ func (w *worker) ProcessJob(ctx context.Context, job models.Job, jobArgs models.
...
+	}
 	for _, f := range files {
 		oldPath := fmt.Sprintf("%s/%s", tempDir, f.Name())
 		newPath := fmt.Sprintf("%s/%s", stagingDir, f.Name())
-		err := os.Rename(oldPath, newPath) //#nosec G304
+		inputFile, err := os.Open(oldPath) //#nosec G304
```
kyeah, line 183: I think if you use `filepath.Clean` you can avoid the G304 flags

bcdaworker/worker/worker.go
```diff
@@ -164,20 +166,41 @@ func (w *worker) ProcessJob(ctx context.Context, job models.Job, jobArgs models.
...
+		if err != nil {
+			return err
+		}
+		defer inputFile.Close() //#nosec G307
```
kyeah, line 187: I know this is essentially safe but would prefer to avoid `#nosec` markers if possible from an auditing perspective. Wonder if we want to create a helper utility for these scenarios?

```golang
// usage: defer CloseOrLogError(inputFile)
func CloseOrLogError(f *os.File) {
	if f == nil {
		return
	}
	if err := f.Close(); err != nil {
		logger.Warnf("Error closing file: %v", err)
	}
}
```

kyeah (commented):
Looks good from a functional perspective! Added a few comments about logging and nosec flags.

bcdaworker/worker/worker.go
```diff
@@ -164,20 +166,41 @@ func (w *worker) ProcessJob(ctx context.Context, job models.Job, jobArgs models.
...
 
-func moveFiles(tempDir string, stagingDir string) error {
+func compressFiles(tempDir string, stagingDir string) error {
 	// Open the input file
 	files, err := os.ReadDir(tempDir)
 	if err != nil {
 		err = errors.Wrap(err, "Error reading from the staging directory for files for Job")
 		return err
 	}
+	gzipLevel, err := strconv.Atoi(os.Getenv("COMPRESSION_LEVEL"))
```
austincanada, line 176: Did we want to have a error thrown for when the compression level is < 1 or > 9 for feedback?

austincanada: Oop, sorry for the double-dip.

bcdaworker/worker/worker.go
alex-dzeda: Good idea; I modified this function slightly, to utilize the snippet above (and passing a field logger to enable the logging).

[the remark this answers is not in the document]
alex-dzeda: In this case, there's a max of two files that could be open at once, but... I think I have a simple way to address this, by putting the core of the loop within a function, adding a layer of indirection that accomplishes this without explicitly closing the files (rather relying on the defers)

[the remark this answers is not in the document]
austincanada: Looks good! Didn't know Go had IIFE's.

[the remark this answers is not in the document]
alex-dzeda: Resolved this by moving that block into an anonymous function

alex-dzeda:
@kyeah and @austincanada made updates utilizing y'alls feedback.

austincanada (approved):
LGTM, not sure if you want to wait on Kevin's as well

bcdaworker/worker/worker.go
```diff
@@ -164,25 +168,64 @@ func (w *worker) ProcessJob(ctx context.Context, job models.Job, jobArgs models.
...
+	if err != nil || gzipLevel < 1 || gzipLevel > 9 { //levels 1-9 supported by BCDA.
+		gzipLevel = gzip.DefaultCompression
+		logger.Warnf("COMPRESSION_LEVEL not set to appropriate value.")
```
kyeah, line 182:
```suggestion
		logger.Warnf("COMPRESSION_LEVEL not set to appropriate value; using default.")
```

bcdaworker/worker/worker.go
```diff
@@ -164,25 +168,64 @@ func (w *worker) ProcessJob(ctx context.Context, job models.Job, jobArgs models.
...
+	}
 	for _, f := range files {
 		oldPath := fmt.Sprintf("%s/%s", tempDir, f.Name())
 		newPath := fmt.Sprintf("%s/%s", stagingDir, f.Name())
-		err := os.Rename(oldPath, newPath) //#nosec G304
+		err := func() error {
```
kyeah, line 187: Can we add a comment here about why we have an anon func? Doesn't seem very intuitive at first glance 😄

kyeah (approved):
LGTM overall, thank you!

bcdaworker/worker/worker.go
alex-dzeda: Updated!

bcdaworker/worker/worker.go
alex-dzeda: Added a comment, agree with you there. Seemed like the least-intensive way to ensure the defers run per-loop 😆 .

PR #959, opened 2024-06-21
Fix lambda handler paths

kyeah:
## 🎫 Ticket

N/A

## 🛠 Changes

- Remove `/bin/` directories from function zips using `zip -j`
- Rename files to `bootstrap`

## ℹ️ Context

Was looking at the lambda logs and saw some errors with trying to find the entrypoint. After reading the AWS documentation, I realized that it expects the zip file to have exactly one binary, which is specifically named `bootstrap` 😅

This was an oversight on my part — I thought if the handler was specified as `cclf-import` in the lambda config, it would look for that binary filename instead.

## 🧪 Validation

Verified in logs (doesn't complete successfully, we have another network issue to address..)

CleanShot 2024-06-21 at 13 35 58@2x

### DIFF
--- a/.github/workflows/cclf-import-dev-deploy.yml
+++ b/.github/workflows/cclf-import-dev-deploy.yml
@@ -23,6 +23,8 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
       - name: Build cclf-import zip file
+        env:
+          CGO_ENABLED: 0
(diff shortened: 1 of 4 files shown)


laurenkrugen-navapbc (approved):

austincanada (approved):
LGTM, thanks Kevin for the fix!

PR #960, opened 2024-06-24
BCDA-8208: Fix S3 path parsing

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8208

## 🛠 Changes

- Fix S3 path parsing

## ℹ️ Context

I observed failures with the lambdas attempting to list files in S3 within the bfdeft01 prefix. This is not the correct prefix — the team-specific IAM roles can only access specific subfolders (i.e. bfdeft01/bcda, bfdeft01/dpc, etc.) so the lambdas _should_ be trying to list files within those subfolders.

When parsing the S3 prefix from the SQS event, it needs to split from the last separator instead of the first separator.

## 🧪 Validation

Deployed to Dev and verified successful import by uploading a file (I basically did this part of the integration test manually: https://github.com/CMSgov/bcda-app/blob/fabde15cb4f460d7398dcdfcf61205840258afcd/.github/workflows/opt-out-import-test-integration.yml#L47-L48 but fixed the filename since it's using the DPC filename)

**Successful lambda logs**

### DIFF
--- a/bcda/aws/parsers.go
+++ b/bcda/aws/parsers.go
@@ -2,6 +2,8 @@ package bcdaaws
 
 import (
 	"encoding/json"
+	"fmt"
+	"strings"
 
 	"github.com/aws/aws-lambda-go/events"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


austincanada (approved):
LGTM

PR #962, opened 2024-06-25
BCDA-8197: Prevent duplicate job keys from queue re-processing

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8197

## 🛠 Changes

Core Changes:
- Add job_keys.que_job_id column and (job_id, que_job_id) index
- Avoid re-processing a que job if we've already created job keys for the job

Additional Changes:
- Create multiple job keys in an single database operation instead of multiple sequential operations
- Return error if we fail to compress/move files from temp --> staging
- Return error if we fail to create job keys

## ℹ️ Context

The core change in this PR is the addition of the job_keys.que_job_id column, which is used to avoid re-processing a queue job if something happens between the job keys being created and the queue job being fully processed and deleted.

The additional changes are minor tweaks to avoid any of the potential errors that may have occurred previously:

- One job key being created without the error
- File retrieval errors due to job keys being created without the files being moved to staging/payload directory
- Queue jobs being processed and deleted without the related job keys being created
- Queue jobs being processed and deleted without the parent job being marked completed

## 🧪 Validation

- [x] Existing unit tests
- [x] New unit tests


bcdaworker/worker/worker.go
```diff
@@ -151,15 +160,20 @@ func (w *worker) ProcessJob(ctx context.Context, job models.Job, jobArgs models.
...
 	if err != nil {
 		logger.Error(err)
+		return err
 	}
 
 	// Not critical since we use the job_keys count as the authoritative list of completed jobs.
 	// CompletedJobCount is purely information and can be off.
+	// TODO: Where does this make sense to live?
+	//       Should this just sync the # of job keys into the field since it's the authoritative source?
+	//       Should we remove this field since it's misleading? Is it used anywhere (and is that a problem?)
```
kyeah, line 176: This is an open question that should be answered and resolved before merging. I was not sure if que job re-processing should still attempt to...

1. Increment the completed job count (It seems possible this could cause the completed_job_count to be incremented multiple times by the same que job ID)
2. Update the completed job count based on the current job key count (this seems most reliable if we still want this field)
3. Not do this (could cause completed_job_count to be undercounted if a que job is reprocessed)
4. Remove the completed_job_count column (seems best if this column is not accurate and not being used)

laurenkrugen-navapbc: I think number four sounds like the best option. Based off a quick search through the repo, that count isn't used to verify anything and "it can be off". Because it's inaccurate and we don't use it, getting rid of it would clean up the database and the code overall and make things more understandable.

kyeah: Let's plan for that. Since the column will be removed, I will remove this logic and then we can have a followup ticket to remove the column in the next sprint / after the next deployment (to avoid removing the column before the application is updated.)

bcdaworker/queueing/manager/que.go
```diff
@@ -119,7 +119,7 @@ func (q *queue) processJob(job *que.Job) error {
 	ctx, _ = log.SetCtxLogger(ctx, "job_id", jobArgs.ID)
 	ctx, logger := log.SetCtxLogger(ctx, "transaction_id", jobArgs.TransactionID)
 
-	exportJob, err := q.worker.ValidateJob(ctx, jobArgs)
+	exportJob, err := q.worker.ValidateJob(ctx, job.ID, jobArgs)
```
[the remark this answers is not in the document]
kyeah: yeah it was pretty confusing for me too, I'll update it for clarity while we're here.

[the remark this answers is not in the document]
laurenkrugen-navapbc: Thank you!

laurenkrugen-navapbc:
This looks good to me! I am doing some QA on this really quick to verify that:

- a job that does not have a completed-type status with job keys that have que_ids will **not** attempt to recreate those keys again. My plan to recreate:

1. manually creating a job in the jobs table 
2. create job_keys with que_ids
3. manually queueing the job keys in the que_jobs table
4. running the app locally, verifying that job_keys with existing que_ids do not get processed again 
5. verify that the job count matches the number of job keys (maybe we should rename job count to be job keys while we are at it?)

bcdaworker/worker/worker.go
kyeah: More changes needed than I thought — splitting into a separate PR here: https://github.com/CMSgov/bcda-app/pull/964

laurenkrugen-navapbc:
### Validation
Alrighty, did some validation on my end running it locally:

- I manually inserted a job with two job keys, one of which had a que_job_id set in the entry
- I manually inserted a queue job (so I could set the que_job_id appropriately)
- After letting the worker pick up the job, I confirmed that it not only removed the job_key from the que_jobs table, but it also did not process it again.

success output:
`{"application":"worker","environment":"local","file":"/go/src/github.com/CMSgov/bcda-app/bcdaworker/queueing/manager/que.go:149","func":"github.com/CMSgov/bcda-app/bcdaworker/queueing/manager.(*queue).processJob","job_id":1,"level":"warning","msg":"Queue job 2 already processed for job 1. Checking completion status and removing queuejob from que.","time":"2024-06-27T20:27:36.226428887Z","transaction_id":"b4b64b71-c30e-4588-ae7f-6a44d69ee95b","version":"latest"}`

I have one final request to do a slight rename in this log message above; can we update?

Previous: `Queue job 2 already processed for job 1. Checking completion status and removing queuejob from que.`
Suggested: `Queue job (que_jobs.job_id) 2 already processed for job_key.job_id 1. Checking completion status and removing queuejob from que.`

laurenkrugen-navapbc (approved):

PR #963, opened 2024-06-25
BCDA-7900: Add integration test validation for opt-out import

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7900

## 🛠 Changes

Integration testing for opt-out lambda has had additional steps added:
- queries the database to see if the opt out file was ingested
- queries the database to see if suppression entries were created from the ingested file

## ℹ️ Context

Integration testing is required to ensure that future changes will not break existing functionality and that the opt-out lambda is working as expected.

## 🧪 Validation

Ran using happy and unhappy paths. Validated that sensitive data is not printed to stderr or stdout in the event of an issue.


laurenkrugen-navapbc:
Do we want to add another layer of validation to the database: do the number of MBIs in the file for opt-out and opt-in match?

Or do we feel like this covers our bases for now?

austincanada (approved):
LGTM

kyeah:
> Do we want to add another layer of validation to the database: do the number of MBIs in the file for opt-out and opt-in match?

I think it would be good to validate this (or really — validate as much as possible that the data in the DB matches the rows in the file. I know we're working with shell script though so it's a bit difficult to write complex logic 🙃)

laurenkrugen-navapbc:
> > Do we want to add another layer of validation to the database: do the number of MBIs in the file for opt-out and opt-in match?
> 
> I think it would be good to validate this (or really — validate as much as possible that the data in the DB matches the rows in the file. I know we're working with shell script though so it's a bit difficult to write complex logic 🙃)

Is this something that we would want to get into this script before merging or at a later date?

.github/workflows/opt-out-import-test-integration.yml
```diff
@@ -43,9 +46,46 @@ jobs:
...
+          echo "FILENAME=T\#EFT.ON.ACO.NGD1800.DPRF.D$(date +'%y%m%d').T$(date +'%H%M%S')" >> "$GITHUB_OUTPUT"
           aws s3 cp --no-progress ../shared_files/synthetic1800MedicareFiles/test/T\#EFT.ON.ACO.NGD1800.DPRF.D181120.T1000009 \
-            s3://bfd-test-eft/bfdeft01/bcda/in/T.NGD.DPC.RSP.D$(date +'%y%m%d').T$(date +'%H%M%S')1.IN
-
-  # TODO Check bucket for response file
-  # TODO Run another job to check database for update
+            s3://bfd-test-eft/bfdeft01/bcda/in/$FILENAME
+  verify:
+    needs: trigger
+    runs-on: self-hosted
```
gsf, line 56: Add the following here to continue to use node16 instead of node20 for the time being on the runner:

```
    env:
      ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: "true"
```

laurenkrugen-navapbc:
functionality is working as expected, but the dev env lambda is occasionally getting to the test file before the test lambda does, which is resulting in integration test failures. Success run: https://github.com/CMSgov/bcda-app/actions/runs/9879612164/job/27286247165

PR #964, opened 2024-06-27
BCDA-8215: Remove completed job count application logic

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8215

## 🛠 Changes

- Remove all application logic referencing completed job count

## ℹ️ Context

This column, by it's own commented definition, is not a reliable indicator of the number of actually completed sub-jobs for an export job. It appears that its only use is in the in-progress job status header to indicate roughly how far along the job processing is.

Instead of attempting to maintain this field, we should be relying on the source of truth (the number of completed job keys). To avoid confusion and to reduce code complexity, we are removing this column.

Before removing the column from the database, we must stop reading it from the application.

## 🧪 Validation

- Unit testing


laurenkrugen-navapbc (approved):
had one question, but other than that, LGTM. For the addition of `CountUniq`, were there possibility of duplicates being reported before?


Validation:

- cleared docker locally and ran process to load fixtures and seed database, with no errors
- ran all tests (unit, postman, etc) with no errors
- did a command search for instances of `completedjobcount` and `completed_job_count`, found no references except for one string in ALR that does not need to be changed.

kyeah:
> For the addition of CountUniq, were there possibility of duplicates being reported before?

Yes - previously it was relying on completed_job_count so it was potentially inaccurate (although didn't really matter much assuming folks weren't using that response header for critical logic!)

PR #853, opened 2023-06-22
BCDA-5355: Disable ALR export endpoints

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-5355

## 🛠 Changes

- Disable ALR endpoints in deployed environments

## ℹ️ Context for reviewers

The ALR export endpoints were never completed and officially released, and they should be disabled to prevent confusion.

Since the default is false, I didn't add env vars for ENABLE_ALR_ENDPOINTS in our deployed environments. However, if folks feel we should have those, I can add that separately.

## ✅ Acceptance Validation

Verified locally:

```sh
bcda-app: curl -X GET 'http://localhost:3000/api/v2/alr/$export' \                                                                                                                  
                        -H "accept: application/fhir+json" \
                        -H "Prefer: respond-async" \
                        -H "Authorization: Bearer $token"
404 page not found

bcda-app: curl -X GET 'http://localhost:3000/api/v1/alr/$export' \
                        -H "accept: application/fhir+json" \ 
                        -H "Prefer: respond-async" \
                        -H "Authorization: Bearer $token"
404 page not found
```

## 🔒 Security Implications

- [ ] This PR adds a new software dependency or dependencies.
- [ ] This PR modifies or invalidates one or more of our security controls.
- [ ] This PR stores or transmits data that was not stored or transmitted before.
- [ ] This PR requires additional review of its security implications for other reasons.

PR #966, opened 2024-07-01
BCDA-8227: Remove usage of completed_job_count in ALR jobs

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8227

## 🛠 Changes

- Replace completed_job_count with a count of unique job keys (based on job_key.que_job_id)
- Avoid re-processing ALR job if que job was already processed

## ℹ️ Context

As part of #964, I discovered that the completed_job_count was still being used as part of ALR jobs. This applies the same fix from the "normal" export jobs in #962.

Note that many different job keys can be produced by an ALR job. To simplify the determination of whether a que job was processed, I added a new "getUniqueJobKeyCount" method.

For some historical context, the ALR endpoints are not enabled in deployed environments - see #853 for more details. To avoid making a decision on whether to remove the code entirely, I am applying these changes for now to remove completed_job_count.

## 🧪 Validation

Unit Testing

### DIFF
--- a/bcdaworker/worker/alr.go
+++ b/bcdaworker/worker/alr.go
@@ -59,7 +59,7 @@ func NewAlrWorker(db *sql.DB) AlrWorker {
 }
 
 func goWriterV1(ctx context.Context, a *AlrWorker, c chan *alr.AlrFhirBulk, fileMap map[string]*os.File,
-	result chan error, resourceTypes []string, id uint) {
+	result chan error, resourceTypes []string, id uint, queJobID int64) {
 
 	writerPool := make([]*bufio.Writer, len(fileMap))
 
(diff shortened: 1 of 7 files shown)


bcdaworker/worker/alr.go
```diff
@@ -200,10 +207,10 @@ func (a *AlrWorker) ProcessAlrJob(
 		return err
 	}
 
-	// If we did not have an ALR data to write, we'll write a specific file name that indicates that the
+	// If we did not have any ALR data to write, we'll write a specific file name that indicates that
```
kyeah, line 210: Note that this seems a little too early for the "blank filename" job key to be created, since there is other logic further down in the code that is still run after this job key is created. Wasn't 100% sure if this makes sense to move, decided to keep it as is unless people feel confident about moving this down.

laurenkrugen-navapbc: yeah, this is a little odd here. Not sure I fully understand why we are continuing job processing if there's no ALR data + the creation of a key here and then more keys later on. If the tests are passing, I am OK to leave this where it is at for now.

When/if this code is reintroduced and/or que-go is replaced, we can do a deeper dive on this?

laurenkrugen-navapbc (approved):
Approved. Pulled branch locally and ran full test suite + app setup with changes for validation.

PR #967, opened 2024-07-03
Increase export job waiting period for smoke tests

kyeah:
## 🎫 Ticket

N/A

## 🛠 Changes

- Update maxRetries from 10 --> 20 everywhere
- Autoformat postman collection JSON files (oops sorry..)

## ℹ️ Context

After https://github.com/CMSgov/bcda-app/releases/tag/r226 was released, it looks like jobs are taking a _little_ bit longer to run in production. It's not clear if it's because of something in the release or tangential.

This updates maxRetries everywhere (previously, a handful of maxRetries were already updated to 20 due to intermittent issues.)

## 🧪 Validation

Successful smoke test on production with r226 after a series of intermittent failures.

### DIFF
--- a/test/postman_test/BCDA_PAC_Postman_Smoke_Tests.postman_collection.json
+++ b/test/postman_test/BCDA_PAC_Postman_Smoke_Tests.postman_collection.json
@@ -1,1502 +1,1484 @@
 {
-	"info": {
-		"_postman_id": "9cc2977d-ced3-4492-80d0-cee9169ec057",
-		"name": "BCDA PAC Postman Smoke Tests",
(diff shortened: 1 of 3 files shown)


laurenkrugen-navapbc (approved):
LGTM! Did a quick ctrl+f locally against the maxretries and verified we updated all of the instances where it was set to 10.

alex-dzeda:
@laurenkrugen-navapbc @kyeah @austincanada fyi, intermittent failures can happen when a customer is running jobs at the same time as a release. The smoke test jobs are not prioritized, so it's a bit of a round robin if a lot of jobs are in the queue during a release.

PR #968, opened 2024-07-04
BCDA-8234: Import a single CCLF file zip when triggered by BFD S3 notifications

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8234

## 🛠 Changes

- Update lambda entrypoint to only import the CCLF file that caused the lambda instance to start
- Consolidate "valid2" test zips into a single zip to match production format (CCLF0 and CCLF8 files come in a single zip, not split across multiple zips)

## ℹ️ Context

To avoid race conditions, we would like to ensure that each lambda instance imports only the CCLF file that started the lambda instance via S3 notifications. This ensures that each lambda instance has exactly one file responsibility, and there are no conflicts between lambdas attempting to import the same file.

Note that this technically allows us to still process entire directories at once by manually sending events in the AWS Console with just the directory name, instead of the path for a single file.

## 🧪 Validation

- Updated unit tests
- Integration tests?


austincanada (approved):
LGTM

PR #970, opened 2024-07-05
Remove stdout print statements

kyeah:
## 🎫 Ticket

N/A

## 🛠 Changes

Remove print statements

## ℹ️ Context

These were causing the error alarm to go off. We used to duplicate messages to stdout and to the formatted JSON Logger since only stdout would be printed in Jenkins logs. Since we are running in lambda now, we can remove that.

## 🧪 Validation

Unit Testing


austincanada (approved):
LGTM

PR #972, opened 2024-07-09
BCDA-8170: Remove NFS-related application code

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8170

## 🛠 Changes

- Removes nfs/hcc_crosswalk.tsv references
- Removes delete-dir-contents which was only used for deleting NFS contents (I think?)

## ℹ️ Context

We no longer need this code since the related NFS server is deprecated. The build and cleanup workflows that utilize this code will be removed first.

## 🧪 Validation

Unit Testing


laurenkrugen-navapbc (approved):
LGTM - ran tests locally and also ran smoke tests against the branch.

PR #978, opened 2024-07-22
BCDA-7900: Add verification to CCLF import integration test

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7900

## 🛠 Changes

- Copied opt-out import integration test to CCLF workflow
- Made some edits to support CCLF imports

## ℹ️ Context

After getting the opt-out import integration tests merged, realized that it wasn't applied to the CCLF import workflow. Now that we are relying on the CCLF lambda, we want to ensure future changes do not break expected functionality.

## 🧪 Validation

Updated integration test works successfully.

### DIFF
--- a/.github/workflows/cclf-import-test-integration.yml
+++ b/.github/workflows/cclf-import-test-integration.yml
@@ -47,9 +47,9 @@ jobs:
       - name: Upload test file to the BFD bucket to trigger lambda function via SNS message
         id: createfile
         run: |
-          fname=T.BCD.A5451.ZCY24.D$(date +'%y%m%d').T$(date +'%H%M%S')1
+          fname=T.BCD.A0001.ZCY24.D$(date +'%y%m%d').T$(date +'%H%M%S')1
           echo "FILENAME=$fname" >> "$GITHUB_OUTPUT"
-          aws s3 cp --no-progress ../shared_files/cclf/archives/valid/T.BCD.ACOB.ZC0Y18.D181120.T0001000 \
+          aws s3 cp --no-progress ../shared_files/cclf/archives/valid2/T.BCD.A0001.ZCY18.D181120.T1000000 \
             s3://bfd-test-eft/bfdeft01/bcda/in/test/$fname
 
   verify:
--- a/.github/workflows/cclf-import-test-integration.yml
+++ b/.github/workflows/cclf-import-test-integration.yml
@@ -56,11 +56,11 @@ jobs:
           unzip ../shared_files/cclf/archives/valid2/T.BCD.A0001.ZCY18.D181120.T1000000
 
           # Clone the synthetic zip file with updated years
-          new_dir="${year}.D${date}.T${time}"
+          new_dir="D${date}.T${time}"
           mkdir $new_dir
-          mv T.BCD.A0001.ZC8Y18.D181120.T1000009 $newdir/T.BCD.A0001.ZC8Y${year}.D${date}.T${time}
-          mv T.BCD.A0001.ZC9Y18.D181120.T1000010 $newdir/T.BCD.A0001.ZC9Y${year}.D${date}.T${time}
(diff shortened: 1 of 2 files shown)


.github/workflows/cclf-import-test-integration.yml
```diff
@@ -43,9 +47,73 @@ jobs:
...
+            CONNECTION_INFO=/bcda/test/api/DATABASE_URL
+      - name: Verify CCLF file was ingested
+        env:
+          FILENAME: ${{needs.trigger.outputs.filename}}
+        # CAUTION: if changing the script below, validate that sensitive information is not printed in the workflow
+        run: |
+          HOST=$(aws rds describe-db-instances --db-instance-identifier bcda-test-rds 2>&1 | jq -r '.DBInstances[0].Endpoint.Address' 2>&1)
+          CONNECTION_URL=$(echo $CONNECTION_INFO 2>&1 | sed -E "s/@.*\/bcda/\@$HOST\/bcda/" 2>&1)
+
+          # Verify that we a record of the CCLF file in the database
```
laurenkrugen-navapbc, line 101: small: think we might be missing a word here?

.github/workflows/cclf-import-test-integration.yml
```diff
@@ -43,9 +47,73 @@ jobs:
...
+          else
+            
+            # Verify that the correct number of benes were imported into the database.
+            CCLF_BENES=`psql -t "$CONNECTION_URL" -c "SELECT count(mbi) FROM cclf_beneficiaries WHERE file_id = $CCLF_FILE" 2>&1`
+              if [[ $? -ne 0 || -z $CCLF_BENES ]]; then
+                echo "CCLF beneficiaries query returned zero results or command failed"
+                exit 1
+              fi
+              if [[ $(echo $CCLF_BENES | xargs) != "6" ]]; then
+                echo "expected 6 beneficiaries imported from file, received $CCLF_BENES".
```
laurenkrugen-navapbc, line 115: is it possible that this will output the stderr here? I know when the connection fails, it can print the connection string.

.github/workflows/cclf-import-test-integration.yml
```diff
@@ -43,9 +47,73 @@ jobs:
...
+          # Clone the synthetic zip file with updated years
+          new_dir="D${date}.T${time}"
+          mkdir $new_dir
+          mv T.BCD.A0001.ZC8Y18.D181120.T1000009 $new_dir/T.BCD.A0001.ZC8Y${year}.D${date}.T${time}1
+          mv T.BCD.A0001.ZC9Y18.D181120.T1000010 $new_dir/T.BCD.A0001.ZC9Y${year}.D${date}.T${time}1
+          mv T.BCD.A0001.ZC0Y18.D181120.T1000011 $new_dir/T.BCD.A0001.ZC0Y${year}.D${date}.T${time}1
+          zip -jr $fname $new_dir
+
+          aws s3 cp --no-progress $fname \
+            s3://bfd-test-eft/bfdeft01/bcda/in/test/$fname
```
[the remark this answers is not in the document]
kyeah: yep, they both get triggered but there are checks in each process to ignore any unrecognized files.

.github/workflows/cclf-import-test-integration.yml
kyeah: yes, english would be great here 😄

.github/workflows/cclf-import-test-integration.yml
kyeah: ah, I figured the previous `$?` check would be an indicator that the query was successful (so we'd always be getting some sort of count here.) Do you think that may not be the case?

kyeah: Can remove the "received ..." part here if there's any doubt!

laurenkrugen-navapbc: oh that's true - if it only hits this condition on success, then I think we are fine!

PR #982, opened 2024-08-05
BCDA-8230: Remove hashed MBI

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8230

## 🛠 Changes

- Switch to POST endpoints with unhashed MBI


shared_files/encrypted/local.env
```diff
@@ -1,125 +1,98 @@
 $ANSIBLE_VAULT;1.1;AES256
```
kyeah, line 1: Removed BB_HASH_ITER and BB_HASH_PEPPER

bcda/client/bluebutton.go
```diff
@@ -66,7 +64,7 @@ type APIClient interface {
 	GetExplanationOfBenefit(jobData models.JobEnqueueArgs, patientID string, claimsWindow ClaimsWindow) (*fhirModels.Bundle, error)
 	GetPatient(jobData models.JobEnqueueArgs, patientID string) (*fhirModels.Bundle, error)
 	GetCoverage(jobData models.JobEnqueueArgs, beneficiaryID string) (*fhirModels.Bundle, error)
-	GetPatientByIdentifierHash(jobData models.JobEnqueueArgs, hashedIdentifier string) (string, error)
+	GetPatientByMbi(jobData models.JobEnqueueArgs, mbi string) (string, error)
```
carlpartridge, line 67: Any argument to keep `GetPatientByIdentifierHash` functionality?  Or is this a known permanent change?

bcda/client/bluebutton.go
```diff
@@ -190,48 +186,38 @@ func (bbc *BlueButtonClient) GetClaim(jobData models.JobEnqueueArgs, mbi string,
...
-
 	params := GetDefaultParams()
-	params.Set("mbi", mbiHash)
 	params.Set("excludeSAMHSA", "true")
 
 	updateParamWithServiceDate(&params, claimsWindow)
 	updateParamWithLastUpdated(&params, jobData.Since, jobData.TransactionTime)
 
-	u, err := bbc.getURL("Claim", params)
+	u, err := bbc.getURL("Claim/_search", params)
```
carlpartridge, line 195: We use getURL above as well.  Some nitpicks/thoughts:
- We should return the same thing each time (we return empty string above and nil here)
- Do we want to return anything more useful than an empty string?  Looking at getURL it is already catching an error, maybe getURL doesnt need to return an error?

.tool-versions
```diff
@@ -0,0 +1 @@
+golang 1.19
```
carlpartridge, line 1: Used by asdf

carlpartridge (approved):

PR #984, opened 2024-08-07
(CCLF Import) Fix large object S3 trigger + update logging

kyeah:
## 🛠 Changes

- Look for `ObjectCreated:CompleteMultipartUpload` events
- Remove non-JSON logging statements
- Add additional log fields (i.e. the import_filename and import_count)

## ℹ️ Context

Noticed that most files received yesterday were quietly skipped. After investigation, when a large file is uploaded, it triggers the multipart upload event instead of the ObjectCreated:Put event.

This blog post indicates that there are several different events that can trigger an ObjectCreated:* event: https://aws.amazon.com/blogs/aws/s3-event-notification/

In addition, the error alert was triggered due to non-JSON logs being picked up, and several "warning-only" logs being logged at the `ERROR` level. This updates that to ensure only real errors are logged at the ERROR level.

## 🧪 Validation

- Integration testing
- Running this in prod with one of the large file uploads

### DIFF
--- a/bcda/cclf/cclf.go
+++ b/bcda/cclf/cclf.go
@@ -58,18 +58,15 @@ type CclfImporter struct {
 
 func (importer CclfImporter) importCCLF0(ctx context.Context, fileMetadata *cclfFileMetadata) (map[string]cclfFileValidator, error) {
 	if fileMetadata == nil {
-		fmt.Println("File CCLF0 not found.")
 		err := errors.New("file CCLF0 not found")
 		log.API.Error(err)
 		return nil, err
 	}
 
-	fmt.Printf("Importing CCLF0 file %s...\n", fileMetadata)
 	log.API.Infof("Importing CCLF0 file %s...", fileMetadata)
 
 	r, closeReader, err := importer.FileProcessor.OpenZipArchive(fileMetadata.filePath)
 	if err != nil {
-		fmt.Printf("Could not read CCLF0 archive %s.\n", fileMetadata)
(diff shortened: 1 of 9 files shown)


bcda/cclf/cclf.go
```diff
@@ -360,14 +335,12 @@ func (importer CclfImporter) ImportCCLFDirectory(filePath string) (success, fail
 		defer c()
 		return importer.FileProcessor.CleanUpCCLF(ctx, cclfMap)
 	}(); err != nil {
-		fmt.Println(err.Error())
-		log.API.Error(err)
+		importer.Logger.Error(err)
 	}
 
 	if failure > 0 {
 		err = errors.New("one or more files failed to import correctly")
```
laurenkrugen-navapbc, line 342: another logging suggestion/non-blocking issue: would it be helpful to include the # of files that failed to import vs a generic message?

laurenkrugen-navapbc:
LGTM! Were there any files that were too large for ingestion that were cleaned up by BFD that we may need to retrieve to try and process again?

laurenkrugen-navapbc (approved):

kyeah:
> LGTM! Were there any files that were too large for ingestion that were cleaned up by BFD that we may need to retrieve to try and process again?

Good question! The timing is such that it's unlikely there were files cleaned up before I noticed, but I can double-check since we're still waiting for A* model entity files this month (last import was July 8th looking at the prod DB).

PR #985, opened 2024-08-09
BCDA-8302: Improve memory usage for CCLF Imports

kyeah:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8302

## 🛠 Changes

I apologize in advance for the size of this PR 🥲 

### Core Changes

1. **Move CCLF0-based validation logic inline to avoid opening and parsing CCLF8 file twice**

    The `validate` function has been removed, and the logic for verifying the maxRecordLength and expected record count is now part of the actual CCLF8 import flow.

2. **Keep zip files open and pass CCLF0 and CCLF8 file references into `importCCLF0` and `importCCLF8` to avoid having to re-download**
 
    When we call LoadCclfFiles, we are now searching for and maintaining references to the CCLF0 and CCLF8 files for use by the import functions later on.

3. **Pre-allocate download buffers by retrieving HEAD information before downloading files**

    This greatly reduces memory usage for S3 downloads and improves download speed. Even with only a single file download for 500-700MB files, I was seeing lambda memory usage up to 3GB and 5 minute downloads. Making this change ensures we cap the memory allocation and allows the process to take on the order of seconds instead of minutes.

### Supplementary Changes

4. **Move some archive validation logic into the initial zip file processing.** 

    In particular, I am now searching for the zip file's inner CCLF0 and CCLF8 files when loading it in `LoadCclfFiles`. This also allows us to be more strict in the inputs for importCCLF0 and importCCLF8 and remove null checks. This also prevents an issue we have had in the past where we process the CCLF0 and then run into an unexpected error when calling `importCclf8` afterwards.

1. **Organize synthetic CCLF file archives to reduce duplication and match expected production archive formats**

    Previously, the synthetic file archives split CCLF0 and CCLF8 files across different zip files. This does not match how we receive data in production. In addition, there were several different folders that essentially duplicated files (e.g. `bcd`, `valid`, and `valid2` were all essentially the same thing.)

### Behavioral Changes

This PR includes some behavioral changes due to the changes above.

1. **We no longer support CCLF files split across multiple archives.**

    Previously, since we imported the entire directory at once, it was possible to combine a CCLF0 file from one archive and a CCLF8 file from another archive. With S3, we are triggering one lambda run for an individual archive, so it is not possible to import across multiple archives.

    This change is now "official" in that we validate for the presence of both a CCLF0 and CCLF8 file within a single archive before attempting to import the archive. This was necessary so we can enforce that both the CCLF0 and CCLF8 file references could be stored up front within CclfZipMetadata.

2. **We fail fast if we find multiple CCLF0 or CCLF8 files within a single archive.**

    A single archive is expected to only contain one CCLF0 and one CCLF8 file. Multiple of these files was undefined behavior previously and I believe only the latest CCLF0 + CCLF8 would be imported.

## 🧪 Validation

Tested on `Test` environment with a ~600MB zip file. Out of memory on the main branch after 5 minutes vs. processed with ~800MB in 23 seconds on this branch.

CleanShot 2024-08-10 at 15 29 55@2x


.github/workflows/cclf-import-test-integration.yml
```diff
@@ -57,7 +57,7 @@ jobs:
           cclf8_fname=T.BCD.A0001.ZC8Y24.D${date}.T${time}1
           echo "FILENAME=$cclf8_fname" >> "$GITHUB_OUTPUT"
 
-          unzip ../shared_files/cclf/archives/valid2/T.BCD.A0001.ZCY18.D181120.T1000000
+          unzip ../shared_files/cclf/archives/valid/T.BCD.A0001.ZCY18.D181120.T1000000
```
kyeah, line 60: `valid2` has been removed since it's the same premise as `valid`.

bcda/cclf/cclf.go
```diff
@@ -56,23 +66,10 @@ type CclfImporter struct {
 	FileProcessor CclfFileProcessor
 }
 
-func (importer CclfImporter) importCCLF0(ctx context.Context, fileMetadata *cclfFileMetadata) (map[string]cclfFileValidator, error) {
-	if fileMetadata == nil {
```
kyeah, line 60: This nil check is now part of `LoadCclfFiles`. cclfZipMetadata is typed to ensure that both the CCLF0 and CCLF8 file exists before calling `importCCLF0` and `importCCLF8`.

bcda/cclf/cclf.go
```diff
@@ -82,42 +79,24 @@ func (importer CclfImporter) importCCLF0(ctx context.Context, fileMetadata *cclf
...
 		b := sc.Bytes()
 		if len(bytes.TrimSpace(b)) > 0 {
 			filetype := string(bytes.TrimSpace(b[fileNumStart:fileNumEnd]))
 
 			if filetype == "CCLF8" {
-				if validator == nil {
-					validator = make(map[string]cclfFileValidator)
-				}
-
-				if _, ok := validator[filetype]; ok {
```
kyeah, line 120: Replaced the map of validators with just a single validator instance, since we only care about the CCLF8 validation.

bcda/cclf/cclf.go
```diff
@@ -286,54 +246,42 @@ func (importer CclfImporter) ImportCCLFDirectory(filePath string) (success, fail
 	}
 
 	if len(cclfMap) == 0 {
-		importer.Logger.Info("Failed to find any CCLF files in directory")
-		return 0, 0, skipped, nil
+		importer.Logger.Info("Failed to find any valid CCLF files in directory")
+		return 0, failure, skipped, err
```
kyeah, line 250: Returning failures here since there are a couple of different failure modes from LoadCclfFiles that don't explicitly return an error, i.e.

- Zip file with correct filename, but missing CCLF0 or CCLF8 within the zip file itself
- Unable to open zip file

It was kinda like this already but I realize maybe the sensible thing here is to actually return an error if failures > 0. Open to opinions!

bcda/cclf/local_fileprocessor.go
```diff
@@ -123,13 +128,52 @@ func (p *processor) walk(path string, info os.FileInfo, err error) error {
...
+		} else {
+			readError = fmt.Errorf("Unexpected CCLF num %d processed (%s)", metadata.cclfNum, path)
+			break
 		}
-		sub[key] = append(sub[key], &metadata)
+	}
+
+	if readError != nil {
+		p.failure++
+		println(readError.Error())
```
laurenkrugen-navapbc, line 153: will this println cause any issues with our log output? ie printing in non-json output?

bcda/cclf/local_fileprocessor.go
```diff
@@ -123,13 +128,52 @@ func (p *processor) walk(path string, info os.FileInfo, err error) error {
 			continue
 		}
 
-		key := metadataKey{perfYear: metadata.perfYear, fileType: metadata.fileType}
-		sub := p.cclfMap[metadata.acoID]
-		if sub == nil {
-			sub = make(map[metadataKey][]*cclfFileMetadata)
-			p.cclfMap[metadata.acoID] = sub
+		if metadata.cclfNum == 0 {
+			if cclf0Metadata != nil {
```
laurenkrugen-navapbc, line 132: Where do the values come from for `cclf0Metadata` and `cclf8Metadata`?

I think I'm missing a step for the comparison here and at line 139; these values are pointers to the type `cclfFileMetadata`, per line 115. They must be getting a value somewhere or they would always be nil at this comparison?  My eyes are probably looking right at it and I'm missing it. 😅

bcda/cclf/local_fileprocessor_test.go
```diff
@@ -189,6 +186,29 @@ func (s *LocalFileProcessorTestSuite) TestProcessCCLFArchives_CorruptedFile() {
...
+	defer conf.SetEnv(s.T(), "CCLF_REF_DATE", cclfRefDate)
+	defer os.RemoveAll(dir)
+
+	// Multiple CCLF0s
+	createZip(s.T(), dir, "T.BCD.A9990.ZCY20.D201113.T0000000", "T.BCD.A9990.ZC0Y20.D201113.T0000010", "T.BCD.A9990.ZC0Y20.D201113.T0000011", "T.BCD.A9990.ZC8Y20.D201113.T0000010")
+
+	// Multiple CCLF8s
+	createZip(s.T(), dir, "T.BCD.A9990.ZCY20.D201114.T0000000", "T.BCD.A9990.ZC0Y20.D201114.T0000010", "T.BCD.A9990.ZC8Y20.D201114.T0000010", "T.BCD.A9990.ZC8Y20.D201114.T0000011")
+
+	m, skipped, f, err := processCCLFArchives(dir)
```
laurenkrugen-navapbc, line 205: Can you point me to where the duplicates are getting caught and returning a failure? Is it part of the `walkfunc`?

bcda/cclf/local_fileprocessor.go
kyeah: This is the local fileprocessor which is only used locally and in tests (and used to be used in Jenkins smoke tests for importing synthetic CCLF files.) Shouldn't end up in Splunk for the most part.

bcda/cclf/local_fileprocessor.go
kyeah: yes! It's right after the conditional checks on line 136 and 143 😄 

```go
cclf0Metadata = &metadata
```

bcda/cclf/local_fileprocessor_test.go
kyeah: yes! in the area you commented on:

```
readError = fmt.Errorf("Multiple CCLF0 files found in zip (%s)", path)
```

This will get bubbled up and increment the file failure count.

bcda/cclf/local_fileprocessor.go
laurenkrugen-navapbc: if the value is assigned after, wouldn't it always be nil? (sorry)

bcda/cclf/local_fileprocessor_test.go
laurenkrugen-navapbc: ty!

bcda/cclf/local_fileprocessor.go
kyeah: This is run in a for loop which iterates over the files in the zip, so we could find a CCLF0 and then later on find another one, which would be an error :)

laurenkrugen-navapbc: that's what I wasn't seeing - thank you! 😄

laurenkrugen-navapbc (approved):
LGTM! ran tests locally and reviewed what you did to verify. 🫡

PR #993, opened 2024-09-18
BCDA-8351: Exclude MD TCoC requests from suppression

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8351

## 🛠 Changes

- When preparing beneficiaries for a request, if the source of the request was from TCoC, then the query to get suppressed beneficiaries will be skipped.
- github.com/go-testfixtures is being used in combination with github.com/golang-migrate/migrate/v4 so that tests that utilize the application database are idempotent. There is [existing code](https://github.com/CMSgov/bcda-app/blob/main/bcda/database/databasetest/databasetest.go#L22) that will create a test database that can have migrations and fixtures applied to it. The long term goal (besides idemtpotence) is to eventually remove all dependency on init() calls, since those prohibit our ability to run singular tests in packages that import another package that uses init().
- Service_test.go has had test name updates, so that we can eventually separate our unit and integration tests in our `make unit-test` command.
- protobuf version was bumped from v1.5.3 -> v1.5.4 due to failing build
-  test fixtures was bumped from v3.5.0 -> v3.7.0 to utilize additional library features that weren't available the previous library
- the service folder contains a directory for data fixtures that was renamed from `testdata` to `fixtures`. Old files that were not in use were removed and new files have been added

	

## ℹ️ Context

To onboard TCOC, we must exclude beneficiary suppression with their requests.

## 🧪 Validation

Added integration test will seeded database for suppression data.


bcda/service/service.go
```diff
@@ -496,10 +496,15 @@ func (s *service) getBenesByFileID(ctx context.Context, cclfFileID uint, conditi
 			upperBound = time.Now()
 		}
 
-		ignoredMBIs, err = s.repository.GetSuppressedMBIs(ctx, s.sp.lookbackDays, upperBound)
-		if err != nil {
-			return nil, fmt.Errorf("failed to retreive suppressedMBIs %s", err.Error())
+		if cfg, ok := s.GetACOConfigForID(conditions.CMSID); ok {
```
laurenkrugen-navapbc, line 499: We already get the config in our middleware when a request is made, so this feels a little redundant. Having something similar to line 493 or adding it as a request condition may have been better, so I am open to discussing this one!

bcda/service/service_test.go
```diff
@@ -170,7 +173,7 @@ func (s *ServiceTestSuite) TearDownTest() {
 	conf.SetEnv(s.T(), "PRIORITY_ACO_REG_EX", s.priorityACOsEnvVar)
 }
 
-func (s *ServiceTestSuite) TestIncludeSuppressedBeneficiaries() {
+func (s *ServiceTestSuite) TestIncludeSuppressedBeneficiariesIntegration() {
```
laurenkrugen-navapbc, line 176: I really liked the strategy outlined in the second response in this [post](https://stackoverflow.com/questions/25965584/separating-unit-tests-and-integration-tests-in-go) about how to discern between unit and integration tests when running tests, hence the name updates.

bcda/service/service_test.go
```diff
@@ -225,7 +228,7 @@ func (s *ServiceTestSuite) TestIncludeSuppressedBeneficiaries() {
 	}
 }
 
-func (s *ServiceTestSuite) TestGetNewAndExistingBeneficiaries() {
+func (s *ServiceTestSuite) TestGetNewAndExistingBeneficiariesIntegration() {
```
laurenkrugen-navapbc, line 231: There are 2-3 tests that I had to make minor updates to; because we are filtering by the ACO model now, the configured test CMS ID of `cmsID` was not being parsed and these tests also weren't loading the config files to do the parsing.

bcda/service/service_test.go
```diff
@@ -1170,6 +1178,201 @@ func (s *ServiceTestSuite) TestGetACOConfigForID() {
...
+			for i := 0; i < len(actualBeneCount); i++ {
+				assert.Equal(t, test.mbis[i], actualBeneCount[i].MBI)
+			}
+		})
+	}
+
+}
+
+func (s *ServiceTestSuiteWithDatabase) TestGetNewAndExistingBeneficiaries_RecentSinceParameterDatabaseIntegration() {
+	db := database.Connection
```
laurenkrugen-navapbc, line 1268: this was an already existing test that it utilizing the database's package `init()`. I moved it under the new test suite since we should eventually remove the dependency on init() and instead use the 'CreateDatabase` function to stand up our test databases. I will continue to work on this if it fits into the sprint.

bcda/service/service.go
```diff
@@ -496,10 +496,15 @@ func (s *service) getBenesByFileID(ctx context.Context, cclfFileID uint, conditi
...
+			if cfg.Model != "TCOCMD" {
```
carlpartridge, line 500: In `NewService` Im seeing:
```
return &service{
		...
		sp: suppressionParameters{
			includeSuppressedBeneficiaries: false,
			lookbackDays: cfg.SuppressionLookbackDays,
		},
		...
```
Is there a reason we cant use `includeSuppressedBeneficiaries` as opposed to a hardcode check of "TCOCMD"?

This means when we create the MD TCoC service (or link it to the ACO?) we will have to override the suppression parameters?

bcda/service/service_test.go
carlpartridge: Quick thought: What about `TestIncludeSuppressedBeneficiaries_Integration`?  Makes it pop out more to me.

Is there a case to be made for adding the `t.Short` clause or do we pretty much always want to run everything?

bcda/service/service_test.go
```diff
@@ -1170,6 +1178,201 @@ func (s *ServiceTestSuite) TestGetACOConfigForID() {
...
+	}
+
+	test_cases := []struct {
+		name      string
+		cmsID     string
+		beneCount int
+		mbis      []string
+	}{
+		{"TCOCMD request with active suppressions", "TCOCMD", 7, []string{"MBI00000001", "MBI00000002", "MBI00000003", "MBI00000004", "MBI00000005", "MBI00000006", "MBI00000007"}},
+		{"Non TCOCMD request with active suppressions", "A0001", 4, []string{"MBI00000004", "MBI00000005", "MBI00000006", "MBI00000007"}},
```
carlpartridge, line 1236: Related to trying to move the check for "TCOC" to a var, could we rename these to something like:
"ACO with ignored suppressions request with active suppressions" and
"ACO request with active suppressions"
I think keeping specifics out keeps things a bit cleaner down the road.

bcda/service/service_test.go
carlpartridge: Im not fully following all of this, maybe we can talk offline and would love to get Kevins perspective here.

go.mod
```diff
@@ -15,9 +15,9 @@ require (
 	github.com/go-chi/chi/v5 v5.0.7
 	github.com/go-chi/render v1.0.1
 	github.com/go-gota/gota v0.12.0
-	github.com/go-testfixtures/testfixtures/v3 v3.5.0
+	github.com/go-testfixtures/testfixtures/v3 v3.7.0
 	github.com/golang-migrate/migrate/v4 v4.14.1
-	github.com/golang/protobuf v1.5.3 // indirect
+	github.com/golang/protobuf v1.5.4 // indirect
```
carlpartridge, line 20: Can we close out this PR: https://github.com/CMSgov/bcda-app/pull/923?

laurenkrugen-navapbc: that PR references a different package (weirdly), so we can check in to see if that PR can be merged.

bcda/service/service_test.go
laurenkrugen-navapbc: yes and yes! I like the snake_case to make it more visible and eventually, we would want to include the `if t.short; t.skip()` line in the integration tests!

bcda/service/service.go
laurenkrugen-navapbc: yeah, [I saw that too](https://github.com/CMSgov/bcda-app/pull/993#discussion_r1765839218) and would much more prefer to use that but it's part of the service structure that's instantiated with our handlers (called on service start instead of per request).

carlpartridge: Ok so a service is instantiated on app start.  We pass something in as part of each request (ACO ID?) to get the specific request ACO out of all of the available ACOConfigs of the service.  It looks like we are pulling ACOConfigs from files in conf/config/{env}.yml?  Is there a reason we cant add onto these a new field eg `ignoreSuppressed`?

Somewhat related question how do these ACOConfigs compare with what we have in the acos DB table?  Looks like the DB table is just identifiers and web/auth related, no config?

laurenkrugen-navapbc: This has been updated to retrieve a setting from the config file!

bcda/api/alr.go
```diff
@@ -19,13 +19,13 @@ func (h *Handler) alrRequest(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 
 	// retrieve ACO & cms_id data from the context through value
-	ad, err := readAuthData(r)
+	ad, err := GetCtxAuthData(r)
```
carlpartridge, line 22: Like the new name!  Any thoughts on name: `GetAuthDataFromCtx`?  I feel like its a bit more explicit/clear, but its also a bit longer...  not sure myself which one I like more so thought I would throw it out there.

bcda/api/alr_test.go
```diff
@@ -57,7 +57,7 @@ func (s *AlrTestSuite) TestAlrRequest() {
 	ad := auth.AuthData{ACOID: s.acoID.String(), CMSID: *aco.CMSID, TokenID: uuid.NewRandom().String()}
 
 	ctx := context.WithValue(req.Context(), auth.AuthDataContextKey, ad)
-	ctx = middleware.NewRequestParametersContext(ctx, middleware.RequestParameters{ResourceTypes: []string{"Patient", "Observation"}})
+	ctx = middleware.NewCtxRequestParams(ctx, middleware.RequestParameters{ResourceTypes: []string{"Patient", "Observation"}})
```
carlpartridge, line 60: Following the `Get` name convention Id like to see this as `Set`, ie `SetCtxRequestParams` (or `SetRequestParamsInCtx`).

bcda/service/service.go
```diff
@@ -496,10 +496,15 @@ func (s *service) getBenesByFileID(ctx context.Context, cclfFileID uint, conditi
 			upperBound = time.Now()
 		}
 
-		ignoredMBIs, err = s.repository.GetSuppressedMBIs(ctx, s.sp.lookbackDays, upperBound)
-		if err != nil {
-			return nil, fmt.Errorf("failed to retreive suppressedMBIs %s", err.Error())
+		if cfg, ok := GetCtxACOCfg(ctx); ok {
+			if !cfg.IgnoreSuppressions {
```
carlpartridge, line 500: I assume this is covered by your tests as well as the bool type but just in case: if `cfg.IgnoreSuppressions` returns a string of either "true"/"false" then this block will always fail.

bcda/service/service.go
```diff
@@ -673,3 +678,17 @@ var (
 	ErrJobNotCancelled   = goerrors.New("Job was not cancelled due to internal server error.")
 	ErrJobNotCancellable = goerrors.New("Job was not cancelled because it is not Pending or In Progress")
 )
+
+type CtxACOCfgType string
```
carlpartridge, line 682: Is there a better place for this type and the following?  Thinking maybe models?

bcda/service/service_test.go
```diff
@@ -1170,6 +1186,206 @@ func (s *ServiceTestSuite) TestGetACOConfigForID() {
...
+	}
+
+	test_cases := []struct {
+		name      string
+		cmsID     string
+		beneCount int
+		mbis      []string
+	}{
+		{"ACO config ignore suppressions is true", "CT000001", 7, []string{"MBI00000001", "MBI00000002", "MBI00000003", "MBI00000004", "MBI00000005", "MBI00000006", "MBI00000007"}},
+		{"ACO config ignore suppressions is false", "A0001", 4, []string{"MBI00000004", "MBI00000005", "MBI00000006", "MBI00000007"}},
```
carlpartridge, line 1247: Looks like we are testing the first two but would like to see the third as well:
- ACO config with ignoreSuppressions = true
- ACO config with ignoreSuppressions = false
- ACO config with no mention of ignoreSuppressions (should default to false)

bcda/api/requests.go
```diff
@@ -470,13 +470,17 @@ func (h *Handler) bulkRequest(w http.ResponseWriter, r *http.Request, reqType se
...
-	if ad, err = readAuthData(r); err != nil {
+	if ad, err = GetCtxAuthData(r); err != nil {
 		logger.Error(err)
 		h.RespWriter.Exception(r.Context(), w, http.StatusUnauthorized, responseutils.TokenErr, "")
 		return
 	}
 
-	rp, ok := middleware.RequestParametersFromContext(ctx)
+	if cfg, ok := h.Svc.GetACOConfigForID(ad.CMSID); ok {
+		ctx = service.NewCtxACOCfg(ctx, cfg)
```
carlpartridge, line 480: `Set`?

bcda/web/middleware/validation.go
```diff
@@ -30,11 +30,11 @@ type requestkey int
 
 const rk requestkey = 0
 
-func NewRequestParametersContext(ctx context.Context, rp RequestParameters) context.Context {
+func NewCtxRequestParams(ctx context.Context, rp RequestParameters) context.Context {
```
carlpartridge, line 33: This function doesnt look like its doing a whole lot for us, do you think its useful?  As far as I can tell the two are equivalent:
`ctx := NewCtxRequestParams(r.Context(), rp)`
`ctx := context.WithValue(ctx, 0, rp)`
The difference is the requestkey?  I didnt understand the comment on it.

conf/configs/dev.yml
```diff
@@ -34,3 +34,9 @@ aco_config:
       data:
         - "adjudicated"
         - "partially-adjudicated"
+    - model: "TCOCMD"
```
carlpartridge, line 37: Do we want to call this TCOCMD or just something like TESTIGNORE?  Same goes for test and opensbx env's.  Thinking that if we put up example code on our static site that referencing a specific company might not be ideal.

bcda/api/alr.go
laurenkrugen-navapbc: yeah, I came back to that rename with fresh eyes and felt conflicted about the name, since the order is reversed from what it should be. I am always in favor of explicit naming, even if it is longer. I think I'd rather switch this around, so I will make that update!

bcda/service/service.go
laurenkrugen-navapbc: Yeah, I agree. I spent quite a bit of time debating this yesterday and messing around with placement and struggled to settle on where to put these. 

In my mind, this key,value would be set in the middleware after the auth data is added to the context, but perhaps the types could live in a more common package, such as models. 

Since GetACOConfigByID is a method of Service and our package structure might benefit from combining some packages and moving things around, I wanted to get y'alls thoughts on that before making changes. Maybe we could discuss more?

bcda/service/service_test.go
laurenkrugen-navapbc: Good point - I will do an explicit set to false, since the last bullet point is applied to all the other acos in the config.

bcda/api/requests.go
laurenkrugen-navapbc: Not sure I follow - do you mean this is where we are setting the ACO config in the context?

conf/configs/dev.yml
laurenkrugen-navapbc: Yeah, that's true. Because this is technically a new APM, I wanted to follow the format of the config file; the shorthand name of the model with the regex for the possible ID it may have. We can chat more about this!

bcda/api/alr_test.go
laurenkrugen-navapbc: I switched these to `set`, although it seems that this is a heavily debated topic in the go community. Some folks say that the usage of getters and setters aren't really necessary unless we are using along an interface, etc. For the sake of readability, we can just keep them as is for now.

bcda/service/service.go
laurenkrugen-navapbc: Per our convo, we are going to keep this here. It really should live elsewhere (ideally middleware), but this is consistent with other k,v pairs that get added to the context in other parts of the project. We are going to discuss how we can do some minor restructuring of packages to make changes like this cleaner in the future.

bcda/web/middleware/validation.go
laurenkrugen-navapbc: it just checks the context for that particular key; it's not really doing anything for us at this point. It looks like it's used in about 11 places; shall I update those or do we want to keep this PR the current size?

bcda/service/service.go
laurenkrugen-navapbc: IgnoreSuppressions will always be a bool type, so we should be OK here!

bcda/web/middleware/validation.go
```diff
@@ -30,11 +30,11 @@ type requestkey int
 
 const rk requestkey = 0
 
-func NewRequestParametersContext(ctx context.Context, rp RequestParameters) context.Context {
+func SetRequestParamsCtx(ctx context.Context, rp RequestParameters) context.Context {
```
carlpartridge, line 33: Nit: If we dont want to change this everywhere can we add a TODO comment?

bcda/service/service.go
```diff
@@ -673,3 +678,20 @@ var (
...
+
+const CtxACOCfg CtxACOCfgType = "ctxACOCfg"
+
+// TODO: this should live within middleware, models, or another common package.
+//
+//	We should move this when we do package cleanup.
+func NewACOCfgCtx(ctx context.Context, cfg *ACOConfig) context.Context {
+	newctx := context.WithValue(ctx, CtxACOCfg, cfg)
+	return newctx
```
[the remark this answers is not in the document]
laurenkrugen-navapbc: oh, good catch! was debugging and broke this statement out to look at the context. I'll update now!

PR #1004, opened 2024-10-16
Add env var feature flag to enable queueing library

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8426

## 🛠 Changes

Enable swapping of queueing libraries based on env var.

## 🧪 Validation

Full test suite and linting pass.


bcda/api/requests.go
```diff
@@ -632,6 +632,7 @@ func (h *Handler) bulkRequest(w http.ResponseWriter, r *http.Request, reqType se
 		sinceParam := (!rp.Since.IsZero() || conditions.ReqType == service.RetrieveNewBeneHistData)
 		jobPriority := h.Svc.GetJobPriority(conditions.CMSID, j.ResourceType, sinceParam) // first argument is the CMS ID, not the ACO uuid
 
+		logger.Infof("Adding jobs using %T", h.Enq)
```
carlpartridge, line 635: Sanity check that we are are using the proper queueing library.  I dont totally understand why we seem to have different loggers and different access to them?

bcdaworker/queueing/enqueue.go
```diff
@@ -20,7 +21,13 @@ type Enqueuer interface {
...
 
 func NewEnqueuer() Enqueuer {
-	return queEnqueuer{que.NewClient(database.QueueConnection)}
+	if conf.GetEnv("QUEUE_LIBRARY") == "que-go" {
+		return queEnqueuer{que.NewClient(database.QueueConnection)}
+	} else if conf.GetEnv("QUEUE_LIBRARY") == "river" {
+		return riverEnqueuer{}
+	} else {
+		panic("NO QUEUE LIBRARY SET")
+	}
```
carlpartridge, line 30: Not sure if panic-ing is better than just defaulting to a specific library?

bcdaworker/queueing/enqueue.go
```diff
@@ -67,3 +74,13 @@ func (q queEnqueuer) AddAlrJob(job models.JobAlrEnqueueArgs, priority int) error
 
 	return q.Enqueue(j)
 }
+
+type riverEnqueuer struct{}
```
carlpartridge, line 78: Dummy type and interface for now.

conf/configs/dev.env
```diff
@@ -10,6 +10,7 @@ BCDA_WORKER_ERROR_LOG=/var/log/worker/error.log
 DEBUG=true
 DEPLOYMENT_TARGET=dev
 EXPIRED_THRESHOLD_HR=24
+QUEUE_LIBRARY=que-go
```
carlpartridge, line 13: These files are the non-sensitive env vars used for each environment?

bcdaworker/queueing/enqueue.go
austincanada: If we're going to a green-blue solution for the libraries I'd just default to que-go.

laurenkrugen-navapbc: yeah, I think defaulting to que-go would probably be best here in the event of a failure, since that's the existing behavior

PR #1008, opened 2024-11-05
Draft: CCLF lambda changes

austincanada:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8405

## 🛠 Changes

Added the main entry to the CCLF lambda.
Added the logic flow for checking if the SQS event contains a CSV file path, or CCLF.
Renamed CCLF lambda functions to attribution agnostic.
Added parser functions for checking CSV file.
Added testing for CSV parser functions.

## ℹ️ Context

These changes will be the entry for the lambda changes in order to make the CCLF lambda more agonistic to an attribution lambda.

## 🧪 Validation

This PR is a part of a larger group of PR's that will come later. Some of these entries don't function properly, such as the CSVImporter. Therefore this PR is more-or less a check for making sure the functionality is up to standards for the team.

### DIFF
--- a/bcda/cclf/parser_test.go
+++ b/bcda/cclf/parser_test.go
@@ -323,6 +323,7 @@ func TestCheckIfAttributionCSVFile(t *testing.T) {
 		{name: "Is not an Attribution CSV File path (incorrect fourth)", path: "P.PCPB.M2014.D003022.T2420001", testIsCSV: false},
 		{name: "Is not an Attribution CSV File path (incorrect fifth)", path: "P.PCPB.M2014.D00302.T24200011", testIsCSV: false},
 		{name: "Is not an Attribution CSV File path (CCLF file)", path: "T.BCD.A0001.ZCY18.D181121.T1000000", testIsCSV: false},
+		{name: "Is not an Attribution CSV File path (opt-out file)", path: "T#EFT.ON.ACO.NGD1800.DPRF.D181120.T1000009", testIsCSV: false},
 	}
 
 	for _, test := range tests {
--- a/bcda/cclf/parser_test.go
+++ b/bcda/cclf/parser_test.go
@@ -91,7 +91,8 @@ func TestGetCCLFMetadata(t *testing.T) {
 		{"Invalid date (no 13th month)", sspID, "T.BCD.A0001.ZC0Y18.D181320.T0001000", "failed to parse date", cclfFileMetadata{}},
 		{"CCLF file too old", sspID, gen(sspProd, startUTC.Add(-365*24*time.Hour)), "out of range", cclfFileMetadata{}},
 		{"CCLF file too new", sspID, gen(sspProd, startUTC.Add(365*24*time.Hour)), "out of range", cclfFileMetadata{}},
-		{"Production SSP file", sspID, sspProdFile, "",
+		{
+			"Production SSP file", sspID, sspProdFile, "",
 			cclfFileMetadata{
 				env:       "production",
 				name:      sspProdFile,
@@ -102,7 +103,8 @@ func TestGetCCLFMetadata(t *testing.T) {
 				fileType:  models.FileTypeDefault,
 			},
 		},
-		{"Test SSP file", sspID, sspTestFile, "",
+		{
+			"Test SSP file", sspID, sspTestFile, "",
 			cclfFileMetadata{
 				env:       "test",
 				name:      sspTestFile,
@@ -125,7 +127,8 @@ func TestGetCCLFMetadata(t *testing.T) {
 				fileType:  models.FileTypeRunout,
 			},
 		},
-		{"Production CEC file", cecID, cecProdFile, "",
+		{
+			"Production CEC file", cecID, cecProdFile, "",
 			cclfFileMetadata{
 				env:       "production",
 				name:      cecProdFile,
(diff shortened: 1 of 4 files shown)


bcda/lambda/cclf/main.go
```diff
@@ -60,14 +60,56 @@ func cclfImportHandler(ctx context.Context, sqsEvent events.SQSEvent) (string, e
...
+		FileProcessor: &cclf.S3FileProcessor{
+			Handler: optout.S3FileHandler{
+				Logger:        logger,
+				Endpoint:      os.Getenv("LOCAL_STACK_ENDPOINT"),
+				AssumeRoleArn: s3AssumeRoleArn,
+			},
+		},
+	}
+
+	success, failure, skipped, err := importer.ImportCSV(s3AssumeRoleArn, s3ImportPath)
```
carlpartridge, line 92: It looks like you have left out the app and DB psuedo-code.  Is that because Lauren has it and is actively working on it?

bcda/cclf/parser_test.go
```diff
@@ -293,3 +293,30 @@ func TestGetCCLFMetadata(t *testing.T) {
...
 }
+
+func TestCheckIfAttributionCSVFile(t *testing.T) {
+	tests := []struct {
+		name      string
+		path      string
+		testIsCSV bool
+	}{
+		{name: "Is an Attribution CSV File path", path: "P.PCPB.M2014.D00302.T2420001", testIsCSV: true},
+		{name: "Is not an Attribution CSV File path (incorrect first)", path: "T.PCPB.M2014.D00302.T2420001", testIsCSV: false},
```
carlpartridge, line 304: By incorrect first do you mean the first part of the file name (in this case the "T")?  A T is actually an acceptable character here as far as I understand (it can be a "T" or a "P" for test and prod).

bcda/cclf/parser_test.go
```diff
@@ -293,3 +293,30 @@ func TestGetCCLFMetadata(t *testing.T) {
...
+		{name: "Is not an Attribution CSV File path (incorrect second)", path: "P.BFD.N2014.D00302.T2420001", testIsCSV: false},
+		{name: "Is not an Attribution CSV File path (incorrect third)", path: "P.PCPB.M2014.D00302.T2420001", testIsCSV: false},
+		{name: "Is not an Attribution CSV File path (incorrect fourth)", path: "P.PCPB.M2014.D003022.T2420001", testIsCSV: false},
+		{name: "Is not an Attribution CSV File path (incorrect fifth)", path: "P.PCPB.M2014.D00302.T24200011", testIsCSV: false},
+		{name: "Is not an Attribution CSV File path (CCLF file)", path: "T.BCD.A0001.ZCY18.D181121.T1000000", testIsCSV: false},
```
carlpartridge, line 309: We should add an opt-out filename as well.  I am fairly sure we are using the same s3 bucket for attribution files as well as opt-out files.

If we have tests that do similar things for the CCLF file and the opt out file, we should add a CSV file example to them as well to make sure they dont pick up this filename schema.  Might be a bit overkill but I think is worth the effort.

bcda/lambda/cclf/main_test.go
```diff
@@ -14,15 +14,15 @@ import (
 	"github.com/stretchr/testify/suite"
 )
 
-type CclfImportMainSuite struct {
+type AttributionImportMainSuite struct {
```
carlpartridge, line 17: I like that we are changing these names.  Do you think it would cause a lot of trouble to also rename the package and directory name to attribution as well?  Would we have to update terraform?

carlpartridge:
I think we should create a new epic branch based off of main (eg ~ `epic/attribution-file-generalization`, then compare this PR to that branch.  That would allow us to merge this PR in (as well as Lauren and anyone elses work in when ready).

bcda/lambda/cclf/main.go
austincanada: Yes, that's correct

bcda/lambda/cclf/main_test.go
austincanada: I'm not sure about updating the terraform, but we'd have to update the other workflows associated with cclf.

austincanada: Yea, we can do that if it conflicts with your changes.

bcda/lambda/cclf/main.go
```diff
@@ -60,14 +60,56 @@ func cclfImportHandler(ctx context.Context, sqsEvent events.SQSEvent) (string, e
 			// importing the one file that was sent in the trigger.
 			filepath := fmt.Sprintf("%s/%s", e.S3.Bucket.Name, e.S3.Object.Key)
 			logger.Infof("Reading %s event for file %s", e.EventName, filepath)
-			return handleCclfImport(s3AssumeRoleArn, filepath)
+			if cclf.CheckIfAttributionCSVFile(e.S3.Object.Key) {
```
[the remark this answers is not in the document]
austincanada: Yea, that sounds like a good future addition.

bcda/cclf/parser_test.go
austincanada: Updated to add an opt-out file, but couldn't find where we did tests for opt-out file names.

bcda/cclf/parser_test.go
austincanada: Updated.

bcda/cclf/parser_test.go
```diff
@@ -268,7 +283,8 @@ func TestGetCCLFMetadata(t *testing.T) {
 				fileType:  models.FileTypeDefault,
 			},
 		},
-		{"Test SBX file", sbxID, sbxTestFile, "",
+		{
+			"Test SBX file", sbxID, sbxTestFile, "",
```
carlpartridge, line 287: We were going to add a csv file name to these tests right?

carlpartridge (approved):

PR #1009, opened 2024-11-13
BCDA-7514: Update and throw OpOutcome on absence of CCLF8

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-7514

## 🛠 Changes

Currently, we throw an error when CCLF8 files are not found using `CCLFNotFoundError`. We also check if we are throwing this error for `/runout` or `/all`. This has been updated to throw an `OperationOutcome` error that is FHIR compliant. Message details [here](https://confluence.cms.gov/display/BCDA/BCDA-7791%3A+Refine+EOY+Mode+Proposals#:~:text=Proposal%201%20%2D%20Gracefully%20handle%20ACOs%20without%20current%20attributions)

A check for `/runout` has been removed. In any scenario where we do not have attribution, we will be throwing the same error creating consistency between `/runout` and `/all`.

## ℹ️ Context

During E.O.Y. mode, we have a hardcoded freeze and multiple manual activities. This is one of the initiatives to move away from E.O.Y. mode.

### DIFF
--- a/bcda/api/requests.go
+++ b/bcda/api/requests.go
@@ -396,10 +396,10 @@ func (h *Handler) AttributionStatus(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Retrieve the most recent cclf 8 file we have successfully ingested
+	group := chi.URLParam(r, "groupId")
 	asd, err := h.getAttributionFileStatus(ctx, ad.CMSID, models.FileTypeDefault)
-	if err != nil {
-		logger.Error(errors.Wrap(err, "Failed to retrieve recent CCLF8 file"))
-		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+	if ok := goerrors.As(err, &service.CCLFNotFoundError{}); ok {
+		h.RespWriter.Exception(r.Context(), w, http.StatusInternalServerError, responseutils.NotFoundErr, fmt.Sprintf("Unable to perform export operations for this Group. No up-to-date attribution information is available for Group '%s'. Usually this is due to awaiting new attribution information at the beginning of a Performance Year.", group))
 		return
 	}
 	if asd != nil {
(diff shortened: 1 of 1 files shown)


bcda/api/requests_test.go
```diff
@@ -117,8 +117,8 @@ func (s *RequestsTestSuite) TestRunoutEnabled() {
 	}{
 		{"Successful", nil, http.StatusAccepted, apiVersionOne},
 		{"Successful v2", nil, http.StatusAccepted, apiVersionTwo},
-		{"No CCLF file found", service.CCLFNotFoundError{}, http.StatusNotFound, apiVersionOne},
-		{"No CCLF file found v2", service.CCLFNotFoundError{}, http.StatusNotFound, apiVersionTwo},
+		{"No up-to-date attribution information", CCLFNotFoundOperationOutcomeError{}, http.StatusInternalServerError, apiVersionOne},
+		{"No up-to-date attribution information", CCLFNotFoundOperationOutcomeError{}, http.StatusInternalServerError, apiVersionTwo},
```
carlpartridge, line 121: Nit: I would add a v2 to the description string to make it easier to distinguish when trying to figure out why one of these failed.

carlpartridge (commented):
Code looks good to me but I would like @laurenkrugen-navapbc experience here for an approval as well.

laurenkrugen-navapbc:
Changes look good to me; the only other place that also needs an update is the [attribution endpoint](https://github.com/CMSgov/bcda-app/blob/main/bcda/api/requests.go#L399C1-L415C3). We are returning an error using `http.error` instead of the operation outcome there when a CCLF file is not found.

laurenkrugen-navapbc:
Also minor nit: can we please rename the title so it starts with BCDA-XXXX?

laurenkrugen-navapbc (approved):

PR #1011, opened 2024-11-14
Carl/bcda 8456 add river queue library

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8456

## 🛠 Changes

Added river queue library.  Added pgx v5.

## 🧪 Validation

Test suite (with additions), linting, gosec all passing.

PR #1013, opened 2024-12-04
Revert "BCDA-8456: add river queue library"

bhagatparwinder:
Reverts CMSgov/bcda-app#1011

Please see thread https://cmsgov.slack.com/archives/CHG7Q7XNH/p1732670116209879

Workers are stalling at 70% usage and we have narrowed it down to two issues:
- Nov Image
- This change

Reverting queue change for now to isolate the root cause.


laurenkrugen-navapbc (approved):
LGTM, thank you for reverting!

laurenkrugen-navapbc:
Looks like downgrading our PGX library is going to be an issue.

PR #1015, opened 2024-12-06
BCDA-8548: Update Attribution Integration Test

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8548

## 🛠 Changes

- github workflow for attribution integration test has been updated to include the csv file type

## ℹ️ Context

Made to support csv attribution types for models that utilize csv attribution files.

## 🧪 Validation

Integration test github action has been updated to include steps to generate a csv attribution file and ingest it. Test is passing and manually validated that database entries looks OK.


.github/workflows/cclf-import-test-integration.yml
```diff
@@ -118,3 +129,24 @@ jobs:
...
+            CSV_BENES=`psql -t "$CONNECTION_URL" -c "SELECT count(mbi) FROM cclf_beneficiaries WHERE file_id = $CSV_FILE" 2>&1`
+              if [[ $? -ne 0 || -z $CSV_BENES ]]; then
+                echo "CSV beneficiaries query returned zero results or command failed"
+                exit 1
+              fi
+              if [[ $(echo $CSV_BENES | xargs) != "5" ]]; then
+                echo "expected 5 beneficiaries imported from file, received $CSV_BENES".
+                exit 1
+              fi
+          fi
```
carlpartridge, line 150: Just out of curiosity what are the pros/cons of testing both cclf and csv file in one run as opposed to two separate runs?

bcda/cclf/parser.go
```diff
@@ -41,8 +41,13 @@ type CSVParser struct {
 func getACOConfigs() ([]service.ACOConfig, error) {
 	configs, err := service.LoadConfig()
 	if err != nil {
+		log.API.Errorf("error when fetching configs: %s", err)
 		return []service.ACOConfig{}, err
 	}
+	if configs == nil {
```
carlpartridge, line 47: Should this logic check also return an error and an empty service config?

bcda/cclf/parser.go
```diff
@@ -55,25 +60,28 @@ func GetCSVMetadata(path string) (csvFileMetadata, error) {
...
+	if acos == nil {
+		return csvFileMetadata{}, errors.New("No ACO configs found.")
 	}
 
 	for _, v := range acos {
 		filenameRegexp := regexp.MustCompile(v.AttributionFile.NamePattern)
 		parts := filenameRegexp.FindStringSubmatch(path)
 		if len(parts) == v.AttributionFile.MetadataMatches {
 			metadata, err = validateCSVMetadata(parts)
+			log.API.Error("regex for aco (%s): %s", v.Model, v.AttributionFile.NamePattern)
```
carlpartridge, line 74: This is in development debugging (not expected to go to prod)?

.github/workflows/cclf-import-test-integration.yml
laurenkrugen-navapbc: Seems simpler to keep these two together, but if there's a case for them being separated, then I am open to it!

bcda/cclf/parser.go
laurenkrugen-navapbc: These commits are purely for testing/logging (since I cannot recreate locally). I will be reverting these -- left the PR in draft mode and hadn't request review yet since I would be adding things like the above if I needed to troubleshoot.

bcda/cclf/parser.go
laurenkrugen-navapbc: yup, same as above!

bcda/cclf/parser_test.go
```diff
@@ -109,6 +109,13 @@ func TestValidateCCLFFileName(t *testing.T) {
 			fileType:  models.FileTypeDefault,
 		},
 		},
+		{"valid csv test filename", "T.PCPB.M2411." + fileDateTime, nil, csvFileMetadata{
+			env:       "test",
+			timestamp: validTime,
+			perfYear:  24,
+			fileType:  models.FileTypeDefault,
+		},
+		},
```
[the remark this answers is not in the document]
laurenkrugen-navapbc: yup. there are definitely cases for invalid files, I just caught this one that needed an update for a case I missed in the previous PR.

.github/workflows/cclf-import-test-deploy.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           CGO_ENABLED: 0
         run: |
           go build -o bin/bootstrap ./lambda/cclf/main.go
-          zip -j function.zip bin/bootstrap
+          zip -j function.zip bin/bootstrap ../conf/configs/dev.yml ../conf/configs/test.yml ../conf/configs/prod.yml
```
carlpartridge, line 25: Sandbox missing?  Also can we do `../conf/configs/*.yml`?

laurenkrugen-navapbc: the only env that we don't do attribution in is sandbox, so this is deliberate! I can make the changes for the wildcard if we are OK with having sandbox in there but not needed. Thoughts?

carlpartridge (approved):
Looks good to me!

PR #1028, opened 2025-01-13
BCDA-8626:  Add aco deny admin lambda

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8626

## 🛠 Changes

Add lambda and deploy workflows.  Lambda code implements logic to deny an array of ACOs access to BCDA.

## ℹ️ Context

This is one of the administrative tasks we are moving off of jenkins.

## 🧪 Validation

Local testing and linting.


bcda/lambda/admin_aco_deny/README.md
```diff
@@ -0,0 +1,6 @@
+The ACO Deny administrative task lambda will add an existing ACO to our deny list.  It should be called via AWS's lambda interface (see: https://confluence.cms.gov/display/BCDA/How+To+deny+an+ACO+From+Generating+Credentials).
```
laurenkrugen-navapbc, line 1: Can we get this documentation page updated with the new method of adding an ACO to the deny list? We can keep the soon-to-be-deprecated steps on there for now.

laurenkrugen-navapbc:
I don't see any invocations for this at all since it was deployed; are you seeing the same? It also looks like there are no log groups for it either under the monitoring tab.

carlpartridge:
> I don't see any invocations for this at all since it was deployed; are you seeing the same? It also looks like there are no log groups for it either under the monitoring tab.

I have run it a few times in dev env and verified the results: https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/bcda-dev-admin-aco-deny?tab=monitoring
I am seeing the same thing as you are seeing for test and prod though.

.github/workflows/admin-aco-deny-test-integration.yml
```diff
@@ -0,0 +1,51 @@
...
+  workflow_dispatch:
+
+# Ensure we have only one integration test running at a time
+concurrency:
+  group: admin-aco-deny-integration-test
+
+jobs:
+  # Deploy first if triggered by pull_request
+  deploy:
+    if: ${{ github.event_name == 'pull_request' }}
```
austincanada, line 20: Could we use the `on: pull_request:` trigger for this?

carlpartridge: As I understand this (and reminder that I am copy-pasting this from existing other lambda integration tests), we are saying here: 
step 1: if this is a pull_request, run the dev deploy workflow
So basically we want to make sure the latest lambda code is deployed before we run tests against it.

laurenkrugen-navapbc:
> > I don't see any invocations for this at all since it was deployed; are you seeing the same? It also looks like there are no log groups for it either under the monitoring tab.
> 
> I have run it a few times in dev env and verified the results: https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/bcda-dev-admin-aco-deny?tab=monitoring I am seeing the same thing as you are seeing for test and prod though and I think thats because they havent properly deployed yet (and Im not sure I can until this branch is merged without going through the hassle of changing all the on: push options).

got it - I will verify in the dev env! thanks for clarifying.

.github/workflows/admin-aco-deny-test-integration.yml
austincanada: If we're just mirroring other tests that's fine, but if we needed to do anything more specific to the event triggers there's a pull_request trigger itself for workflows.

https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request

carlpartridge: Yup, we are actually using that already on line 4 above.

PR #1029, opened 2025-01-17
BCDA-8529: Fix logic for getResponse method

austincanada:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8529

## 🛠 Changes

Updated logic within getResponse. Removed a unnecessary nil check. Reevaluated some error checking with request responses.

## ℹ️ Context

Needed to try and see if getResponse logic was responsible for potential memory leaks. 

## 🧪 Validation

Unit-test suite runs.

### DIFF
--- a/bcda/client/bluebutton.go
+++ b/bcda/client/bluebutton.go
@@ -463,7 +463,7 @@ func (h *httpLogger) logRequest(req *http.Request) {
 
 func (h *httpLogger) logResponse(req *http.Request, resp *http.Response) {
 	h.l.WithFields(logrus.Fields{
-		"resp_code":      resp.StatusCode,
+		"resp_status":    resp.StatusCode,
 		"bb_query_id":    req.Header.Get(constants.BBHeaderOriginQID),
 		"bb_query_ts":    req.Header.Get(constants.BBHeaderTS),
 		"bb_uri":         req.URL.String(),
--- a/bcda/client/fhir/client.go
+++ b/bcda/client/fhir/client.go
@@ -38,7 +38,6 @@ type singleClient struct {
 var _ Client = &singleClient{}
 
 func (c *singleClient) DoBundleRequest(req *http.Request) (bundle *models.Bundle, nextURL *url.URL, err error) {
-
 	// Ensure that we'll receive the entire bundle response in a single request
 	vals := req.URL.Query()
 	vals.Del("_count")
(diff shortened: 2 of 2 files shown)


bcda/client/fhir/client.go
```diff
@@ -129,26 +128,25 @@ func getResponse(c *http.Client, req *http.Request) (body []byte, err error) {
...
+	defer resp.Body.Close()
+	s.Response = resp
+	s.End()
 
 	if resp.StatusCode >= http.StatusBadRequest {
-		// Attempt to read the body in case it offers valuable troubleshooting info
-		body, _ := io.ReadAll(resp.Body)
+		body, err = io.ReadAll(resp.Body)
+		if err != nil {
+			return nil, fmt.Errorf("received incorrect status code %d, failed to parse body %+v", resp.StatusCode, err)
```
laurenkrugen-navapbc, line 144: nit: can we take the opportunity to update this error message and the one on line 146? Maybe something like:
`fmt.Errorf("failed to parse response body %+v", err)`

We wrap the status code in this call in a logging statement up the stack, so I think reporting the status code is 1) not needed here and 2) the status code itself is an indicator of the request so adding that additional detail  of ("status code is incorrect") seems like it could create confusion when troubleshooting.

On a related note: this line, specifically `resp_code` could be updated to `resp_status` - we currently have both configured for log fields and it would be nice to unify those to make searching through the logs easier. Thoughts on adding that one line change?

austincanada: I updated the above error message. I don't think the error message on 146 should be changed due to what that error is catching based on the if statement.

I also updated the `resp_code` entry.

laurenkrugen-navapbc (approved):
Thanks for addressing the changes! I think line 146 could still use a less confusing log message, but maybe we can discuss another time.

bcda/client/bluebutton.go
```diff
@@ -463,7 +463,7 @@ func (h *httpLogger) logRequest(req *http.Request) {
 
 func (h *httpLogger) logResponse(req *http.Request, resp *http.Response) {
 	h.l.WithFields(logrus.Fields{
-		"resp_code":      resp.StatusCode,
+		"resp_status":    resp.StatusCode,
```
carlpartridge, line 466: Will this cause a rift in older logs vs newer logs?  Is that problematic?  @laurenkrugen-navapbc

laurenkrugen-navapbc: nope! have updated many fields and the only thing to note is: when you are searching for historical logs before the date this is merged/deployed for the environment that you are searching in, make sure you use the old field name.

PR #1030, opened 2025-01-22
BCDA-8460-MBI-masking-alert-testing

bhagatparwinder:
## 🎫 Ticket

[BCDA-8460](https://jira.cms.gov/browse/BCDA-8460)

## 🛠 Changes

Two log statements logging made up MBI and HICN numbers

## ℹ️ Context

We mask MBI/HICN numbers in Splunk logs. An alert triggers if there is an unmasked identifier logged. Testing that alert as part of this ticket.

### DIFF
--- a/bcda/main.go
+++ b/bcda/main.go
@@ -93,4 +93,8 @@ func main() {
 		fmt.Printf("Error occurred while executing command %s\n", err)
 		log.API.Fatal(err)
 	}
+
+	// Log a made up MBI number for testing the Splunk alert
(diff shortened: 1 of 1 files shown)


carlpartridge:
> > The goal will be to deploy this to dev, verify the alerts, then revert?
> 
> Correct. Technically, I can deploy my branch in dev and test it out. However, in the ticket I see @greg-pf mentioned testing additional environments
> 
> > run this search against other envs to see if we are getting any false positives
> 
> Hence the PR.

Is it possible to log into a running instance and generate a custom log?

carlpartridge (approved):

PR #1031, opened 2025-01-22
BCDA-8635: Migrate Platinum AMI From Jenkins to Github Workflows

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8635

## 🛠 Changes

Migration of platinum image creation from Jenkins -> Github Workflows.

## ℹ️ Context

Made as part of the larger effort to deprecate Jenkins. This requires another PR in the ops repo that must also be merged: https://github.com/CMSgov/bcda-ops/pull/1146 

## 🧪 Validation

Workflow passes and confirmed AMI is available in EC2. (latest run will fail until ops repo is merged, but successful run can be found here: https://github.com/CMSgov/bcda-app/actions/runs/13039689810/job/36378517357?pr=1031)

### DIFF
--- a/.github/workflows/build-ami.yml
+++ b/.github/workflows/build-ami.yml
@@ -50,7 +50,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: CMSgov/bcda-ops
-          ref: 'main'
+          ref: 'lauren/BCDA-8635'
           token: ${{ env.GITHUB_TOKEN }}
 
       - name: download packer configs
@@ -65,6 +65,6 @@ jobs:
           ls $GITHUB_WORKSPACE/packer-configs/
           ls ./packer-configs/
           pwd
-          packer init packer/platinum.json 2>&1
-          PACKER_LOG=1 packer build -color=false -var 'source_ami=$GOLD_IMAGE' -var 'subnet_id=$SUBNET_ID' packer/platinum.json 2>&1 | tee platinum_packer_output.txt
+          packer init packer/platinum.json.pkr.hcl 2>&1
+          PACKER_LOG=1 packer build -color=false -var 'source_ami=$GOLD_IMAGE' -var 'subnet_id=$SUBNET_ID' packer/platinum.json.pkr.hcl 2>&1 | tee platinum_packer_output.txt
 
--- a/.github/workflows/build-ami.yml
+++ b/.github/workflows/build-ami.yml
@@ -44,6 +44,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
+          clean: false
           repository: CMSgov/bcda-ops
           ref: 'main'
           token: ${{ env.GITHUB_TOKEN }}
--- a/.github/workflows/build-ami.yml
+++ b/.github/workflows/build-ami.yml
@@ -64,7 +64,7 @@ jobs:
       - name: packer build
         env:
           PACKER_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          #TMPDIR: /home/ec2-user/
+          TMPDIR: /home/ec2-user/
         run: |
           # set -euo pipefail to ensure internal packer/ansible failures get passed to the jenkins pipeline and fails that jobstep
           set -euo pipefail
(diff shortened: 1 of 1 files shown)


.github/workflows/build-ami.yml
```diff
@@ -0,0 +1,55 @@
+name: Build AMI
+
+# on:
+#   pull_request:
+#     types:
+#       - closed
+on:
+  pull_request:
+    paths:
+      - .github/workflows/build-ami.yml
```
carlpartridge: Im thinking we will want to call this via `workflow_dispatch` and maybe `workflow_call` as well?  Or is this more of a one time thing that happens per base AMI?

.github/workflows/build-ami.yml
```diff
@@ -0,0 +1,55 @@
+name: Build AMI
```
carlpartridge, line 1: Nit: Slightly more descriptive?  Or maybe a comment describing what the AMI is used for/context?

.github/workflows/build-ami.yml
```diff
@@ -0,0 +1,55 @@
...
+jobs:
+  build_ami:
+    name: build the platinum ami
+    runs-on: self-hosted
+    steps:
+      - uses: hashicorp/setup-packer@v2.0.1
+      - uses: aws-actions/configure-aws-credentials@v3
+        with:
+          aws-region: ${{ vars.AWS_REGION }}
+          role-to-assume: arn:aws:iam::${{ secrets.ACCOUNT_ID }}:role/delegatedadmin/developer/bcda-test-github-actions
```
carlpartridge, line 21: I believe we have `bcda-{env}-github-actions` (per env), do we need to specify that?  Does anything change on this build depending on env?

.github/workflows/build-ami.yml
```diff
@@ -0,0 +1,55 @@
...
+      - uses: cmsgov/ab2d-bcda-dpc-platform/actions/aws-params-env-action@main
+        env:
+          AWS_REGION: ${{ vars.AWS_REGION }}
+        with:
+          params: |
+            GITHUB_TOKEN=/ci/github/token
+            SUBNET_ID=/bcda/workflows/packer_subnet_id
+            GOLD_IMAGE=/bcda/workflows/gold_image_ami
```
carlpartridge, line 29: For my clarity, we build off an AWS gold image and we create what we call a platinum image?

.github/workflows/build-ami.yml
```diff
@@ -0,0 +1,55 @@
...
+      - name: checkout bcda-ops
+        uses: actions/checkout@v4
+        with:
+          repository: CMSgov/bcda-ops
+          ref: 'lauren/BCDA-8635'
```
carlpartridge: Probably want `main` here?  (or remove ref, as I assume that defaults to main?

gfreeman-navapbc (approved):
LGTM!

.github/workflows/build-ami.yml
laurenkrugen-navapbc: closing this comment: had this in here originally so that I could test it when pushing to PR!

.github/workflows/build-ami.yml
laurenkrugen-navapbc: Nope, nothing should change for this; it's the baseline for all of the instances, so it doesn't have anything env specific.

.github/workflows/build-ami.yml
laurenkrugen-navapbc: That's correct! We use the monthly gold image and create an AMI based off of that called the 'platinum image'

.github/workflows/build-ami.yml
laurenkrugen-navapbc: yup - sorry! this is also outdated; I had to remove some unused parts and in order to test, I needed to point it to my other PR temporarily. Didn't catch it until after I requested review.

.github/workflows/build-ami.yml
laurenkrugen-navapbc: done!

carlpartridge (approved):

PR #1032, opened 2025-01-23
BCDA-8523: Adjust query that counts total number of unprocessed jobs

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8523

## 🛠 Changes

Adjust query that counts total unprocessed jobs.

## ℹ️ Context

This query is what determines the cloudwatch alarms for total number of unprocessed jobs.  Previously que-go would just delete finished jobs, while river does not.

## 🧪 Validation

Local testing and linting.

### DIFF
--- a/bcdaworker/queueing/river.go
+++ b/bcdaworker/queueing/river.go
@@ -169,7 +169,7 @@ func updateJobQueueCountCloudwatchMetric(db *sql.DB, log logrus.FieldLogger) {
 }
 
 func getQueueJobCount(db *sql.DB, log logrus.FieldLogger) float64 {
-	row := db.QueryRow(`select count(*) from river_job;`)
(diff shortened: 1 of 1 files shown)


laurenkrugen-navapbc (approved):
Does river require us to do any cleanup to delete the jobs from the que? Is there a possibility of the queue hitting a limit?

carlpartridge:
> Does river require us to do any cleanup to delete the jobs from the que? Is there a possibility of the queue hitting a limit?

It looks like there are some automated processes that clean things up for you: https://riverqueue.com/docs/maintenance-services.  But we will still want these changes in as the cloudwatch number is what determines how many workers are running.  We may want to modify the cleaner to run more often at some point?

PR #1037, opened 2025-01-31
BCDA-8627- add workflows to deploy/promote/test lambda

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8627

## 🛠 Changes

Migration of Jenkins job to GHA. This PR will include workflows and lambda business logic.

## ℹ️ Context

We are migrating away from Jenkins jobs.

### DIFF
--- a/.github/workflows/archive-clean-job-data-opensbx-deploy.yml
+++ b/.github/workflows/archive-clean-job-data-opensbx-deploy.yml
@@ -1,0 +1,35 @@
+name: Archive and Clean Job Data opensbx deploy
+
+on:
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+  test:
+    permissions:
+      contents: read
+      id-token: write
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: bcda
+    environment: opensbx
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+      - name: Build archive-clean-job-data zip file
+        env:
+          CGO_ENABLED: 0
+        run: |
+          go build -o bin/bootstrap ./lambda/archive_clean_job/*.go
+          zip -j function.zip bin/bootstrap
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: ${{ vars.AWS_REGION }}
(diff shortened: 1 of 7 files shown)


.github/workflows/archive-clean-job-data-opensbx-deploy.yml
laurenkrugen-navapbc, line 1: We don't typically deploy lambdas to opensbx, but I think we need to in this case. We should probably keep this in mind with other admin-type lambdas that come up.

carlpartridge: I made a note to create a ticket for this.  I can see wanting other admin tasks in opensbx, but not our other lambdas.

.github/workflows/archive-clean-job-data-opensbx-deploy.yml
```diff
@@ -0,0 +1,35 @@
...
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: ${{ vars.AWS_REGION }}
+          role-to-assume: arn:aws:iam::${{ secrets.ACCOUNT_ID }}:role/delegatedadmin/developer/bcda-dev-github-actions
+      - name: Upload and reload
+        run: |
+          aws s3 cp --no-progress function.zip \
+            s3://bcda-opensbx-archive-clean-job-data-function/function-${{ github.sha }}.zip
+          aws lambda update-function-code --function-name bcda-opensbx-archive-clean-job-data \
+            --s3-bucket bcda-opensbx-archive-clean-job-data-function --s3-key function-${{ github.sha }}.zip
```
carlpartridge, line 35: Arguably we could deploy to opensbx like we do prod (as in copy over test lambda as opposed to rebuilding and deploying).

bcda/lambda/archive_clean_job_data/main.go
```diff
@@ -0,0 +1,11 @@
+package main
+
+import "fmt"
+
+func main() {
+	fmt.Println("Hello, World!")
```
carlpartridge, line 6: I know you arent here yet but just in case, main has to follow aws lambda call patterns, ie:
```
lambda.Start(handlerFunc)
```

PR #1038, opened 2025-02-03
Carl/bcda 8630 admin create aco creds

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8630

## 🛠 Changes

Add Admin Create ACO Credentials lambda as well as relevant supporting files.

## ℹ️ Context

Part of moving off jenkins.  This task finds then creates a new creds file for an ACO.

## 🧪 Validation

Local testing, linting.


.github/workflows/admin-aco-deny-sbx-deploy.yml
```diff
@@ -0,0 +1,25 @@
+name: Admin ACO Deny sbx deploy
```
carlpartridge, line 1: Adding this in because its quick and easy.  This is a separate lambda, but we do want this functionality in sbx env.

bcda/auth/provider.go
```diff
@@ -70,6 +70,9 @@ type CommonClaims struct {
 
 // Provider defines operations performed through an authentication provider.
 type Provider interface {
+	// FindAndCreateACOCredentials takes an ACO ID and calls RegisterSystem, then formats the results
+	FindAndCreateACOCredentials(ACOID string, IPs []string) (string, error)
```
carlpartridge, line 74: This creates a seam in between the lambda and setting up a SSAS client/provider.  This allows us to skip some unnecessary DB queries as well as because the SSAS client's init() does not have access to certain variables at init() time.  We set those env vars, then reset the client.

bcda/bcdacli/cli.go
```diff
@@ -724,20 +724,13 @@ func createACO(name, cmsID string) (string, error) {
...
-	creds, err := auth.GetProvider().RegisterSystem(aco.UUID.String(), "", aco.GroupID, ips...)
+	creds, err := auth.GetProvider().FindAndCreateACOCredentials(acoCMSID, ips)
 	if err != nil {
 		return "", errors.Wrapf(err, "could not register system for %s", acoCMSID)
 	}
 
-	msg := fmt.Sprintf("%s\n%s\n%s", creds.ClientName, creds.ClientID, creds.ClientSecret)
-
-	return msg, nil
+	return creds, nil
```
carlpartridge, line 733: Refactoring this to use the simplified (refactored) method.

bcda/database/connection.go
```diff
@@ -71,7 +72,7 @@ func createDB(cfg *Config) (*sql.DB, error) {
 
 	stdlib.RegisterDriverConfig(&dc)
 
-	db, err := sql.Open("nrpgx", dc.ConnectionString(cfg.DatabaseURL))
+	db, err := sql.Open("nrpgx", dc.ConnectionString(strings.TrimSpace(cfg.DatabaseURL)))
```
carlpartridge, line 75: Not sure why, but the strings coming back from the aws param store were including line returns (\n).  Not sure why that would happen in a lambda and not in app code?

bcda/auth/ssas.go
```diff
@@ -27,6 +27,28 @@ type SSASPlugin struct {
 // validates that SSASPlugin implements the interface
 var _ Provider = SSASPlugin{}
 
+func (s SSASPlugin) FindAndCreateACOCredentials(ACOID string, ips []string) (string, error) {
```
laurenkrugen-navapbc, line 30: Was this created because the logic in `[ResetSecret](https://github.com/CMSgov/bcda-app/pull/1038/files#diff-0ff831a4d1ac470c0af6137692614fa873897b4bcc080ce3acc583f43f062e75R95)` wasn't meeting the criteria for what we need to do?

carlpartridge: See above comment, but also the original CLI command being referenced from the old jenkins script was calling `RegisterSystem` so I kept with that.  ResetSecret sounds more appropriate but I didnt want to rock the boat too much.

bcda/lambda/admin_create_aco_creds/aws.go
```diff
@@ -0,0 +1,174 @@
...
+
+	bcdaaws "github.com/CMSgov/bcda-app/bcda/aws"
+	"github.com/CMSgov/bcda-app/conf"
+)
+
+var destBucket = "bcda-aco-credentials"
+var kmsAliasName = "alias/bcda-aco-creds-kms"
+var pemFilePath = "/tmp/BCDA_CA_FILE.pem"
+
+func getAWSParams(session *session.Session) (awsParams, error) {
```
austincanada, line 25: Did we want to make a utils file for the lambdas and add this to it? Seems like we're using something like this in a lot of them.

.github/workflows/admin-create-aco-creds-test-integration.yml
```diff
@@ -0,0 +1,52 @@
...
+    uses: ./.github/workflows/admin-create-aco-creds-dev-deploy.yml
+    secrets: inherit
+
+  trigger:
+    if: ${{ always() }}
+    needs: deploy
+    permissions:
+      contents: read
+      id-token: write
+    runs-on: ubuntu-latest
```
austincanada, line 30: Not sure but I remember @laurenkrugen-navapbc having problems with using `ubuntu` vs `ubuntu-latest`. Have you had any problems with this?

.github/workflows/admin-create-aco-creds-test-integration.yml
```diff
@@ -0,0 +1,52 @@
...
+        with:
+          aws-region: ${{ vars.AWS_REGION }}
+          role-to-assume: arn:aws:iam::${{ secrets.ACCOUNT_ID }}:role/delegatedadmin/developer/bcda-dev-github-actions
+      - name: Send event to trigger lambda function
+        id: invoke-lambda
+        run: |
+          echo "STARTTIME=`date +%s`" >> "$GITHUB_OUTPUT"
+          aws lambda invoke --function-name bcda-dev-admin-create-aco-creds \
+            --cli-binary-format raw-in-base64-out \
+            --payload '{"aco_id":"A0012","ips":["1.2.3.4", "1.2.3.5"]}' test-result.txt
```
austincanada, line 48: Did we want to test this with an IPv6 as well?

bcda/bcdacli/cli.go
laurenkrugen-navapbc: Not sure I am following on the refactor here; are we still utilizing the cli command or the function it calls elsewhere in the code? Might be nice to eliminate the dependency on it so we can remove this package/no longer needed commands later on.

carlpartridge: The refactor just simplifies the logic (less lines of code).  Nothing dramatic.  Happy to revert that or just wholesale remove this CLI functionality.  I was generally under the preference of leaving CLI functionality in place, then wholesale removing all of CLI when the time came.  But also dont mind doing it piece by piece if that is preferred.

bcda/lambda/admin_create_aco_creds/aws.go
```diff
@@ -0,0 +1,174 @@
...
+			// Print the error, cast err to awserr.Error to get the Code and Message from an error.
+			log.Error(err.Error())
+		}
+		return "", err
+	}
+
+	return result.String(), nil
+}
+
+func adjustedEnv() string {
```
laurenkrugen-navapbc, line 168: seems like we are moving away from `opensbx` and more towards `sbx`, maybe we should get a ticket in the backlog to update the name on our end eventually so we don't have to make workarounds like this one?

carlpartridge: I felt like we were making strides towards standardizing, but so much of our existing infra/resources is still using opensbx.  Ive also heard greenfield might add a new/different sandbox env name in there as well...

PR #1041, opened 2025-02-04
BCDA-8637: Smoke tests github workflow

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8637

## 🛠 Changes

Adding smoke tests as github action.

- removed EOY mode, since we no longer use that

## ℹ️ Context

Part of migration away from jenkins. 

## 🧪 Validation

Verified that it works with hard-coded inputs.
 https://github.com/CMSgov/bcda-app/pull/1058#pullrequestreview-2679888587

Once merged, will test the inputs for all envs to validate.


.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,142 @@
...
+  smoke_tests: 
+    name: run smoke tests
+    runs-on: self-hosted
+    env:
+      DOCKER_BUILDKIT: 1
+      COMPOSE_DOCKER_CLI_BUILD: 1
+    steps:
+      - name: Modify Credentials For Dev Input
+        run: |
+          if [[ inputs.test-aco == "dev" ]]; then
```
laurenkrugen-navapbc, line 76: each env has a default client; this makes sure that if dev is chosen, that it gets the default for that env. Default -> env name.

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,150 @@
+name: 'BCDA Smoke Tests'
+
+on:
+  pull_request:
+    paths:
+      - .github/workflows/smoke-tests.yml
+  workflow_dispatch:
```
carlpartridge, line 7: We will want workflow_call as well (can be called from another workflow).  I want to call this from my deploy workflow for example.

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,149 @@
...
+    inputs:
+      app-branch:
```
carlpartridge, line 9: Nit: Im using `release_version` in other workflows to refer to the branch/tag being used, might be good to standardize.  (And `ops_release_version` for bcda-ops, `ssas_release_version` for bcda-ssas-app).  Also how strongly do you feel about -'s instead of _'s for these variable names?

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,149 @@
...
+          - 'dev'
+          - 'paca'
+        default: 'dev'
+      smoke-tests:
+        description: Flag which indicates if smoke integration tests should be run
+        required: true
+        type: boolean
+        default: true
+      postman-tests:
+        description: Flag which indicates if smoke integration tests should be run
```
carlpartridge, line 47: Nit: Description: `postman integration tests` instead of `smoke integration tests`?

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,149 @@
...
+    name: run smoke tests
+    runs-on: self-hosted
+    env:
+      DOCKER_BUILDKIT: 1
+      COMPOSE_DOCKER_CLI_BUILD: 1
+    steps:
+      - name: Modify Credentials For Dev Input
+        run: |
+          if [[ inputs.test-aco == "dev" ]]; then
+            echo "creds=${{ inputs.env }}" >> $GITHUB_OUTPUT
```
carlpartridge, line 77: Nit: I think its preferred to use GITHUB_ENV for env vars, I believe GITHUB_OUTPUT is more for job outputs.

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,149 @@
...
+            HOST=/bcda/workflows/${{ inputs.env }}/ssas_elb
+            CLIENT_CREDENTIALS_PARAMS=/bcda/${{ inputs.env }}/${{ steps.vars.outputs.creds }}_client_credentials
+            DENYLIST_CLIENT_CREDENTIALS_PARAMS=/bcda/${{ inputs.env }}/denylisted_client_credentials
+            PACA_CLIENT_CREDENTIALS_PARAMS=/bcda/${{ inputs.env }}/paca_client_credentials
+      - name: checkout bcda-ssas
+        uses: actions/checkout@v4
+        with:
+          repository: CMSgov/bcda-ssas-app
+          ref: 'main'
+      - name: Install docker compose manually
```
carlpartridge, line 97: Nit: I dont think we need this step anymore (see PLT-905)

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,149 @@
...
+        type: boolean
+        default: true
+      postman-tests:
+        description: Flag which indicates if smoke integration tests should be run
+        required: true
+        type: boolean
+        default: true
+
+
+concurrency:
```
carlpartridge, line 53: This concurrency relates to multiple job runs at a time I imagine?  You can have individual jobs run concurrently, eg you could have all tests, smoke tests, and postman tests run as individual jobs and at the same time.

.github/workflows/smoke-tests.yml
laurenkrugen-navapbc: definitely would like to standardize. For the variable names not being called in the param section, happy to! for the names of params in param stores, is it ok to leave those? param store won't allow updates to parameter names once they've been created, so they would all need to be deleted and recreated. Could probably do that with a script, but they get stored in a GHA variable immediately anyway.

.github/workflows/smoke-tests.yml
laurenkrugen-navapbc: you got it! will update!

.github/workflows/smoke-tests.yml
laurenkrugen-navapbc: oh, good catch! I previously had multiple jobs but now that it's just one, I will take this out!

.github/workflows/smoke-tests.yml
laurenkrugen-navapbc: Would these technically be integration tests? or more like e2e?

carlpartridge (approved):

PR #1039, opened 2025-02-03
BCDA-8631: Create group lambda

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8631

## 🛠 Changes

Lambda has been added so that we can call it to create groups, ad hoc.

## ℹ️ Context

Done as part of the effort to migrate away from jenkins.

## 🧪 Validation

Tests & workflows passing; integration test works and verified expected database entries and lambda logs.

PR #1042, opened 2025-02-05
Parwinder/bcda-8627-archive cleanup job migration

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8627

## 🛠 Changes

Updated River Client to set archive cleanup periodic job.

## ℹ️ Context

- We are migrating away from Jenkins and moving to GHA.
- Archive/Cleanup Job has been moved to the bcda-app.
- Job is scheduled using River Go

### DIFF
--- a/bcdaworker/queueing/river_test.go
+++ b/bcdaworker/queueing/river_test.go
@@ -169,3 +169,22 @@ func TestCleanupJobWorker_Work(t *testing.T) {
 	mockCleanupJob.AssertExpectations(t)
 	mockArchiveExpiring.AssertExpectations(t)
 }
+
+func TestGetCutOffTime(t *testing.T) {
+	// Save and set environment variable using conf.SetEnv and defer to reset it
+	defer func(origValue string) {
+		conf.SetEnv(t, "ARCHIVE_THRESHOLD_HR", origValue)
+	}(conf.GetEnv("ARCHIVE_THRESHOLD_HR"))
+
+	// Test with default value
+	conf.SetEnv(t, "ARCHIVE_THRESHOLD_HR", "")
+	expectedCutoff := time.Now().Add(-24 * time.Hour)
+	actualCutoff := getCutOffTime()
+	assert.WithinDuration(t, expectedCutoff, actualCutoff, time.Second, "Cutoff time should be 24 hours ago by default")
+
+	// Test with custom value
+	conf.SetEnv(t, "ARCHIVE_THRESHOLD_HR", "48")
+	expectedCutoff = time.Now().Add(-48 * time.Hour)
+	actualCutoff = getCutOffTime()
+	assert.WithinDuration(t, expectedCutoff, actualCutoff, time.Second, "Cutoff time should be 48 hours ago")
+}


bcdaworker/queueing/river.go
```diff
@@ -30,14 +34,35 @@ func StartRiver(numWorkers int) *queue {
 	workers := river.NewWorkers()
 	river.AddWorker(workers, &JobWorker{})
 
+	schedule, err := cron.ParseStandard("0 11,23 * * *")
```
carlpartridge, line 37: I think after line 35 we need:
`river.AddWorker(workers, &CleanupJobWorker{})`

bcdaworker/queueing/river.go
```diff
@@ -29,15 +33,37 @@ import (
...
 
 	riverClient, err := river.NewClient(riverpgxv5.New(database.Pgxv5Pool), &river.Config{
 		Queues: map[string]river.QueueConfig{
 			river.QueueDefault: {MaxWorkers: numWorkers},
 		},
 		// TODO: whats an appropriate timeout?
-		JobTimeout: -1, // default for river is 1m, que-go had no timeout, mimicking que-go for now
-		Logger:     getSlogLogger(),
-		Workers:    workers,
+		JobTimeout:   -1, // default for river is 1m, que-go had no timeout, mimicking que-go for now
```
laurenkrugen-navapbc, line 63: Based on the last week or so, it looks like the length of time for the clean has varied between 1m - 5m. Maybe we can go with 6-7m to start, especially since we expect the amount of data to eventually go up?

bcdaworker/queueing/river.go
```diff
@@ -99,6 +125,61 @@ type JobWorker struct {
...
 }
 
+type CleanupJobArgs struct {
+}
+
+type CleanupJobWorker struct {
+	river.WorkerDefaults[CleanupJobArgs]
+}
+
+func (args CleanupJobArgs) Kind() string {
```
austincanada, line 135: Can you explain what's going on here? Not sure I understand what's happening with this chunk.

carlpartridge: River is pretty verbose, probably worth a quick read through: https://riverqueue.com/docs

bcdaworker/queueing/river.go
```diff
@@ -178,3 +260,91 @@ func getQueueJobCount(db *sql.DB, log logrus.FieldLogger) float64 {
 
 	return float64(count)
 }
+
+func ArchiveExpiring(maxDate time.Time) error {
```
carlpartridge, line 264: Id much prefer moving these functions to a new/separate module(s).  
ArchiveExpiring seems to:
```
Connect to DB
Get jobs by time and status
For each job
- check if dir exists, if so move file to archived dir
- update job status to archived
```
CleanupJob seems to:
```
Connect to DB
Get jobs by time and status
For each job
- clean up data (wipe job dirs/files)
- update job status
```
This feels like a unique ~ file clean up domain that is unrelated to river queueing.

Overall my concerns are:
- Mixing business logic domains.  I think this is largely how we got to the tangled web of dependencies we are at now.  Would love for modules to be small, unique, have known domain boundaries, and be well tested.
- Removing cli functionality prematurely.  If we remove this and our jobs arent as we expect we have no backup besides fixing and redeploying.  Probably not a big deal but Id feel more comfortable with all of this.

carlpartridge: My thoughts were that we have both river Work call the new module functions as well as cli would call the new module functions.  So we get both bullet points covered.  As I mentioned Im not sure how much the second bullet point is really worth.  I feel like we will be doing a full sweep cleanup of cli at the very end of our migration work but its probably also fine to do some ahead of time as well.  So overall Ill defer to your and other dev preferences.
Just one nitpick on your last bullet point is that we would **not** be able to  manually clean these files via CLI.  (A redeploy isnt the end of the world or a custom script that can connect to the EFS could work as well).

carlpartridge: Not that I am aware of, but we might!

bcda/bcdacli/cli.go
```diff
@@ -286,75 +283,6 @@ func setUpApp() *cli.App {
 				return nil
 			},
 		},
-		{
-			Name:     "archive-job-files",
```
austincanada, line 290: Did we want to remove these?

carlpartridge:
What is our testing strategy going to be for making sure these files are being cleaned up properly?  Can we stick a bunch of files into EFS to test?  They will need to match up with expired jobs right?

bcdaworker/cleanup/cleanup.go
```diff
@@ -0,0 +1,174 @@
...
+		cleanupJob:      cleanupJob,
+		archiveExpiring: archiveExpiring,
+	}
+}
+
+func (args CleanupJobArgs) Kind() string {
+	return "CleanupJob"
+}
+
+func (w *CleanupJobWorker) Work(ctx context.Context, rjob *river.Job[CleanupJobArgs]) error {
```
carlpartridge, line 44: Hmm, wondering if it makes sense to leave these above structs in river and just have the business logic in this module?  Feels like we are mixing domains a bit (river & cleanup).

bcdaworker/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,372 @@
...
+	"github.com/CMSgov/bcda-app/bcda/models"
+	"github.com/CMSgov/bcda-app/bcda/models/postgres/postgrestest"
+	"github.com/CMSgov/bcda-app/conf"
+	"github.com/pborman/uuid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+	"github.com/urfave/cli"
+)
+
+type CleanupTestSuite struct {
```
laurenkrugen-navapbc, line 22: Are the tests under this suite running with make unit-test? tests that utilize a suite usually require a Run(..) statement with go -test, I think?
```
func TestCleanupTestSuite(t *testing.T) {
    suite.Run(t, new(CleanupTestSuite))
}
```

laurenkrugen-navapbc: I don't think the suite or test setup/teardowns are required, but might be worth checking that one too.

bcdaworker/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,372 @@
...
+	suite.Suite
+	testApp *cli.App
+
+	testACO models.ACO
+
+	db *sql.DB
```
laurenkrugen-navapbc, line 28: A lot of our tests rely on the database that gets spun up with `init()` and the database persists across all tests; using this [test helper](https://github.com/CMSgov/bcda-app/blob/f8d1d0f9319ce220f9c8878fe8ec8f31e2ad4606/bcda/database/databasetest/databasetest.go#L22) in combination with [testfixtures](https://github.com/CMSgov/bcda-app/pull/1039/files#diff-696149897963f3142f0dc4f6dcfde3d63cf0ba5b10efb8ae7becb4c06a167efcR33)  would create a copy of our database, apply migrations to it to setup the schema + tables, and then tear it down when the test is over, making it a clean data set for each test run.

In this case, I don't see database.connection being used but outright, but wondering if it gets set somewhere?

bcdaworker/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,372 @@
...
+	f, err := os.Create(fmt.Sprintf("%s/fake.ndjson", path))
+	if err != nil {
+		s.T().Error(err)
+	}
+	defer f.Close()
+
+	// condition: normal execution
+	// execute the test case from CLI
+	args = []string{"bcda", constants.ArchJobFiles}
+	err = s.testApp.Run(args)
```
laurenkrugen-navapbc, line 108: Since we removed the cli commands from the bcdacli package, is it necessary to run the arguments through this test app here and in other tests?

carlpartridge:
> > What is our testing strategy going to be for making sure these files are being cleaned up properly? Can we stick a bunch of files into EFS to test? They will need to match up with expired jobs right?
> 
> To test it in DEV/TEST environment, I believe that is the route we will have to take. We will have to find out:
> 
> * How to get files in EFS
> * What are the correct files (for expired, canceled, and failed jobs)

Thinking about this a bit.  One way we could test is that we generate some jobs at 10:30 and 11:30.  Then the next day (24 hrs later) at ~11:01 the 10:30 jobs should be cleaned up and marked appropriately right?  And the 11:30 jobs should not be touched?  The issue with this approach is that it requires a 24hr testing window, which means that it will be slow to get a deploy out (not the end of the world or anything).

It also occurs to me that we cant disable the current jenkins script as it is doing all environments.  At least not without deploying bcda-ops changes.  This could also make it difficult to test?  Also the jenkins script will start failing as the CLI commands have been removed.

I think with all these difficulties to test it might be safer/easier to bring back the CLI commands for now?  @laurenkrugen-navapbc @austincanada How do you feel about all this?

bcdaworker/cleanup/cleanup_test.go
laurenkrugen-navapbc: yup, I think the helper you linked could be used to queue up the job, or you could put it in a `SetupTest()` if the all the tests require a db connection?

laurenkrugen-navapbc:
> > > What is our testing strategy going to be for making sure these files are being cleaned up properly? Can we stick a bunch of files into EFS to test? They will need to match up with expired jobs right?
> > 
> > 
> > To test it in DEV/TEST environment, I believe that is the route we will have to take. We will have to find out:
> > 
> > * How to get files in EFS
> > * What are the correct files (for expired, canceled, and failed jobs)
> 
> Thinking about this a bit. One way we could test is that we generate some jobs at 10:30 and 11:30. Then the next day (24 hrs later) at ~11:01 the 10:30 jobs should be cleaned up and marked appropriately right? And the 11:30 jobs should not be touched? The issue with this approach is that it requires a 24hr testing window, which means that it will be slow to get a deploy out (not the end of the world or anything).
> 
> It also occurs to me that we cant disable the current jenkins script as it is doing all environments. At least not without deploying bcda-ops changes. This could also make it difficult to test? Also the jenkins script will start failing as the CLI commands have been removed.
> 
> I think with all these difficulties to test it might be safer/easier to bring back the CLI commands for now? @laurenkrugen-navapbc @austincanada How do you feel about all this?

I think having a written strategy + expectations of the cleanup would be good -- I also think one (or more) of us can help QA this. We could temporarily update the jenkins script to not clean up files in our dev env (remove dev as default value); that way we can still manually cleanup with the old script if needed, but it won't automatically cleanup files before we can test the new script.

bcdaworker/queueing/river.go
```diff
@@ -178,3 +320,28 @@ func getQueueJobCount(db *sql.DB, log logrus.FieldLogger) float64 {
 
 	return float64(count)
 }
+
+func getCutOffTime() time.Time {
```
carlpartridge, line 324: Can we get some tests on this?

laurenkrugen-navapbc (approved):
Thanks for all the updates on this! Before you merge, would you mind updating the title to be in the format of: `BCDA-XXXX: <title>`?

Other than that, LGTM! 🎉

PR #1043, opened 2025-02-05
Create ACO Admin Task (Lambda and Workflows)

austincanada:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8629

## 🛠 Changes

Added lambda for Create ACO tasks.
Added workflows to run integration tests and deploy lambda to envs.
Made cli action, CreateACO, an export.
Linter added cli updates.

## ℹ️ Context

Since we are moving off of Jenkins, we need to create lambdas for certain actions and tasks.

## 🧪 Validation

Verifying through workflows.
```"32c0751e-***"	"TESTACO"	"2025-02-24 19:58:04.512983+00"	"2025-02-24 19:58:04.512983+00"	1382	"32c0751e-***"	"TEST623"			"{""CutoffDate"": ""0001-01-01T00:00:00Z"", ""DenylistType"": 0, ""ClaimsStrategy"": 0, ""OptOutStrategy"": 0, ""TerminationDate"": ""0001-01-01T00:00:00Z"", ""AttributionStrategy"": 0}"```
Entry added to the database.

### CODE REVIEW

File: bcda/lambda/admin_create_aco/main.go [hunk 1]
```diff
@@ -0,0 +1,129 @@
...
+
+func handleCreateACO(ctx context.Context, data payload, notifier Notifier) error {
+	_, _, err := notifier.PostMessageContext(ctx, slackChannel, slack.MsgOptionText(
+		fmt.Sprintf("Started Create ACO lambda in %s env.", os.Getenv("ENV")), false),
+	)
+	if err != nil {
+		log.Errorf("Error sending notifier start message: %+v", err)
+	}
+
+	// TODO: acoUuid potentially needs to be returned
```
laurenkrugen-navapbc, line 81: Is the UUID currently returned when creating an ACO using the cli command? Would this only impact the internal team that uses this if it's not returned?

carlpartridge:
Cant fully review until the checks are all passing but looking good so far!

bcda/lambda/admin_create_aco/main.go
```diff
@@ -0,0 +1,129 @@
...
+		fmt.Sprintf("Success: Create ACO lambda in %s env.", os.Getenv("ENV")), false),
+	)
+	if err != nil {
+		log.Errorf("Error sending notifier success message: %+v", err)
+	}
+
+	return nil
+}
+
+func getAWSParams() (awsParams, error) {
```
laurenkrugen-navapbc, line 106: @carlpartridge is it worthwhile to break this out into a package that can just be imported instead? I think we are recreating this in a lot of our lambdas.

carlpartridge: Its already mostly reusing `bcdaaws` package.  There might be a good way to refactor that a bit for better use with lambdas.  For the most part the logic is along the lines of:
```
Get env name
Get param A
Get param B
Return params{A, B}
```
So Im not entirely sure how much more we could get out of it without getting more complex then it needs to be?

bcda/lambda/admin_create_aco/main.go
austincanada: Yes, currently the UUID is being returned when using the CLI command. Currently the only place where the returned UUID is being set is within the smoke tests, but it is also not being used.

.github/workflows/admin-create-aco-prod-deploy.yml
```diff
@@ -0,0 +1,24 @@
+name: Admin Create ACO prod deploy
```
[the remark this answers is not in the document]
austincanada: Just for clarification, for prod deploy, we want to copy the test lambda over to prod env and for sbx env we want to do the same?

bcda/lambda/admin_create_aco/README.md
```diff
@@ -0,0 +1,6 @@
+The Create ACO administrative task lambda will create an ACO on list. It should be called via AWS's lambda interface (see: <https://confluence.cms.gov/display/BCDA/How+To+deny+an+ACO+From+Generating+Credentials>).
```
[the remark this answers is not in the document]
austincanada: Updated

.github/workflows/admin-create-aco-test-integration.yml
```diff
@@ -0,0 +1,51 @@
...
+        with:
+          aws-region: ${{ vars.AWS_REGION }}
+          role-to-assume: arn:aws:iam::${{ secrets.ACCOUNT_ID }}:role/delegatedadmin/developer/bcda-dev-github-actions
+      - name: Send event to trigger lambda function
+        id: invoke-lambda
+        run: |
+          echo "STARTTIME=`date +%s`" >> "$GITHUB_OUTPUT"
+          aws lambda invoke --function-name bcda-dev-admin-create-aco \
+            --cli-binary-format raw-in-base64-out \
+            --payload '{"name":"Test ACO","cms_id":"testcms"}' test-result.txt
```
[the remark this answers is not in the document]
austincanada: Looking into it currently.

.github/workflows/admin-create-aco-prod-deploy.yml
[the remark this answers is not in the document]
austincanada: Added

bcda/lambda/admin_create_aco/main.go
```diff
@@ -0,0 +1,129 @@
...
+	if env == "local" {
+		return awsParams{conf.GetEnv("DATABASE_URL"), ""}, nil
+	}
+
+	bcdaSession, err := bcdaaws.NewSession("", os.Getenv("LOCAL_STACK_ENDPOINT"))
+	if err != nil {
+		return awsParams{}, err
+	}
+
+	dbURL, err := bcdaaws.GetParameter(bcdaSession, fmt.Sprintf("/bcda/%s/api/DATABASE_URL", env))
```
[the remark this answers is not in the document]
austincanada: Updated to remove dbURL, and also removed AWS params since we're only using the Slack token from this function.

bcda/lambda/admin_create_aco/main.go
```diff
@@ -0,0 +1,118 @@
...
+	)
+	if err != nil {
+		log.Errorf("Error sending notifier start message: %+v", err)
+	}
+
+	_, err = bcdacli.CreateACO(data.Name, data.CMSID)
+	if err != nil {
+		log.Errorf("Error creating ACO: %+v", err)
+
+		_, _, err := notifier.PostMessageContext(ctx, slackChannel, slack.MsgOptionText(
```
carlpartridge, line 80: This just happened to me so be aware you will want to change the above line to: `_, _, err := notifier...` to `_, _, slackErr := notifier...`.  The reason being that on line 76 we are setting the `err` var, then on line 80 we were resetting it.  So then on line 87 we are returning nil, which results in a successful failure...  Not sure how to phrase that better...

.github/workflows/admin-create-aco-opensbx-deploy.yml
```diff
@@ -0,0 +1,24 @@
...
+      contents: read
+      id-token: write
+    runs-on: ubuntu-latest
+    environment: opensbx
+    steps:
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: ${{ vars.AWS_REGION }}
+          role-to-assume: arn:aws:iam::${{ secrets.ACCOUNT_ID }}:role/delegatedadmin/developer/bcda-prod-github-actions
+      - name: Promote lambda code from test to opensbx
```
carlpartridge, line 18: Sorry to be annoying but can we stick with `sbx` over `opensbx`?  Trying to standardize this as much as possible.  This applies to name of file, role we are assuming, bucket names, etc.

austincanada: Yep, do we want to rename the entire files' names as well?

.github/workflows/admin-create-aco-sbx-deploy.yml
```diff
@@ -0,0 +1,24 @@
...
+    permissions:
+      contents: read
+      id-token: write
+    runs-on: ubuntu-latest
+    environment: sbx
+    steps:
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: ${{ vars.AWS_REGION }}
+          role-to-assume: arn:aws:iam::${{ secrets.ACCOUNT_ID }}:role/delegatedadmin/developer/bcda-prod-github-actions
```
carlpartridge, line 17: Role should be `bcda-sbx-github-actions`

.github/workflows/admin-create-aco-sbx-deploy.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: ${{ vars.AWS_REGION }}
+          role-to-assume: arn:aws:iam::${{ secrets.ACCOUNT_ID }}:role/delegatedadmin/developer/bcda-sbx-github-actions
+      - name: Promote lambda code from test to sbx
+        run: |
+          aws s3 cp --no-progress \
+            s3://bcda-test-admin-create-aco-deny-function/function-${{ github.sha }}.zip \
+            s3://bcda-test-admin-create-aco-function/function-${{ github.sha }}.zip
```
carlpartridge, line 22: This should be sbx here instead of test right?  as we are copying from test bucket to sbx bucket?

.github/workflows/admin-create-aco-sbx-deploy.yml
```diff
@@ -0,0 +1,24 @@
...
+          aws lambda update-function-code --function-name bcda-test-admin-create-aco \
```
carlpartridge, line 23: And here

.github/workflows/admin-create-aco-sbx-deploy.yml
```diff
@@ -0,0 +1,24 @@
...
+            --s3-bucket bcda-test-admin-create-aco-function --s3-key function-${{ github.sha }}.zip
```
carlpartridge, line 24: And here

.github/workflows/admin-create-aco-prod-deploy.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: ${{ vars.AWS_REGION }}
+          role-to-assume: arn:aws:iam::${{ secrets.ACCOUNT_ID }}:role/delegatedadmin/developer/bcda-prod-github-actions
+      - name: Promote lambda code from test to prod
+        run: |
+          aws s3 cp --no-progress \
+            s3://bcda-test-admin-create-aco-function/function-${{ github.sha }}.zip \
+            s3://bcda-test-admin-create-aco-function/function-${{ github.sha }}.zip
```
carlpartridge, line 22: These should be prod

austincanada: The first instance should be kept as test, no?

carlpartridge: Yup, cp is copying {from thing} {to thing } 
https://docs.aws.amazon.com/cli/latest/reference/s3/cp.html

bcda/lambda/admin_create_aco/main.go
laurenkrugen-navapbc: hm, maybe we update to TODO with the usage for the smoke tests?

bcda/bcdacli/cli.go
```diff
@@ -45,8 +45,10 @@ import (
 )
 
 // App Name and usage.  Edit them here to prevent breaking tests
-const Name = "bcda"
-const Usage = "Beneficiary Claims Data API CLI"
+const (
```
laurenkrugen-navapbc, line 48: we should remove this cli command, since it will no longer be used.

bcda/bcdacli/cli_test.go
```diff
@@ -527,14 +532,18 @@ func (s *CLITestSuite) TestDenylistACO() {
 	notDenylistedCMSID := testUtils.RandomHexID()[0:4]
 	notFoundCMSID := testUtils.RandomHexID()[0:4]
 
-	denylistedACO := models.ACO{UUID: uuid.NewUUID(), CMSID: &denylistedCMSID,
+	denylistedACO := models.ACO{
+		UUID: uuid.NewUUID(), CMSID: &denylistedCMSID,
```
laurenkrugen-navapbc, line 536: the changes for this test are not needed; instead of removing these tests, let's just revert your changes. this is probably something we will eventually want to cleanup, but OK with leaving for now.

bcda/lambda/admin_create_aco/main.go
```diff
@@ -0,0 +1,148 @@
...
+	bcdaaws "github.com/CMSgov/bcda-app/bcda/aws"
+	"github.com/CMSgov/bcda-app/bcda/models"
+	"github.com/CMSgov/bcda-app/bcda/service"
+
+	log "github.com/sirupsen/logrus"
+)
+
+var (
+	slackChannel = "C034CFU945C" // #bcda-alerts
+	id           = uuid.NewRandom()
```
laurenkrugen-navapbc, line 26: This is a global variable, which we should try and avoid where we can. Can we move this declaration to `handler` since that's where it's initially used?

bcda/bcdacli/cli.go
```diff
@@ -626,7 +628,7 @@ func createGroup(id, name, acoID string) (string, error) {
 	return ssasID, nil
 }
 
-func createACO(name, cmsID string) (string, error) {
+func CreateACO(name, cmsID string) (string, error) {
```
carlpartridge, line 631: I think we should undo this as we are no longer using this (no longer needed as an exported function)..

bcda/lambda/admin_create_aco/main.go
```diff
@@ -0,0 +1,148 @@
...
+
+	bcdaaws "github.com/CMSgov/bcda-app/bcda/aws"
+	"github.com/CMSgov/bcda-app/bcda/models"
+	"github.com/CMSgov/bcda-app/bcda/service"
+
+	log "github.com/sirupsen/logrus"
+)
+
+var (
+	slackChannel = "C034CFU945C" // #bcda-alerts
```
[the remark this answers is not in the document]
carlpartridge: Seems slightly excessive to put it in param store (as it takes a few more steps to get out), maybe we can just put it Constants?

bcda/lambda/admin_create_aco/main.go
```diff
@@ -0,0 +1,148 @@
...
+		log.Errorf("Failed to handle Create ACO: %+v", err)
+		return err
+	}
+
+	log.Info("Completed Create ACO administrative task")
+
+	return nil
+}
+
+func handleCreateACO(ctx context.Context, conn PgxConnection, data payload, id uuid.UUID, notifier Notifier) error {
```
carlpartridge, line 87: Seems excessive to pass in a id/uuid here since we want a random uuid each time.

bcda/lambda/admin_create_aco/main_test.go
```diff
@@ -0,0 +1,92 @@
...
+type mockString struct{}
+
+func (ms mockString) Match(v any) bool {
+	_, ok := v.(string)
+	return ok
+}
+
+type mockUuid struct{}
+
+func (muuid mockUuid) Match(v any) bool {
```
carlpartridge, line 33: I feel like we probably dont need these mock strings/uuid structs?  Cant they just match on generic string?

bcda/lambda/admin_create_aco/db.go
```diff
@@ -0,0 +1,30 @@
...
+	QueryRow(context.Context, string, ...any) pgx.Row
+	Query(context.Context, string, ...any) (pgx.Rows, error)
+	Prepare(context.Context, string, string) (*pgconn.StatementDescription, error)
+}
+
+func createACO(ctx context.Context, conn PgxConnection, aco models.ACO) error {
+	query := `INSERT INTO acos (uuid, cms_id, client_id, name, termination_details) VALUES($1, $2, $3, $4, $5) RETURNING id;`
+	_, err := conn.Exec(ctx, query, aco.UUID, aco.CMSID, aco.ClientID, aco.Name, &models.Termination{})
+	if err != nil {
+		log.Errorf("Error running update query: %+v", err)
```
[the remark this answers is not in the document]
austincanada: Good catch, updated to be more descriptive.

bcda/lambda/admin_create_aco/db_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+func TestCreateACOSuccess(t *testing.T) {
+	ctx := context.Background()
+	mockConn, err := pgxmock.NewConn()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer mockConn.Close(ctx)
+
+	expectedSQL := "INSERT INTO acos (uuid, cms_id, client_id, name, termination_details) VALUES($1, $2, $3, $4, $5) RETURNING id;"
```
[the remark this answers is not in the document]
austincanada: Will also use the const for the other tests as well.

bcda/lambda/admin_create_aco/main_test.go
```diff
@@ -0,0 +1,92 @@
...
+	expectedSQL := "INSERT INTO acos (uuid, cms_id, client_id, name, termination_details) VALUES($1, $2, $3, $4, $5) RETURNING id;"
+	literalRegex := regexp.QuoteMeta(expectedSQL)
+
+	mockConn.ExpectExec(literalRegex).
+		WithArgs(
+			mockUuid{},
+			testACO.CMSID,
+			mockString{}, // mocking random clientid
+			testACO.Name,
+			testACO.TerminationDetails,
```
carlpartridge, line 56: Why cant we use testACO.uuid and testACO.clientID here?  We can set them in testACO above if we need to right?

bcda/lambda/admin_create_aco/db_test.go
```diff
@@ -0,0 +1,58 @@
...
+	literalRegex := regexp.QuoteMeta(expectedSQL)
+
+	mockConn.ExpectExec(literalRegex).
+		WithArgs(testACO.UUID, testACO.CMSID, testACO.ClientID, testACO.Name, testACO.TerminationDetails).
+		WillReturnError(errors.New("test error"))
+
+	err = createACO(ctx, mockConn, *testACO)
+	assert.ErrorContains(t, err, "test error")
+}
```
carlpartridge, line 58: Id love to see some tests that actually use the DB.  We can use a transaction to make sure we dont actually affect the DB.  There should be some examples in ACO Deny lambda db_test.

bcda/lambda/admin_create_aco/main.go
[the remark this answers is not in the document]
laurenkrugen-navapbc: you, me, and Parwinder all commented on this. Shall you and I resolve our comments and leave Parwinder's open?

bcda/lambda/admin_create_aco/main_test.go
laurenkrugen-navapbc: the type that get's passed in is an actual UUID (and the column in the DB is also UUID), so in addition to the generic mock type `string`, the`uuid` mock type was also needed!

bcda/lambda/admin_create_aco/main_test.go
laurenkrugen-navapbc: UUID is randomly generated with each invocation of the lambda and the clientId is just a stringified version of the UUID; to meet the expectations of the query being mocked, we used [pgxmock's recommendation](https://pkg.go.dev/github.com/pashagolub/pgxmock#section-readme:~:text=sql%20parsing%20matchers.-,Matching%20arguments%20like%20time.Time,-There%20may%20be) to create a mock type that just checks the type is being met but the actual value can be anything within that type. Ideal for situations when mocking the specific value is difficult or requires some uglier code changes to work.

bcda/lambda/admin_create_aco/main.go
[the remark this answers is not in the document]
laurenkrugen-navapbc: We're already retrieving values via param store in other places, so I would be game for putting this behind some more security.

bcda/lambda/admin_create_aco/main.go
```diff
@@ -0,0 +1,148 @@
...
+	var cmsIDPt *string
+	if data.CMSID != "" {
+		match := service.IsSupportedACO(data.CMSID)
+		if !match {
+			return errors.New("ACO CMS ID is invalid")
+		}
+		cmsIDPt = &data.CMSID
+	}
+
+	aco := models.ACO{Name: data.Name, CMSID: cmsIDPt, UUID: id, ClientID: id.String()}
```
carlpartridge, line 108: We are using the same `id` (randomized uuid) here for both UUID as well as ClientID?  That sounds wrong at a glance.

bcda/lambda/admin_create_aco/main_test.go
carlpartridge: Ok I see, right now we are passing in to handCreateACO() an `id` so we know ahead of time what this uuid is, so we should not need these mocks (I think).  But we are likely to remove that.  Then the id (uuid) will be randomized each time, so we will need these mocks.

austincanada: Same explanation as above comment.

bcda/lambda/admin_create_aco/main.go
austincanada: Pulled from original `createACO` cli action where the ClientID is just a stringified version of uuid. 

https://github.com/CMSgov/bcda-app/blob/main/bcda/bcdacli/cli.go#L644

bcda/bcdacli/cli.go
laurenkrugen-navapbc: yup, I would vote to remove entirely, tbh. If we can cleanup the cli package while we are doing this work instead of later, I think that'd be a big bonus.

bcda/lambda/admin_create_aco/db_test.go
austincanada: Do we accomplish those integration tests with the workflows that are available?

carlpartridge: Not sure I totally follow, do you mean the existing tests?  If so not really, just due to the fact that we are totally stubbing them out.  Or do you mean the integration test?  I would prefer beefed up unit tests and a few integration tests if that makes sense.

austincanada: The integration test workflow currently runs on the db. If we want tests using the actual db, would it be better to add them into the workflows instead of within the unit tests?

carlpartridge: I prefer robust unit tests.  Unit tests are definitely the place to be doing all of the various possible verifications of individual function functionality, as mentioned in my bullet point list above.  Integration tests are more suited to verify an entire process, as it goes through its various pieces.  So integration tests are generally a lot more set up and can leave artifacts so generally are just testing the happy path, whereas unit tests should try to account for all of a functions possible logic paths.

.github/workflows/admin-create-aco-test-integration.yml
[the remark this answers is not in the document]
austincanada: Added a flag to cleanup the DB once the integration test runs.

bcda/lambda/admin_create_aco/db_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+	params, err := getAWSParams()
+	if err != nil {
+		assert.FailNow(c.T(), "Failed to get AWS Params")
+	}
+
+	conn, err := pgx.Connect(c.ctx, params.dbURL)
+	if err != nil {
+		assert.FailNow(c.T(), "Failed to setup pgx connection")
+	}
```
carlpartridge, line 36: Do we need to make sure this thing closes properly? `defer conn.Close(c.ctx)`

bcda/lambda/admin_create_aco/db_test.go
```diff
@@ -0,0 +1,138 @@
...
+	assert.Equal(c.T(), count, 1)
+}
+
+func (c *CreateACOTestSuite) TestCreateACO_DupCMSID_Integration() {
+	// put original ACO into DB
+	_, err := c.tx.Exec(c.ctx, "INSERT INTO acos (name, uuid, cms_id) VALUES ('TEST ACO 2', $1, 'TEST501') RETURNING id;", uuid.New())
+	if err != nil {
+		assert.FailNow(c.T(), "Failed to insert ACO into database")
+	}
```
carlpartridge, line 122: Nit: This could just be a one liner: `assert.Nil(t, err)`.

bcda/lambda/admin_create_aco/db_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// insert ACO with duplicate CMS ID
+	c.T().Run("Duplicate CMSID", func(t *testing.T) {
+		err = createACO(c.ctx, c.tx, createACOModel("Test ACO 3", "TEST501"))
+		assert.Contains(t, err.Error(), "duplicate key")
+	})
+}
```
carlpartridge, line 129: I dont see a test checking for duplicate name, does the business logic and DB structure not care about that?

austincanada: That's correct, business logic and DB structure don't care about the name.

bcda/lambda/admin_create_aco/db_test.go
austincanada: Mainly using `FailNow()` to assert about errors on setup within the test and `Nil()` to assert about the test itself.

bcda/lambda/admin_create_aco/db_test.go
```diff
@@ -0,0 +1,138 @@
...
+	err = createACO(ctx, mockConn, *testACO)
+	assert.ErrorContains(t, err, "test error")
+}
+
+func (c *CreateACOTestSuite) TestCreateACO_Integration() {
+	var (
+		name  = "Test ACO 1"
+		cmsId = "TEST501"
+	)
+	testACO := createACOModel(name, cmsId)
```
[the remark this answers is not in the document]
austincanada: Updated the function name, but the `cmsId` for example is a different type and seems like a more readable way to know the variables are shared.

laurenkrugen-navapbc:
This looks good! The only thing that caught my eye was the integration test is passing but when I check the database, I don't see a record for the entry. Can you validate that it's writing to the database?

laurenkrugen-navapbc (approved):

PR #1055, opened 2025-02-20
BCDA-8760: First pass simplified migrate DB workflow

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8760

## 🛠 Changes

Added migrate DB workflow

## ℹ️ Context

Part of moving off of jenkins.

## 🧪 Validation

Local testing, linting

### DIFF
--- a/.github/workflows/migrate-db.yml
+++ b/.github/workflows/migrate-db.yml
@@ -1,0 +1,93 @@
+name: Migrate DB
+
+on:
+  push: # TEMP
+    branches:
+      carl/BCDA-8760-migrate-dbs
+  workflow_call:
+    inputs:
+      release_version:
+        description: 'Release version (or branch name)'
+        required: true
+        type: string
+      env:
+        description: 'Environment you want to migrate for (dev, test, opensbx, prod)'
+        required: true
+        type: string
+  workflow_dispatch:
+    inputs:
+      release_version:
+        description: 'Release version (or branch name)'
+        required: true
+        type: string
+      env:
+        description: 'Environment you want to migrate for (dev, test, opensbx, prod)'
+        required: true
+        default: dev
+        type: choice
+        options:
+          - dev
+          - test
+          - opensbx
+          - prod
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  migrate_db:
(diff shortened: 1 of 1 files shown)


.github/workflows/migrate-db.yml
```diff
@@ -0,0 +1,109 @@
...
+        with:
+          method: chat.postMessage
+          token: ${{ secrets.SLACK_BOT_TOKEN }}
+          # Sends to bcda-alerts
+          payload: |
+            channel: "C034CFU945C"
+            attachments:
+              - color: danger
+                text: "FAILURE: DB Migrations in ${{ inputs.env }} (run: <${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|${{ github.run_id }})>"
+                mrkdown_in:
```
austincanada, line 108: Is this a typo?

carlpartridge: Dont think so: https://github.com/slackapi/slack-github-action?tab=readme-ov-file#usage-2 (this is working elsewhere, just copy pasted these slack alerts around)

.github/workflows/migrate-db.yml
```diff
@@ -0,0 +1,115 @@
...
+        type: string
+      env:
+        description: 'Environment you want to migrate for (dev, test, opensbx, prod)'
+        required: true
+        default: dev
+        type: choice
+        options:
+          - dev
+          - test
+          - opensbx
```
austincanada, line 28: Did we want to change this to `sbx`?

carlpartridge: Unfortunately a majority of the related resources use `opensbx` over `sbx` :(

austincanada (approved):
LGTM

laurenkrugen-navapbc:
I think we briefly discussed this already but: this workflow looks like it does not take a migration as input; are we just going to run the latest generated migration and apply?

carlpartridge:
> I think we briefly discussed this already but: this workflow looks like it does not take a migration as input; are we just going to run the latest generated migration and apply?

Yup, no real reason why we would ever want to be messing around with schema versions in a long standing application.

PR #1070, opened 2025-03-19
BCDA-8637: refresh attribution data GHA

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8637

## 🛠 Changes

creating github action to refresh attribution data weekly.

## ℹ️ Context

Pulled the refresh out of smoke tests to reduce number of times it runs and so we can expand upon this for PY transitions or run it ad-hoc without smoke tests.

## 🧪 Validation

Works in all envs, verified in database and multiple successful test runs: https://github.com/CMSgov/bcda-app/actions/runs/13996655667/job/39193385477?pr=1070

### DIFF
--- a/.github/workflows/opt-out-import-test-integration.yml
+++ b/.github/workflows/opt-out-import-test-integration.yml
@@ -69,7 +69,7 @@ jobs:
           role-to-assume: arn:aws:iam::${{ secrets.ACCOUNT_ID }}:role/delegatedadmin/developer/bcda-test-github-actions
       - name: Install psql
         run: |
-          sudo amazon-linux-extras install postgresql14
+          sudo dnf install postgresql15 -y
       - name: Get database credentials
         uses: cmsgov/ab2d-bcda-dpc-platform/actions/aws-params-env-action@main
         env:
--- a/.github/workflows/opt-out-import-test-integration.yml
+++ b/.github/workflows/opt-out-import-test-integration.yml
@@ -69,7 +69,7 @@ jobs:
           role-to-assume: arn:aws:iam::${{ secrets.ACCOUNT_ID }}:role/delegatedadmin/developer/bcda-test-github-actions
       - name: Install psql
         run: |
-          sudo dnf install postgresql15 -y
+          sudo dnf install postgresql16 -y
       - name: Get database credentials
         uses: cmsgov/ab2d-bcda-dpc-platform/actions/aws-params-env-action@main
         env:
--- a/.github/workflows/refresh_test_attribution.yml
+++ b/.github/workflows/refresh_test_attribution.yml
@@ -53,14 +53,14 @@ jobs:
               HOST=$(aws rds describe-db-instances --query 'DBInstances[*].[Endpoint.Address]' --output text | grep bcda-$key 2>&1)
               if [ $? -ne 0 ]; then
                   echo "Error: failed to retrieve database host."
-                  exit
+                  exit 1
               fi
               CONNECTION_URL=$(echo $value 2>&1 | sed -E "s/@.*\/bcda/\@$HOST\/bcda/" 2>&1)
               ./scripts/refresh_test_attribution.sh -c "${CONNECTION_URL}" -f > /dev/null 2>&1
 
               if [ $? -ne 0 ]; then
                   echo "Error: Refresh failed; run script locally for more details."
-                  exit
+                  exit 1
               else 
                   echo "Success: refresh complete."
               fi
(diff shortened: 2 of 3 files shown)


.github/workflows/opt-out-import-test-integration.yml
```diff
@@ -69,7 +69,7 @@ jobs:
           role-to-assume: arn:aws:iam::${{ secrets.ACCOUNT_ID }}:role/delegatedadmin/developer/bcda-test-github-actions
       - name: Install psql
         run: |
-          sudo amazon-linux-extras install postgresql14
+          sudo dnf install postgresql15 -y
```
laurenkrugen-navapbc, line 72: copied this command originally and discovered it no longer worked; updating it now.

carlpartridge: I believe we are on postgres 16 now, and soon to be on 17, does that affect this?

.github/workflows/refresh_test_attribution.yml
```diff
@@ -0,0 +1,81 @@
...
+          environments["opensbx"]=$SBX_DB
+          environments["prod"]=$PROD_DB
+
+          for key in "${!environments[@]}"; do
+              value="${environments[$key]}"
+
+              HOST=$(aws rds describe-db-instances --query 'DBInstances[*].[Endpoint.Address]' --output text | grep bcda-$key 2>&1)
+              if [ $? -ne 0 ]; then
+                  echo "Error: failed to retrieve database host."
+                  exit
```
carlpartridge, line 56: I think we probably want to mark a specific code here, like exit 1, otherwise it defaults to 0 which denotes success.

.github/workflows/refresh_test_attribution.yml
```diff
@@ -0,0 +1,81 @@
...
+              fi
+              CONNECTION_URL=$(echo $value 2>&1 | sed -E "s/@.*\/bcda/\@$HOST\/bcda/" 2>&1)
+              ./scripts/refresh_test_attribution.sh -c "${CONNECTION_URL}" -f > /dev/null 2>&1
+
+              if [ $? -ne 0 ]; then
+                  echo "Error: Refresh failed; run script locally for more details."
+                  exit
```
carlpartridge, line 63: ~`Exit 1` here too

.github/workflows/refresh_test_attribution.yml
```diff
@@ -0,0 +1,81 @@
...
+          for key in "${!environments[@]}"; do
+              value="${environments[$key]}"
+
+              HOST=$(aws rds describe-db-instances --query 'DBInstances[*].[Endpoint.Address]' --output text | grep bcda-$key 2>&1)
+              if [ $? -ne 0 ]; then
+                  echo "Error: failed to retrieve database host."
+                  exit
+              fi
+              CONNECTION_URL=$(echo $value 2>&1 | sed -E "s/@.*\/bcda/\@$HOST\/bcda/" 2>&1)
+              ./scripts/refresh_test_attribution.sh -c "${CONNECTION_URL}" -f > /dev/null 2>&1
```
carlpartridge, line 59: SuperNit: If we already have a script we are running to do the work, wondering if we should ship a bunch of this steps logic into that script?

.github/workflows/refresh_test_attribution.yml
```diff
@@ -0,0 +1,81 @@
...
+
+              if [ $? -ne 0 ]; then
+                  echo "Error: Refresh failed; run script locally for more details."
+                  exit
+              else 
+                  echo "Success: refresh complete."
+              fi
+              done
+      - name: Failure Alert
```
carlpartridge, line 68: Do we want a success alert?  The main advantage I see is that we would know for sure that this was running regularly (at the cost of some extra noise).

.github/workflows/refresh_test_attribution.yml
laurenkrugen-navapbc: kept this in here because I wanted to keep the script simple to run ad-hoc and locally and also did not want it to update all envs when it runs locally. imo, the benefits of moving more of this to the script vs keeping it here are low/none, since this will be automated 99% of the time.

.github/workflows/refresh_test_attribution.yml
laurenkrugen-navapbc: I'm OK with keeping it as a failure only, since this would make the smoke tests in dev generate some noise during our daily deploy.

.github/workflows/opt-out-import-test-integration.yml
laurenkrugen-navapbc: upgrading to 16 works - will hold off until platform has completed all upgrades.

PR #1077, opened 2025-03-28
BCDA-8795: Add CDAC CSV attribution logic

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8795

## 🛠 Changes

Added CDAC CSV attribution handling and config.

## ℹ️ Context

Part of onboarding CDAC and handling their attribution files.

## 🧪 Validation

Local linting and testing


.github/workflows/cclf-import-test-integration.yml
```diff
@@ -55,8 +55,8 @@ jobs:
           date=$(date +'%y%m%d')
           time=$(date +'%H%M%S')
 
-          fname=T.BCD.A0001.ZCY24.D${date}.T${time}1
-          cclf8_fname=T.BCD.A0001.ZC8Y24.D${date}.T${time}1
+          fname=T.BCD.A0001.ZCY${year}.D${date}.T${time}1
+          cclf8_fname=T.BCD.A0001.ZC8Y${year}.D${date}.T${time}1
```
carlpartridge, line 59: These were failing as later on we were trying to find files set to PY24 but were being set to the current year.  This likely has been failing since Jan 1, however we only run this integration test on changes to the cclf import dirs.

laurenkrugen-navapbc: thanks for catching this!

carlpartridge: I meant to mention that I did not add a new test case to the integration tests.  I figure that we are currently testing both a cclf file and a csv file, and the only difference with the new code is the filename, which I believe is covered well enough via unit-testing.

laurenkrugen-navapbc:
Have we done a verification that the new file to be ingested can be yet? if not, can we verify before merge?

laurenkrugen-navapbc (approved):

PR #1086, opened 2025-04-10
Replace watch.sh with docker watch

michaeljvaldes:
**Closed in favor of https://github.com/CMSgov/bcda-app/pull/1087**

## 🎫 Ticket

## 🛠 Changes

A script that performs hot reloading, watch.sh, was replaced with docker watch. More changes were added to support quicker docker builds.
Also, the docker stop grace period has been reduced to 1s from 10s so that containers are force-stopped sooner (since they are designed to ignore graceful stop signals anyway).

## ℹ️ Context

The current entrypoint to the docker containers is a script called watch.sh, which checks whether the environment is CI, and if not, it immediately reinstalls the go module once and then reinstalls again on any changes to the source code (which are mounted from the parent file system). Since the first reinstall doesn't have access to the go build cache within the docker build cache, it introduces > 2 minutes of build time to re-download go dependencies every time the docker containers are run.

This PR removes the watch.sh script and functionally replaces it with docker watch. Now when the docker containers are run with `--watch` they rebuild and rerun automatically. Since the rebuilds have access to the docker build cache (and therefore the go build cache), they don't introduce that extra > 2 min  of go dependencies re-downloading. Unfortunately, this means that hot reloads are slightly slower (~6s from about ~3s with the watch.sh script).

ℹ️ **Note: Since watch.sh does a CI environment check, the changes from this PR should only impact local development.**

### Pros of this PR:
- significantly faster first build of api/worker docker containers (> 2 min faster)
- clearer architecture with no extra script that has to do an environment check while the application is running (entrypoint is now the app cli itself)
- source code is no longer mounted to the docker container, which brings us a step closer to being able to use our docker images properly in a deployed environment
- docker watch rebuilds on dependency/tool changes too, not just code changes

### Cons of this PR:
- hot reloads upon source code changes are slower (~6s from about ~3s with the watch.sh script)

## 🧪 Validation

To test this PR:
1. start the containers `docker compose up`
2. monitor the container logs when they start (perhaps via docker desktop) and verify that they don't include logs from `go install` (since we are no longer re-installing and re-downloading all the go dependencies on boot)
3. stop the services (this happens quicker now too), then run them again with the watch flag `docker compose up --watch` -- since the docker build was cached in step 1, this build and run should take mere seconds. verify that you can interact with the api (perhaps via the _health endpoint) soon after you run the containers, certainly long before 2 minutes it took previously
4. make a code change to any bcda api code file (perhaps by adding a commented newline) and verify that the services rebuild quickly.
5. make a change to a dependency in go.mod (perhaps by upgrading a dependency version) and verify that the services rebuild (albeit a little more slowly than before)
6. repeat above steps for worker code

**In particular, pay attention to how long it _feels_ like it takes for the containers to rebuild. The developer experience is very important here -- is the extra time a deal-breaker, and not worth the pros of this PR?**


Dockerfiles/Dockerfile.bcdaworker_prod
```diff
@@ -18,14 +17,18 @@ COPY . .
...
-# syntax = docker/dockerfile:1-experimental
+# ------------------------------------------------------------------------
 FROM golang:1.23.1-alpine3.20
 ARG ENVIRONMENT
 # only add dev packages if the environment argument was set to development
 RUN [ "$ENVIRONMENT" != "development" ] || apk add git openssl entr bash
 WORKDIR /go/src/github.com/CMSgov/bcda-app/bcdaworker
+
+# This file is needed on the worker container by hcc.go -- similar requirement in build_and_package.sh
+COPY --from=builder /go/src/github.com/CMSgov/bcda-app/bcda/models/fhir/alr/utils/hcc_crosswalk.tsv /etc/sv/worker/hcc_crosswalk.tsv
```
michaeljvaldes, line 30: While this looks a little ugly, this file from the bcda api is required on the worker file system. This mirrors a requirement of our current non-docker packaging process in `build_and_package.sh`

docker-compose.yml
```diff
@@ -23,37 +23,47 @@ services:
...
       - queue
+    develop:
+      watch:
+        - action: rebuild
+          path: ./bcda
+          ignore: ["!**/*.go"]
+        - action: rebuild
+          path: .
+          ignore: ["!go.mod", "!go.sum", "!tools.go"]
+    stop_grace_period: 1s
```
michaeljvaldes, line 44: As an additional change, whenever we ask the docker containers to stop, the length of time docker waits before force stopping has been reduced from 10s to 1s. I made this a default because the api services are designed to ignore gentle stop commands.

Dockerfiles/Dockerfile.bcdaworker_prod
```diff
@@ -1,4 +1,3 @@
-# syntax = docker/dockerfile:1-experimental
```
michaeljvaldes, line 1: Removed references to a different docker frontend, which slightly speeds up the image build. My theory is that this was originally added to enable run cache mounting.

carlpartridge (commented):
Looks like we might need to include more files, seeing this error: ```sh: ../scripts/bulk_import_synthetic_cclf_package.sh: not found```

Generally looking good!  Once you get a ticket up you should rename the title of this PR to "BCDA-{ticket#}: Replace..." as well as the branch name as well (possibly easier to just create a new branch?)

PR #1087, opened 2025-04-10
BCDA-8975: Replace watch.sh with docker watch

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8975

## 🛠 Changes

- A script that performs hot reloading, watch.sh, was replaced with docker watch. More changes were added to support quicker docker builds.
- The entire source code repo was previously mounted onto the container. Now the dockerfile instructions accurately capture the filesystem of the resulting container.
- The docker stop grace period has been reduced to 1s from 10s so that containers are force-stopped sooner (since they are designed to ignore graceful stop signals anyway).

## ℹ️ Context

The current entrypoint to the docker containers is a script called watch.sh, which checks whether the environment is CI, and if not, it immediately reinstalls the go module once and then reinstalls again on any changes to the source code (which are mounted from the parent file system). Since the first reinstall doesn't have access to the go build cache within the docker build cache, it introduces > 2 minutes of build time to re-download go dependencies every time the docker containers are run.

This PR removes the watch.sh script and functionally replaces it with docker watch. Now when the docker containers are run with `--watch` they rebuild and rerun automatically. Since the rebuilds have access to the docker build cache (and therefore the go build cache), they don't introduce that extra > 2 min  of go dependencies re-downloading. Unfortunately, this means that hot reloads are slightly slower (~6s from about ~3s with the watch.sh script).

ℹ️ **Note: Since watch.sh does a CI environment check, the changes from this PR should only impact local development.**

### Pros of this PR:
- significantly faster first build of api/worker docker containers (> 2 min faster)
- clearer architecture with no extra script that has to do an environment check while the application is running (entrypoint is now the app cli itself)
- source code is no longer mounted to each docker container; this clarifies exactly which directories are included/needed in the final container and brings us a step closer to being able to use our docker images properly in a deployed environment
- to add to the previous point, more directories have been _removed_ from the final docker container that maybe shouldn't be there, including `/scripts`
- docker watch rebuilds on dependency/tool changes too, not just code changes

### Cons of this PR:
- hot reloads upon source code changes are slower (~6s from about ~3s with the watch.sh script)
- `make load-synthetic-cclf-data` takes slightly longer now since it was refactored to make more docker run calls (though I imagine it can be made more efficient without too much effort)

## 🧪 Validation

To test this PR:
1. start the containers `docker compose up`
2. monitor the container logs when they start (perhaps via docker desktop) and verify that they don't include logs from `go install` (since we are no longer re-installing and re-downloading all the go dependencies on boot)
3. stop the services (this happens quicker now too), then run them again with the watch flag `docker compose up --watch` -- since the docker build was cached in step 1, this build and run should take mere seconds. verify that you can interact with the api (perhaps via the _health endpoint) soon after you run the containers, certainly long before 2 minutes it took previously
4. make a code change to any bcda api code file (perhaps by adding a commented newline) and verify that the services rebuild quickly.
5. make a change to a dependency in go.mod (perhaps by upgrading a dependency version) and verify that the services rebuild (albeit a little more slowly than before)
6. repeat above steps for worker code

### Things to pay special attention to
- In particular, pay attention to how long it _feels_ like it takes for the containers to rebuild. The developer experience is very important here -- is the extra time a deal-breaker, and not worth the pros of this PR?
- Additionally, please scrutinize the instructions for the makeup of the container file system. Since we've removed the bind mound, we've removed a layer of obfuscation and we need to specifically define what will go in the final image. Are we including all of the directories that should be necessary?


Makefile
```diff
@@ -124,26 +124,32 @@ load-fixtures: reset-db
...
+	docker run --rm --network bcda-app-net willwill/wait-for-it ssas:3003 -t 100
 
 	# Additional fixtures for postman+ssas
 	docker compose run db psql -v ON_ERROR_STOP=1 "postgres://postgres:toor@db:5432/bcda?sslmode=disable" -f /var/db/postman_fixtures.sql
 
 load-synthetic-cclf-data:
 	$(eval ACO_SIZES := dev dev-auth dev-cec dev-cec-auth dev-ng dev-ng-auth dev-ckcc dev-ckcc-auth dev-kcf dev-kcf-auth dev-dc dev-dc-auth small medium large extra-large)
 	# The "test" environment provides baseline CCLF ingestion for ACO
-	docker compose run --rm api sh -c "../scripts/bulk_import_synthetic_cclf_package.sh test ' ' $(ACO_SIZES)"
+	for ACO_SIZE in $(ACO_SIZES) ; do \
```
carlpartridge, line 136: If Im reading this right, we are now looping in here as opposed to passing in an ~array to the script itself?

Makefile
```diff
@@ -124,26 +124,32 @@ load-fixtures: reset-db
 
 	# Ensure components are started as expected
 	docker compose up -d api worker ssas
-	docker run --rm --network bcda-app-net willwill/wait-for-it api:3000 -t 300
-	docker run --rm --network bcda-app-net willwill/wait-for-it ssas:3003 -t 300
+	docker run --rm --network bcda-app-net willwill/wait-for-it api:3000 -t 100
+	docker run --rm --network bcda-app-net willwill/wait-for-it ssas:3003 -t 100
```
carlpartridge, line 128: Nit: Do we still need 100 here?  Can/should we lower it?

Makefile
michaeljvaldes: That's right; while this is less efficient since we now have to loop docker runs instead of performing a loop within the container itself, I preferred to pull the /scripts fully out of the docker container since they seem out of place there.
I think if we felt that this needed to be more efficient, my first inclination would be to update the CLI to accept arrays instead of adding /scripts back into the container.

Makefile
michaeljvaldes: Lowered it to 30s, but I welcome suggestions for specific wait times.

scripts/watch.sh
laurenkrugen-navapbc, line 1: excited to see one less script!

PR #1088, opened 2025-04-11
BCDA-8750: Make bulkRequest asynchronous

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8750

## 🛠 Changes

Requests:
- getQueJobs removed from bulkRequest() to queueing package under PrepareWorker.

Service:
- slight refactoring of GetLatestCCLF file so it can be used to validate that an ACO has attribution before starting a job. Used to be validated in getQueJobs, which is a performance issue.
-  getbeneficiaries and getnewbeneficiaries had duplicate code which now lives in Service.GetLatestCCLF().
- Updated Mock Service to match updates to GetLatestCCLF

Queueing:
- Creation of new worker and job "kind" to build sub jobs for parent job. "PrepareJobWorker".
- Moved Mock Enqueuer to queueing package to reduce circular dependencies and updated mock methods.
- Moved workers into their own files for easier reading.

Other:
- created a constants file for bcdaworker. for the sake of keeping this already large PR small, only put one in there that was in use during this work.
- Some unexported structs/variables have been made exported for access. 

Dependencies:
- updated river to 0.20 so that rivertest can be used to verify Work() and successful enqueueing of jobs.
- the above required an update to PGX, which is now on 5.7.4

## ℹ️ Context

These changes have been made due to performance issues for jobs with a large amount of benes. By building the job asynchronously, the initial request for a job should be much faster. 

Side notes: 

## 🧪 Validation

Updated and modified tests; still ongoing.


bcda/api/requests.go
```diff
@@ -570,79 +563,23 @@ func (h *Handler) bulkRequest(w http.ResponseWriter, r *http.Request, reqType se
...
-		Resources: resourceTypes,
-
-		CMSID: ad.CMSID,
-		ACOID: newJob.ACOID,
-
-		JobID:           newJob.ID,
-		Since:           rp.Since,
-		TransactionTime: newJob.TransactionTime,
-		CreationTime:    time.Now(),
+	pjob := queueing.PrepareJobArgs{
```
carlpartridge, line 566: Theres a lot of complexity and comments above here about setting up a transaction in order to create jobs and subjobs.  I think we could probably cut a lot of that out?

bcdaworker/queueing/enqueue.go
```diff
@@ -18,9 +24,35 @@ import (
 // Enqueuer only handles inserting job entries into the appropriate table
 type Enqueuer interface {
 	AddJob(ctx context.Context, job models.JobEnqueueArgs, priority int) error
+	AddPrepareJob(ctx context.Context, job PrepareJobArgs) error
 	AddAlrJob(job models.JobAlrEnqueueArgs, priority int) error
 }
 
+/* Creates a river client for the PrepareJob queue. If the client does not call .Start(), then it is insert only.*/
+func NewPrepareEnqueuer() Enqueuer {
```
carlpartridge, line 32: I dont think we need this?  If its insert only then any enqueuer should be able to handle this work.  In the existing Enqueuer I doubt we need to define workers as well.  All of our jobs currently share the same (default) queue, which is just a column name defined in `river_job` table.

The reason we have two river clients is because we run river in two different ec2 instances (one the API, which is insert only, and one on Worker, where it actually handles the work).

bcdaworker/queueing/enqueue.go
```diff
@@ -95,6 +133,7 @@ func (q riverEnqueuer) AddJob(ctx context.Context, job models.JobEnqueueArgs, pr
 	// This could also be refactored to a batch insert: riverClient.InsertManyTx or riverClient.InsertMany
 	_, err := q.Insert(ctx, job, &river.InsertOpts{
 		Priority: priority,
+		Queue:    river.QueueDefault,
```
carlpartridge, line 136: Nit: Likely defaults to this anyways without this line.

bcdaworker/queueing/enqueue.go
```diff
@@ -106,3 +145,14 @@ func (q riverEnqueuer) AddJob(ctx context.Context, job models.JobEnqueueArgs, pr
 func (q riverEnqueuer) AddAlrJob(job models.JobAlrEnqueueArgs, priority int) error {
 	return nil
 }
+
+func (q riverEnqueuer) AddPrepareJob(ctx context.Context, job PrepareJobArgs) error {
+	_, err := q.Insert(ctx, job, &river.InsertOpts{
```
carlpartridge, line 150: Nit: Probably just need `q.Insert(ctx, job, nil)` here as well (defaults to default queue).

bcdaworker/queueing/prepare_worker.go
```diff
@@ -0,0 +1,198 @@
+/*
+Prepare Worker takes all of the arguments of a bulk export request and prepares all of the
```
carlpartridge, line 2: Nit: ~ "Prepare Worker takes all of the arguments of a bulk export request to BCDA API and asynchronously prepares and creates (enqueues?) all of the subjobs needed for the requests bulk export main job."

bcdaworker/queueing/prepare_worker.go
```diff
@@ -0,0 +1,198 @@
...
+	TransactionID   string
+}
+
+func (args PrepareJobArgs) Kind() string {
+	return constants.PrepareJobKind
+}
+
+// PrepareJobWorker has two BFD clients because it depends on a configuration variable that is not available until Work() is called.
+// There were other discussed methods of injecting the client and overwriting the the basepath but ruled out due to the risk and time constraints.
+// Many of the Service's functionality is used solely in this PrepareJob functionality and should eventually be migrated when time allows.
```
carlpartridge, line 45: Nit: What is "the Service" (can we add the path to it)?

bcdaworker/queueing/enqueue.go
```diff
@@ -18,9 +24,35 @@ import (
...
+			Workers: workers,
+		})
+		if err != nil {
+			panic(err)
+		}
+
+		return riverEnqueuer{riverClient}
+	}
+
+	return queEnqueuer{que.NewClient(database.QueueConnection)}
```
carlpartridge, line 52: QueueConnection seems wrong, doesnt that reference the connection to the old queue-go DB?  I would imagine we would want the Pgxv5Connection here?

bcdaworker/queueing/prepare_worker.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (w *PrepareJobWorker) Work(ctx context.Context, rjob *river.Job[PrepareJobArgs]) error {
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	ctx = log.NewStructuredLoggerEntry(log.Worker, ctx)
+	_, logger := log.SetCtxLogger(ctx, "transaction_id", rjob.Args.TransactionID)
+	ctx = context.WithValue(ctx, m.CtxTransactionKey, uuid.New())
```
carlpartridge, line 83: Why do we need this?  Is it that different from transaction_id?

bcdaworker/queueing/prepare_worker.go
```diff
@@ -0,0 +1,198 @@
...
+
+	exports, since, err := w.prepareExportJobs(ctx, rjob.Args)
+	if err != nil {
+		// TODO update job in jobs table as failed
```
carlpartridge, line 87: Just tagging these TODOs for myself as they seem important.

bcdaworker/queueing/prepare_worker.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	client := river.ClientFromContext[pgxv5.Tx](ctx)
+	q := riverEnqueuer{client}
+	err = w.queueExportJobs(ctx, q, rjob.Args, exports, since)
+	if err != nil {
+		// TODO update job in jobs table as failed
+		logger.Errorf("failed to add jobs to the main queue: %s", err)
+		return err
+	}
```
carlpartridge, line 99: Looks like we are setting up a mini, insert only client here?

bcdaworker/queueing/prepare_worker.go
```diff
@@ -0,0 +1,198 @@
...
+
+	return nil
+
+}
+
+// prepareExportJobs builds a list of jobs to be processed based on the parent job.
+func (p *PrepareJobWorker) prepareExportJobs(ctx context.Context, args PrepareJobArgs) ([]*models.JobEnqueueArgs, time.Time, error) {
```
carlpartridge, line 106: Im guessing this is mostly copy/pasted out of the service?

bcdaworker/queueing/worker_processjob.go
```diff
@@ -0,0 +1,68 @@
+package queueing
```
carlpartridge, line 1: Nit: Filename: `worker_process_job.go`

bcdaworker/queueing/river.go
```diff
@@ -1,3 +1,17 @@
+/*
+Package queueing implements "job" processing for bulk export requests
```
carlpartridge, line 2: I think I would broaden this.  River is our job handling package.  It handles the following:
- ProcessJob: Main job, ie bulk export requests
- PrepareJob: Handles logic dedicated to creating subjobs for ProcessJob
- CleanupJob: Handles cleaning up old/archived bulk export job files.

Then maybe move some of the more specific descriptions to the appropriate worker file?

bcdaworker/queueing/worker_cleanup.go
```diff
@@ -0,0 +1,131 @@
+package queueing
```
carlpartridge, line 1: Love moving these jobs to their own files.

bcda/api/requests.go
laurenkrugen-navapbc: comment-wise or code-wise?

bcdaworker/queueing/enqueue.go
laurenkrugen-navapbc: Nope, I can rename it back to NewEnqueue() -- this was left in because I was testing the client but will remove.

bcdaworker/queueing/enqueue.go
laurenkrugen-navapbc: this is copy + paste from the existing NewEnqueuer(), so this is how it is currently setup for the old que-go library.

bcda/api/requests.go
carlpartridge: Both, feels like lines ~ 518 - 560 are unnecessary?

bcdaworker/queueing/enqueue.go
carlpartridge: ohhhh, yes, ok that makes sense.  the old que-go library should be using the queue-db...

bcdaworker/queueing/worker_processjob.go
michaeljvaldes: How do we feel about moving the workers to the `worker` package and dropping 'worker' from their file names?

michaeljvaldes (commented):
I love the refactoring you were able to add to this PR, and I'm excited to remove the que-go stuff soon 🎉

bcda/lambda/admin_aco_deny/db_test.go
```diff
@@ -50,63 +47,64 @@ func TestDenyACOsQueryFailure(t *testing.T) {
...
-	}
-
-	// double check we are finding the appropriate amount of rows
-	assert.Equal(t, i, 5)
-
-	// cleanup
-	err = tx.Rollback(ctx)
-	assert.Nil(t, err)
-}
+// TODO <ticket number> this test is failing due to pgx version update needed for riverqueue tests
```
carlpartridge, line 50: Nit: TODO

bcda/service/service.go
```diff
@@ -345,6 +346,7 @@ func (s *service) getNewAndExistingBeneficiaries(ctx context.Context, conditions
 	// Retrieve beneficiaries from either:
 	// - The newest CCLF file in the last X days (where X is the environment's configured cutoff duration)
 	// - OR the newest CCLF file prior to the requested attributionDate
+	// TODO: service has a method called GetLatestCCLFFile; update
```
carlpartridge, line 349: Nit: TODO

bcda/service/service.go
```diff
@@ -661,6 +663,41 @@ func (s *service) GetLatestCCLFFile(ctx context.Context, cmsID string, fileType
 	return cclfFile, nil
 }
 
+func (s *service) GetLatestCCLFFilet(ctx context.Context, rc RequestConditions, lowerBound time.Time, upperBound time.Time, fileType models.CCLFFileType) (*models.CCLFFile, error) {
```
carlpartridge, line 666: Is this on purpose? GetLatestCCLFFile**t**()
Is this called anywhere else or is it a duplicate of the above?

bcdaworker/queueing/enqueue.go
carlpartridge: Any trouble changing this back to the original?

bcdaworker/queueing/worker_prepare_test.go
```diff
@@ -0,0 +1,295 @@
...
+	svc := &service.MockService{}
+	c := new(client.MockBlueButtonClient)
+	c.On("GetPatient", mock.Anything, "0").Return(&fhirModels.Bundle{}, nil)
+	worker := &PrepareJobWorker{svc: svc, v1Client: c, v2Client: c, r: s.r}
+	_, err := worker.GetBundleLastUpdated(basepath, models.JobEnqueueArgs{})
+	assert.Nil(s.T(), err)
+}
+
+func (s *PrepareWorkerIntegrationTestSuite) TestQueueExportJobs() {
+}
```
carlpartridge, line 295: TODO or unneded?

bcdaworker/queueing/worker_prepare.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	args.Job.JobCount = len(exports)
+
+	return exports, args.Since, err
+}
+
+// GetBundleLastUpdated requests a fake patient in order to acquire the bundle's lastUpdated metadata.
+func (p *PrepareJobWorker) GetBundleLastUpdated(basepath string, jobData models.JobEnqueueArgs) (time.Time, error) {
+	switch basepath {
+	case "/v1/fhir":
```
carlpartridge, line 172: "/v1/fhir" probably better to make a constant?  Im seeing 9 references to it.  (Same with v2)

bcdaworker/queueing/enqueue.go
laurenkrugen-navapbc: yeah, I did have some issues with que-go related tests breaking. I was going to ask how we felt about keeping these separate, since we were going to deprecate/remove que-go eventually.

bcda/service/service.go
laurenkrugen-navapbc: This is my mistake - I missed this during cleanup. I was updating the service function because I needed a way to check attribution before starting an export job. I will update this and push the changes.

bcdaworker/queueing/worker_prepare_test.go
laurenkrugen-navapbc: Missed this one -- adding it!

bcdaworker/queueing/prepare_worker.go
laurenkrugen-navapbc: correct!

bcdaworker/queueing/prepare_worker.go
laurenkrugen-navapbc: also correct!

bcdaworker/queueing/worker_processjob.go
laurenkrugen-navapbc: filename updated. I do think the worker package is a misnomer; the workers are defined in queueing. Carl and I discussed possibly reserving the worker package for business type logic, but it seems to be spread out across the queueing package. We should discuss more!

bcda/service/service.go
```diff
@@ -661,6 +662,41 @@ func (s *service) GetLatestCCLFFile(ctx context.Context, cmsID string, fileType
...
+	// 		cutoffTime = time.Now().Add(-1 * s.stdCutoffDuration)
+	// 	} else if rc.fileType == models.FileTypeRunout && s.rp.cutoffDuration > 0 {
+	// 		cutoffTime = time.Now().Add(-1 * s.rp.cutoffDuration)
+	// 	}
+	// }
+
+	// // new bene request
+	// if s.stdCutoffDuration > 0 && rc.attributionDate.IsZero() {
+	// 	cutoffTime = time.Now().Add(-1 * s.stdCutoffDuration)
+	// }
```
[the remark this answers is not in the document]
laurenkrugen-navapbc: yeah, I thought I had updated the code to use this commented out function but missed this as well. This will be the final change needed (🤞 ) other than PR feedback changes.

laurenkrugen-navapbc:
Had some discrepancies between tests on my machine and github actions so I reverted the docker changes temporarily to see if that was a contributing factor. Doesn't look like it was, so I will be removing those from the PR tomorow!

bcdaworker/constants/constants.go
```diff
@@ -0,0 +1,5 @@
+package constants
+
+const PrepareJobKind = "PrepareJob"
+const BFDV1Path = "/v1/fhir"
+const BFDV2Path = "/v2/fhir"
```
carlpartridge, line 5: Super Nit: Should these constants be in our BFD client (bluebutton.go)?

bcdaworker/queueing/enqueue.go
carlpartridge: Its basically the old thing but with some slight modifications right?  Where else do we use the old one?

bcdaworker/queueing/worker_prepare_test.go
```diff
@@ -0,0 +1,295 @@
...
+func TestCleanupTestSuite(t *testing.T) {
+	suite.Run(t, new(PrepareWorkerIntegrationTestSuite))
+}
+
+func (s *PrepareWorkerIntegrationTestSuite) SetupTest() {
+	db, _ := databasetest.CreateDatabase(s.T(), "../../db/migrations/bcda/", true)
+	tf, err := testfixtures.New(
+		testfixtures.Database(db),
+		testfixtures.Dialect("postgres"),
+		testfixtures.Directory("testdata/"),
```
carlpartridge, line 49: Nit: Where does it make sense to keep these fixtures?  Close to the domain or in a shared fixture space?

laurenkrugen-navapbc: what we've been doing so far is each package has a test fixtures folder, since they fixtures may vary by testing needs. this also reduces circular dependencies, but I wouldn't be opposed to changing this in the future.

bcdaworker/constants/constants.go
laurenkrugen-navapbc: I moved them into the worker to reduce the circular dependency issue and to stay consistent with the structure we currently have, which is a singular constants package for the `bcda` folder and one for `bcdaworker`. also not opposed to changing this in the future.

bcdaworker/queueing/enqueue.go
laurenkrugen-navapbc: yes, and as far as I am aware, it's just breaking que-go tests. I'll update with a comment!

bcdaworker/constants/constants.go
carlpartridge: This feels like a shared thing, both BCDA API and Workers likely want to know this info, so a third place inbetween seems like a good place for this.  However its also fine for now.

bcdaworker/queueing/worker_prepare_test.go
carlpartridge: As far as I can tell this function enqueues an array of jobs, do we want/need to test that here?  Or is that tested elsewhere?  Or is that not necessary?

bcda/service/service.go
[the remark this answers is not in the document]
laurenkrugen-navapbc: marking this comment as closed, since we will be doing this as part of a separate ticket.

bcdaworker/constants/constants.go
laurenkrugen-navapbc: an alternative is a constants package that is at the root level, similar as we do for some other packages.

laurenkrugen-navapbc: I will leave for now so the PR doesn't get any larger but will make a ticket.

bcda/api/requests.go
```diff
@@ -515,9 +509,17 @@ func (h *Handler) bulkRequest(w http.ResponseWriter, r *http.Request, reqType se
...
+	// _, err = h.Svc.GetLatestCCLFFile(ctx, ad.CMSID, time.Time{}, time.Time{}, models.CCLFFileType(reqType))
+	// if ok := goerrors.As(err, &service.CCLFNotFoundError{}); ok {
+	// 	logger.Error("No update to date attribution information for: %s", ad.CMSID)
+	// 	h.RespWriter.Exception(r.Context(), w, http.StatusInternalServerError, responseutils.NotFoundErr, "Unable to perform export operations for this Group. No up-to-date attribution information is available. Usually this is due to awaiting new attribution information at the beginning of a Performance Year.")
+	// 	return
+	// } else if err != nil {
+	// 	logger.Error("failed to get latest cclf file: %s", err)
+	// 	h.RespWriter.Exception(r.Context(), w, http.StatusInternalServerError, responseutils.InternalErr, "")
+	// 	return
+	// }
```
carlpartridge, line 521: Still need this commented out code?

carlpartridge:
Looks good.  Holding off on approving/merging until we get through the current large set of major changes currently in the pipeline.

carlpartridge (approved):

PR #1094, opened 2025-04-18
BCDA-8509: InProgress JobStatus response always shows 0%

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8509

## 🛠 Changes

- the method `GetJobAndKeys` in `bcda/service/service.go` now fetches job keys even when the job does not have the "Completed" status
- tests are added to validate the behavior of `GetJobAndKeys`
- tests are added to validate the behavior of the "X-Progress" header

## ℹ️ Context

The method `GetJobAndKeys` in `bcda/service/service.go` only fetched and returned job keys if the job had the "Completed" status. This contradicted the comment just above the status check, which implied that the keys should only be fetched if the job _didn't_ have the "Completed" status. Since the keys seemed to be required for jobs with the Completed status and otherwise, I simply removed the check.

## 🧪 Validation

- I created tests that check the conditions of this ticket, as well as other conditions
- I created a new job request with 3 sub jobs on my local environment, paused the bcda worker after it completed one sub job, and validated that the job status displayed "33%" progress

### DIFF
--- a/bcda/service/service.go
+++ b/bcda/service/service.go
@@ -190,10 +190,6 @@ func (s *service) GetJobAndKeys(ctx context.Context, jobID uint) (*models.Job, [
 		return nil, nil, err
 	}
 
-	// No need to look up job keys if the job is complete
-	if j.Status != models.JobStatusCompleted {
-		return j, nil, nil
-	}
 	keys, err := s.repository.GetJobKeys(ctx, jobID)
 	if err != nil {
 		return nil, nil, err
--- a/bcda/api/requests_test.go
+++ b/bcda/api/requests_test.go
@@ -819,6 +819,51 @@ func (s *RequestsTestSuite) TestJobStatusErrorHandling() {
 	}
 }
 
+func (s *RequestsTestSuite) TestJobStatusProgress() {
+	tests := []struct {
+		testName         string
+		status           models.JobStatus
+		expectedProgress string
+	}{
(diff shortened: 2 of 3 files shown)


bcda/service/service.go
```diff
@@ -190,10 +190,6 @@ func (s *service) GetJobAndKeys(ctx context.Context, jobID uint) (*models.Job, [
 		return nil, nil, err
 	}
 
-	// No need to look up job keys if the job is complete
-	if j.Status != models.JobStatusCompleted {
-		return j, nil, nil
-	}
```
carlpartridge, line 196: I was originally a bit concerned that this generic function might be used in different areas of the codebase for different reasons, thereby removing this could have unintended consequences.  That doesnt seem to be the case, it seems to only be used in the JobStatus request.  Having said that, this early return could be a performance adjustment (if checking on a job that is completed then we dont need to find all jobKeys).  I think we can ship this with your adjustments and if we do see some performance issues then we can take that up then (likely make changes in JobStatus, not here).

carlpartridge (approved):
Love the unit and integration tests!

bcda/service/service.go
michaeljvaldes: I had the same concerns, though based on the contradiction between the code and the comment it was difficult to tell exactly which scenarios were intended to be more efficient 😕

PR #1095, opened 2025-04-18
BCDA-8975: Fix worker docker container and enable debugging

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8975

## 🛠 Changes

- fixed a problem with the worker docker container by including the `shared_files` directory in its file system
- added a docker volume to share files in the FHIR_ PAYLOAD_DIR between the api and worker containers
- adjusted the debug docker compose file to enable debugging with our new docker setup, and updated docs to make it user-friendly

## ℹ️ Context

The changes from the first BCDA-8975 PR more clearly defined the file system of the docker containers, and the bind mount to the host file system was removed. I missed that the `share_files` directory was required by the worker container, which caused it to fail when it picked up jobs. Adding the directory explicitly into the container should address this problem. **On a related note, we should keep an eye out for any other directories or files that may have been missed.** This should give us the opportunity to better organize our code and the requirements of the program in general.

Another issue we were seeing in smoke tests is that the api and worker containers expect to share a `FHIR_PAYLOAD_DIR` directory, which was previously being provided by default when the entire repo directory was a bind mount. Now the single directory is shared explicitly via a docker volume. Ideally we would like to define the directory path within the containers via configuration, but since the filepath is specified in many tests, it will be more work to untangle it.

I also (re?)enabled docker container bcda app debugging. Follow the readme "Debug" instructions to test it out! Try debugging the api, attaching to it with vscode, and making a request to an endpoint like `_metadata` to view the local variables at execution. Let me know if the instructions are unclear, or if you think the `make debug-<>` commands should work differently. On the make commands, I assumed that if you're debugging, you probably want to enable watch as well.

## 🧪 Validation

- I tested new jobs on my local environment and verified that the worker completed them as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,7 +225,7 @@ make performance-test
 ### **Update Database Seed Data**
 
 After the user has finished updating the Postgres database used for unit testing with the new data, the user can update
-the seed data by running the following comamnd:
+the seed data by running the following command:
 
 ```sh
 make unit-test-db-snapshot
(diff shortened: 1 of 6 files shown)


carlpartridge (approved):
Looks good to me!  I would get Laurens approval as well as her current work adjusted some of this stuff.

laurenkrugen-navapbc (approved):

README.md
```diff
@@ -309,3 +309,31 @@ docker exec -it bcda-app-api-1 sh -c 'bcda -h'
 
 Follow installing go + vscode [setup guide](https://marketplace.visualstudio.com/items?itemName=golang.go#getting-started).
 Additional settings found under `.vscode/settings.json` allow tests to be run within vscode.
+
+## Debugging
```
laurenkrugen-navapbc, line 313: thanks for updating the readme with the debugging steps! 🎉

PR #1099, opened 2025-04-23
BCDA-8959: Run inferno tests script

colby-seyferth-nava:
## 🎫 Ticket

### DIFF
--- a/inferno/config.json
+++ b/inferno/config.json
@@ -1,0 +1,52 @@
+{
+  "bulk_url": "https://sandbox.bcda.cms.gov/api/v2/",
+  "auth": {
+    "token_endpoint": "https://sandbox.bcda.cms.gov/auth/token",
+    "client_id": "",
+    "client_secret": ""
+  },
+  "required_tests": [
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_server_tests_stu2-bulk_data_server_tls_version_stu2",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_group_export_v200-bulk_data_group_export_group_stu2-bulk_data_group_export_operation_support",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_group_export_v200-bulk_data_group_export_group_stu2-bulk_data_group_no_auth_reject",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_group_export_v200-bulk_data_group_export_group_stu2-bulk_data_group_kick_off",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_group_export_v200-bulk_data_group_export_group_stu2-bulk_data_group_status_check",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_group_export_v200-bulk_data_group_export_group_stu2-bulk_data_group_output_check",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_group_export_v200-bulk_data_group_export_validation_stu2-bulk_file_server_tls_version",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_group_export_v200-bulk_data_group_export_validation_stu2-bulk_data_group_ndjson_download",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_group_export_v200-bulk_data_group_export_cancel_group_stu2-bulk_data_export_group_cancel",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_group_export_v200-bulk_data_group_export_parameters_group-output_format_in_group_export_response",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_group_export_v200-bulk_data_group_export_parameters_group-since_in_group_export_response",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_patient_export_v200-bulk_data_patient_export_group_stu2-bulk_data_patient_export_operation_support",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_patient_export_v200-bulk_data_patient_export_group_stu2-bulk_data_patient_no_auth_reject",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_patient_export_v200-bulk_data_patient_export_group_stu2-bulk_data_patient_kick_off",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_patient_export_v200-bulk_data_patient_export_group_stu2-bulk_data_patient_status_check",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_patient_export_v200-bulk_data_patient_export_group_stu2-bulk_data_patient_output_check",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_patient_export_v200-bulk_data_patient_export_validation_stu2-bulk_file_server_tls_version",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_patient_export_v200-bulk_data_patient_export_validation_stu2-bulk_data_patient_ndjson_download",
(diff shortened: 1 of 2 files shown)


inferno/run-inferno.sh
```diff
@@ -0,0 +1,161 @@
...
+fi
+
+### TODO: Cloning the bulk test repo, running the docker compose/build, and
+### breaking things down at the end will all be handled outside of this script.
+### As part of automating the tests, we will remove these steps from here. This
+### is just for making running the script locally a little easier.
+
+# Clone test kit repo
+if [ ! -d "bulk-data-test-kit" ] ; then
+  git clone https://github.com/inferno-framework/bulk-data-test-kit.git
```
michaeljvaldes, line 33: I don't know if it's conventional to clone another repo into our repo, even if we're removing it later. I would add this subdirectory to .gitignore just in case.
I'll brainstorm alternatives; it would be very convenient if images built by this repo were published.

inferno/run-inferno.sh
```diff
@@ -0,0 +1,161 @@
...
+git checkout 957190d889b544d573766d1c1178da22496ce62a
+
+# Build / Compose Inferno Docker container
+./setup.sh
+docker compose build
+docker compose up -d
+sleep 10
+
+# Stop the hl7 validator one - it takes a lot of time / memory to run, and those tests are not applicable
+docker stop bulk-data-test-kit-hl7_validator_service-1
```
michaeljvaldes, line 46: If we know what services we want, we can specify them and combine a few steps here.
```suggestion
docker compose up --build -d inferno worker nginx redis
```
```

File: inferno/run-inferno.sh [hunk 3]
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+# Get Config
+CONFIG=`cat config.json`
+BULK_URL=$(echo $CONFIG | jq -r ".bulk_url")
+TOKEN_URL=$(echo $CONFIG | jq -r ".auth.token_endpoint")
+CLIENT_ID=$(echo $CONFIG | jq -r ".auth.client_id")
+CLIENT_SECRET=$(echo $CONFIG | jq -r ".auth.client_secret")
+TESTS=$(echo $CONFIG | jq -r ".required_tests.[]")
```
michaeljvaldes, line 10: This line only worked on my machine without a period
```suggestion
TESTS=$(echo $CONFIG | jq -r ".required_tests[]")
```

inferno/run-inferno.sh
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env bash
```
michaeljvaldes, line 1: It might be useful to add a command in our Makefile to run script. In the future, we could also have it depend on another step that sets up test data to run this locally (perhaps as part of a CI test).

michaeljvaldes (changes requested):
Left a few comments; some are requested changes, others are just food for thought.
Generally this is a really cool feature that I hope we can integrate into our automated testing at some point.

inferno/run-inferno.sh
michaeljvaldes: Possible alternative: build the docker image directly from the github repo and use our own docker-compose file! Something like 
` docker build -t test_inferno https://github.com/inferno-framework/bulk-data-test-kit.git#957190d889b544d573766d1c1178da22496ce62a`

colby-seyferth-nava: Good idea. Ya, I plan on removing this altogether in the next PR, but I'll:
 - add to .gitignore
 - move the `rm` command up earlier so we are less likely to quit the process before deleting the folder

inferno/config.json
```diff
@@ -0,0 +1,52 @@
...
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_system_export_v200-bulk_data_system_export_validation_stu2-bulk_data_system_ndjson_download",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_system_export_v200-bulk_data_system_export_validation_stu2-bulk_data_system_valid_resources",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_system_export_v200-bulk_data_system_export_validation_stu2-bulk_data_system_multiple_patients",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_system_export_v200-bulk_data_system_export_cancel_group_stu2-bulk_data_system_export_cancel",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_system_export_v200-bulk_data_system_export_cancel_group_stu2-bulk_data_system_export_cancel_stu2",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_system_export_v200-bulk_data_system_export_parameters_group-output_format_in_system_export_response",
+    "bulk_data_v200-bulk_data_export_tests_v200-bulk_data_system_export_v200-bulk_data_system_export_parameters_group-since_in_system_export_response"
```
carlpartridge, line 50: Where do these test names come from?  Can we wildcard them?  Would it be helpful to have a list of tests we dont run?

I think it would be helpful to add a comment at the top describing what and how this file is used.

inferno/config.json
```diff
@@ -0,0 +1,52 @@
+{
```
carlpartridge, line 1: Wondering if we could change the base name of the directory from `inferno` to `fhir_testing` or something along those lines.  Generally prefer things to be less focused on the specific tool and more what the goal is.

inferno/run-inferno.sh
colby-seyferth-nava: ya, agree on both parts: script in a docker image; inferno server in a docker image; automated by GHA on merge. I was planning on doing that in the next ticket. 

This ticket was supposed to just focus on the inferno test part. going thru the steps run the tests and process the results. Ideally for the next ticket we can keep that logic, and re-home it to the proper spot

inferno/run-inferno.sh
```diff
@@ -0,0 +1,161 @@
...
+echo "
+SUMMARY:
+ - Tests Passed: ${#PASS_RESULTS[@]}
+ - Tests Failed: ${#FAIL_RESULTS[@]}
+ "
+
+for i in "${FAIL_RESULTS[@]}"; do
+   echo "FAIL - $i"
+done
+
```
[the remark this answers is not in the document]
colby-seyferth-nava: Yaaa, pretty much. a couple of early exit fails if things go wrong. so exit success means hey we got to the end and successfully ran the tests. 

but also success/fail doesn't mean anything yet because it's not plugged in to the workflow or anything 😅

inferno/run-inferno.sh
```diff
@@ -0,0 +1,161 @@
...
+### As part of automating the tests, we will remove these steps from here. This
+### is just for making running the script locally a little easier.
+
+# Clone test kit repo
+if [ ! -d "bulk-data-test-kit" ] ; then
+  git clone https://github.com/inferno-framework/bulk-data-test-kit.git
+fi
+
+cd bulk-data-test-kit
+git checkout 957190d889b544d573766d1c1178da22496ce62a
```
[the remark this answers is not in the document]
colby-seyferth-nava: ya, just wanted to have *a* consistent one, but it doesn't have to be *this* specific one

inferno/config.json
colby-seyferth-nava: That is just a list of all the tests that are run. 
- The "required" list is the baseline for what we support today. 
- The "other" tests are the ones that the framework supports, but we do not pass them (either they are N/A, or we actually fail because we need to update our API)

we could probably just use the last bit (except for the TLS ones)

Maybe a readme would help?

inferno/config.json
colby-seyferth-nava: nice, i like that

inferno/run-inferno.sh
colby-seyferth-nava: this is how I had it, then it wouldn't work unless i put the dot. then it was fine, but i probably changed something else too. but anyways, I removed the dot because ya I think this is correct. good find!

carlpartridge (approved):

michaeljvaldes (approved):

PR #1104, opened 2025-04-30
BCDA-8998: check for attribution status

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8998

## 🛠 Changes

Refactor to allow for verification of valid attribution files before accepting a job request.

## ℹ️ Context

This was an issue we found as we refactored to async job request processing.

## 🧪 Validation

Local linting and testing.


.mockery.yml
```diff
@@ -0,0 +1,22 @@
+all: false
```
carlpartridge, line 1: Mockery v2 upgrade.  Specifically spells out the interfaces we want to mock at the bottom of this file.

bcda/utils/datetime_helpers.go
```diff
@@ -0,0 +1,8 @@
+package utils
```
carlpartridge, line 1: Maybe a more specific `datetime_helpers` package would be better than lumping this in generic utils?

bcdaworker/constants/constants.go
```diff
@@ -1,5 +0,0 @@
-package constants
-
-const PrepareJobKind = "PrepareJob"
-const BFDV1Path = "/v1/fhir"
-const BFDV2Path = "/v2/fhir"
```
carlpartridge, line 5: There was a clash of constants across bcda/constants package and bcdaworker/constants package.  Extracted this out into bcda/constants, it probably best lives somewhere inbetween.

bcdaworker/queueing/worker_prepare.go
```diff
@@ -124,11 +109,11 @@ func (p *PrepareJobWorker) prepareExportJobs(ctx context.Context, args PrepareJo
...
-	id, err := safecast.ToInt(args.Job.ID) // NOSONAR
-	if err != nil {                        // NOSONAR
-		logger.Error(err)               // NOSONAR
-		return exports, args.Since, err // NOSONAR
-	} // NOSONAR
+	id, err := safecast.ToInt(args.Job.ID)
+	if err != nil {
+		logger.Error(err)
+		return exports, args.Since, err
+	}
```
carlpartridge, line 116: 99% sure the `// NOSONAR` didnt do anything...

bcdaworker/queueing/worker_types/worker_types.go
```diff
@@ -0,0 +1,30 @@
+package worker_types
```
carlpartridge, line 1: Moved this out to its package to avoid circular dependencies.  I think generally speaking these types of structs should be in their own package?

docker-compose.yml
```diff
@@ -26,6 +26,8 @@ services:
     command: ["bcda", "start-api"]
     env_file:
       - ./shared_files/decrypted/local.env
+    environment:
+      - FHIR_STAGING_DIR=/go/src/github.com/CMSgov/bcda-app/bcdaworker/data/tmpdata
```
carlpartridge, line 30: This was breaking worker processing, as it was failing to write between two different volumes (FHIR_STAGING_DIR vs FHIR_PAYLOAD_DIR).  This env var is generally set in your local.env but I think this issue was hard enough to find that it makes the most sense to force it here.

michaeljvaldes: Something to maybe look into (either for this ticket or another day): do we need to worry about `FHIR_TEMP_DIR` and `FHIR_ARCHIVE_DIR` as well?

michaeljvaldes (approved):
I would definitely wait for an approval from Lauren, but at a high level this looks good to me!

docker-compose.yml
carlpartridge: Thats a good point, I think the default env var is set to bcda-app/bcdaworker/data, which is what the volume we mount is set to, but forcing that in the env var makes sense as well.

bcda/api/requests.go
```diff
@@ -31,35 +31,30 @@ import (
...
+	supportedStatuses      map[models.JobStatus]struct{}
+	bbBasePath             string
+	apiVersion             string
+	RespWriter             fhirResponseWriter
+	// cfg                    *service.Config
 
 	// Needed to have access to the repository/db for lookup needed in the bulkRequest.
 	// TODO (BCDA-3412): Remove this reference once we've captured all of the necessary
 	// logic into a service method.
 	r  models.Repository
```
laurenkrugen-navapbc, line 56: nit: looks like we don't actually use Handler.r anywhere. might be useful to remove?

bcda/api/requests.go
```diff
@@ -551,23 +568,34 @@ func (h *Handler) bulkRequest(w http.ResponseWriter, r *http.Request, reqType se
...
+	// lots of things needed for downstream logic!
+	prepJob := worker_types.PrepareJobArgs{
+		Job:                    newJob,
+		ACOID:                  acoID,
+		CMSID:                  ad.CMSID,
+		CCLFFileNewID:          cclfFileNew.ID,
+		CCLFFileOldID:          cclfFileOldID,
+		BFDPath:                h.bbBasePath,
+		RequestType:            reqType,
+		ComplexDataRequestType: complexDataRequestType,
```
laurenkrugen-navapbc, line 580: Do we need to add this additional field or could we potentially use `RequestType`, since that looks to be the [ same set of info?](https://github.com/CMSgov/bcda-app/blob/main/bcda/service/service.go#L50)?

laurenkrugen-navapbc: Also - I don't see this in use anywhere after it's added as a job argument. Does it need to be a job arg or can it just be scoped to a `bulkRequest` variable?

bcda/api/requests.go
```diff
@@ -509,37 +502,61 @@ func (h *Handler) bulkRequest(w http.ResponseWriter, r *http.Request, reqType se
...
+
+	cclfFileNew, err := h.Svc.GetLatestCCLFFile(
+		ctx,
+		ad.CMSID,
+		cutoffTime,
+		timeConstraints.AttributionDate,
+		fileType,
+	)
+	if err != nil {
+		logger.Error("error finding latest cclf file")
```
laurenkrugen-navapbc, line 532: should we add the error message to the logger here?

`logger.Error("error finding latest cclf file: %s", err)`

bcda/api/requests.go
```diff
@@ -509,37 +502,61 @@ func (h *Handler) bulkRequest(w http.ResponseWriter, r *http.Request, reqType se
...
+	performanceYear := utils.GetPY()
+	if fileType == models.FileTypeRunout {
+		performanceYear -= 1
+	}
 
-		if err = tx.Commit(); err != nil {
-			logger.Error(err.Error())
+	// validate cclffile and PY
+	if cclfFileNew == nil || performanceYear != cclfFileNew.PerformanceYear {
+		logger.Error("failed to validate cclf file or performance year of found cclf file")
```
laurenkrugen-navapbc, line 544: same as above comment!

bcda/api/requests.go
```diff
@@ -509,37 +502,61 @@ func (h *Handler) bulkRequest(w http.ResponseWriter, r *http.Request, reqType se
...
 	if err != nil {
-		err = fmt.Errorf("failed to start transaction: %w", err)
 		logger.Error(err)
 		h.RespWriter.Exception(r.Context(), w, http.StatusInternalServerError, responseutils.InternalErr, "")
 		return
 	}
-	// Use a transaction backed repository to ensure all of our upserts are encapsulated into a single transaction
-	rtx := postgres.NewRepositoryTx(tx)
 
-	defer func() {
```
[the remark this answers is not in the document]
carlpartridge: That should still be the case, just happens that there is now only 1 step that adds/modifies the DB which is creating the job itself.  If that fails we return err.

bcda/bcdacli/cli_test.go
```diff
@@ -406,7 +406,6 @@ func (s *CLITestSuite) TestImportCCLFDirectory() {
 	}
 
 	for _, tc := range tests {
-		fmt.Printf("\n-----path %+v,\n   expected: %+v\n", tc.path, tc.expectedLogs)
```
laurenkrugen-navapbc, line 409: just cleanup?

bcda/api/requests.go
carlpartridge: Ooh nice, at one point it was used deeper down, but I guess at some point that was refactored away.  Will remove!

bcda/api/requests.go
carlpartridge: Its used to create the job on 559: `newJob.ID, err = h.r.CreateJob(ctx, newJob)`.  Could/should be refactored to go through service?

bcda/service/config.go
```diff
@@ -41,14 +41,14 @@ type Config struct {
 	ACOConfigs []ACOConfig `conf:"aco_config"`
 
 	// Un-exported fields that are computed using the exported ones above
```
laurenkrugen-navapbc, line 43: nit: can we update the comment or remove since this field is now exported?

bcda/bcdacli/cli_test.go
carlpartridge: Yeah, I had left a bunch in requests_test.go earlier, so did a project wide search.  Found a few in a few different files that I had added in a few months ago.

bcda/service/service.go
```diff
@@ -282,32 +297,33 @@ func (s *service) createQueueJobs(ctx context.Context, conditions RequestConditi
 						// data ingestion timelines don't line up, therefore for all
 						// partially-adjudicated jobs we will just use conditions.CreationTime as an
 						// upper bound
+						// fmt.Printf("\n--- setting transaction time: %+v, %+v, %+v", dataType, args.CreationTime, args.Job.TransactionTime)
```
laurenkrugen-navapbc, line 300: nit: can we remove this?

bcdaworker/queueing/worker_types/worker_types.go
laurenkrugen-navapbc: nit 1: the worker is defined under the queueing package but the args are defined here. Possibly rename to `arg_types`? `job_args`?

nit 2: if we do move the args here, should we do the same for the clean up worker and process job worker?

bcdaworker/queueing/worker_prepare_test.go
```diff
@@ -94,18 +96,23 @@ func (s *PrepareWorkerIntegrationTestSuite) TestPrepareExportJobsDatabase_Integr
...
 			}
 
-			if tt.qErr {
+			if tt.bfdErr {
+				// code returns before GetQueJobs
+			} else if tt.qErr {
 				svc.On("GetQueJobs", testUtils.CtxMatcher, mock.Anything).Return([]*models.JobEnqueueArgs{}, errors.New("an error occurred"))
 			} else {
-				svc.On("GetQueJobs", testUtils.CtxMatcher, mock.Anything).Return([]*models.JobEnqueueArgs{{ID: 1}}, nil)
+				svc.On("GetQueJobs", testUtils.CtxMatcher, mock.Anything).Return([]*models.JobEnqueueArgs{{ID: 52}}, nil)
```
laurenkrugen-navapbc, line 115: why ID = 52 out of curiosity?

carlpartridge: I spent a long time trying to figure out some test that was leaving data in the DB, so I changed all the IDs I could find.  I think I may have changed some back?

bcdaworker/queueing/worker_types/worker_types.go
carlpartridge: I think I prefer 'worker' as we have 'worker' files and types and as far as I can tell that is also the preferred name used by these things by river itself.  I think I also prefer 'types' over 'args' as it feels a bit broader.  Happy to update if you have strong preferences.

laurenkrugen-navapbc: oof, yeah I'm more than OK than holding off on that. Should we make a TODO placeholder for that work in the interim? As for preferences, not super strong so whatever you think makes sense.

bcda/api/requests.go
laurenkrugen-navapbc: got it - if it's used later on, that works for me!

bcda/api/requests.go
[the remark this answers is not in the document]
laurenkrugen-navapbc: We have logging from the create queue jobs, so I think as long as we keep our logging frequent and correlated (which it should all be correlated via the transaction_id), then it should be fine?

bcdaworker/queueing/worker_prepare_test.go
laurenkrugen-navapbc: hmmm in theory the ids should be starting at 1, although I think a fair amount of tests don't do database cleanup. if tests are passing then I think we are good and that indicates that there is some form of cleanup happening.

laurenkrugen-navapbc (approved):

PR #1114, opened 2025-05-07
BCDA-9061: update config file regex

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9061

## 🛠 Changes

Test case was added for "DA\d{4}" and config file was modified so that "A\d{4}" is now "^A\d{4}". Latter now matches "A####" exactly rather than a substring.

## ℹ️ Context

Regex matching for ACO names was inadvertently capturing substrings instead of the exact match, resulting in incorrect match being returned.

## 🧪 Validation

Added test case for unit test. Configuration file needs to be updated in all env configuration management: devs - can we verify the config file reflects the update in lower envs?


carlpartridge (commented):
Can we update the repos config files as well?  (I know they arent directly used but I think its useful to keep things in sync if possible).  (conf/configs/{env}.yml)

carlpartridge:
Seeing some (seemingly unrelated) attr file test failures as well?

carlpartridge (approved):

PR #1115, opened 2025-05-08
BCDA-9065: Increase retries, fail on max retries, fix flaky test

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9065

## 🛠 Changes

Force failure on reaching max retries.  Add another fail status to a known (very) flaky smoke test.

## ℹ️ Context

Previously smoke tests would pass even if we were not getting successful job responses.  Meaning we were not fully testing whether or not jobs were being processed.  This could be a sign that the worker instance isnt up and running properly.

## 🧪 Validation

Local smoketest run.


juliareynolds-nava (approved):

laurenkrugen-navapbc:
nit: ticket number in title if there is one?

laurenkrugen-navapbc (approved):
Thanks for updating! 🎉

carlpartridge:
> nit: ticket number in title if there is one?

Good catch, ty!

PR #1119, opened 2025-05-13
BCDA-9067: Adjust loggers to include transaction_id for SSAS

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9067

## 🛠 Changes

Added transaction id across a few different loggers as well as to auth requests to SSAS.

## ℹ️ Context

Currently our logging of auth requests is lacking, often times we are seeing nothing when auth requests fail.

## 🧪 Validation

Local linting and testing.


bcda/auth/mock_provider.go
```diff
@@ -1,8 +1,7 @@
-// Code generated by mockery v2.14.0. DO NOT EDIT.
```
carlpartridge, line 1: Not happy about doing this (removing this line of code) however it wont let me update the mocked function without removing this line.  I am updating mockery to v3 in the large refactoring to async requests and did not want to replicate those changes in this PR due to conflict worries.  The end result will be that this file will need to be regenerated after (which I am tracking).

log/logger.go
```diff
@@ -122,8 +122,11 @@ func GetCtxEntry(ctx context.Context) *StructuredLoggerEntry {
 
 // Gets the logrus.FieldLogger from a context
 func GetCtxLogger(ctx context.Context) logrus.FieldLogger {
-	entry := ctx.Value(CtxLoggerKey).(*StructuredLoggerEntry)
-	return entry.Logger
+	entry := ctx.Value(CtxLoggerKey)
+	if entry != nil {
+		return entry.(*StructuredLoggerEntry).Logger
+	}
+	return API
```
carlpartridge, line 129: This should pretty much always be available (its set at the end of the pre-request middlewares).  However those middlewares are skipped in a lot of tests meaning this is unset, which causes a panic.  Catching this is probably still useful for prod, as panic-ing when its not found isnt great either.  This will now just return the baseline logger if not found.

bcda/auth/client/ssas.go
```diff
@@ -309,6 +311,9 @@ func (c *SSASClient) GetToken(credentials Credentials, r http.Request) (string,
 		return "", &customErrors.InternalParsingError{Err: err, Msg: constants.RequestStructErr}
 	}
 
+	req.Header.Add(client.TransactionIDHeader, r.Context().Value(middleware.CtxTransactionKey).(string))
+	// the following is more or less a duplicate of the above, however it does not match what SSAS is expecting
```
michaeljvaldes, line 315: I can't find any use of the existing `transaction-id` header, so I'm a fan of just removing it.

michaeljvaldes (approved):
Looks good to me! Left one small nitpick, and I think the only thing to be concerned about is the removal of the mockery comment, which you're already on top of 🚀

PR #1121, opened 2025-05-15
BCDA-9059: add make fhir_testing command

colby-seyferth-nava:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9059
https://jira.cms.gov/browse/BCDA-9058

## 🛠 Changes

 - Added a command to the Makefile to run the FHIR inferno tests on your local BCDA deployment
 - increased the inferno timeout from 180 seconds to 300 seconds
 - remove config that is no longer used
 - add FHIR testing readme

## ℹ️ Context

These changes add a command to run the inferno tests locally. before, the only way to run is by manually setting it up and running it, or by running the FHIRScan workflow, but that doesn't point to local deployment. With this command, you can make sure the FHIR conformance still passes before you merge / deploy your changes. 

## 🧪 Validation

run `make fhir_testing` and see that it runs all the way thru and shows you your results


carlpartridge (approved):
Looks good to me!  Love the detailed docs!

PR #1128, opened 2025-05-16
BCDA-9064: Add aco config to prepare job data

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9064

## 🛠 Changes

Add ACO config values (data types) to prepare job args

## ℹ️ Context

Currently prepare jobs are failing on not being able to find ACO configs.  I believe this is due to worker instance not having aco config.yml available.  This would also explain how it got past unit testing, docker composet smoke tests, but failed on actual public url smoke tests.

## 🧪 Validation

Local linting and testing.

### DIFF
--- a/bcda/service/service.go
+++ b/bcda/service/service.go
@@ -288,55 +288,50 @@ func (s *service) createQueueJobs(ctx context.Context, args worker_types.Prepare
 			rowCount++
 			jobIDs = append(jobIDs, fmt.Sprint(b.ID))
 			if len(jobIDs) >= maxBeneficiaries || rowCount >= len(beneficiaries) {
-				if acoConfig, ok := s.GetACOConfigForID(args.CMSID); ok {
-					// Create separate jobs for each data type if needed
-					for _, dataType := range acoConfig.Data {
-						// conditions.TransactionTime references the last time adjudicated data
-						// was updated in the BB client. If we are queuing up a partially-adjudicated
-						// data job, we need to assume that the adjudicated and partially-adjudicated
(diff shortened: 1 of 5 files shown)


bcda/service/service.go
```diff
@@ -288,55 +288,50 @@ func (s *service) createQueueJobs(ctx context.Context, args worker_types.Prepare
...
-									TransactionID:   ctx.Value(middleware.CtxTransactionKey).(string),
-									TransactionTime: transactionTime,
-									BBBasePath:      args.BFDPath,
-									DataType:        dataType,
-								}
-
-								s.setClaimsDate(&enqueueArgs, args)
-
-								jobs = append(jobs, &enqueueArgs)
+				// Create separate jobs for each data type if needed
```
carlpartridge, line 291: Almost all of the changes here are indentation.  Removed the `if acoConfig, ok := s.GetACOConfigForID(args.CMSID); ok {` block and its else check, now using `args.ACOConfigDataTypes` now.

bcda/service/service_test.go
```diff
@@ -1058,9 +1063,10 @@ func (s *ServiceTestSuite) TestGetQueJobsByDataType_Integration() {
 		expTxTime          time.Time
 		resourceTypes      []string
 		terminationDetails *models.Termination
+		dataTypes          []string
```
carlpartridge, line 1066: This is the test that checks Data Types.

laurenkrugen-navapbc (approved):

PR #1130, opened 2025-05-21
BCDA-9102: Make sure config is available for worker instance

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9102

## 🛠 Changes

Not represented in this PR:
- Added param store values for BCDA_WORKER_CONFIG_PATH to point at (cloned) worker config.yml files (also in param store).  This doubles the config.yml files we have to keep in sync but was the easiest path forward.  (The alternative is messing around with an archaic script in bcda-ops, something that we have spent a lot of time trying to refactor as part of our AL2023 work).

In this PR:
- Add error checks for when we fail to load ACO Config appropriately.
- Add tests for the above and update existing.
- Some cleanup of now unused functions.

## ℹ️ Context

Previously if we failed to load ACO Configs we would just silently ignore the related logic blocks.  This included important logic checks.

## 🧪 Validation

Local linting and testing.

### DIFF
--- a/bcda/service/service.go
+++ b/bcda/service/service.go
@@ -456,7 +456,7 @@ func (s *service) getBenesByFileID(ctx context.Context, cclfFileID uint, args wo
 				}
 			}
 		} else {
-			return nil, errors.New("failed to access ACO Config, possibly failing to ignore suppressed MBIs")
+			return nil, fmt.Errorf("failed to load or match ACO config (or potentially no ACO Configs set), CMS ID: %+v", args.CMSID)
 		}
 
 	}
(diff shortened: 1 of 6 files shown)


bcda/service/service.go
```diff
@@ -474,11 +481,11 @@ func (s *service) setClaimsDate(args *worker_types.JobEnqueueArgs, prepareArgs w
...
 	// Applies the lower bound from the first matching ACOConfig
-	for _, cfg := range s.acoConfigs {
-		if cfg.patternExp.MatchString(prepareArgs.CMSID) {
-			args.ClaimsWindow.LowerBound = cfg.LookbackTime()
-			break
-		}
+	cfg, ok := s.GetACOConfigForID(prepareArgs.CMSID)
+	if ok {
+		args.ClaimsWindow.LowerBound = cfg.LookbackTime()
+	} else {
```
laurenkrugen-navapbc, line 487: does this need to fail if we can't set the claims date? the only other reference I see to this function doesn't do anything sort of error handling and there's no return. The claims window also determines the level of access for bene data, so maybe we should fail the job if this fails?

bcdaworker/queueing/worker_prepare.go
```diff
@@ -45,6 +45,9 @@ func NewPrepareJobWorker() (*PrepareJobWorker, error) {
 	if err != nil {
 		logger.Fatalf("failed to load service config. Err: %v", err)
 	}
+	if len(cfg.ACOConfigs) == 0 {
+		logger.Fatalf("no ACO configs found, these are required for downstream processing")
```
laurenkrugen-navapbc, line 49: does this also need a panic/fatal here?

carlpartridge: logger.Fatalf should log the message then call exit for us.

laurenkrugen-navapbc (approved):

PR #1131, opened 2025-05-21
[PLT-1052] Build-and-package-all-gf

jscott-nava:
## 🎫 Ticket

https://jira.cms.gov/browse/PLT-1052

## 🛠 Changes

Greenfield specific GHA files were added; significant changes include:

- references to the rpms bucket are now pulled for parameter store and passed to child build files since bucket prefixes are now used; the default 'bcda-rpms' bucket name has been retained to avoid breaking the legacy build.
- a matrix has been introduced in order to support running parallel workflows on both the prod and non-prod accounts; at this time the bucket module has not been run against the prod account so that part of matrix has been commented out.

## ℹ️ Context

These changes have been made to support the greenfield migration initiative. Once the greenfield migration is complete the legacy build files will be deleted and the greenfield specific files will be renamed to remove the '_gf' suffix.

## 🧪 Validation

Various runs have succeeded ( for example: https://github.com/CMSgov/bcda-app/actions/runs/15117572777 ) and have created api and worker AMIs in the non-prod account. The api instances are currently failing with db init errors that do not seem related to the build process.


carlpartridge (approved):
Some comments but overall I think its good enough for now!

PR #1145, opened 2025-06-17
BCDA-9148: Add cleanup step to create group lambda

michaeljvaldes:
## 🎫 Ticket

### DIFF
--- a/.github/workflows/admin-create-group-integration-test-GF.yml
+++ b/.github/workflows/admin-create-group-integration-test-GF.yml
@@ -12,6 +12,10 @@ on:
 concurrency:
   group: create-group-integration-test
 
+env:
+  ACO: A5555
+  GROUP: A5555-group
+
 jobs:
   # Deploy first if triggered by pull_request
   deploy:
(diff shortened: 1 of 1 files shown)


.github/workflows/admin-create-group-integration-test-GF.yml
```diff
@@ -46,7 +48,46 @@ jobs:
...
+            echo "number of groups with group_id = ${{ env.GROUP }} not equal to 1"
+            echo "matching number of groups: $GROUP_COUNT"
+            exit 1
+          fi
+          DELETED_COUNT=`psql -tA "$CONNECTION_URL" -c "DELETE FROM groups WHERE group_id = '${{ env.GROUP }}'" 2>&1`
+          if [[ $? -ne 0 || $DELETED_COUNT != "DELETE 1" ]]; then
+            echo "group deletion command failed"
+            echo "deletion command results: $DELETED_COUNT"
+            exit 1
+          fi
```
carlpartridge, line 93: If there is an existing group in the DB with that group_id, wont this delete that?

carlpartridge (approved):

.github/workflows/admin-create-group-integration-test-GF.yml
michaeljvaldes: I updated the cleanup step to soft delete the group instead of deleting the row.
For extra context: the cleanup step only runs if the trigger step succeeds, and the trigger step should fail if there already exists a group with that group id (for a group that has a null `deleted_at` column). The cleanup step also checks that there is only one group to delete, though this is just an extra precaution.
This integration also only ever runs on the dev environment and is already dependent on the state of the dev database having a particular ACO.

PR #1149, opened 2025-06-23
BCDA-9145: Remove old workflows

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9145

## 🛠 Changes

- removed legacy workflows and renamed greenfield workflows

PR #1150, opened 2025-06-24
BCDA-9166-post-greenfield-cleanup

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9166

## 🛠 Changes

- Renamed references to opensbx and sbx. We decided to switch to sandbox.
- Removed conf for opensbx and postman collection for it.
- Most of these files will be deleted as part of https://github.com/CMSgov/bcda-app/pull/1149. However, this branch existed and instead of doing all renaming from scratch I have used this branch as base.

### DIFF
--- a/.github/workflows/refresh_test_attribution.yml
+++ b/.github/workflows/refresh_test_attribution.yml
@@ -4,7 +4,7 @@ name: Refresh Attribution For Test ACOs
 on:
   workflow_call:
   schedule:
-    - cron: "0 12 * * 1" 
+    - cron: "0 12 * * 1"
 
 permissions:
   id-token: write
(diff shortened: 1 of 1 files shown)


laurenkrugen-navapbc (approved):
LGTM! The only thing I am not sure we have updated is the `env` variable the application uses and the parameter paths in parameter store. If those two are mismatched, then we won't be able to retrieve env variables for sandbox.

Do we need a script to update that path and do we need to update the env var so the application and parameters uses `sandbox`, too?

carlpartridge:
Do we want to use this branch in order to get changes required for https://github.com/CMSgov/bcda-ops/pull/1230 in?  Or create a new PR for those?

Everything else looks good here.

It looks like smoke tests are failing on PACA data (401) which is generally one of three things:
- bad creds in param store
- outdated CCLF files (I will check and update these in dev now)
- flakiness

carlpartridge (approved):

PR #1153, opened 2025-06-30
Carl/bcda 8912 add ssas admin migration

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8912

## 🛠 Changes

Updated deploy-all workflow to include call to migrate SSAS DB.
Add unique constraint on acos cms_id.
Add manual sql script to set up ssas schema_migrations table.

## ℹ️ Context

Currently we have two separate repos both modifying the same DB.  Because of this any changes/migrations can be a bit tangled up.  This is the cleanest approach I could think of that did not include a larger overhaul.

## 🧪 Validation

Ran migrations locally


db/migrations/manual/20250627-add-ssas-schema-migrations.sql
```diff
@@ -0,0 +1,13 @@
...
+
+-- DROP TABLE public.schema_migrations_ssas;
+
+CREATE TABLE public.schema_migrations_ssas (
+	"version" int8 NOT NULL,
+	dirty bool NOT NULL,
+	CONSTRAINT schema_migrations_ssas_pkey PRIMARY KEY (version)
+);
+
+INSERT INTO public.schema_migrations_ssas (version) VALUES (9);
```
carlpartridge, line 13: Hardcoded to 9 as that is the current migration number in bcda-ssas-app.  Maybe it would be best to add a comment to this file stating that is intended for ssas app migrations as well as being a one off?

PR #1155, opened 2025-07-01
BCDA-9225: Reconfigure golang linting

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9225


Makefile
```diff
@@ -19,7 +19,7 @@ setup-tests:
 LINT_TIMEOUT ?= 3m
 lint: setup-tests
 	docker compose -f docker-compose.test.yml run \
-	--rm tests golangci-lint run --timeout=$(LINT_TIMEOUT) --verbose
+	--rm tests golangci-lint run --timeout=$(LINT_TIMEOUT) --verbose --new-from-merge-base=main
```
carlpartridge, line 22: Theoretically we do sometimes branch off of other PRs but overall this is probably fine for now.

Dockerfiles/Dockerfile.tests
```diff
@@ -5,7 +5,7 @@ RUN apk update upgrade
 
 RUN apk add bash build-base curl
 
-RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin
+RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.2.1
```
carlpartridge, line 8: HEAD can sometimes include alpha like content.  Do we want to stick to a specific version?  Is that what the second part of this script is doing?

Is this the only dockerfile that we add golangci-lint to?

carlpartridge (approved):

PR #1164, opened 2025-07-08
BCDA-9107: use ssas health endpoint in app health check

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9107

## 🛠 Changes

- pointed the bcda-app `/_health` check endpoint to hit the ssas `_/health` check endpoint instead of `/introspect`

## 🧪 Validation

Tested locally by curling `localhost:3000/_health`

### DIFF
--- a/bcda/auth/client/ssas.go
+++ b/bcda/auth/client/ssas.go
@@ -510,7 +510,7 @@ func (c *SSASClient) GetVersion() (string, error) {
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusOK {
-		return "", errors.New("SSAS server failed to return version ")
+		return "", errors.New("SSAS server failed to return version")
 	}
 
 	type ssasVersion struct {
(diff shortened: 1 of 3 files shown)


bcda/auth/client/ssas.go
```diff
@@ -525,3 +488,21 @@ func (c *SSASClient) GetVersion() (string, error) {
...
+	}
+	if err := c.setAuthHeader(req); err != nil {
+		return err
+	}
+	resp, err := c.Do(req)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return errors.New("SSAS server health check failed")
```
carlpartridge, line 505: I wonder how we might have something watching for these errors in order to restart the service.  But that is still probably just be better off happening at the ALB -> SSAS level instead of the API -> SSAS level...

carlpartridge (approved):

bcda/auth/client/ssas.go
michaeljvaldes: For what it's worth, the ssas load balancer is already checking the same health endpoint for load-balancing and asg replacement purposes 👍

bcda/health/health.go
```diff
@@ -50,9 +50,9 @@ func IsSsasOK() (result string, ok bool) {
 		log.Auth.Errorf("no client for SSAS. no provider set; %s", err.Error())
 		return "No client for SSAS. no provider set", false
 	}
-	if err := c.Ping(); err != nil {
-		log.API.Error("Health check: ssas ping error: ", err.Error())
-		return "Cannot authenticate with SSAS", false
+	if err := c.GetHealth(); err != nil {
```
laurenkrugen-navapbc, line 53: would it be better to create a client and persist it, vs creating a client for each health check?

laurenkrugen-navapbc: if we feel like the overhead is low, then I'm OK to keep it in here!

laurenkrugen-navapbc (approved):

PR #1166, opened 2025-07-08
BCDA-8703-remove-que-go-library

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8703

## 🛠 Changes

- Dependency removal from `connection.go` (and removal of `que.PrepareStatements`)
- Removed conditional logic for `QUEUE_LIBRARY` environment variable. Now only uses `StartRiver()` and `StopRiver()`
- Removed `QUEUE_LIBRARY=river` from all configuration files
- Removed `QUEUE_LIBRARY` environment variable setting/unsetting from `bcda/api/requests_test.go`. Test only uses River now
- Remove ALR route and implementation

## ℹ️ Context

We have migrated away from [que-go](https://github.com/bgentry/que-go) and switched to [river](https://github.com/riverqueue/river) for all queuing tasks and periodic jobs. This is cleanup of the que-go library.

## 🧪 Validation

- Ran linter, build, test suite


carlpartridge (commented):
This is awesome!  I think its worth a run through of bcda-ssas-app to see if there is anything related to que-go or ALR there.  Both repos (packages in each repo) reference eachothers packages and many interfaces have changed.

I am also curious how best we can handle dropping the bcda_queue DB, do we want to create a migration in db/migrations/bcda to do that (with an if exists)?  Or maybe we just leave it alone for now?  Maybe we need to keep the data around for compliance?

The things I think are missing right now are
- deleting db/migrations/bcda_queue/*
- adjusting dbdocs workflow (make dbdocs second line prints out bcda_queue docs)
- adjusting docker-compose.yml remove queue service (maybe just a larger pass of finding/reviewing everything referencing `bcda_queue`)

michaeljvaldes (approved):

PR #1171, opened 2025-07-23
BCDA-9314: refactor API db connection globals

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9314
https://jira.cms.gov/browse/BCDA-9287

## 🛠 Changes

- moved database connection init logic to `connect` functions, which establish new connections to the database and  replace init and global connection variables
-  split db connection and pool into separate `connect` functions with separate health checks
- refactored files that relied on global vars (e.g. routers, middleware, apis) to structs with explicitly defined dependencies on database connections in apis, routers, and services
- configured database connections at code entrypoints (e.g. cli setup) and passed connections down to services 
- refactored `auth/Provider.go` to not be a global, either, and instead be configured at setup and passed down like the database connections

To-Do:

- [X] refactor references to global connection in lambdas
- [X] refactor references to global connection in various other places (middleware, health check, etc.)
- [X] remove global connection var
- [x] add a unit test for pool connection check

## ℹ️ Context

This PR demonstrates a strategy to refactor the database Connection -- specifically, to remove the global Connection variables defined in /database/connection.go which are created in an init() function, and instead do the following:
1. Create the database connection(s) once at each code entrypoint (e.g. cli.go, each lambda main.go, etc.)
2. Pass each connection as a dependency to the routers and services that need it. 
3. For each router/service that needs a connection, create a type with state that stores the connection, as well as a constructor that accepts its required dependencies, if necessary.

As a result, many other files needed to be changed to store the db connection within its state, including the API files and middleware. Notably, the Provider, which also depends on the db connection, was also being used as a global across many files, so it too had to be updated to a stateful type which is constructed at entrypoints.
Finally, many tests were updated to account for these changes. The majority of the test updates were completely innocuous -- replacing references to a global connection with references to a constructed connection -- but others involved untangling hidden dependencies on the database, provider, etc. that became very obvious after they were refactored.


bcda/auth/middleware.go
```diff
@@ -31,6 +31,9 @@ var (
 	AuthDataContextKey = &contextKey{"ad"}
 )
 
+type Middleware struct {
+}
```
carlpartridge, line 35: ??

bcda/database/connection.go
```diff
@@ -23,24 +23,37 @@ var (
 	Pgxv5Pool       *pgxv5Pool.Pool
 )
 
+type Connections struct {
+	Connection      *sql.DB
+	QueueConnection *pgx.ConnPool
+	Pgxv5Pool       *pgxv5Pool.Pool
+}
```
carlpartridge, line 30: I dont think a group of different connections is a useful thing.  It feels like going further down a messy situation.  We have:
- Connection: used by all of our ORM
- QueueConnection: deprecated
- Pgxv5Pool: used by our workers, lambdas, etc?
Ultimately we will only have/want 1 connection type right?  Can we not just use/pass a specific connection into where its needed?

This article has some useful insights and examples: https://www.alexedwards.net/blog/organising-database-access

bcda/bcdacli/cli.go
```diff
@@ -586,20 +581,34 @@ func createACO(name, cmsID string) (string, error) {
...
-	creds, err := auth.GetProvider().FindAndCreateACOCredentials(acoCMSID, ips)
+	creds, err := provider.FindAndCreateACOCredentials(acoCMSID, ips)
 	if err != nil {
 		return "", errors.Wrapf(err, "could not register system for %s", acoCMSID)
 	}
 
 	return creds, nil
 }
 
+func resetClientCredentials(acoCMSID string) (string, error) {
```
michaeljvaldes, line 592: Pulling some logic out to a separate function to simplify testing

bcda/auth/middleware_test.go
```diff
@@ -56,21 +62,27 @@ func (s *MiddlewareTestSuite) CreateRouter() http.Handler {
...
+	// s.server = httptest.NewServer(s.CreateRouter(auth.NewProvider(s.connection)))
 	s.rr = httptest.NewRecorder()
 }
 
-func (s *MiddlewareTestSuite) TearDownTest() {
-	s.server.Close()
-}
+// func (s *MiddlewareTestSuite) TearDownTest() {
+// 	s.server.Close()
+// }
```
carlpartridge, line 76: Remove commented out?

bcda/bcdacli/cli.go
```diff
@@ -586,20 +581,34 @@ func createACO(name, cmsID string) (string, error) {
...
+	aco, err := r.GetACOByCMSID(context.Background(), acoCMSID)
```
carlpartridge, line 593: r is a global var here which makes this function a bit more complex to read.  Could be better if we just passed in r (repo) as a param?  Looks like the same with provider.

bcda/database/connection.go
```diff
@@ -17,37 +17,50 @@ import (
...
+	defer cancel()
+
+	startConnectionHealthCheck(
+		ctx,
+		conn,
+		time.Duration(cfg.HealthCheckSec)*time.Second,
+	)
+
+	return conn
+}
```
carlpartridge, line 41: We arent returning an error with this Connect() function, is that because any issues force a panic?
Can `conn` ever be a nil sql.DB object?  Like it returns properly but the actual connection isnt present (ie is more or less failing silently)?

bcdaworker/queueing/river.go
```diff
@@ -48,14 +48,16 @@ type Notifier interface {
 }
 
 // TODO: better dependency injection (db, worker, logger).  Waiting for pgxv5 upgrade
-func StartRiver(numWorkers int) *queue {
+func StartRiver(db *sql.DB, numWorkers int) *queue {
+	pool := database.ConnectPool()
```
carlpartridge, line 52: We are passing in `db` but forcing `pool` to be default connection?  Is that because we dont want to muck around with rivers inner workings?

bcda/api/requests.go
```diff
@@ -63,14 +62,14 @@ type fhirResponseWriter interface {
 	JobsBundle(context.Context, http.ResponseWriter, []*models.Job, string)
 }
 
-func NewHandler(dataTypes map[string]service.DataType, basePath string, apiVersion string) *Handler {
-	return newHandler(dataTypes, basePath, apiVersion, database.Connection)
+func NewHandler(dataTypes map[string]service.DataType, basePath string, apiVersion string, db *sql.DB, pool *pgxv5Pool.Pool) *Handler {
```
carlpartridge, line 65: I see all handlers and APIvX passing around pool, but I dont see it being used anywhere?  Is this preparation for swapping out db?  Or did I miss it somewhere?

bcda/auth/middleware_test.go
michaeljvaldes: Whoops, forgot to push that one. Fixed!

bcda/api/requests.go
michaeljvaldes: Yes, I think the handler passes it to its enqueuer, which then passes it to its river client. As far as I can tell, I only replaced existing instances of the global pool with the pool passed as a dependency.

bcdaworker/queueing/enqueue.go
```diff
@@ -24,16 +25,16 @@ type Enqueuer interface {
 
 // Creates a river client for the Job queue. If the client does not call .Start(), then it is insert only
 // We still need the workers and the types of workers to insert them
-func NewEnqueuer() Enqueuer {
+func NewEnqueuer(db *sql.DB, pool *pgxv5Pool.Pool) Enqueuer {
```
michaeljvaldes, line 28: @carlpartridge here the pool is passed down to the the enqueuer via its constructor and subsequently to the river client that it stores in its state.

bcda/api/requests.go
michaeljvaldes: I added a PR comment on `enqueue.go` to make it a little clearer.

bcda/database/connection.go
michaeljvaldes: This is a great question.
The current state: we don't return an error because issues force a process exit. This is how the previous `init()` worked and I didn't intend to change that.
The ideal state: _do_ we want to allow the connection to be nil? Is having a nil database ever a valid state? I would say no, since we are calling this function to connect to a database and I don't recall any instances in which we would ever want the result to be anything but a valid connection and not immediately panic. But this is super interesting and I'm open to exploring that.

bcda/bcdacli/cli.go
michaeljvaldes: I think that makes it much more readable. I originally intended to change as little as I could in this file, but virtually all of the functions like this one all use the "global" variables (service, repo, provider). Should I update them all to be more readable?

bcda/database/connection.go
michaeljvaldes: Yeah, it looks like the health check does just log warnings, but in the event of an error in the `Connect()` and `ConnectPool() functions the process is exited via `logrus.Fatalf()`

michaeljvaldes: I think that makes sense. If we are interested in exploring alternatives, I would also vote to do it in a separate PR/effort since I tried not to change too much of the _behavior_ of bcda in this PR.

bcda/health/health.go
```diff
@@ -1,34 +1,41 @@
...
-	if err := db.Ping(); err != nil {
+type HealthChecker struct {
+	db *sql.DB
+}
+
+func NewHealthChecker(db *sql.DB) HealthChecker {
+	return HealthChecker{db: db}
+}
+
+func (h HealthChecker) IsDatabaseOK() (result string, ok bool) {
```
laurenkrugen-navapbc, line 20: do we think it's necessary to check have a worker and non worker health check, since they connect to the same database?

bcda/web/middleware/ratelimit.go
```diff
@@ -21,58 +21,57 @@ import (
...
 	jobTimeout   time.Duration
 	retrySeconds int
-)
+}
 
-func init() {
-	repository = postgres.NewRepository(database.Connection)
-	jobTimeout = time.Hour * time.Duration(utils.GetEnvInt("ARCHIVE_THRESHOLD_HR", 24))
-	retrySeconds = utils.GetEnvInt("CLIENT_RETRY_AFTER_IN_SECONDS", 0)
+func NewRateLimitMiddleware(config *service.Config, db *sql.DB) RateLimitMiddleware {
```
laurenkrugen-navapbc, line 31: 🙌

bcda/web/middleware/ratelimit_test.go
```diff
@@ -77,38 +91,40 @@ func TestHasConcurrentJobs(t *testing.T) {
...
-			)
-			repository = mockRepo
-
-			rr := httptest.NewRecorder()
-			middleware := CheckConcurrentJobs(cfg)
-			middleware(nil).ServeHTTP(rr, getRequest(tt.rp))
-			assert.Equal(t, http.StatusTooManyRequests, rr.Code)
-			assert.NotEmpty(t, rr.Header().Get("Retry-After"))
-		})
+		mockRepo := &models.MockRepository{}
```
laurenkrugen-navapbc, line 94: Think we need the line:

`t.Run(tt.name, func(t *testing.T) {`

in here so that each loop is run as its own test!

bcda/web/router_test.go
```diff
@@ -344,11 +352,11 @@ func createExpectedAuthData(cmsID string, aco models.ACO) auth.AuthData {
 	}
 }
 
-func createConfigsForACOBlacklistingScenarios(s *RouterTestSuite) (configs []struct {
+func createConfigsForACOBlacklistingScenarios(s *RouterTestSuite, p auth.Provider) (configs []struct {
```
laurenkrugen-navapbc, line 355: unrelated to this PR, but could we update this function name to:

`createConfigsForACODenylistingScenarios`

bcdaworker/queueing/river.go
laurenkrugen-navapbc: I would be +1 to waiting for the pgx upgrade, especially since this is a large set of changes and there doesn't seem to be any sort of immediate benefit, like you said.

bcdaworker/queueing/worker_cleanup.go
```diff
@@ -19,14 +20,16 @@ import (
 
 type CleanupJobWorker struct {
 	river.WorkerDefaults[worker_types.CleanupJobArgs]
-	cleanupJob      func(time.Time, models.JobStatus, models.JobStatus, ...string) error
-	archiveExpiring func(time.Time) error
+	cleanupJob      func(*sql.DB, time.Time, models.JobStatus, models.JobStatus, ...string) error
```
laurenkrugen-navapbc, line 23: probably not suited for changes in this PR, but if we are updating the DB depenedencies, would `cleanupJob` and `archiveExpiring` be better suited as methods with `CleanupJobWorker` being the receiver? they would have access to db via `CleanupJobWorker.db` instead of having to pass it in. Thoughts on maybe leaving a comment on here?

bcda/health/health.go
michaeljvaldes: I think it makes sense since they are technically two separate connections.
I tried not to change the behavior too much but there is some subtlety here. We had a check for both of them previously, but the difference is now the checks are running in separate goroutines.
I suspect that with the pgx upgrade, this will simplify again.

bcda/auth/router.go
```diff
@@ -7,11 +7,13 @@ import (
 	"github.com/go-chi/chi/v5"
 )
 
-func NewAuthRouter(middlewares ...func(http.Handler) http.Handler) http.Handler {
+func NewAuthRouter(provider Provider, middlewares ...func(http.Handler) http.Handler) http.Handler {
+	baseApi := NewBaseApi(provider)
```
laurenkrugen-navapbc, line 11: I am certain this is user error, but I don't see the definition for this anywhere with cmd+f - does this func definition exist?

bcdaworker/queueing/worker_cleanup.go
michaeljvaldes: Absolutely, I actually thought the same thing and wondered why they were in separate files.
I left a comment for now but I'm open to changing it if folks feel strongly about it.

bcda/bcdacli/cli.go
```diff
@@ -61,8 +64,11 @@ func setUpApp() *cli.App {
 	app.Usage = Usage
 	app.Version = constants.Version
 	app.Before = func(c *cli.Context) error {
-		db = database.Connection
+		db = database.Connect()
```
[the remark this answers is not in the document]
michaeljvaldes: I think this is a great idea, and I would vote to do this in a separate tech debt ticket that would remove the cli, if that works for you.

carlpartridge (approved):

carlpartridge:
The two failing CI checks likely just need to update to latest changes on main.

bcda/health/health.go
laurenkrugen-navapbc: that's a good point - probably best not to change too much behavior given the size of the set of changes.

bcdaworker/queueing/worker_cleanup.go
laurenkrugen-navapbc: Maybe we can leave a TODO + a ticket for this one for future work to minimize risk?

laurenkrugen-navapbc:
> * cclf-import test integration / verify (pull_request)

these failing on the aurora name?

laurenkrugen-navapbc (approved):
Super happy to remove these inits from the codebase - thanks so much for knocking this out! 🎉

carlpartridge (approved):

PR #1175, opened 2025-08-04
BCDA-8777: Update alert destinations for workflows and lambdas

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8777

## 🛠 Changes

**Admin Lambdas:**
- Updated slack alerts to be triggered in the `handle` function
- All status updates will go to operations channel

**Workflows**:
- Deploy related workflows have had destination channels updated to deploy channel

**General**:
- Added slack utils package and constants for slack messages to reduce duplicate code

## ℹ️ Context

Alerting channels are currently getting status and informational messages that don't pertain to alerts. To better manage alerts and critical infrastructure related monitoring, our alerting channel should be free of messages that do not indicate an issue. Deploy related updates will only be in the deploy channel and admin lambdas will report status updates to the operations channel, including run failures, since those are expected.

## 🧪 Validation

Integration tests are passing, workflows in expected channels. Looking to verify:

- lambdas that run on a schedule (non-admin) route failures to the alerts channel but route success to operations
- admin lambdas route all statuses to operations
- deploy workflows and related workflows route only to deploy channel
- `Starting..` messages removed to reduce noise

### DIFF
--- a/bcda/api/requests.go
+++ b/bcda/api/requests.go
@@ -237,10 +237,10 @@ func (h *Handler) JobStatus(w http.ResponseWriter, r *http.Request) {
 	job, jobKeys, err := h.Svc.GetJobAndKeys(r.Context(), uint(jobID))
 	if err != nil {
 		if errors.Is(err, sql.ErrNoRows) {
-			log.API.Info("Requested job not found.", err.Error())
+			logger.Infof("Requested job not found.", err.Error())
 			h.RespWriter.Exception(r.Context(), w, http.StatusNotFound, responseutils.DbErr, "Job not found.")
 		} else {
-			log.API.Error("Error attempting to request job. Error:", err.Error())
+			logger.Errorf("Error attempting to request job. Error:", err.Error())
 			h.RespWriter.Exception(r.Context(), w, http.StatusInternalServerError, responseutils.InternalErr, "Error trying to fetch job. Please try again.")
 		}
 
--- a/bcda/api/v1/api.go
+++ b/bcda/api/v1/api.go
@@ -267,14 +267,19 @@ Responses:
 	500: errorResponse
 */
 func ServeData(w http.ResponseWriter, r *http.Request) {
+
 	dataDir := conf.GetEnv("FHIR_PAYLOAD_DIR")
 	fileName := chi.URLParam(r, "fileName")
 	jobID := chi.URLParam(r, "jobID")
 	filePath := fmt.Sprintf("%s/%s/%s", dataDir, jobID, fileName)
 
+	logger := log.GetCtxLogger(r.Context())
+
 	encoded, err := isGzipEncoded(filePath)
 	if err != nil {
-		writeServeDataFailure(err, w)
+		logger.Errorf("failed when checking if file is gzip encoded: %s", err)
+		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+
 	}
 
 	var useGZIP bool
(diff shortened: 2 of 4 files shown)


.github/workflows/build-and-package-all.yml
```diff
@@ -135,9 +135,8 @@ jobs:
         with:
           method: chat.postMessage
           token: ${{ secrets.SLACK_BOT_TOKEN }}
-          # Sends to bcda-alerts
           payload: |
-            channel: "C034CFU945C"
+            channel: "C03S23MJFJS"
```
laurenkrugen-navapbc, line 139: This is the deploy channel - the five workflows updated here will all report to the deploy channel regardless of pass/fail.

bcda/slack/slack.go
```diff
@@ -0,0 +1,31 @@
...
+	AlertsChannel     = "C034CFU945C" // #bcda-alerts
+	SuccessMsg        = "SUCCESS"
+	FailureMsg        = "FAILURE"
+)
+
+func SendSlackMessage(sc *slack.Client, channel string, msg string, status bool) {
+	color := "danger"
+	if status {
+		color = "good"
+	}
```
carlpartridge, line 21: Nit: Dont think we need to have Slack in the name of the function (SendSlackMessage -> SendMessage).

Nit: Not a big fan of status (bool) param type, it seems to exist just to differentiate between the color?  I know its more verbose but Id prefer a SendSuccess(...) function and a SendFailure(...) function but interested to hear what other people think.  I think at the very least we could change the param name to something more appropriate.

Overall this feels like a fairly thin wrapper around the actual slack client and its interface.  To some degree I feel like its a bit excessive but it does also slim various use cases down a fair amount.  I think this concern is shown by what happens if we need to add a bunch of customization/configuration to the messages we want to send.  This wrapper then gets built up until we more or less just have the slack client all over again.  However as our use case is now, this does save us quite a bit of verbose references elsewhere.

Part of the goal of `notifier` was to decouple the thing we want to happen (notification of success/failure) from the implementation tool (slack).  If/when we need to change implementations it should be a bit more straightforward.  I dont think the `notifier` was a good implementation of this, but maybe theres a good middle ground?  One possibility would be to change this slack_utils package to a notifier package, with a slack implementation/wrapper.

.github/workflows/build-and-package-all.yml
carlpartridge: We could theoretically put these channel IDs in param store, then reference them as such, but that doesnt feel like its worth the effort?

bcda/lambda/admin_create_aco/main.go
```diff
@@ -73,12 +68,16 @@ func handler(ctx context.Context, event json.RawMessage) error {
 	id := uuid.NewRandom()
 
 	if data.CleanUp == nil {
+
 		// run the regular logic (non-rollback transaction)
-		err = handleCreateACO(ctx, conn, data, id, slackClient)
+		err = handleCreateACO(ctx, conn, data, id)
 		if err != nil {
+			slUtls.SendSlackMessage(slackClient, slUtls.OperationsChannel, fmt.Sprintf("%s: Create ACO lambda in %s env.", slUtls.FailureMsg, os.Getenv("ENV")), false)
```
carlpartridge, line 75: One concern about moving notification messages out of the business logic (eg handleCreateACO) is that we can no longer test/verify that we are sending notifications as part of the process.  Im not sure if that matters too much overall.

My general thinking around lambdas is that the handler should do as little as possible since its very difficult to properly test.  Basically it should be setting up the requirements/environment in which to run the actual business logic.  This allows us to verify as much as possible with unit tests and such.

laurenkrugen-navapbc: that's fair - there's definitely a gap in testing with slack if we are to make changes in the future, some even outside the scope of testing here. Ex: while moving some of these alerts to a different channel, the slackbot didn't have access to the channel yet. The error was logged, but that was it. 

possible solutions:

- revert these changes altogether and put the notifiers back in
- do some error handling for the slack messages (which didn't exist before - just logging it like it is now)
- manual review every so often of alerts to verify that they aren't missed from our critical services
- nothing; if modifications are made to these lambdas, reviewers and PR owner should be validating that integration tests work, and lambda runs aren't logging errors, etc

carlpartridge: I believe the reason we only log an error when slack fails to send a message is that we dont want to fail the function on an issue with slack.  Overall I see the slack messages in these lambda contexts as a nice to have.  Keeps us more aware of whats happening, but losing some isnt the end of the world.

bcda/slack/slack.go
carlpartridge: I think things are fine as is.  I think one or some of the following could help though:
- rename bool var (status -> success_msg? color?)
- change bool var to string and just pass in the color option
- default bool value to success instead of danger
- create a SendSuccess() and a SendFailure() function that handle the bool
- instead of passing a bool in pass in an options object, with color as one of the options (again this is rewriting the slack package so in addition to being more work is probably not worth it)

laurenkrugen-navapbc:
> I would like some tests for slack_utils, other than that things look good!

sounds good, adding tests now!

michaeljvaldes (approved):

PR #1176, opened 2025-08-06
BCDA-8395-pgx-driver-update

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-8395

## 🛠 Changes

- We have updated to pgx v5 to support River. However, certain parts of the application still use v3 and v4. Updating the codebase to rely on v5.
- Update nrpgx to v5 (from v4)
- Multiple dependency updates
- Resolve the driver registered twice error (properly)
- CCLF bulk upload process has changed. We do not use SQL TX or SQL DB anymore. Everything is done in one pgx transaction.
- CSV bulk upload has changed (like CCLF). We had to update databasetest to pass back pgx pool, ensuring csv_test connects to the same DB for SQL and pgx-based operations.
- New PGX-based repository

## ℹ️ Context

- We had PGX v3, 4, and 5 in our application. We want to bring the application up to date and only use PGX v5.

## 🧪 Validation

Local lint, build, test.

### DIFF
--- a/bcda/models/postgres/pgx_repository_test.go
+++ b/bcda/models/postgres/pgx_repository_test.go
@@ -50,7 +50,7 @@ func (s *PgxRepositoryTestSuite) TearDownSuite() {
 	}
 }
 
-func (s *PgxRepositoryTestSuite) SetupTest() {
+func (s *PgxRepositoryTestSuite) TearDownTest() {
 	s.cleanupTestData("TEST123")
 }
 


bcda/database/databasetest/databasetest_test.go
```diff
@@ -5,7 +5,7 @@ import (
 	"testing"
 
 	"github.com/CMSgov/bcda-app/bcda/database"
-	_ "github.com/jackc/pgx"
```
carlpartridge, line 8: Why this change?  Could this be jackc/pgxv5 instead?

carlpartridge: Totally fair to break off the work in a separate ticket and maybe it would be best to wait until we switch over to DataDog to do this in general.  My thoughts are that we wouldnt be getting rid of our NewRelic tracing capability, just our custom nrpgx file set up.  So get rid of our nrpgx file, then in connection.go (in either ConnectDB() or createDB()) we set up the NewRelic connection.

bcda/cclf/cclf_test.go
```diff
@@ -246,11 +257,11 @@ func (s *CCLFTestSuite) TestImportCCLF8DBErrors() {
 	}
 
 	//Send an invalid context to fail DB check
-	ctx, function := context.WithCancel(context.TODO())
+	ctx2, function := context.WithCancel(ctx)
```
carlpartridge, line 260: The comment above sounds weird, "send an invalid context to fail db check" but we are passing in a valid context (to create a WithCancel context?).  I guess the context we pass into the WithCancel doesnt matter?

bcda/models/postgres/pgx_repository.go
```diff
@@ -0,0 +1,167 @@
...
+		return err
+	}
+
+	affected := result.RowsAffected()
+	if affected == 0 {
+		return fmt.Errorf("failed to delete file entry %d, no entry found", fileID)
+	}
+
+	return nil
+}
```
carlpartridge, line 167: Opinions (no right answer with these but interested to hear how people feel):
- I think we should have the names be more like:
`DeleteCCLFFileWithPool` -> `DeleteCCLFFile`
`DeleteCCLFFile` -> `DeleteCCLFFileTx`
I think the Tx version is the more specific, lesser used variant.

- I think some of these functions are too specific in what I would want from an ORM layer.  I think of the basic CRUD functions.  So `UpdateCCLFFileImportStatusWithPool` -> `UpdateCCLFFile` and we would pass it in a `CCLFFile{ImportStatus: "newStatus"}`.  I think if we see a ton of use of a specific variant (eg updating importstatus) then sure lets make a function.  But generally speaking open ended, re-usable, simple, and slim interfaces seems better than having a whole bunch of functions for every possible thing we want to modify in the DB.  I know this is pushing the argument a bit but would we want a separate function for updating each of ACOCMSID, PerformanceYear, Name, TimeStamp, Type, etc?  And success, failure, and edge case tests for each as well?

- This might just be my Ruby/Rails background but I kinda like the idea of having each models ORM functions be in their own file.  Eg we would have a cclf_files.go file which would include all of the above functions and a cclf_beneficiaries file, etc.  Even if they ended up all being roughly the same general CRUD interfaces.  (Not really sure how I feel about this).

- Do we want to start using an ORM library?  We use gORM in SSAS.  I think we use a few different things in bcda?  That could greatly simplify this file and these calls?  And handles a lot of our testing?  It does lock us in a bit, but maybe seems worth it?

bcda/models/postgres/repository.go
```diff
@@ -219,6 +214,28 @@ func (r *Repository) UpdateCCLFFileImportStatus(ctx context.Context, fileID uint
 	return nil
 }
 
+func (r *Repository) DeleteCCLFFile(ctx context.Context, fileID uint) error {
```
carlpartridge, line 217: Do we still need this?  This was for the sql version right?
(Same for the tests and the interface and likely the mocks as well)

bcda/models/postgres/pgx_repository_test.go
```diff
@@ -0,0 +1,24 @@
...
+	assert.NotNil(t, repo)
+}
+
+func TestPgxRepository_NewPgxRepository(t *testing.T) {
+	repo := NewPgxRepository()
+	require.NotNil(t, repo)
+
+	_, ok := interface{}(repo).(*PgxRepository)
+	assert.True(t, ok, "NewPgxRepository should return *PgxRepository")
+}
```
carlpartridge, line 24: We need tests for all these DB interactions (eg `DeleteCCLFFileWithPool`) or are they somewhere else?

bcda/models/postgres/pgx_repository_test.go
```diff
@@ -0,0 +1,338 @@
...
+		Name:            name,
+		ACOCMSID:        acoCMSID,
+		Timestamp:       time.Now().Round(time.Millisecond),
+		PerformanceYear: 2024,
+		ImportStatus:    "PENDING",
+		Type:            models.FileTypeDefault,
+	}
+}
+
+func (s *PgxRepositoryTestSuite) cleanupTestData(cmsID string) {
```
carlpartridge, line 298: Should we be calling this either at the end of all of these tests or after each individual test?

carlpartridge (approved):

michaeljvaldes (approved):
@bhagatparwinder Thanks for the plethora of detail and comments/conversations -- they helped a lot in understanding the changes in this pr.

bcda/cclf/csv.go
```diff
@@ -113,29 +115,22 @@ func (importer CSVImporter) ProcessCSV(csv csvFile) error {
...
+		return fmt.Errorf("failed to start pgx transaction: %w", err)
 	}
-
-	rtx := postgres.NewRepositoryPgxTx(tx)
-	var records int
 	defer func() {
 		if err != nil {
-			if err1 := tx.Rollback(); err1 != nil {
-				importer.Logger.Errorf("Failed to rollback transaction: %s, %s", err.Error(), err1.Error())
+			if err1 := pgxTx.Rollback(ctx); err1 != nil {
```
laurenkrugen-navapbc, line 125: nit: could we update `err1` to `rollbackErr` or something similar?

PR #1188, opened 2025-08-15
BCDA-9360: Refactor logger away from init

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9360

## 🛠 Changes

Logger was refactored away from init.  It should now be set at various code entry points but defaults to a global default logger.

## ℹ️ Context

Refactoring away from init functions.  Init functions create complex dependencies and should generally be avoided.

## 🧪 Validation

Local testing and linting.

### DIFF
--- a/bcda/bcdacli/cli.go
+++ b/bcda/bcdacli/cli.go
@@ -64,11 +64,11 @@ func setUpApp() *cli.App {
 	app.Usage = Usage
 	app.Version = constants.Version
 	app.Before = func(c *cli.Context) error {
+		log.SetupLoggers()
 		db = database.Connect()
 		pool = database.ConnectPool()
 		repository = postgres.NewRepository(db)
 		provider = auth.NewProvider(db)
-		log.SetupLoggers()
 		log.API.Info(fmt.Sprintf(`Auth is made possible by %T`, provider))
 		return nil
 	}


bcda/api/v1/api_test.go
```diff
@@ -167,6 +167,7 @@ func (s *APITestSuite) TestJobStatusCompleted() {
 		Status:     models.JobStatusCompleted,
 	}
 	postgrestest.CreateJobs(s.T(), s.db, &j)
+	defer postgrestest.DeleteJobByID(s.T(), s.db, j.ID)
```
carlpartridge, line 170: These defers are all cleanup for tests adding things to the DB.

bcda/bcdacli/cli_test.go
```diff
@@ -391,34 +392,50 @@ func (s *CLITestSuite) TestImportCCLFDirectory() {
 	}
 
 	tests := []test{
-		{path: "../../shared_files/cclf/archives/valid/", err: errors.New("files skipped or failed import. See logs for more details"), expectedLogs: []string{"Successfully imported 6 files.", "Failed to import 0 files.", "Skipped 0 files."}},
-		{path: "../../shared_files/cclf/archives/invalid_bcd/", err: errors.New("failed to import 1 files"), expectedLogs: []string{"missing CCLF0 or CCLF8 file in zip", "", ""}},
-		{path: "../../shared_files/cclf/archives/skip/", err: errors.New("files failed to import or no files were imported. See logs for more details."), expectedLogs: []string{"Successfully imported 0 files.", "Failed to import 0 files.", "Skipped 1 files."}},
+		{
```
carlpartridge, line 395: Just spaced this out to make it easier to read.

log/logger.go
```diff
@@ -14,47 +14,32 @@ import (
...
-	Health   logrus.FieldLogger
+	API     logrus.FieldLogger = defaultLogger()
+	Auth    logrus.FieldLogger = defaultLogger()
+	BBAPI   logrus.FieldLogger = defaultLogger()
+	Request logrus.FieldLogger = defaultLogger()
+	SSAS    logrus.FieldLogger = defaultLogger()
+
+	Worker   logrus.FieldLogger = defaultLogger()
+	BBWorker logrus.FieldLogger = defaultLogger()
+	Health   logrus.FieldLogger = defaultLogger()
```
carlpartridge, line 25: Set up a default global logger which adds the customized fields that are available and outputs to stdout.  This seems useful to have everywhere and I think is a step in the direction of fargate as well.  I think one of the few things that can be argued to have global default presence are logging capabilities.

log/logger.go
```diff
@@ -14,47 +14,32 @@ import (
...
+func SetupLoggers() {
+	API = logger(logrus.New(), conf.GetEnv("BCDA_ERROR_LOG"), "api")
+	Auth = logger(logrus.New(), conf.GetEnv("AUTH_LOG"), "api")
+	BBAPI = logger(logrus.New(), conf.GetEnv("BCDA_BB_LOG"), "api")
+	Request = logger(logrus.New(), conf.GetEnv("BCDA_REQUEST_LOG"), "api")
+	SSAS = logger(logrus.New(), conf.GetEnv("BCDA_SSAS_LOG"), "api")
+
+	Worker = logger(logrus.New(), conf.GetEnv("BCDA_WORKER_ERROR_LOG"), "worker")
+	BBWorker = logger(logrus.New(), conf.GetEnv("BCDA_BB_LOG"), "worker")
+	Health = logger(logrus.New(), conf.GetEnv("WORKER_HEALTH_LOG"), "worker")
```
carlpartridge, line 38: This is called at API and worker entry points and points logs to files instead of stdout (which is what is currently happening).

laurenkrugen-navapbc: Do we think any of these files can be combined or modified? or should we leave them as they are for now and discuss separating once we are moving to fargate?

bcda/main.go
```diff
@@ -48,6 +48,7 @@ import (
 )
 
 func init() {
+	log.SetupLoggers()
```
michaeljvaldes, line 51: Do you think we can call `log.SetupLoggers()` from either here or cli.go instead of both places, since this main.go immediately calls cli.go's setupApp() ?

laurenkrugen-navapbc (approved):

log/logger.go
carlpartridge: Nice, I have main (app and worker) and cli covered.  And as far as I can tell the lambdas should be outputting to stdout and generally seem set up that way.  We could probably refactor them a bit if we wanted to re-use some of our logger customization but Im not sure if we would get a lot out of that (and also out of scope of this ticket).

PR #1197, opened 2025-08-20
BCDA-9247: update docker files

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9247

## 🛠 Changes

- refactored the various dockerfiles in bcda-app to conform to formatting best practices
- reordered and/or combined commands in dockerfiles to better leverage build caching
- added an app user in the bcda and worker dockerfiles for increased security
- explicitly defined a volume in the bcda and worker dockerfiles for sharing fhir data files, which can easily work together with a fargate task definition
- renamed packaging env vars as they are flagged by docker as being secrets when they are actually just pointing at a file path (BCDA_GPG_PUB_KEY and BCDA_GPG_SEC_KEY)

## ℹ️ Context

See our [confluence documentation](https://confluence.cms.gov/spaces/BCDA/pages/1353058967/2025-07-30+EC2+configuration+needed+for+Fargate) for more information and a high-level idea of what the deployed dockerfiles might need to do.

## 🧪 Validation

1. stop all running containers `docker stop $(docker ps -a -q)`
2. prune docker containers `docker system prune`
3. rebuild docker images/containers `make docker-bootstrap`
4. run tests `make test`


Dockerfiles/Dockerfile.bdt
```diff
@@ -2,7 +2,7 @@ FROM node:current-slim
 
 WORKDIR /
 
-RUN apt update -y && apt install git -y
+RUN apt update upgrade -y && apt install git -y
```
carlpartridge, line 5: Do we use this image anywhere?  If I remember right BDT was an old tool we used to do fhir testing (which is now replaced by our fhir_testing image?)?

Dockerfiles/Dockerfile.bcda
```diff
@@ -1,8 +1,16 @@
...
-
+# install dev packages if the environment argument was set to development
+ARG ENVIRONMENT
+RUN [ "$ENVIRONMENT" != "development" ] || apk update upgrade && \
+    apk add \
+        bash \
+        entr \
+        git \
+        openssl && \
+    go install github.com/go-delve/delve/cmd/dlv@latest
```
carlpartridge, line 12: Curious why we only need these tools in dev?

Dockerfiles/Dockerfile.package
```diff
@@ -2,14 +2,24 @@ FROM golang:1.24.4-alpine3.22
...
+    apk add \
+        build-base \
+        gcc \
+        gcompat \
+        git \
+        gpg \
+        gpg-agent \
+        rpm \
+        ruby \
+        ruby-dev
```
carlpartridge, line 21: I was able to remove ruby and ruby-dev from the ssas package image, could we try here?  I imagine that would save us quite a large download and time.  Was also able to remove the below command to `gem install ...`.

Dockerfiles/Dockerfile.ssas
```diff
@@ -1,8 +1,8 @@
 # syntax = docker/dockerfile:1-experimental
 FROM golang:1.24.4-alpine3.22 AS base
 
-RUN apk update upgrade
-RUN apk add git
+RUN apk update upgrade && \
```
carlpartridge, line 4: Where is this image used?  Is it used in local testing of bcda-api?  If so is there a way to publish an image from bcda-ssas repo and pull down here?  

How is this different than the ssas image being built in our ssas repo?  Should we unify them?

Is this all another PRs concerns?

Dockerfiles/Dockerfile.tests
```diff
@@ -1,14 +1,18 @@
 # syntax = docker/dockerfile:1-experimental
 FROM golang:1.24.4-alpine3.22
 
-RUN apk update upgrade
-
-RUN apk add bash build-base curl
+RUN apk update upgrade && \
```
carlpartridge, line 4: We arent setting user/group in this image.  I was having problems myself with this because it was ultimately calling a shell script in our CI and make file.  Curious if there is a good/better approach to this?

Dockerfiles/Dockerfile.bcda
michaeljvaldes: - entr: removed; I think this used to be used in part to replicate the functionality of docker watch
- git: removed; I don't think this is used
- bash: used for some scripts and makefile commands
- openssl: removed; not sure about this one
- dlv: used for debugging

Dockerfiles/Dockerfile.ssas
laurenkrugen-navapbc: it's very rare, but it is possible to get a rate limit via github when cloning the ssas repo. pulling it from ECR, might be better in the long run but also may not be applicable in this ticket/until we are moving forward with fargate.

Dockerfiles/Dockerfile.tests
```diff
@@ -1,14 +1,18 @@
...
+    apk add \
+        bash \
+        build-base \
+        curl
 
 RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.2.1
 
-WORKDIR /go/src/github.com/CMSgov/bcda-app
+RUN go install github.com/securego/gosec/v2/cmd/gosec@v2.21.4
```
laurenkrugen-navapbc, line 12: also might be outside the scope of this ticket: worthwhile to check if we need to upgrade gosec or gotestsum versions?

Dockerfiles/Dockerfile.fhir_testing
```diff
@@ -1,13 +1,18 @@
 FROM alpine:3.22
```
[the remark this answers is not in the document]
michaeljvaldes: I don't think so -- this particular dockerfile uses just an alpine image, no Golang required.

Dockerfiles/Dockerfile.tests
michaeljvaldes: I honestly could go either way since this isn't a deployed docker image

Dockerfiles/Dockerfile.package
michaeljvaldes: I think we need ruby to install fpm, which we're using in `build_and_package.sh` to package our binaries into rpm packages

Dockerfiles/Dockerfile.tests
laurenkrugen-navapbc: did the user lack the permissions to run the script?

carlpartridge: The script runs mkdir commands at the beginning which the user did not have permissions for, had trouble creating those directories as part of the docker build process, and I imagine later parts of the script as well would require some root privileges as well...

docker-compose.yml
```diff
@@ -44,18 +44,21 @@ services:
...
-      - FHIR_PAYLOAD_DIR=/go/src/github.com/CMSgov/bcda-app/bcdaworker/data
-      - FHIR_STAGING_DIR=/go/src/github.com/CMSgov/bcda-app/bcdaworker/data/tmpdata
+      # subdirectories of mounted volume
+      FHIR_ARCHIVE_DIR: /var/fhir/archive
+      FHIR_PAYLOAD_DIR: /var/fhir/data
+      FHIR_STAGING_DIR: /var/fhir/tmpdata
+      FHIR_TEMP_DIR: /var/fhir/TEMP
     volumes:
-      - fhir_payload_dir:/go/src/github.com/CMSgov/bcda-app/bcdaworker/data
+      - fhir:/var/fhir
```
carlpartridge, line 61: Nit: I kind of feel like `fhir` isnt the clearest descriptor of what this volume is for.  Ultimately this shared volume is used for fhir data response payloads from BFD, shared between the worker and API.  FHIR to me seems a lot more open ended (standards, data, schemas, etc).

carlpartridge (approved):

PR #1213, opened 2025-09-26
BCDA-9412: rename demo to v3

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9412

## 🛠 Changes

- Renamed demo endpoint to v3
- Added environment variables to disable endpoint in prod

## ℹ️ Context

As we prep for pilot of the v3 endpoint we are renaming what was presented for FHIR presentation from demo to v3.

## 🧪 Validation

Local lint, build and test

### DIFF
--- a/bcda/web/middleware/validation.go
+++ b/bcda/web/middleware/validation.go
@@ -136,7 +136,7 @@ func validateResourceTypes(r *http.Request, rw fhirResponseWriter, w http.Respon
 
 func validateTypeFilterParameter(r *http.Request, rw fhirResponseWriter, w http.ResponseWriter, version string) ([][]string, bool) {
 	params, ok := r.URL.Query()["_typeFilter"]
-	if version != "v3" || !ok {
+	if version != constants.V3Version || !ok {
 		return nil, true
 	}
 
@@ -310,7 +310,7 @@ func getRespWriter(version string) (fhirResponseWriter, error) {
 	case "v2":
 		return responseutilsv2.NewFhirResponseWriter(), nil
 	case constants.V3Version:
-		return responseutilsv2.NewFhirResponseWriter(), nil // TODO: V3
+		return responseutilsv2.NewFhirResponseWriter(), nil
 	default:
 		return nil, fmt.Errorf("unexpected API version: %s", version)
 	}


bcda/web/middleware/validation.go
```diff
@@ -235,7 +235,7 @@ func getKeys(kv map[string]struct{}) []string {
 	return keys
 }
 
-var versionExp = regexp.MustCompile(`\/api\/(v\d+|demo)\/`) // TODO: V3
+var versionExp = regexp.MustCompile(`\/api\/(v\d+|v3)\/`)
```
carlpartridge, line 238: I think you dont need this part of the regex as (v\d) should cover v3.

colby-seyferth-nava (approved):
nice modularization of param validation

bcda/web/middleware/validation.go
```diff
@@ -46,6 +46,152 @@ func GetRequestParamsFromCtx(ctx context.Context) (RequestParameters, bool) {
...
+			rw.Exception(r.Context(), w, http.StatusBadRequest, responseutils.RequestErr, errMsg)
+			return nil, false
+		}
+	}
+	return resourceTypes, true
+}
+
+func validateTypeFilterParameter(r *http.Request, rw fhirResponseWriter, w http.ResponseWriter, version string) ([][]string, bool) {
+	params, ok := r.URL.Query()["_typeFilter"]
+	if version != "v3" || !ok {
```
colby-seyferth-nava, line 139: nitpick: you could use the `constants.V3Version` constant instead of the `"v3"` string here

bcda/web/middleware/validation.go
```diff
@@ -46,6 +46,152 @@ func GetRequestParamsFromCtx(ctx context.Context) (RequestParameters, bool) {
...
+			} else {
+				errMsg := fmt.Sprintf("Invalid _typeFilter subquery parameter: %s", paramName)
+				log.API.Error(errMsg)
+				rw.Exception(r.Context(), w, http.StatusBadRequest, responseutils.RequestErr, errMsg)
+				return nil, false
+			}
+		}
+	}
+	return typeFilterParams, true
+}
```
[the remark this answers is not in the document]
carlpartridge: These functions are used for all our endpoints right?  Just a bit worried that there might be some kind of unseen effect here.  Is our existing testing for this area comprehensive enough as is?  My guess is that you just ripped blocks of logic out into its own function, then passing in anything required, so it should be pretty safe?

colby-seyferth-nava (approved):

PR #1282, opened 2025-12-19
BCDA-9461: Add typeFilter tests to v3 smoke tests

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9461

## 🛠 Changes

Add typeFilter smoke tests.  Add missing unit tests.

## ℹ️ Context

With v3 we are adding new advanced filtering options.  These new smoke tests are designed to do some amount of making sure they are working as expected.  The heavy duty verification is done via unit-tests.

## 🧪 Validation

Postman smoke test run.


test/postman_test/BCDA_V3_Tests.postman_collection.json
```diff
@@ -2028,6 +2028,369 @@
...
+													"    ",
+													"    pm.test(\"Schema is valid\", function() {",
+													"        pm.expect(tv4.validate(respJson, schema)).to.be.true;",
+													"    });",
+													"",
+													"    pm.test(\"Contains Required Resources\", () => {",
+													"        const requiredResources = [\"ExplanationOfBenefit\"];",
+													"        const partiallyAdjudicatedResources = [\"Patient\", \"Coverage\", \"Claim\", \"ClaimResponse\"];",
+													"        const returnedResources = respJson.output.map(r => r.type);",
+													"",
```
carlpartridge: @colby-seyferth-nava This part is failing and Im not sure I understand why.  Here is the error log from the test:
```
 1.  AssertionError  Contains Required Resources                                                                            
                     ExplanationOfBenefit is required: expected [ 'Patient', 'Coverage' ] to include 'ExplanationOfBenefit' 
                     at assertion:2 in test-script                                                                          
                     inside "Authenticated Endpoints / With Token / Check Filters / Get EOB export v3 job status"   
```
I copied much of the previous end to end smoke test for v3 EOB, specifically this `"name": "Get EOB export v3 job status",` test, including this check.

carlpartridge: > for v3, the claim and claimresponse resources won't be returned. the explanationOfBenefit resource is a partiallyAdjudicatedResource. so the list should be Patient, Coverage, and EOB?

Thinking about it now a better question would probably be, what/how would you want to verify a successful v3 request with typeFilter?

colby-seyferth-nava: I think the best would be to do a search for PartiallyAdjudicated claims (typefilter param with _tag = partiallyAdjudicated subquery), and then check the EOB resource content to make sure each of the EOBs in the response are partially adjudicated.

Not sure the best way to check the response content using postman / newman. I think Lauren looked into that in a SPIKE before maybe?

PR #1352, opened 2026-04-03
BCDA-9735: _typeFilter Parsing Updates

colby-seyferth-nava:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9735

## 🛠 Changes

Change the structure of the _typeFilter parameter representation from a 2d string array to a struct. This way, we will be able to more accurately convey the _typeFilter params when we query BFD. 

This change is just the initial parsing, but then we flatten the struct back down to a 2d string array before we pass the data to the job. So there will be no functional change. 

## ℹ️ Context

Initial _typeFilter param parsing was quick for connectathon, and was knowingly creating this technical debt in order to get the /demo endpoint out in time. This will eliminate the tech debt and allow an API consumer (in the follow-up PR) to make an $export request with multiple _typeFilter parameters

## 🧪 Validation

No functional change, just validated with the Unit test suite.


bcda/api/requests.go
```diff
@@ -945,6 +948,21 @@ func GetAuthDataFromCtx(r *http.Request) (data auth.AuthData, err error) {
 	return
 }
 
+// Flatten the param from the struct into the 2D array that is already being used.
+// TODO: remove this, and pass as a struct. Then update the downstream logic to handle the struct
+func FlattenTypefilterParams(typefilterParams []middleware.TypeFilterParameter) [][]string {
+	var typeFilterParams [][]string
+	for _, typefilterParam := range typefilterParams {
+		if typefilterParam.ResourceType == "ExplanationOfBenefit" {
```
bhagatparwinder, line 956: Previously, we grabbed all query parameters regardless of the resource type provided in the query string. Enforcing `ExplanationOfBenefit` here, any `_typeFilter` parameters configured for another resource (like `Patient` or `Coverage`) will now be silently dropped and ignored by the downstream logic.

Confirming if BCDA's FHIR implementation strictly uses `_typeFilter` only for `ExplanationOfBenefit`

colby-seyferth-nava: We only accept EOB typefilter subqueries right now. We fail in validation.go if there is a non EOB subquery: https://github.com/CMSgov/bcda-app/blob/6c7dcea2a8a13b487f99b50885877d391ebb03c4/bcda/web/middleware/validation.go#L215-L225

So this my even just be overkill. Current behavior is only _typeFilter for EOB

bhagatparwinder (approved):

laurenkrugen-navapbc (approved):

PR #1358, opened 2026-04-09
BCDA-9962: update postman test assertion

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9962

## 🛠 Changes

- Updated test assertion

## ℹ️ Context

We are appropriately getting an error back from the API that the CCLF file is out of date, but we are getting a different error string in the response than what we are expecting in the smoke test. Smoke tests are updated to match the new string for missing attribution files

## 🧪 Validation

Executed postman tests in dev


bhagatparwinder:
This is happening due to timestamps/cutoff date in the DB. This is why it doesn't fail in all environments, but only in test and prod. If I make this change, it will work in test and prod but start failing in dev and sandbox. Closing this PR and addressing the database.

PR #1378, opened 2026-05-05
BCDA-9988: Update _typeFilter Validation and Structure

colby-seyferth-nava:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9988

## 🛠 Changes

Added Parameter Validation:
 - Only allow one _typeFilter parameter to be passed. Return an error if a user makes a request like `\$export?_typeFilter=subquery1&_typeFilter=subquery2
 - Validate service-date _typefilter subquery param. Only accept valid FHIR date param
 - Update _typeFilter param structure

## ℹ️ Context

 - Our initial support did not correctly handle the logical OR scenario of a user passing multiple _typeFilter parameters. it would just quietly combine all the _typeFilter subquery params and treat them like a logical AND. Instead of supporting the OR logic, we are just returning an error if a user tries to pass multiple _typeFilter params. 

 - validating the service-date param at the time of the initial export allows us to catch any errors early, before we fail when querying BFD.

 - Changing the internal structure of how we represent the _typeFilter param from a [][]String to a Struct  is cleaner and will allow the support to be more flexible going forward, if we want to support other resourceTypes, Modifiers, or the logical OR multiple _typeFilter Params

## 🧪 Validation

Unit tests pass

Deploy locally and manually test:
 - Multiple _typeFilter params: return OperationOutcome
 - valid and invalid service-date subqueries
 - valid and invalid _tag queries


bcda/web/middleware/validation_test.go
```diff
@@ -175,6 +176,45 @@ func TestValidateOutcomeSubqueryParameter(t *testing.T) {
...
+		fhirDate string
+		expected error
+	}{
+		// Valid
+		{"yearOnly", "2024", nil},
+		{"yearMonth", "2024-01", nil},
+		{"fullDate", "2024-01-15", nil},
+		{"dateTimeNoTZ", "2024-01-15T10:30:00", nil},
+		{"dateTimeWithZ", "2024-01-15T10:30:00Z", nil},
+		{"dateTimeWithOffset", "2024-01-15T10:30:00-05:00", nil},
```
bhagatparwinder: @colby-seyferth-nava could we add a valid test case with a positive offset (like `2024-01-15T10:30:00+07:00`) since we have that in implementation?

bcda/api/requests.go
```diff
@@ -961,7 +972,7 @@ func GetAuthDataFromCtx(r *http.Request) (data auth.AuthData, err error) {
 
 // Flatten the param from the struct into the 2D array that is already being used.
 // TODO: remove this, and pass as a struct. Then update the downstream logic to handle the struct
-func FlattenTypefilterParams(typefilterParams []middleware.TypeFilterParameter) [][]string {
+func FlattenTypefilterParams(typefilterParams []utils.TypeFilterParameter) [][]string {
```
bhagatparwinder: Do we use this anymore? I see we removed the only call to it above (and we directly pass a structured parameter to the queue).

bcda/api/requests.go
```diff
@@ -945,7 +942,21 @@ func (h *Handler) omitSharedSystemForNonPAC(ctx context.Context, typeFilter [][]
...
 	// This function is only called when ExplanationOfBenefit is in the resource types
 	tagValue := "https://bluebutton.cms.gov/fhir/CodeSystem/System-Type|NationalClaimsHistory,https://bluebutton.cms.gov/fhir/CodeSystem/System-Type|DDPS"
-	typeFilter = append(typeFilter, []string{"_tag", tagValue})
+	subqueryParam := utils.TypeFilterSubqueryParam{
+		Name:  "_tag",
+		Value: tagValue,
+	}
+
+	// if there is no _typeFilter param passed, create a new one and add this _tag filter
+	if typeFilter.QueryParameters == nil {
```
bhagatparwinder: I think we can make this check _safer_
```suggestion
	if len(typeFilter.QueryParameters) == 0 {
```

bcda/api/requests.go
colby-seyferth-nava: ope, yeah, definitely not needed anymore, should delete, good catch!

bcda/web/middleware/validation_test.go
colby-seyferth-nava: good idea 👍

bcda/api/requests.go
```diff
@@ -945,7 +942,21 @@ func (h *Handler) omitSharedSystemForNonPAC(ctx context.Context, typeFilter [][]
 	// This tag filters response data by System-Type=NCH OR System-Type=DDPS
 	// This function is only called when ExplanationOfBenefit is in the resource types
 	tagValue := "https://bluebutton.cms.gov/fhir/CodeSystem/System-Type|NationalClaimsHistory,https://bluebutton.cms.gov/fhir/CodeSystem/System-Type|DDPS"
-	typeFilter = append(typeFilter, []string{"_tag", tagValue})
+	subqueryParam := utils.TypeFilterSubqueryParam{
+		Name:  "_tag",
+		Value: tagValue,
+	}
+
+	// if there is no _typeFilter param passed, create a new one and add this _tag filter
```
carlpartridge, line 951: Is this akin to a default?

bcda/utils/common.go
```diff
@@ -11,6 +11,16 @@ import (
...
 
+type TypeFilterParameter struct {
+	ResourceType    string
+	QueryParameters []TypeFilterSubqueryParam
+}
+
+type TypeFilterSubqueryParam struct {
+	Name  string
+	Value string
+}
```
carlpartridge: Not a big fan of having these in utils/common.  How about moving them to a new types file under bcda/client/fhir?  It kind of looks like we dont use that area of the code much but I think we are going to be expanding on our fhir types in the near future so having an independent and specifc location seems useful.

colby-seyferth-nava: Nice, I was trying to find a place this could live and it felt bad in utils/common. Thanks for the suggestion, I'll make the change

bcda/api/requests.go
colby-seyferth-nava: Yeah, this PR shouldn't change the existing behavior. this `omitSharedSystemForNonPAC` just makes sure to specify the Non-SharedSystems tags if you are making a request and don't have access to SS data.

bcda/utils/common.go
bhagatparwinder: Thanks for moving it! I wasn't sure if you were still addressing the PR, but it looks like we can delete the `TypeFilterParameter` and `TypeFilterSubqueryParam` from here.

bhagatparwinder (dismissed):

bcda/utils/common.go
colby-seyferth-nava: ope, thanks, removed!

carlpartridge (approved):

bhagatparwinder (approved):

PR #1380, opened 2026-05-05
BCDA-9952: Remove databasetest package

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9952

## 🛠 Changes

- databasetest package removed
- tests using databasetest package updated to use database containers package instead
- `testdata/` files updated to be sql instead of yml (to be compatible with database containers package)  
- tests utilizing database containers package and table driven test format updated to use the suite.Run instead of suite's `t testing.T` reference to run, since it lacks the subtest setup and teardown
- added a `GetCCLFFiles` to `repository.go`, since it is likely going to be used in many other tests and we will eventually deprecate `postgrestest`.

## ℹ️ Context

Made as part of effort to move more towards idempotent tests, and standardizing database integration testing. We have another custom database testing postgrestest, which will also need to be removed.

## 🧪 Validation

All tests passing.


bcda/api/requests_test.go
```diff
@@ -1217,7 +1212,9 @@ func (s *RequestsTestSuite) genPatientRequest(rp middleware.RequestParameters) *
 
 func (s *RequestsTestSuite) genASRequest() *http.Request {
 	req := httptest.NewRequest("GET", "http://bcda.cms.gov/api/v1/attribution_status", nil)
-	aco := postgrestest.GetACOByUUID(s.T(), s.db, s.acoID)
+	aco, err := s.r.GetACOByUUID(context.Background(), s.acoID)
+	require.NoError(s.T(), err)
+	//aco := postgrestest.GetACOByUUID(s.T(), s.db, )
```
carlpartridge: Nit: remove unused line.

bcda/models/postgres/repository.go
```diff
@@ -604,6 +604,38 @@ func (r *Repository) GetJobKey(ctx context.Context, jobID uint, fileName string)
 	return jk, nil
 }
 
+func (r *Repository) GetCCLFFiles(ctx context.Context, field, value string) ([]models.CCLFFile, error) {
```
carlpartridge: Im hesitant to add a new method to the ORM that seems to only be used for testing purposes.  Is there another way to write your tests?  I think even if its a bit more verbose it would be preferable to keep the ORM on the slimmer side.

bcda/service/service.go
```diff
@@ -152,12 +152,7 @@ func (s *service) FindOldCCLFFile(ctx context.Context, cmsID string, since time.
 		return 0, err
 	}
 
-	if cclfFileOld == nil {
-		s.logger.Infof("Unable to find CCLF8 File for cmsID %s prior to date: %s; all beneficiaries will be considered NEW", cmsID, since)
-		return 0, nil
-	} else {
-		return cclfFileOld.ID, nil
-	}
+	return cclfFileOld.ID, nil
```
carlpartridge, line 155: How come we are losing the logger info here?  Also it seems like this is a change of functionality (in the nil case we are now returning nil instead of 0)?  (or does Go return a 0 in the case of a nil int?)

bcda/models/postgres/repository.go
laurenkrugen-navapbc: Yeah, I was also on the fence about this. I ultimately put it in here because I thought there was a chance that this could be called in other test packages, especially as we cleanup and remove postgrestest. By putting it in this package, it's still very much accessible, but it's not great having methods used for testing blended with our actual methods. 

We have a few options:

- put this in another package with no dependencies, so it may be called from all test packages (this was pulled from postgrestest, and it used in many places
- create a utility/test file in the repository package, that contains methods for testing needs
- manually create these methods in the package in which it's used, which could result in duplication

bcda/service/service.go
laurenkrugen-navapbc: I meant to leave a comment here; this is unreachable code. it's not possible for oldcclffile to be nil, because if it's not found, that results in an error. The else portion is reachable, but the initial if will never be reached.

bcda/api/requests_test.go
```diff
@@ -85,39 +78,40 @@ func TestRequestsTestSuite(t *testing.T) {
...
-	db, _, _ := databasetest.CreateDatabase(s.T(), "../../db/migrations/bcda/", true)
-	s.db = db
-	s.pool = database.ConnectPool()
-	tf, err := testfixtures.New(
-		testfixtures.Database(db),
-		testfixtures.Dialect("postgres"),
-		testfixtures.Directory("testdata/"),
-	)
+	var err error
+	s.dbContainer, err = db.NewTestDatabaseContainer()
```
bhagatparwinder, line 85: I ran these tests locally. We create this test db container, but we never terminate it. On my local it left a test container running in background

```
CONTAINER ID   IMAGE                COMMAND                  STATUS          PORTS                                           NAMES
1deb91e5ef56   postgres:16-alpine   "docker-entrypoint.s…"   Up 12 seconds   0.0.0.0:51086->5432/tcp, [::]:51086->5432/tcp   infallible_chatterjee
```


I think we need to introduce a tiny `Container.Terminate(ctx)`

bcda/api/requests_test.go
```diff
@@ -85,39 +78,40 @@ func TestRequestsTestSuite(t *testing.T) {
...
+	require.NoError(s.T(), err)
+	s.r = postgres.NewRepository(s.db)
+	require.NoError(s.T(), err)
+
 }
 
 func (s *RequestsTestSuite) TearDownTest() {
 	err := conf.SetEnv(s.T(), "BCDA_ENABLE_RUNOUT", s.runoutEnabledEnvVar)
 	assert.Empty(s.T(), err)
+	s.db.Close()
```
bhagatparwinder, line 123: Should we also have `s.pool.Close()`?

bcda/models/postgres/repository.go
carlpartridge: Independent utility/test package sounds like a lesser version of what we are removing :(.  Is there a way to run these tests by just grabbing things using existing methods?  If all its doing is getting a bunch of CCLF files we could just run a straight sql query?  That could be shared pretty easily as its just a string?  Can you think of a situation where the app itself would want to get numerous cclf files in this manner?

Interested in others thoughts on this as well!  @bhagatparwinder @colby-seyferth-nava @juliareynolds-nava

bcda/api/requests_test.go
laurenkrugen-navapbc: There's another container called ryuk which is like a sidecar container that cleans up the test container once the tests are done. I don't have a container still running on mine once it's done, which is odd. I will introduce the terminate for the `SuiteTearDown` so we don't have stale containers hanging around.

bcda/api/requests_test.go
laurenkrugen-navapbc: Great question! closing the database connection will automatically also close the pool. The only reason that we close the connection is to restore the base migrations (initial data set, table set up, migrations, etc). The database container cannot be restored to that "fresh" state if the pool/db is still open. 

It doesn't feel graceful to close both the pool and db using `db.close()`, but I am not seeing much evidence to support or contradict that. I haven't noticed any odd behaviors during testing or performance problems, so maybe we keep an eye on it?

bcda/models/postgres/repository.go
laurenkrugen-navapbc: That's the main question -- we need sql methods in various test places to get what we need. Do we keep it isolated within test files, knowing that SQ will complain about possible duplication _or_ it starts getting import across too many packages which could introduce more circular dependencies?

If we aren't sure, this isn't necessarily adding _more_ mess at the moment (mostly moving it around) so we can take some time to discuss how we want to tackle testing functionalities that may be shared across packages more, if needed.

bcda/api/requests_test.go
laurenkrugen-navapbc: oop, missed this one, thank you!

bcda/models/postgres/repository.go
bhagatparwinder: I lean a bit towards Carl on this one. We should avoid adding test-only methods to our prod `Repository` interface. SQ might flag duplication in test code, but I don't think that's bad (especially for Go, "[A little copying is better than a little dependency](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s)")

I think we should:

1. Duplicate the raw SQL query directly in the `_test.go` files that need it, or share the query as a string constant.
    - If SonarQube gets too noisy about it, we can adjust its rules, maybe.
2. If we find that this exact query is used in a dozen different places, we could create a standalone test helper function inside our new database container package.

carlpartridge (approved):

bhagatparwinder (approved):

PR #1381, opened 2026-05-05
BCDA-9679 move terraform into bcda-app monorepo

juliareynolds-nava:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9679

## 🛠 Changes

moved terraform code and scripts from the bcda-ops repository
adapted github action workflows to reference local terraform and scripts

## ℹ️ Context

These changes are to centralize bcda code repositories into one

## 🧪 Validation

See tests below
Post-deployment to dev, smoke tests have passed:  https://github.com/CMSgov/bcda-app/actions/runs/25447236776/job/74654568124


.github/workflows/tag-and-release.yml
```diff
@@ -77,7 +61,7 @@ jobs:
         uses: actions/download-artifact@v4.2.1
         with:
           name: release-scripts
-          path: scripts/tag_and_release
+          path: ./scripts/scripts/tag_and_release
```
bhagatparwinder: I think this should be
```suggestion
          path: ./scripts/tag_and_release
```

terraform/dev/main.tf
```diff
@@ -0,0 +1,1558 @@
...
+  name = "/bcda/${var.env}/sensitive/db_admin_password"
+}
+
+module "platform" {
+  source    = "github.com/CMSgov/cdap.git//terraform/modules/platform?ref=ff2ef539fb06f2c98f0e3ce0c8f922bdacb96d66"
+  providers = { aws = aws, aws.secondary = aws.secondary }
+
+  app         = "bcda"
+  env         = var.env
+  root_module = "github.com/CMSgov/bcda-ops/tree/main/terraform/${var.env}"
```
bhagatparwinder: These are still referencing `bcda-ops`

terraform/sandbox/main.tf
```diff
@@ -0,0 +1,1575 @@
...
+  name = "/bcda/${var.env}/sensitive/db_admin_password"
+}
+
+module "platform" {
+  source    = "github.com/CMSgov/cdap.git//terraform/modules/platform?ref=ff2ef539fb06f2c98f0e3ce0c8f922bdacb96d66"
+  providers = { aws = aws, aws.secondary = aws.secondary }
+
+  app         = "bcda"
+  env         = var.env
+  root_module = "github.com/CMSgov/bcda-ops/tree/main/terraform/${var.env}"
```
bhagatparwinder: similar bcda-ops reference here

terraform/test/main.tf
```diff
@@ -0,0 +1,1575 @@
...
+  name = "/bcda/${var.env}/sensitive/db_admin_password"
+}
+
+module "platform" {
+  source    = "github.com/CMSgov/cdap.git//terraform/modules/platform?ref=ff2ef539fb06f2c98f0e3ce0c8f922bdacb96d66"
+  providers = { aws = aws, aws.secondary = aws.secondary }
+
+  app         = "bcda"
+  env         = var.env
+  root_module = "github.com/CMSgov/bcda-ops/tree/main/terraform/${var.env}"
```
bhagatparwinder: another one for bcda-ops

terraform/prod/main.tf
```diff
@@ -0,0 +1,1704 @@
...
+  name = "/bcda/${var.env}/sensitive/db_admin_password"
+}
+
+module "platform" {
+  source    = "github.com/CMSgov/cdap.git//terraform/modules/platform?ref=ff2ef539fb06f2c98f0e3ce0c8f922bdacb96d66"
+  providers = { aws = aws, aws.secondary = aws.secondary }
+
+  app         = "bcda"
+  env         = var.env
+  root_module = "github.com/CMSgov/bcda-ops/tree/main/terraform/${var.env}"
```
bhagatparwinder: here as well (bcda-ops)

bhagatparwinder (commented):
We also need to update the README.md in the root to not reference `bcda-ops` (See section for environment variables).

bhagatparwinder (dismissed):

laurenkrugen-navapbc:
I see we did a deployment to verify things are working as expected; should we anticipate any issues with the terraform state files?

Do we need to check ssas and verify if there are any workflows that reference bcda-ops?

.github/workflows/ci-checks.yml
```diff
@@ -109,7 +109,7 @@ jobs:
         uses: sonarsource/sonarqube-scan-action@299e4b793aaa83bf2aba7c9c14bedbb485688ec4
         with:
           args:
-            -Dsonar.scanner.arch=aarch64
```
carlpartridge: This is wrong?  This seems like its resurfaced again?

.github/workflows/deploy-all.yml
```diff
@@ -12,10 +12,6 @@ on:
         description: 'Ref of bcda-app repo to apply migrations from (e.g. r123, main)'
         required: true
         type: string
-      ops_ref:
-        description: 'Ref of bcda-ops repo to deploy (e.g. r123, main)'
-        required: true
-        type: string
```
carlpartridge, line 18: We cant get rid of these until sandbox and prod are moved over as well right?

.github/workflows/deploy-all.yml
```diff
@@ -187,12 +179,6 @@ jobs:
         working-directory: ops/services/10-config
         run: |
           tofu apply bcda-release-sops.tfplan
-      - name: Checkout bcda-ops
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.3.0
-        with:
-          repository: CMSgov/bcda-ops
-          ref: ${{ inputs.ops_ref }}
-          token: ${{ env.GITHUB_TOKEN }}
```
carlpartridge, line 198: We will need to create some forking logic to deploy dev/test from bcda-app and sandbox/prod from bcda-ops.  At least until we move everything over.  Alternatively we could create a second workflow that just handles dev/test in the new style, but that could be more complex then the logic fork.

scripts/tag_and_release/github_release.py
```diff
@@ -0,0 +1,61 @@
...
+    )
+ 
+    parser.add_argument(
+        '--repo', dest='repo', type=str,
+        help='The repository of the release (i.e., /repos/CMSgov/bcda-app/releases)'
+    )
+
+    args = parser.parse_args()
+
+    main(args.release, args.release_file, args.repo)
```
carlpartridge: Im fairly sure these release related scripts raised security alarms.  Is there a good way to verify that before merging these in (and thus springing security alarms against ourselves)?

scripts/refresh_test_attribution.sh
```diff
@@ -0,0 +1,82 @@
+# This file refreshes attribution data (CCLF & CSV) for all test ACOs.
```
carlpartridge: A bit worried about moving this to public as it exposes some testing ACO IDs as well as some inner workings of our business logic.

terraform/insights/queries/get_active_acos.sql
```diff
@@ -0,0 +1,113 @@
+select
```
carlpartridge: This whole insights dir I think is no longer in use?  Im also a bit worried about including the sql queries as they go somewhat in depth into our inner db workings and business logic.

terraform/static_site/modules/cloudfront/main.tf
```diff
@@ -0,0 +1,94 @@
+# Fetch the region-agnostic WAF ACL meant specifically for Cloudfront
```
carlpartridge: These static site modules are related to bcda static site right?  If so we should not move them as part of this ticket I think?

carlpartridge (changes requested):

carlpartridge:
> I see we did a deployment to verify things are working as expected; should we anticipate any issues with the terraform state files?
> 
> Do we need to check ssas and verify if there are any workflows that reference [bcda-ops](https://jira.cms.gov/browse/BCDA-ops)?

My understanding is that we arent removing anything out of bcda-ops yet, so nothing should break yet...  This is also a big announcement to the team as the longer the two structures exist, the more likely we are to find them out of sync...

carlpartridge:
Looks like sonarqube is failing on needing node installed to analyze js code.  The js code seems to be coming from cloudfront (part of the static site) and insights (unused).  I recommend we remove those (as suggested in my other PR comments) over adding node to our workflow.

juliareynolds-nava:
> I see we did a deployment to verify things are working as expected; should we anticipate any issues with the terraform state files?
> 
> Do we need to check ssas and verify if there are any workflows that reference [bcda-ops](https://jira.cms.gov/browse/BCDA-ops)?

I don't expect state file issues.
Good point about ssas!  Here's the PR:  https://github.com/CMSgov/bcda-ssas-app/pull/318

.github/workflows/deploy-all.yml
```diff
@@ -177,9 +177,6 @@ jobs:
         uses: cmsgov/cdap/actions/setup-sops@f4c14d47cc20e7f6de9112d7155af1213c9bca5a
       - name: Install YQ
         uses: cmsgov/cdap/actions/setup-yq@0b65099ef9232cfa2dfd4eea89cf8b9fa6052013
-      - name: Install OpenTofu for SOPS
-        working-directory: ops/services/10-config
-        run: tenv tofu install
```
carlpartridge, line 182: is there a reason we are getting rid of this?

.github/workflows/refresh_test_attribution.yml
carlpartridge, line 61: Looks like we need a /scrips/refresh_test_attribution.sh script for this to work out of bcda-app instead of bcda-ops.

.github/workflows/tag-and-release.yml
```diff
@@ -8,10 +8,6 @@ on:
         description: 'Release version (or branch name)'
         required: true
         type: string
-      ops_release_version:
-        description: 'Release version for bcda-ops (or branch name)'
-        required: true
-        type: string
```
carlpartridge, line 14: I think until we have moved completely out of all envs we should keep this.  Might be useful to tag it with a comment so we can find it later on more easily.

scripts/mark_deployment.py
```diff
@@ -0,0 +1,77 @@
+import argparse
```
carlpartridge: Seems like we are moving up our removal of NewRelic timeline so could be argued to just remove this script and references to it.

terraform/dev/main.tf
```diff
@@ -0,0 +1,1558 @@
...
+  cluster_identifier    = "bcda-${var.env}-aurora"
+  deletion_protection   = !module.platform.is_ephemeral_env
+  instance_class        = "db.r8g.large"
+  maintenance_window    = "sun:23:08-sun:23:38"
+  monitoring_interval   = var.monitoring_interval
+  monitoring_role_arn   = aws_iam_role.db_monitoring.arn
+  username              = "bcda"
+  password              = data.aws_ssm_parameter.db_admin_password.value
+  platform              = module.platform
+  subnet_group_override = "bcda-${var.env}-rds-subnets"
```
carlpartridge, line 971: Im not seeing the latest changes to the tf code from bcda-ops here (engine_version).  Or did we update the source version sha to match the latest cdap changes?
Looks like there are 3 changes since you started this PR: https://github.com/CMSgov/bcda-ops/pulls?q=is%3Apr

terraform/modules/insights_data_event_processor/main.tf
```diff
@@ -0,0 +1,67 @@
+locals {
```
carlpartridge: I dont think we need these insights related things?

terraform/prod/local.tf
carlpartridge, line 1: I thought we were going to just move dev and test first?  Or did that make things too complicated?

terraform/prod/local.tf
```diff
@@ -0,0 +1,264 @@
...
+  worker_asg_service_name = "worker"
+  worker_asg_processes    = "worker"
+
+  # Scaling
+  api_desired_min    = 2
+  api_desired_max    = 4
+  ssas_desired_min   = 4
+  ssas_desired_max   = 10
+  worker_desired_min = 2
+  worker_desired_max = 12
```
carlpartridge: These do not have the latest from bcda-ops.

.github/workflows/deploy-all-from-monorepo.yml
carlpartridge, line 1: Is the goal of this to allow for parallel deploy workflows?  The old still running tf through bcda-ops and this new one running tf through bcda-app?

terraform/dev/main.tf
juliareynolds-nava: Good catch.

.github/workflows/deploy-all-from-monorepo.yml
juliareynolds-nava: Yes

terraform/prod/local.tf
juliareynolds-nava: I'm moving all trying to stay in sync with changes in bcda-ops.

.github/workflows/deploy-all.yml
```diff
@@ -94,10 +94,6 @@ on:
         type: string
         default: 'dev'
 
-permissions:
-  id-token: write
-  contents: read
-
```
carlpartridge, line 100: Why these changes?

.github/workflows/deploy-all.yml
```diff
@@ -239,13 +247,11 @@ jobs:
           aws ecs update-service --cluster bcda-${{ env.RELEASE_ENV }} --service bcda-${{ env.RELEASE_ENV }}-api --force-new-deployment
           aws ecs update-service --cluster bcda-${{ env.RELEASE_ENV }} --service bcda-${{ env.RELEASE_ENV }}-ssas --force-new-deployment
           aws ecs update-service --cluster bcda-${{ env.RELEASE_ENV }} --service bcda-${{ env.RELEASE_ENV }}-worker --force-new-deployment
-      - name: Upload notify script
-        uses: actions/upload-artifact@v4
-        with:
-          name: notify-script
-          path: ./scripts/mark_deployment.py
```
carlpartridge, line 246: Why are we getting rid of this?  Any changes to this file are a bit worrying as this is our current, functional deploy pipeline.  Would prefer this remain 100% functional/reliable until we have vetted the new pipeline deploy process.

.github/workflows/deploy-all.yml
```diff
@@ -334,54 +349,13 @@ jobs:
...
-        if: ${{ env.RELEASE_ENV == 'dev' }}
-        run: |
-          aws ecs wait services-stable --cluster bcda-${{ env.RELEASE_ENV }} --services bcda-${{ env.RELEASE_ENV }}-api
-          API_ARN=$(aws ecs list-tasks --cluster=bcda-${{ env.RELEASE_ENV }} --service-name=bcda-${{ env.RELEASE_ENV }}-api --query 'taskArns[0]' --output text)
-          API_IMAGE=$(aws ecs describe-tasks --cluster=bcda-${{ env.RELEASE_ENV }} --tasks=$API_ARN --query="tasks[0].containers[?name == 'api'].image" --output=text)
-          export API_VERSION=${API_IMAGE##*:}
-          python3 scripts/mark_deployment.py \
-            --app_id ${{ env.NEWRELIC_APP_ID }} \
-            --api_key ${{ env.NEWRELIC_API_KEY }} \
-            --version ${API_VERSION}
```
carlpartridge: This might break existing alarms so we should be a bit careful with removing New Relic related things.  Probably too much for this PR.

.github/workflows/refresh_test_attribution.yml
```diff
@@ -41,10 +41,10 @@ jobs:
       - name: Install psql
         run: |
           sudo dnf install postgresql16 -y
-      - name: checkout bcda-ops
```
carlpartridge, line 44: I am leaning towards the following steps:
- move everything over, but everything still runs on old workflows (via bcda-ops)
- start swapping to new workflows (ops -> app)

This would mean more minimal changes over a slightly longer period of time.  While also leaving us a bit safer in case things go wrong.  The big downside is having to keep the tf between the two repos in sync.

So following this preference we would undo most changes to these existing workflows for now.  If you think that would be too much effort then we can see how other developers feel and have a larger conversation.

scripts/refresh_test_attribution.sh
carlpartridge: I think you re-added this on account of another one of my PR comments.  I think overall I would prefer this out of here and we should probably have a conversation with Brian about exposure of things like this.

carlpartridge:
It looks like there are 14 new security issues raised by moving some of these scripts into bcda-app :(
https://sonarqube.cloud.cms.gov/project/issues?branch=420aab72cda8b7ee3de08c0ae9e0032aa12a766f&issueStatuses=OPEN%2CCONFIRMED&inNewCodePeriod=true&id=bcda-aco-api
I think we could leave the scripts out for now and just concentrate on the tf?

.github/workflows/deploy-all.yml
juliareynolds-nava: sonarqube complained that this should be at a job level

.github/workflows/ci-checks.yml
```diff
@@ -109,7 +109,7 @@ jobs:
         uses: sonarsource/sonarqube-scan-action@299e4b793aaa83bf2aba7c9c14bedbb485688ec4
         with:
           args:
-            -Dsonar.scanner.arch=arm64
+            -Dsonar.scanner.arch=aarch64
```
carlpartridge, line 112: Ive lost track at this point of which is the correct one :(
Maybe it accepts either?  Sonarqube has been passing and it seems to be passing in this branch as well...

terraform/shared-services/main.tf
carlpartridge, line 1: Where is shared-services used?  I dont see any references in either bcda-app nor bcda-ops?

terraform/test/local.tf
```diff
@@ -0,0 +1,239 @@
...
+  aco_ms_cidr_blocks = [
+    "10.247.240.0/22",
+    "10.247.252.0/22",
+    "10.131.64.0/22",
+    "10.242.104.0/21",
+    "10.223.38.0/25",
+    "10.242.100.0/22",
+    "10.223.38.128/25",
+    "10.149.62.0/23",
+  ]
```
carlpartridge: Are these CIDR ranges ok to have publicly?

.github/workflows/ci-checks.yml
juliareynolds-nava: aarch64 is the right one.

terraform/shared-services/main.tf
juliareynolds-nava: I will delete that.

terraform/test/local.tf
juliareynolds-nava: I'll ask Brian.

terraform/dev/variables.tf
```diff
@@ -0,0 +1,54 @@
...
+  description = "KMS Key ID for the ACO Creds KMS Key"
+}
+
+variable "eni_ips" {
+  type        = list(string)
+  description = "Private IP used to create the ENI"
+
+  #FIXME
+  default = [
+    "10.234.241.202",
```
carlpartridge: Where is this used?  Do we need to put this in a param as well?

terraform/dev/variables.tf
```diff
@@ -0,0 +1,54 @@
...
+variable "ssl_policy" {
+  type        = string
+  description = "The AWS predefined SSL policy for the ALB"
+  default     = "ELBSecurityPolicy-TLS13-1-2-Res-2021-06"
+}
+
+variable "kms_key_id" {
+  #FIXME
+  type        = string
+  default     = "1fca833c-6d08-44d2-8088-9be2c37155f7"
```
carlpartridge: Is this ok to be public?

juliareynolds-nava: Good catch.  I can't find that key in non-prod but it should be referenced via the alias.  I think this is leftover from ec2.  Removing it and I'll run smoke tests in dev to make sure we don't need it.

carlpartridge: It looks like it relates to the ACO Creds lambda (which drops a creds file into an s3 bucket).  Its possible that we just havent needed to run this lambda yet in non-prod so the resources havent been created/tested?

carlpartridge:
It looks like you removed the kms key related to aco creds lambda as well as an IP address related to ENI?  I would prefer those were handled first in bcda-ops where we can prove that things dont break.

PR #1384, opened 2026-05-07
BCDA-10030: tofu install issue

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10030

## 🛠 Changes

Explicitly install tofu

## ℹ️ Context

The `setup-tenv` action in our workflow sets `TENV_AUTO_INSTALL=true`. This should automatically download and install OpenTofu the first time a `tofu` command (like `tofu init`) is called, rather than relying on a separate install step.

However, tenv's auto-installation might silently time out or fail if the runner doesn't stick around to cache the result. 
If the idea is to guarantee tofu is always installed, we might as well do an explicit install rather than a tenv flag.

Example of failure: https://github.com/CMSgov/bcda-app/actions/runs/25008204748


.github/workflows/deploy-all.yml
```diff
@@ -177,6 +177,9 @@ jobs:
         uses: cmsgov/cdap/actions/setup-sops@f4c14d47cc20e7f6de9112d7155af1213c9bca5a
       - name: Install YQ
         uses: cmsgov/cdap/actions/setup-yq@0b65099ef9232cfa2dfd4eea89cf8b9fa6052013
+      - name: Install OpenTofu for SOPS
+        working-directory: ops/services/10-config
+        run: tenv tofu install
```
carlpartridge, line 182: How does this handle where tofu is already installed?  I believe this has been an issue in the past.

laurenkrugen-navapbc: we could turn it into a check instead?

bhagatparwinder: `tenv tofu install` is fully idempotent. It is completely safe to run even if OpenTofu is already installed. If you have an example in the past where it failed when we did an explicit install, I would be interested to see it. 

`tenv` used to face this issue during initial install when multiple install attempts happened at the same time (only for fresh tenv initialization), and their team fixed that with a lock file: https://github.com/tofuutils/tenv/issues/116

bhagatparwinder: If the exact version is already installed, `tenv` simply prints that it's already installed and immediately exits with a success code (0) without re-downloading anything or throwing an error

image

carlpartridge (approved):

juliareynolds-nava (approved):

PR #1387, opened 2026-05-14
BCDA-9961: fix id logging in for Splunk

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9961

## 🛠 Changes

- Add `GetCMSIDByClientID` interface and implementation 
- Replace `GetACOByClientID` call with `GetCMSIDByClientID`
- Add repository method and associated unit tests
- Only return `AuthData` required

## ℹ️ Context

We were incorrectly fetching the CMS ID from the acos table when we should have been targeting the systems table. Addressing the 

## 🧪 Validation

- Unit tests
- Manually executed query from query builder to ensure we are getting the correct ID


bcda/models/repositorytest.go
laurenkrugen-navapbc, line 1: can rename this file to `repository_test.go`?

bcda/models/postgres/repository_test.go
laurenkrugen-navapbc, line 1: should we also have an integration test for this in addition to the mocks? just to verify the repository func `GetCMSIDByClientID` behavior?

bcda/auth/ssas.go
```diff
@@ -266,15 +266,12 @@ func checkTokenExpiration(bytes []byte) error {
...
 		return AuthData{}, err
 	}
 	return AuthData{
-		ACOID:       aco.UUID.String(),
-		CMSID:       *aco.CMSID,
-		ClientID:    clientID,
-		SystemID:    aco.SystemID,
-		Blacklisted: aco.Denylisted(),
+		CMSID:    cmsID,
+		ClientID: clientID,
```
carlpartridge, line 275: I think AuthData is used in a bunch of places?  Makes me a bit nervous to change its fields, although Id imagine we would see some unit failures?

bcda/models/repositorytest.go
bhagatparwinder: We can delete this file altogether. Changing to `_test` will have a Go implication that this file won't be available outside the test binary for that specific package.

It is currently used in `worker_prepare_test.go` and `ssas_test.go`. The structs being tested here both natively accept `repository`.

bcda/auth/ssas.go
bhagatparwinder: This method `GetAuthData` is totally isolated and used only for logging. The rest of the app uses a different method and obtains an AuthData object from `GetAuthDataFromCtx(r)`. This is the reason you don't see a bunch of test failures. Returning a full object here is unnecessary when all we log is the CMS ID.

bcda/models/postgres/repository_test.go
bhagatparwinder: I had been thinking of this. Adding one.

Edit: Or maybe not 😭 We are working on `systems` and `groups` table. This will spin out into a multi step change (create schemas for test, add seed data, integration test etc)

carlpartridge (approved):

laurenkrugen-navapbc (approved):

PR #1394, opened 2026-05-27
BCDA-10119: Remove new relic

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10119

## 🛠 Changes

Remove New Relic and related metrics, monitors, etc.

## ℹ️ Context

Part of moving away from New Relic.

## 🧪 Validation

Local linting and testing.


carlpartridge:
Snyk is having an issue with a dep of a dep (golang.com/x/image) which doesnt seem relevant to our functionality.

juliareynolds-nava (dismissed):

laurenkrugen-navapbc:
Should we deploy your branch to dev to see what (if any) new relic alarms go off? this looks good to me, but can't hurt to test before merging.

carlpartridge:
> Should we deploy your branch to dev to see what (if any) new relic alarms go off? this looks good to me, but can't hurt to test before merging.

New Relic is such a pain to navigate and every env is so differently configured that I would rather just handle it as we deploy to each env.  I also dont actually expect any issues but this is a good call out.  I have the upcoming deploy ticket and will make that more explicit in the ticket as well.

bcda/cclf/utils/cclfUtils.go
```diff
@@ -26,6 +27,7 @@ import (
 // ImportCCLFPackage will copy the appropriate synthetic CCLF files, rename them,
 // begin the import of those files and delete them from the place they were copied to after successful import.
 func ImportCCLFPackage(db *sql.DB, pgxPool *pgxv5Pool.Pool, acoSize, environment string, fileType models.CCLFFileType) (err error) {
+	ctx := context.Background()
```
michaeljvaldes, line 30: From what I can tell, most of the use of ctx here was relevant to our new relic metrics, and it's also passed to our aws client functions. Do you think it would be possible/better to not pass this context down through all of these functions in the handlers, processors, etc., or do we still need that?

.tool-versions
```diff
@@ -1 +1 @@
-golang 1.26.0
+golang 1.26.2
```
michaeljvaldes, line 1: Nice catch :)

bcda/cclf/utils/cclfUtils.go
carlpartridge: It looks like this cclfUtils package is only called from a CLI action that I dont think Ive ever heard about or used before.  I can create a ticket to further investigate and remove this and related code.

Generally speaking ctx, especially when it comes to lambda and http requests, is pretty important to pass around as it contains timeout/cancellation/lifecycle information.

michaeljvaldes (approved):

laurenkrugen-navapbc (approved):

bhagatparwinder (approved):

PR #1397, opened 2026-06-02
BCDA-10031: import and update lambda tf

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10031

## 🛠 Changes

for five lambdas:
- admin-aco-deny
- admin-create-aco
- admin-create-aco-creds
- admin-create-group
- api-waf-sync

The following changes are being made by this PR:
- update the lambda arch/runtime to al2023/arm64
- remove superfluous bucket params in sops that should just be managed in tf
- update lambda module ref to the latest version
- general updates (e.g. get env from workspace.env instead of a var, as in the 10-config service)
- include a `tf apply` step in each lambda's deployment workflow

## ℹ️ Context

These changes support a general update to five lambdas. The update to the lambda module supports recent datadog additions and other features, and there is a general cleanup of superfluous resources and readability of each lambda service. 
Each lambda's code deployment workflows now also includes a `tf apply` of its tf service to support automatic updates and a well-defined dependency of the code on its supporting architecture. The waf-sync lambda code lives in bcda-ssas-app, so that repo will have similar changes.

**Things that are _not_ covered in this PR:**
- function module liveness check. The newest version of the function module includes a liveness check, which is intended to ensure that the function is configured properly at `tf apply` and must be implemented in the function code itself. I chose to skip this for the time being since implementing it in our lambdas would take a medium amount of refactoring and we're already covering the "test at deploy time" via our workflows pretty comprehensively
- lambdas outside of the five outlined above. There are other lambdas that should be updated, but I ran into some hiccups in applying their tf at all and I expect that there will be a good amount of work there.

## 🧪 Validation

The deployment and integration test workflows for each lambda are passing on this PR.


ops/services/30-admin-aco-deny/main.tf
```diff
@@ -59,3 +67,8 @@ resource "aws_security_group_rule" "function_access" {
   security_group_id        = data.aws_security_group.db.id
   source_security_group_id = module.admin_aco_deny_function.security_group_id
 }
```
mianava, line 66: This might now play better as a aws_security_group_ingress_rule ; recommend putting that in a dedicated "security_group_rules.tf"

michaeljvaldes: Fixed! Thanks for pointing this out -- we have a lot of things like this that should be updated, and direction here is very appreciated.
I had to do something funky with `count` and `depends_on` to ensure that the deletion of the old rule would happen before the creation of the new one to prevent a conflict.

.github/workflows/admin-aco-deny-deploy.yml
```diff
@@ -21,8 +21,43 @@ on:
...
+    runs-on: codebuild-bcda-app-${{ contains(fromJSON('["prod", "sandbox"]'), inputs.deploy_env) && 'prod' || 'non-prod'}}-${{github.run_id}}-${{github.run_attempt}}
+    steps:
+      - uses: aws-actions/configure-aws-credentials@ec61189d14ec14c8efccab744f656cffd0e33f37 # v6.1.0
+        with:
+          aws-region: ${{ vars.AWS_REGION }}
+          role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/delegatedadmin/developer/${{ vars.AWS_ROLE_TO_ASSUME }}
+      - name: Checkout BCDA
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.3.0
+        with:
+          repository: CMSgov/bcda-app
```
carlpartridge: Nit: I dont think we need this with:, repository:, should default to bcda-app.

.github/workflows/admin-aco-deny-deploy.yml
```diff
@@ -21,8 +21,43 @@ on:
...
+        run: |
+          TF_WORKSPACE=default tofu init -var parent_env=${{ env.RELEASE_ENV }} -reconfigure && tofu workspace select -var parent_env=${{ env.RELEASE_ENV }} -or-create ${{ env.RELEASE_ENV }}
+          tofu plan \
+            -out 'bcda-release-lambda.tfplan'
+      - name: OpenTofu Apply
+        working-directory: ops/services/30-admin-aco-deny
+        run: |
+          tofu apply bcda-release-lambda.tfplan
+
+  deploy_function:
```
carlpartridge, line 56: Is there a big advantage to separating this to another job?  Seems like we need to redo a couple of steps (checkout bcda, assume role, get params).  Maybe still preferable as they are mostly different things?

ops/services/30-admin-create-aco-creds/main.tf
```diff
@@ -1,19 +1,17 @@
...
+  app               = "bcda"
+  env               = terraform.workspace
   service           = "admin-create-aco-creds"
-  full_name         = "${var.app}-${var.env}-${local.service}"
-  db_sg_name        = "bcda-${var.env}-db"
+  full_name         = "${local.app}-${local.env}-${local.service}"
+  db_sg_name        = "bcda-${local.env}-db"
   memory_size       = 256
-  creds_bucket_name = "bcda-${var.env}-aco-creds-*"
+  creds_bucket_name = "bcda-${local.env}-aco-creds-*"
```
carlpartridge, line 8: Whats the `-*` at the end for?  Oh... it was previously like that... 😅

.github/workflows/admin-aco-deny-deploy.yml
michaeljvaldes: Fixed!

.github/workflows/admin-aco-deny-deploy.yml
michaeljvaldes: Yeah, I thought the separation was worth it since the overlap is relatively small, we don't run these workflows all the time so efficiency isn't critical, and I wanted some separation between the separate work. I could go either way though, if y'all think I should change it.

ops/services/30-admin-create-aco-creds/main.tf
michaeljvaldes: This is a reference to the bucket that contains the actual generated aco creds, which has a tofu-generated name that is timestamped. The asterisk covers that bucket name suffix.

bhagatparwinder (approved):

.github/workflows/admin-aco-deny-deploy.yml
carlpartridge: Was mostly just curious if there was a specific reason I might have missed, no need to change!

carlpartridge (approved):

PR #1399, opened 2026-06-04
BCDA-10120: Add datadog APM monitoring

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10120

## 🛠 Changes

Add Datadog orchestrion which adds compile time automated APM.

## ℹ️ Context

Part of moving monitoring to datadog.

## 🧪 Validation

Local linting and testing, will deploy to dev from this branch, make sure everything builds/deploys, and also verify in datadog that metrics are happening.


.github/workflows/attribution-import-deploy.yml
```diff
@@ -40,14 +40,18 @@ jobs:
         working-directory: bcda
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.3.0
+        with:
+          ref: carl-10120-add-datadog
```
bhagatparwinder: Is this for testing?

.github/workflows/bene-prefs-import-deploy.yml
```diff
@@ -40,14 +42,18 @@ jobs:
         working-directory: bcda
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.3.0
+        with:
```
bhagatparwinder: Same here

bcda/bcdacli/cli.go
```diff
@@ -95,6 +97,16 @@ func setUpApp() *cli.App {
 			Action: func(c *cli.Context) error {
 				fmt.Fprintf(app.Writer, "%s\n", "Starting bcda...")
 
+				err := profiler.Start(
+					profiler.WithService("BCDA"),
+					profiler.WithEnv(os.Getenv("ENV")),
+					profiler.WithVersion(constants.Version),
+				)
+				if err != nil {
+					log.API.Fatal(err)
```
bhagatparwinder: Am I understanding this correctly? If the DataDog agent is missing, will this completely prevent the API from starting up? Is this a hard requirement for the application?

carlpartridge: Thats a good point.  Its not so much that its missing but that it errors out, but the point stands that we probably just want to make a log of it and not crash the whole service.  Specifically this should affect local dev so I can test out to see what happens there.

Update: It doesnt crash out in local dev, even though I dont have various env vars set, and its failing to reach out to its expectant url.

juliareynolds-nava (dismissed):

.github/workflows/ci-checks.yml
```diff
@@ -71,7 +71,7 @@ jobs:
           bash ops/secrets --decrypt
           mv -fv shared_files/encrypted/* shared_files/decrypted/
       - name: Build the stack
-        run: make docker-bootstrap
+        run: make docker-bootstrap-no-bake
```
carlpartridge, line 74: This runs docker bootstrap without COMPOSE BAKE which runs builds in parallel and saves a bunch of time.  Unfortunately had to remove as it was consistently breaking on maxing out memory.

.github/workflows/bene-prefs-import-deploy.yml
```diff
@@ -6,6 +6,8 @@ on:
       - main
     paths:
       - optout/**
+      - bcda/lambda/bene-prefs/**
```
bhagatparwinder: We've set `bene-prefs` as the path, but I still see the folder as `optout`. I'm a little confused.

bcda/lambda/optout/main.go
```diff
@@ -37,6 +40,16 @@ func optOutImportHandler(ctx context.Context, sqsEvent events.SQSEvent) (string,
 	appName := conf.GetEnv("APP_NAME")
 	logger := configureLogger(env, appName)
 
+	err := profiler.Start(
```
bhagatparwinder, line 35: How often do we run this bene-prefs lambda? Do we want to start/stop every time we get an invocation?

Also, is there enough time for the profiler to get the data and upload it by the time the lambda completes?

Edit: Looks like DD Go Profiler does it every 60 seconds: https://github.com/DataDog/go-profiler-notes/blob/main/datadog.md

Makefile
```diff
@@ -134,6 +134,16 @@ docker-build:
 	COMPOSE_BAKE=true docker compose build --force-rm
 	COMPOSE_BAKE=true docker compose -f compose.test.yml build --force-rm
 
+docker-build-no-bake:
+	docker compose build --force-rm
+	docker compose -f compose.test.yml build --force-rm
+
+docker-build:
```
bhagatparwinder: Is this duplicating the `docker-build` target from above?

.github/workflows/ci-checks.yml
laurenkrugen-navapbc: any option to ask if we can up the memory?

bcda/lambda/optout/main.go
laurenkrugen-navapbc: is it necessary to have dd running in our lambdas? I am obviously all for visibility, but we have lots of baked in monitoring accessible already.

bcda/bcdacli/cli.go
```diff
@@ -95,6 +97,16 @@ func setUpApp() *cli.App {
 			Action: func(c *cli.Context) error {
 				fmt.Fprintf(app.Writer, "%s\n", "Starting bcda...")
 
+				err := profiler.Start(
```
laurenkrugen-navapbc, line 100: any objection to moving this under main? because we only use this cli app for startup but the cli commands aren't in use except for bootstrapping test envs, maybe we start moving away from making more updates to this package so we can eventually deprecate it?

bcda/lambda/optout/main.go
carlpartridge: Looks like it should work?  https://docs.datadoghq.com/serverless/aws_lambda/profiling/

I added it to our core lambdas as I figured it would be nice to be able to see what was going on in them.  Bene-prefs is quite light weight, and so is waf-sync, so maybe not so useful?  But for attr-import Im very curious as there are significant memory issues and Im hoping this can help us refactor that.

juliareynolds-nava (dismissed):

.github/workflows/ci-checks.yml
carlpartridge: I have a faint recollection that no we cant increase them (or maybe just a preference that we dont), but worth a shot, but outside the scope of this imo.

Update: Going to ask CDAP.

Second update: Increased the cpu and memory of the instance cutting certain times in almost half.  With no-bake still active we are still maxing out on CPU for a decent portion but memory is only getting a bit over 50%.

Makefile
carlpartridge: See comment above about needing to remove COMPOSE_BAKE due to codebuilder memory constraints.

.github/workflows/bene-prefs-import-deploy.yml
carlpartridge: Hmmm, good call!

bcda/bcdacli/cli.go
carlpartridge: Id like to get this in at the soonest possible so that we are profiling as much as possible.  I think if we can move away from cli in the future we can move this with that.  But I think for the foreseeable future we are locked into at least the app starting from this cli package.  I think everything else in the cli package can probably go though...

Makefile
bhagatparwinder: I meant this file, [line 133](https://github.com/CMSgov/bcda-app/pull/1399/changes#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L133), I see the same `docker-build`.

bcda/lambda/optout/main.go
bhagatparwinder: How about we move this to the `main()` entrypoint function above? This way, it will only get initialized once during a lambda cold start.

Edit: And same for the attribution-import lambda above.

bcda/bcdacli/cli.go
bhagatparwinder: I don't know how exactly Profiler knows what to collect, so I'm going to ask if we need to add this to the bcdaworker cli file too?

docker/Dockerfile.bcdaworker
```diff
@@ -17,12 +17,14 @@ RUN --mount=type=cache,target=/go/pkg/mod \
 RUN --mount=type=cache,target=/go/pkg/mod \
     cat tools.go | awk '/[\/\/] bcda/{flag=1; next} /[\/\/] end/{flag=0} flag' \
     | awk -F '"' '{print $2}' | xargs -tI % go get %
+RUN go install github.com/DataDog/orchestrion@49ad05fbcde841361e6bcc953f78bb7e72b1ca69
```
bhagatparwinder: Do we want to add caching here?

```
--mount=type=cache,target=/go/pkg/mod
```

docker/Dockerfile.bcda
```diff
@@ -17,12 +17,14 @@ RUN --mount=type=cache,target=/go/pkg/mod \
 RUN --mount=type=cache,target=/go/pkg/mod \ 
     cat tools.go | awk '/[\/\/] bcda/{flag=1; next} /[\/\/] end/{flag=0} flag' | \
     awk -F '"' '{print $2}' | xargs -tI % go get % 
+RUN go install github.com/DataDog/orchestrion@49ad05fbcde841361e6bcc953f78bb7e72b1ca69
```
bhagatparwinder: Maybe caching here too

```
--mount=type=cache,target=/go/pkg/mod
```

compose.yml
```diff
@@ -35,6 +35,10 @@ services:
       FHIR_ARCHIVE_DIR: /var/efs/archive
       FHIR_PAYLOAD_DIR: /var/efs/data
       FHIR_STAGING_DIR: /var/efs/tmpdata
+      # skip datadog setup and tracing in local dev/CI/CD envs
```
bhagatparwinder, line 38: We are removing the DD setup in local/CI, but the profiler starts in our CLI independently/unconditionally. Won't this cause issues, or do we need a check for `profiler.Start` to see if profiling is enabled in the environment it is running?

carlpartridge: This didnt end up affecting anything anyways, but my guess is that how it works is that the profiler will start anyways, but just not actually send anything if the flags are disabled.  I wouldnt think the setup is any major efficiency loss on startup.  I can remove it but feels slightly better to keep it in just to make sure/make it explicit.

bcda/bcdacli/cli.go
carlpartridge: Good catch, thought I had added that...

compose.yml
carlpartridge: The profiler seems good at not breaking.  If various things are not present it just wont run, but also will not error out/cause issues.  At least in my limited experience.

laurenkrugen-navapbc (approved):

juliareynolds-nava (approved):

michaeljvaldes (approved):

PR #1400, opened 2026-06-04
BCDA-10107 Datadog Dashboard

juliareynolds-nava:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10107

## 🛠 Changes

Adds:
- Datadog dashboard module

## ℹ️ Context

We are moving to Datadog, meaning we need to configure dashboards and monitors. This PR aims to set up the initial dashboards along with the automation for deploying and maintaining them.

## 🧪 Validation

Dashboard and monitors are up for the dev and test environments.
https://app.ddog-gov.com/dashboard/stb-a6w-w8x/bcda-metrics-dashboard?fromUser=false&refresh_mode=sliding&from_ts=1780663286076&to_ts=1780666886076&live=true


ops/services/40-datadog-dashboards/main.tf
carlpartridge, line 1: Do we want the datadog dashboard to be in a different directory than the monitors?

ops/services/40-datadog-dashboards/main.tf
```diff
@@ -0,0 +1,75 @@
...
+locals {
+  default_tags = module.platform.default_tags
+  env          = terraform.workspace
+  service      = "dashboard"
+
+  ssm_root_map = {
+    common   = "/bcda/${local.env}/common"
+    core     = "/bcda/${local.env}/core"
+    accounts = "/bcda/mgmt/aws-account-numbers"
+    splunk   = "/bcda/mgmt/splunk"
```
carlpartridge, line 41: Should we use local.app for these "bcda" refs?

juliareynolds-nava: I prefer this for readability.

ops/services/40-datadog-dashboards/main.tf
juliareynolds-nava: This is the way other teams have structured the IaC so it's standard.

ops/services/40-datadog-dashboards/tofu.tf
```diff
@@ -0,0 +1,69 @@
+# This root tofu.tf is symlink'd to by all per-env Terraservices. Changes to this tofu.tf apply to
+# _all_ Terraservices, so be careful!
+
+locals {
+  app              = "bcda"
+  established_envs = ["prod"]
```
laurenkrugen-navapbc, line 6: should we also have the other envs here, or do we not have the infra configured for that yet?

ops/services/40-datadog-dashboards/main.tf
michaeljvaldes: We have a lot of places in our bcda-ops tf that just use the "bcda" string instead of referencing the app generically. I also think it's more readable, especially since we won't ever need many of these modules to be generic.

carlpartridge (approved):

ops/services/40-datadog-dashboards/tofu.tf
juliareynolds-nava: This is per CDAP.  I'm not sure exactly why but there is only one datadog environment so this is more of a label.

michaeljvaldes (approved):

PR #1402, opened 2026-06-09
BCDA-10060 cross account grant

juliareynolds-nava:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10060

## 🛠 Changes

Updated the function module to allow injection of the eft role for cross account access.

## ℹ️ Context

These changes are to map to updated function module changes to enable cross account role assume grants.  We need this so eft can drop files in our s3 bucket.

## 🧪 Validation

See checks below.


ops/services/30-attribution-import/main.tf
```diff
@@ -72,6 +72,7 @@ data "aws_iam_policy_document" "attribution-import_bucket_rw" {
     effect = "Allow"
 
     actions = [
+      "s3:PutObject",
```
bhagatparwinder: Don't we do this on line 94?

ops/services/30-attribution-import/main.tf
bhagatparwinder: All actions are s3, I don't think we can/need `arn:aws:logs:*:*:*`

ops/services/30-attribution-import/main.tf
```diff
@@ -103,26 +104,47 @@ data "aws_iam_policy_document" "attribution-import_bucket_rw" {
...
+    env               = var.env
+    service           = local.service
+    kms_alias_primary = { target_key_arn = module.platform.kms_alias_primary.target_key_arn }
+    primary_region    = { name = module.platform.region_name }
+    account_id        = module.platform.account_id
+  }
+  liveness_check_enabled = false
+
+  additional_admin_role_arns = [module.platform.ssm.attribution-import.misp-eft-role_arn.value]
+  github_actions_repos       = ["bcda-app:*"]
```
bhagatparwinder, line 134: Should this be `["CMSgov/bcda-app"]`?

juliareynolds-nava: No, the latest version of the function module formats the github action repo with the org so you just need the repo name and :*

ops/services/30-attribution-import/main.tf
```diff
@@ -103,26 +104,47 @@ data "aws_iam_policy_document" "attribution-import_bucket_rw" {
...
+
   environment_variables = {
     ENV      = var.env
-    APP_NAME = "${local.app}-${var.env}-attribution-import"
+    APP_NAME = "${local.app}-${var.env}-${local.service}"
+    DB_HOST  = "postgres://${data.aws_rds_cluster.this.endpoint}:${data.aws_rds_cluster.this.port}/bcda"
```
bhagatparwinder: Are we using this anywhere?

ops/services/30-attribution-import/main.tf
juliareynolds-nava: Removed.

ops/services/30-attribution-import/main.tf
juliareynolds-nava: removed

ops/services/30-attribution-import/main.tf
```diff
@@ -98,31 +98,51 @@ data "aws_iam_policy_document" "attribution-import_bucket_rw" {
...
-  description = "Ingests the most recent attribution from BFD"
+  name        = local.service
+  description = "Ingests the most recent attribution from EFT"
 
   handler = "bootstrap"
   runtime = "provided.al2023"
 
   memory_size = 2048
 
+  platform = {
```
carlpartridge, line 123: Cant we just link to `module.plaform`?

juliareynolds-nava: I wish.  The function module now takes the platform settings via this object syntax.

carlpartridge: 😱 Nit: Would it make sense to reference the module vars in the below eg `app = module.platform.app`?  Just thinking it would make it one less place to need to change things in the future.

ops/services/30-attribution-import/security_group_rules.tf
```diff
@@ -0,0 +1,24 @@
...
+resource "aws_vpc_security_group_egress_rule" "default" {
+  security_group_id = module.attribution_import_function.security_group_id
+  cidr_ipv4         = "0.0.0.0/0"
+  ip_protocol       = "tcp"
+  from_port         = 5432
+  to_port           = 5432
+  description       = "Allow DB access"
+}
+
+resource "aws_security_group_rule" "function_access" {
```
carlpartridge, line 14: Looks like this should be `aws_vpc_security_group_egress_rule`.

Also nit: The rule on line 14 is related to the data resource set on line 1, so if we could move the rule on line 5 to below this.  Just try to keep like things a bit more together.

ops/services/30-attribution-import/security_group_rules.tf
```diff
@@ -0,0 +1,24 @@
+data "aws_security_group" "db" {
+  name = local.db_sg_name
+}
+
+resource "aws_vpc_security_group_egress_rule" "default" {
```
carlpartridge: Nit: Instead of "default" can we call this "db_egress" or something?

bhagatparwinder (dismissed):

ops/services/30-attribution-import/security_group_rules.tf
juliareynolds-nava: This is the rule for the db to get our traffic, so needs to be ingress.

carlpartridge: Oh good point, but looks like theres `aws_vpc_security_group_ingress_rule`.

carlpartridge (approved):

bhagatparwinder (approved):

PR #1403, opened 2026-06-09
BCDA-10135: add access model attribution

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10135

## 🛠 Changes

- added ACCESS model csv attribution ingestion (entity id placeholder present; still awaiting)
- updated api.yml with ACCESS model
- minor comment cleanup
- moved IsSupportedACO() from service to config; config package had other `IsXYXACO` methods and this had hardcoded regex expressions that had drift from our aco configuration files. 
- removed some unused methods to convert struct to string and some used methods for the aforementioned
- add test attribution file for ACCESS model.

## ℹ️ Context

*Holding off on merging this PR until we get the entity ID naming convention for ACCESS.*

PR is part of effort to add ACCESS model attribution. Did some minor cleanup for verifying if an ACO ID is valid; we have regex in our config files and we had a function with hardcoded values; there was drift between the two, so the hardcoded func was updated to instead check against the aco configs, since that is our source of truth.

## 🧪 Validation

Unit tests passing. Integration test TBD.


bcda/service/service.go
```diff
@@ -643,25 +643,25 @@ func getMaxBeneCount(requestType string) (int, error) {
...
-		guide   = `^GUIDE-\d{5}$`
-		test    = `^TEST\d{3}$`
-		sandbox = `^SBX[A-Z]{2}\d{3}$`
-		iot     = `^IOTA\d{3}$`
-		pattern = `(` + ssp + `)|(` + ngaco + `)|(` + cec + `)|(` + ckcc + `)|(` + kcf + `)|(` + dc + `)|(` + mdtcoc + `)|(` + cdac + `)|(` + guide + `)|(` + test + `)|(` + sandbox + `)|(` + iot + `)`
-	)
-
-	return regexp.MustCompile(pattern).MatchString(cmsID)
-}
+// func IsSupportedACO(cmsID string) bool {
```
bhagatparwinder: Do we want to delete this?

bcda/bcdacli/cli_test.go
```diff
@@ -291,18 +291,18 @@ func (s *CLITestSuite) TestCreateGroup_InvalidACOID() {
...
 	assert.Empty(s.T(), buf.String())
 	buf.Reset()
 
-	// Valid format, but no matching ACO
-	aUUID := "4e5519cb-428d-4934-a3f8-6d3efb1277b7"
-	args = []string{"bcda", constants.CreateGroupArg, "--id", "invalid-aco-id-group", constants.NameArg, "Invalid ACO ID Group", constants.ACOIDArg, aUUID}
-	err = s.testApp.Run(args)
-	assert.EqualError(s.T(), err, "no ACO record found for "+aUUID)
-	assert.Empty(s.T(), buf.String())
+	// // Valid format, but no matching ACO
```
bhagatparwinder: Is this commented out temporarily till we get the entity naming convention?

.github/workflows/attribution-import-integration-test.yml
```diff
@@ -69,10 +69,15 @@ jobs:
           guidecsvname=T.GUIDE.GUIDE-00001.Y${year}.D${date}.T${time}1
           echo "GUIDEFILENAME=$guidecsvname" >> "$GITHUB_ENV"
 
+          accesscsvname=T.ACCESS.0001.Y${year}.D${date}.T${time}1
```
carlpartridge: Maybe our test ID should be 9999?  I guess it doesnt matter much since this only runs in our test env?

bcda/service/config.go
```diff
@@ -143,19 +127,30 @@ func (config *Config) IsACODisabled(CMSID string) bool {
...
-	for _, aco := range config.V3EnabledACOs {
+	for _, aco := range cfg.V3EnabledACOs {
 		if aco == ACOID {
 			return true
 		}
 	}
 	return false
 }
 
+// IsSupportedACO determines if the particular ACO is supported by checking
```
bhagatparwinder, line 143: Thank you for this! I disliked the hardcoded string regex!

bcda/bcdacli/cli.go
```diff
@@ -419,7 +419,12 @@ func createGroup(r models.Repository, id, name, acoID string) (string, error) {
 		err error
 	)
 
-	if match := service.IsSupportedACO(acoID); match {
+	cfg, err := service.LoadConfig()
```
bhagatparwinder, line 434: I'm ok for now if we leave this as is (thinking we want to retire this cli.go sometime in the future), but this will make unit testing harder. Ideally, we want this as a dep injection.

.github/workflows/attribution-import-integration-test.yml
```diff
@@ -91,6 +96,9 @@ jobs:
 
           aws s3 cp --no-progress ${guidecsvname} \
             s3://$BUCKET/bfdeft01/bcda/in/test/${guidecsvname}
+
+          aws s3 cp --no-progress ${accesscsvname} \
+            s3://bfd-test-eft/bfdeft01/bcda/in/test/${accesscsvname}
```
carlpartridge: I think we need $BUCKET at the start of this.  Its a bit confusing but I believe we kept the path of the file inside the bucket the same to make it easier for clients but means that we are still referring to our path using old names (bfd-test-eft/bfdeft01/...).

.github/workflows/attribution-import-integration-test.yml
```diff
@@ -178,3 +186,24 @@ jobs:
               exit 1
             fi
           fi
+
+           # Verify that we have a record of the ACCESS CSV file in the database
+          CSV_FILE=`psql -t "$CONNECTION_URL" -c "SELECT id FROM cclf_files WHERE name = '$ACCESSFILENAME' LIMIT 1" 2>&1`
```
carlpartridge: Super nit: We are re-using CSV_FILE var name (and CSV_BENES as well) here which I think is fine since we should be done with the previous csv file entry and checks, but might be nice to make it a bit more explicit.  Seeing we do the same with the GUIDE check so marking this as a super nit.

bcda/cclf/parser.go
```diff
@@ -31,10 +31,12 @@ func CheckIfAttributionCSVFile(filePath string) bool {
 	MDTCOCPattern := `(P|T)\.(PCPB)\.(M)([0-9][0-9])(\d{2})\.(D\d{6}\.T\d{6})\d`
 	CDACPattern := `(P|T)\.(BCD)\.(DA)(\d{4})\.(MBIY)(\d{2})\.(D\d{6}\.T\d{6})\d`
 	GUIDEPattern := `(P|T)\.(GUIDE)\.(GUIDE-)(\d{5})\.(Y)(\d{2})\.(D\d{6}\.T\d{6})\d`
+	ACCESSPattern := `(P|T)\.(ACCESS)\.(\d{4})\.(Y)(\d{2})\.(D\d{6}\.T\d{6})\d`
 	MDTCOCRegexp := regexp.MustCompile(MDTCOCPattern)
 	CDACRegexp := regexp.MustCompile(CDACPattern)
 	GUIDERegexp := regexp.MustCompile(GUIDEPattern)
-	return (MDTCOCRegexp.MatchString(filePath) || CDACRegexp.MatchString(filePath) || GUIDERegexp.MatchString(filePath))
+	ACCESSRegexp := regexp.MustCompile(ACCESSPattern)
```
bhagatparwinder: I think I mentioned this in a PR from Carl a while back, but this could definitely use a teensy performance improvement if we make this package-level private variable (pre-compile vs re-compile)

bcda/bcdacli/cli_test.go
```diff
@@ -291,18 +291,18 @@ func (s *CLITestSuite) TestCreateGroup_InvalidACOID() {
...
+	// aUUID := "4e5519cb-428d-4934-a3f8-6d3efb1277b7"
+	// args = []string{"bcda", constants.CreateGroupArg, "--id", "invalid-aco-id-group", constants.NameArg, "Invalid ACO ID Group", constants.ACOIDArg, aUUID}
+	// err = s.testApp.Run(args)
+	// assert.EqualError(s.T(), err, "no ACO record found for "+aUUID)
+	// assert.Empty(s.T(), buf.String())
```
carlpartridge: Why is this failing?

bcda/cclf/parser.go
```diff
@@ -144,6 +145,8 @@ func validateCSVMetadata(attributionFile service.AttributionFile, subMatches []s
 		metadata.acoID = subMatches[3]
 	case "GUIDE":
 		metadata.acoID = fmt.Sprintf("%s%s", subMatches[3], subMatches[4])
+	case "ACCESS":
```
carlpartridge: I think you could make this an OR on line 144 since the what to do is the same for each (ie `case "BCD", "ACCESS":` ...).

.github/workflows/attribution-import-integration-test.yml
laurenkrugen-navapbc: the test is of 0001 is just a placeholder for now! we don't yet have the naming convention for the part of `ACCESS.<entity-id>.Y26`

.github/workflows/attribution-import-integration-test.yml
laurenkrugen-navapbc: good catch, will update this!

.github/workflows/attribution-import-integration-test.yml
laurenkrugen-navapbc: we are using that convention for the guide and non-guide csv file above; do you want me to update all three to use different names?

bcda/bcdacli/cli.go
laurenkrugen-navapbc: I agree - I actually started removing these cli commands but unfortunately, they are used in `unit_test.sh`; I spent some time seeing if I could deprecate most of cli.go in this PR but it was too large to include.

.github/workflows/attribution-import-integration-test.yml
carlpartridge: I think it would be a bit easier to read, but up to you!  Not really a lot of juice from this squeeze.

bcda/bcdacli/cli_test.go
laurenkrugen-navapbc: oop, my mistake. I was trying to remove a bunch of stuff from cli.go (see comment above, too large for this ticket) and forgot to revert this.

bcda/bcdacli/cli_test.go
laurenkrugen-navapbc: not failing! started removing related code in cli.go (see my comments above!) but discovered it was being used in our test setup in `unit_test.sh`, just forgot to revert this change back to how it was.

bcda/cclf/parser.go
laurenkrugen-navapbc: Let me see if I can squeeze that into this PR!

bcda/cclf/parser.go
laurenkrugen-navapbc: can do!

bcda/service/service.go
laurenkrugen-navapbc: good catch, removing!

bcda/cclf/parser.go
bhagatparwinder: Yeah, not necessary, but it will be nice!

bcda/service/service_test.go
```diff
@@ -40,6 +40,9 @@ var (
 )
 
 func TestSupportedACOs(t *testing.T) {
```
michaeljvaldes: Should we just move this test to config_test.go?

bcda/cclf/parser.go
```diff
@@ -14,6 +14,17 @@ import (
 	"github.com/pkg/errors"
 )
 
+var (
+	mdtcocPattern = `(P|T)\.(PCPB)\.(M)([0-9][0-9])(\d{2})\.(D\d{6}\.T\d{6})\d`
```
michaeljvaldes, line 18: I love that you made the config the source of truth for these patterns in `IsSupportedACO`
Does it make sense to do the same here?

bhagatparwinder (dismissed):

bcda/service/service_test.go
laurenkrugen-navapbc: yup, good call! I will update that with the latest changes.

bcda/cclf/parser.go
laurenkrugen-navapbc: Ooh, I missed this section, let me see if I can get that updated in this PR!

bhagatparwinder (dismissed):

bcda/cclf/parser.go
```diff
@@ -28,13 +39,30 @@ func getCMSID(name string) (string, error) {
...
+		}
+	}
+
+	return match
+
+	// mdtcocRegexp = regexp.MustCompile(cfg.ACOConfigs[0].Model)
+	// cdacRegexp = regexp.MustCompile(cdacPattern)
+	// guideRegexp = regexp.MustCompile(guidePattern)
+	// accessRegexp = regexp.MustCompile(accessPattern)
+	// return (mdtcocRegexp.MatchString(filePath) || cdacRegexp.MatchString(filePath) || guideRegexp.MatchString(filePath) || accessRegexp.MatchString(filePath))
```
carlpartridge: Nit: Cleanup.

bcda/service/config_test.go
```diff
@@ -267,6 +268,85 @@ func TestIsACOV3Enabled_EnvironmentBased(t *testing.T) {
...
+
+		{"CDAC too short", "DA999", false},
+		{"CDAC too long", "DA9999999", false},
+		{"CDAC invalid characters", "DA999V", false},
+		{"valid CDAC", "DA9999", true},
+
+		{"GUIDE too short", "GUIDE-999", false},
+		{"GUIDE too long", "GUIDE-9999999", false},
+		{"GUIDE invalid characters", "GUIDE99999", false},
+		{"valid GUIDE", "GUIDE-99999", true},
```
carlpartridge, line 323: Do we want ACCESS tests in here as well?

carlpartridge (approved):

bhagatparwinder (approved):

PR #1404, opened 2026-06-11
BCDA-10031: lambda cleanup

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10031
## 🛠 Changes

- fix permissions in lambda deployment workflows
- add explicit security group rules for egress to db
- remove deleted security group rules

## ℹ️ Context

The deployment workflows need adjusted permissions in order to deploy to environments outside of dev.
Each lambda has a default "allow all" egress security group rule that is not captured in the terraform state; this PR will add a narrower "egress to db" rule and the broader rules can be removed manually afterward.
Other security group rules that were previously deleted (count=0) are now being removed from the tofu.

## 🧪 Validation

Testing in dev with integration tests (including waf-sync, which lives in ssas-app).


carlpartridge (dismissed):
Looks good to me.  Is there anything we need to handle as part of deploy or is that all taken care of?

michaeljvaldes:
> Looks good to me. Is there anything we need to handle as part of deploy or is that all taken care of?

Nothing special, the deploy workflows will need to be run for each environment once more.

laurenkrugen-navapbc (dismissed):

carlpartridge:
> > Looks good to me. Is there anything we need to handle as part of deploy or is that all taken care of?
> 
> Nothing special, the deploy workflows will need to be run for each environment once more.

Preferably this would be done as part of this ticket as opposed to relying on the regular deployment process?

ops/services/30-admin-create-aco-creds/security_group_rules.tf
```diff
@@ -10,5 +10,24 @@ resource "aws_vpc_security_group_ingress_rule" "function_db_access" {
...
+  description = "egress to db"
+
+  security_group_id            = module.admin_create_aco_creds_function.security_group_id
+  referenced_security_group_id = data.aws_security_group.db.id
+}
+
+resource "aws_vpc_security_group_egress_rule" "ssas_admin" {
+  from_port   = 444
+  to_port     = 444
+  cidr_ipv4   = "0.0.0.0/0"
```
bhagatparwinder, line 28: Is it ok to have CIDR this wide open?

michaeljvaldes: That's a great question, and it's certainly something in this PR that can be improved in the future. Ideally we wouldn't define a CIDR range and would just associate this rule with a target security group, that of the SSAS ALB. Unfortunately, due to how our SSAS ALB security groups are configured, that is difficult. 
For now, I think it is okay to specify this generalized egress rule for port 444. For what it's worth, all of our lambdas have a similar rule for port 443 specified in terraform, and most of them have an even more generalized rule for all ports and ip protocols.
In the future, we should limit this to just the SSAS security group, specifically when we replace our ALBs with service connect.

carlpartridge (approved):

bhagatparwinder (approved):

PR #1405, opened 2026-06-11
BCDA-10109 turn off shadow mode

juliareynolds-nava:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10109

## 🛠 Changes

turned shadow mode off for monitors so they will send alerts

## ℹ️ Context

This change is per CDAP now that alerting is configured in datadog.

## 🧪 Validation

see checks below

### DIFF
--- a/ops/services/40-datadog-monitors/config/sandbox.yml
+++ b/ops/services/40-datadog-monitors/config/sandbox.yml
@@ -1,4 +1,4 @@
-shadow_mode: true
+shadow_mode: false
 
 lambda:
   error_rate_threshold: 1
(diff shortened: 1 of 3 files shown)


ops/services/40-datadog-monitors/config/sandbox.yml
```diff
@@ -1,4 +1,4 @@
-shadow_mode: true
+shadow_mode: false
```
carlpartridge, line 1: We could just remove these since the default is now false.  Wondering if it makes sense to have the default be true then explicitly call false in every env?

carlpartridge (approved):

bhagatparwinder (approved):

michaeljvaldes (approved):

PR #1412, opened 2026-06-17
BCDA-10203: utilize config files instead of hard coded values

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10203

## 🛠 Changes

- hardcoded values in `parser.go` removed
- renamed some script variables for clearer error messages
- move `loadBCDAParams` function before checking if file is a csv, since we need access to config files before that call is made

## ℹ️ Context

Changes were made so that config files act as single source of truth, which reduces configuration drifts, redundant regex, and lowers risk of bugs/issues.

## 🧪 Validation

tbd integration test


bcda/lambda/attribution-import/main.go
```diff
@@ -82,6 +82,11 @@ func attributionImportHandler(ctx context.Context, sqsEvent events.SQSEvent) (st
 	pool := database.ConnectPool()
 	defer pool.Close()
 
+	err = loadBCDAParams()
```
laurenkrugen-navapbc, line 85: moved this from the `handle<csv|cclf>import` functions; this needs to happen before checking the filetype or it fails.

bcda/lambda/attribution-import/main_test.go
```diff
@@ -43,14 +42,6 @@ func TestAttributionImportMainSuite(t *testing.T) {
 	suite.Run(t, new(AttributionImportMainSuite))
 }
 
-func TestHandleCSVImportNoACOConfig(t *testing.T) {
```
laurenkrugen-navapbc, line 46: the integration tests validate this now.

carlpartridge: Which integration tests are you referring to?  The one that runs in  our test env as a workflow?  If so I would prefer a test that can be run locally.
Should we add ACCESS to this unit test or is that excessive? https://github.com/CMSgov/bcda-app/blob/e9ab154b3b80cfd9b03c22cae66ef981028ac691/bcda/cclf/parser_test.go#L560

carlpartridge (commented):
Do we need to update any of our env.yaml/api.yml files or were those already taken care of?  Do we need to update our config.yml files in our s3 as part of our next deploy?

bcda/lambda/attribution-import/main_test.go
laurenkrugen-navapbc: I can add those tests in parser.go.  I would also prefer a test that we can run locally, but since that loading of params is required in the lambda call, testing the config loading is tricky without ministack or another alternative. I added some additional logging for when the config fails to load or the length of the aco config is zero.

carlpartridge: Ah yes, good point!

carlpartridge (dismissed):
Seeing some fmt lint issues but besides that looks good!

michaeljvaldes (dismissed):

carlpartridge (dismissed):

carlpartridge (dismissed):

michaeljvaldes (dismissed):

michaeljvaldes (approved):

carlpartridge (approved):

bhagatparwinder (approved):

PR #1415, opened 2026-06-23
BCDA-10050: adjust data metric setters

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10050

## 🛠 Changes

Fixed a few issues with how we are setting certain job related metrics.

## ℹ️ Context

We released these a few months ago to help with various job related data metrics collection and analysis.  Some of these were eventually seen as not working as intended.

## 🧪 Validation

Local linting and testing.

### DIFF
--- a/bcda/models/postgres/repository_test.go
+++ b/bcda/models/postgres/repository_test.go
@@ -844,7 +844,7 @@ func (r *RepositoryTestSuite) TestJobsMethods() {
 
 	failed := models.Job{ACOID: aco.UUID, RequestURL: reqURL, Status: models.JobStatusFailed, JobCount: 10}
 	pending := models.Job{ACOID: aco.UUID, RequestURL: reqURL, Status: models.JobStatusPending, JobCount: 30}
-	completed := models.Job{ACOID: aco.UUID, RequestURL: reqURL, Status: models.JobStatusCompleted, JobCount: 40}
+	completed := models.Job{ACOID: aco.UUID, RequestURL: reqURL, Status: models.JobStatusCompleted, JobCount: 40, BenesAttributedToACO: 20}
 
 	failed.ID, err = r.repository.CreateJob(ctx, failed)
 	assert.NoError(err)
@@ -931,10 +931,13 @@ func (r *RepositoryTestSuite) TestJobsMethods() {
 	assert.True(newFailed.UpdatedAt.After(newCompleted.UpdatedAt))
 
 	// Verify BenesAttributedToACO
+	updatedCompleted, err := r.repository.GetJobByID(ctx, completed.ID)
+	assert.NoError(err)
+	assert.Equal(20, updatedCompleted.BenesAttributedToACO)
 	completed.BenesAttributedToACO = 10
 	err = r.repository.UpdateJob(ctx, completed)
 	assert.Nil(err)
-	updatedCompleted, err := r.repository.GetJobByID(ctx, completed.ID)
+	updatedCompleted, err = r.repository.GetJobByID(ctx, completed.ID)
 	assert.Nil(err)
 	assert.Equal(10, updatedCompleted.BenesAttributedToACO)
 
--- a/bcdaworker/repository/postgres/repository.go
+++ b/bcdaworker/repository/postgres/repository.go
@@ -86,7 +86,7 @@ func (r *Repository) GetCCLFBeneficiaryByID(ctx context.Context, id uint) (*mode
 
 func (r *Repository) GetJobByID(ctx context.Context, jobID uint) (*models.Job, error) {
 	sb := sqlFlavor.NewSelectBuilder()
-	sb.Select("id", "aco_id", "request_url", "status", "transaction_time", "job_count", "created_at", "updated_at")
+	sb.Select("id", "aco_id", "request_url", "status", "transaction_time", "job_count", "created_at", "updated_at", "benes_attributed_to_aco")
 	sb.From("jobs").Where(sb.Equal("id", jobID))
 
 	query, args := sb.Build()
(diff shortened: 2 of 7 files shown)


bcda/models/postgres/repository_test.go
```diff
@@ -931,10 +931,13 @@ func (r *RepositoryTestSuite) TestJobsMethods() {
...
 
 	// Verify BenesAttributedToACO
+	updatedCompleted, err := r.repository.GetJobByID(ctx, completed.ID)
+	assert.NoError(err)
+	assert.Equal(20, updatedCompleted.BenesAttributedToACO)
 	completed.BenesAttributedToACO = 10
 	err = r.repository.UpdateJob(ctx, completed)
 	assert.Nil(err)
-	updatedCompleted, err := r.repository.GetJobByID(ctx, completed.ID)
+	updatedCompleted, err = r.repository.GetJobByID(ctx, completed.ID)
```
carlpartridge, line 940: Updated these tests and some tests in other seemingly unrelated areas of the code base to make sure that we werent losing the data on an update or subsequent call of some kind.

bcdaworker/repository/postgres/repository.go
```diff
@@ -86,7 +86,7 @@ func (r *Repository) GetCCLFBeneficiaryByID(ctx context.Context, id uint) (*mode
 
 func (r *Repository) GetJobByID(ctx context.Context, jobID uint) (*models.Job, error) {
 	sb := sqlFlavor.NewSelectBuilder()
-	sb.Select("id", "aco_id", "request_url", "status", "transaction_time", "job_count", "created_at", "updated_at")
+	sb.Select("id", "aco_id", "request_url", "status", "transaction_time", "job_count", "created_at", "updated_at", "benes_attributed_to_aco")
```
carlpartridge, line 89: This ended up being the cause, getting/updating job in the bcda worker ORM as opposed to the bcda ORM :(

bcdaworker/repository/postgres/repository_test.go
```diff
@@ -3,37 +3,64 @@ package postgres_test
...
 
 func TestRepositoryTestSuite(t *testing.T) {
 	suite.Run(t, new(RepositoryTestSuite))
 }
 
 func (r *RepositoryTestSuite) SetupSuite() {
-	r.db = database.Connect()
+	var err error
+	r.dbContainer, err = db.NewTestDatabaseContainer()
+	require.NoError(r.T(), err)
```
carlpartridge, line 39: Converted bcda worker ORM to use db containers.

bcdaworker/worker/worker.go
```diff
@@ -441,13 +441,13 @@ func appendErrorToFile(ctx context.Context, fileUUID string,
...
 	logger := log.GetCtxLogger(ctx)
+	hasAtLeastOneEntry := false
 
 	for _, entry := range b.Entries {
 		if entry["resource"] == nil {
 			continue
 		}
-
-		*hasEntriesCount++
+		hasAtLeastOneEntry = true
```
carlpartridge, line 454: This was the fix for another issue where we were tallying every benes total number of entries instead of just tallying whether or not the bene had any actual entries.

bcdaworker/worker/worker.go
```diff
@@ -465,6 +465,10 @@ func fhirBundleToResourceNDJSON(ctx context.Context, w *bufio.Writer, b *fhirmod
 				responseutils.InternalErr, fmt.Sprintf("Error writing %s to file for beneficiary %s in ACO %s", jsonType, beneficiaryID, acoID), tmpDir)
 		}
 	}
+
+	if hasAtLeastOneEntry {
+		*hasEntriesCount++
```
michaeljvaldes: Does it make more sense to return a boolean for "hasAtLeastOneEntry" and manage the count outside of this function?

bcdaworker/repository/postgres/repository_test.go
```diff
@@ -196,8 +237,3 @@ func (r *RepositoryTestSuite) TestJobKeyMethods() {
 	_, err = r.repository.GetJobKey(ctx, jobID, -1)
 	assert.EqualError(err, repository.ErrJobKeyNotFound.Error())
 }
-
-func assertJobsEqual(assert *assert.Assertions, expected, actual models.Job) {
-	expected.TransactionTime, actual.TransactionTime = expected.TransactionTime.UTC(), actual.TransactionTime.UTC()
-	assert.Equal(expected, actual)
-}
```
carlpartridge, line 203: Removed this as it strictly checks all fields including automatically created timestamps, etc.  We were also (previously) forced into doing timestamp column manipulation to get this passing.  I am now just explicitly asserting that the columns we care about are equal.  I can see this having a downside in that future column additions could be less covered but I think overall this is a bit simpler and cleaner for now.

bcdaworker/worker/worker.go
carlpartridge: Hmm interesting, the current approach is
parent func:
- set totalCount = 0
- pass pointer to totalCount into this method, let it increment

Alternative approach:
parent func:
- set totalCount = 0
- have this method return an additional bool var
- increment if bool var is true

Difference being +1 passed in param, -1 passed back param, no pointer complexity.  Im kind of torn on which function should care about this more so thats a wash.  So a bit of a tossup for me in which case I side on not having to rewrite this and its tests but can easily be convinced!

michaeljvaldes: I like setting a pattern of relying more on return values and making input params read-only in Go, so I'm on team refactor, but I also agree it isn't a big deal.

michaeljvaldes (approved):

juliareynolds-nava (approved):

PR #1416, opened 2026-06-23
BCDA-10099: propogate rate limiting to clients

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10099

## 🛠 Changes

- New error type `SSASErrorTooManyRequests`
- Updated `GetToken()` to intercept HTTP 429
- Modified the `GetAuthToken` error handler to return an HTTP 429 status code, and forward the Retry-After header

## ℹ️ Context

Following the implementation of token rate limiting in bcda-ssas-app, the SSAS client inside bcda-app needs to recognize HTTP 429 Too Many Requests responses and correctly propagate the 429 status code and Retry-After header back to the calling client.

## 🧪 Validation

Unit testing


bcda/auth/api.go
```diff
@@ -77,6 +77,10 @@ func (a BaseApi) GetAuthToken(w http.ResponseWriter, r *http.Request) {
 		case *customErrors.SSASErrorBadRequest:
 			ctxLogger.WithField("resp_status", http.StatusBadRequest).Errorf("Error making access token - %s | HTTPS Status Code: %v", err.Error(), http.StatusBadRequest)
 			http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
+		case *customErrors.SSASErrorTooManyRequests:
+			w.Header().Set("Retry-After", "300")
```
carlpartridge: This 300 is the same as `SSAS_RATE_LIMIT_DURATION_MINUTES` right?  If so we should use that (and convert to seconds).

bcda/auth/client/ssas.go
```diff
@@ -336,6 +336,8 @@ func (c *SSASClient) GetToken(credentials Credentials, r http.Request) (string,
 				return "", &customErrors.SSASErrorUnauthorized{Err: err, SsasStatusCode: resp.StatusCode, Msg: fmt.Sprintf("%s, Response Body: %s", constants.TokenRequestUnexpectedErr, string(b))}
 			case 400:
 				return "", &customErrors.SSASErrorBadRequest{Err: err, SsasStatusCode: resp.StatusCode, Msg: fmt.Sprintf("%s, Response Body: %s", constants.TokenRequestUnexpectedErr, string(b))}
+			case 429:
+				return "", &customErrors.SSASErrorTooManyRequests{Err: err, SsasStatusCode: resp.StatusCode, Msg: fmt.Sprintf("SSAS rate limit exceeded, Response Body: %s", string(b))}
```
carlpartridge, line 340: It looks like we werent catching for 429 previously so we can probably focus this message more specifically if we wanted to.

bcda/auth/api.go
bhagatparwinder: As I am writing this code, I feel it doesn't look clean. What I am doing is grabbing the value of the SSM param, multiplying it by 60, and setting it here (I also added a default of 5 minutes). I haven't pushed the changes yet. I do think it is ok to do this.

```go
retryMinutesStr := conf.GetEnv("SSAS_RATE_LIMIT_DURATION_MINUTES")
retryMinutes, parseErr := strconv.Atoi(retryMinutesStr)
if parseErr != nil || retryMinutes <= 0 {
    retryMinutes = 5
}
retrySeconds := strconv.Itoa(retryMinutes * 60)
w.Header().Set("Retry-After", retrySeconds)
```

However, another way will be to add the RetryAfter in the `SSASErrorTooManyRequests` struct, capture the header in the SSAS client, and then forward it in the API handler. Maybe this is too much?

bcda/auth/client/ssas.go
bhagatparwinder: Any idea what we would like it to be? I know @michaeljvaldes had tagged @greg-pf in the SSAS PR for this.

bcda/auth/api.go
carlpartridge: One option would be to add a helper method ~ `minutesToSeconds`, that would at least move the code elsewhere and could be more easily unit-tested.  I agree that getting the retry from SSAS is a bit excessive/complex.

Overall though, somewhat stemming from recent conversations, I think its best that our api returns the most accurate information whenever possible.

bcda/auth/client/ssas.go
carlpartridge: Going from this thread: https://cmsgov.slack.com/archives/CHG7Q7XNH/p1781037896611459
Something like "The auth token lasts 20 minutes.  As such we only allow for a certain amount of new tokens generated in {x time frame}.  Any token requests over that and you may see rate limiting take effect."
Maybe thats too verbose for an error message...
We can ask in #bcda, but if you dont hear back I think we can re-visit adjusting this message as part of the ticket where we greatly reduce the threshold.

bcda/auth/api.go
bhagatparwinder: Done

carlpartridge (dismissed):

michaeljvaldes (dismissed):

michaeljvaldes (dismissed):

carlpartridge (dismissed):

bcda/auth/client/ssas.go
bhagatparwinder: Asked in #bcda and I have tweaked it.

carlpartridge (dismissed):

michaeljvaldes (dismissed):

michaeljvaldes (approved):

carlpartridge (approved):

PR #1418, opened 2026-06-26
BCDA-10124: remove fhir go package

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10124

## 🛠 Changes

- Removed `github.com/google/fhir/go` dependency and associated protobuf deps
- Created minimal STU3 and R4 models
- Swapped the protobuf marshaller for native `encoding/json`
- Updated metadata endpoints (`v2/api.go` and `v3/api.go`) and the background worker (`worker.go`) to compile with the new native structs.
- Deleted dead code for `CreateCapabilityStatement` and `WriteCapabilityStatement` code from the `v2` and `v3` writers.
- Refactored tests to use standard JSON unmarshalling and assertions on native Go structs.

## ℹ️ Context

`https://pkg.go.dev/github.com/google/fhir/go` was last updated in 2022. It gets flagged pretty regularly by Snyk package vulns. 
Removed it from bcda-app and implemented native Go Structs, updated tests, and refactored any associated code.

## 🧪 Validation

Unit testing


bcda/api/v3/api.go
```diff
@@ -388,11 +363,11 @@ func (a ApiV3) Metadata(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func restResourceSearchParam(n string, t fhircodes.SearchParamTypeCode_Value, d string) *fhircapabilitystatement.CapabilityStatement_Rest_Resource_SearchParam {
-	p := &fhircapabilitystatement.CapabilityStatement_Rest_Resource_SearchParam{
-		Name:          &fhirdatatypes.String{Value: n},
-		Type:          &fhircapabilitystatement.CapabilityStatement_Rest_Resource_SearchParam_TypeCode{Value: t},
-		Documentation: &fhirdatatypes.Markdown{Value: d},
+func restResourceSearchParam(n string, t string, d string) r4.SearchParam {
```
colby-seyferth-nava: can we add the search param types to our new model and require they use one here? This way we won't have to rely on the engineer (or AI agent) adding the param to the CapabilityStatement having to know the valid types.

bcda/api/v2/api.go
```diff
@@ -79,28 +66,30 @@ func (a ApiV2) BulkPatientRequest(w http.ResponseWriter, r *http.Request) {
 }
 
 /*
-		swagger:route GET /api/v2/Group/{groupId}/$export bulkDataV2 bulkGroupRequestV2
+swagger:route GET /api/v2/Group/{groupId}/$export bulkDataV2 bulkGroupRequestV2
+
+# Start FHIR R4 data export (for the specified group identifier) for all supported resource types
 
-	    Start FHIR R4 data export (for the specified group identifier) for all supported resource types
+Initiates a job to collect data from the Blue Button API for your ACO. The only Group identifier supported by the system are `all` and `runout`.
```
carlpartridge: Nit: Blue Button API -> BFD

bcda/api/requests_test.go
```diff
@@ -1711,3 +1691,31 @@ func makeTypeFilterParam(params [][]string) fhir.TypeFilterParameter {
 	}
 	return typeFilterParam
 }
+
+func getTasksFromSTU3Bundle(t *testing.T, data []byte) (uint32, []*stu3.Task) {
```
carlpartridge, line 1695: Why do we need these helpers?  It looks like it gets all tasks from all of an entries resources?

bcda/api/v3/api.go
bhagatparwinder: This is an excellent suggestion. I have added all valid FHIR search parameter types in the model and updated references to use the defined type-safe constants.

bcda/api/v2/api.go
bhagatparwinder: Done!

bcda/api/v3/api.go
```diff
@@ -206,7 +195,7 @@ func (a ApiV3) DeleteJob(w http.ResponseWriter, r *http.Request) {
 }
 
 /*
-swagger:route GET /api/v3/attribution_status attributionStatusv3 attributionStatus
+swagger:route GET /api/v3/attribution_status attributionStatusv3 attributionStatusv3
```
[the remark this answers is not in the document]
bhagatparwinder: Updated the Swagger route definition for `/api/v3/attribution_status`. Retained the versioned tag (attributionStatusv3) but used the stable operation ID (attributionStatus), matching the naming convention established in v2.

bcdaworker/worker/worker_test.go
```diff
@@ -510,7 +510,7 @@ func (s *WorkerTestSuite) TestWriteEOBDataToFile_BlueButtonIDNotFound() {
 
 func (s *WorkerTestSuite) TestAppendErrorToFile() {
 	appendErrorToFile(s.logctx, s.testACO.UUID.String(),
-		fhircodes.IssueTypeCode_CODE_INVALID,
+		stu3.IssueTypeCode("code-invalid"),
```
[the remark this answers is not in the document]
bhagatparwinder: Replaced the string stu3.IssueTypeCode("code-invalid") with the defined constant stu3.IssueTypeCodeException

bcda/api/requests_test.go
bhagatparwinder: We need these because of how I defined `BundleEntry` in our models. It has the `Resource` type as `any` (so we can hold any FHIR resource type).

When we call `json.Unmarshal` on the outer `Bundle` the decoder has no type info and will end up parsing it as a generic map[string]. The helper functions are there so we can convert it into typed stu3.Task or r4.Task and make test assertions better (type safety).

It could be handled in a different way:

1. Since all bundles in our test cases only contain `Task` resources, we can define a test-only bundle struct that hardcodes `Resource` type to `Task`. I think this will be my preferred way because it removes the helpers and double-marshaling.
2. I looked up some libraries like `mapstructure` that can do this (so basically switch the helpers to a library). I don't think we want to add more bloat.
3. We _could_ create a custom JSON unmarshaler to decode into the correct struct, but that will be making our models fairly complex.

I think I recommend either leaving it as is or switching to the first option.

carlpartridge: I dont think I feel too strongly about this but probably slightly lean towards leave as is for now.

carlpartridge (approved):

colby-seyferth-nava (approved):

PR #1421, opened 2026-06-30
BCDA-10216: update attribution-import roles

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10216

## 🛠 Changes

- create new delivery role for uploading attribution files with limited permissions
- create new kms key for encrypting attribution file bucket
- create new param for storing role arns that can assume the delivery role

## ℹ️ Context

The attribution import process requires external users to be able to deliver attribution files to an s3 bucket owned by BCDA. This PR intends to create a delivery role which can be assumed by external users, and which has very narrow permissions. These permissions include:
- upload to a particular path within the attribution file bucket (the root path)
- encryption/decryption of the kms key used by the attribution file bucket

To support this, a new kms key has been created to encrypt the attribution file bucket, and a param store parameter has been created to store a comma-delimited list of roles that can assume the delivery role.

Currently, the role of the lambda function itself is configured to be assumed by other users. Additionally, we have asked external users to deliver files to certain subdirectories within the bucket, which is no longer necessary. This PR limits the new delivery role to just the root directory of the bucket. 
After these changes are implemented and all users are moved to the new delivery role and bucket path, certain params and permissions will be updated (and the lambda function role will be refactored) in a [future ticket](https://jira.cms.gov/browse/BCDA-10245).

## 🧪 Validation

Tested with attribution-import integration test workflow
Tofu plan for Prod is shared via Box -- please message me for access!


ops/services/30-attribution-import/iam.tf
```diff
@@ -0,0 +1,67 @@
...
+    actions = ["sts:AssumeRole"]
+
+    principals {
+      type = "AWS"
+      identifiers = concat(
+        [
+          data.aws_iam_role.admin.arn,
+          data.aws_iam_role.dasg_admin.arn,
+        ],
+        local.assume_role_arns
```
michaeljvaldes, line 24: External roles, grabbed from a comma-delimited list stored in paramstore

ops/services/30-attribution-import/iam.tf
```diff
@@ -0,0 +1,67 @@
...
+    sid    = "AllowS3Upload"
+    effect = "Allow"
+
+    actions = [
+      "s3:PutObject",
+      "s3:AbortMultipartUpload",
+      "s3:ListMultipartUploadParts"
+    ]
+    resources = [
+      "${module.attribution-import_file_bucket.arn}/bfdeft01/bcda/in/${var.env}/*"
```
michaeljvaldes: Restrict delivery role to a specific path in the file bucket

ops/services/30-attribution-import/main.tf
```diff
@@ -23,6 +23,17 @@ data "aws_kms_alias" "bcda_app_config_kms_key" {
   name = "alias/bcda-${var.env}-app-config-kms"
 }
 
+resource "aws_kms_key" "attribution-import_bucket" {
```
michaeljvaldes, line 26: Create a custom key for encrypting the file bucket since it will be decrypted by a role available to external users.

ops/services/30-attribution-import/main.tf
```diff
@@ -191,11 +208,12 @@ data "aws_iam_policy_document" "sns_send_message" {
 }
 
 module "attribution-import_file_bucket" {
-  source = "github.com/CMSgov/cdap//terraform/modules/bucket?ref=787224b"
+  source = "github.com/CMSgov/cdap//terraform/modules/bucket?ref=6ded520857376f46bb317dca898e5df6a9ecc93b"
 
   app           = local.app
   env           = var.env
   name          = "${local.full_name}-file"
+  kms_key_arn   = join("", ["", aws_kms_key.attribution-import_bucket.arn])
```
michaeljvaldes, line 216: This "join" operation is an attempt to ensure that the value passed as the `kms_key_arn` is not null, since this depends on the key having already been created, and tofu will fail in the planning stage otherwise.

ops/services/30-attribution-import/security_group_rules.tf
```diff
@@ -3,12 +3,13 @@ data "aws_security_group" "db" {
...
-  from_port         = 5432
-  to_port           = 5432
-  description       = "Allow DB access"
+  from_port   = 5432
+  to_port     = 5432
+  ip_protocol = "tcp"
+  description = "Allow DB access"
+
+  security_group_id            = module.attribution_import_function.security_group_id
+  referenced_security_group_id = data.aws_security_group.db.id
```
michaeljvaldes, line 12: Restrict the db egress rule to the db security group; this change was necessary in order to be able to apply the tofu.

ops/services/10-config/values/test.sopsw.yaml
```diff
@@ -150,6 +150,7 @@
 /bcda/${env}/bene-prefs/sensitive/iam_bucket_role_arn: ENC[AES256_GCM,data:x2th+yghg4iJXhHNULY//lv7ttOMFFg1BKkh9QntUiGMDEZ0XZTXwJdHd2w8MOSupTGRh46dALowWWwmsH0A4/AqUsQhvbqeuDd7BWsx8nNbT41T1MpCYAy/u2WzkoiVogac,iv:UIdpM5P4702VWYnm94Bfj76eyw28zPWOItAVb69Xu2k=,tag:eoUk2cWM7MfPAJIDcknOpQ==,type:str]
 /bcda/${env}/bene-prefs/sensitive/sns_topic_arn: ENC[AES256_GCM,data:76QcBFUTw5lQuXntlu1Ttu4MunyP5TIvDzkHqZ5N/r37TBVgPlLmfe/1g+xNmxNxGtE9u4ZyIne1tXt54RsTwDAhBJYpMSw=,iv:kYzRk1BaiGDeTSaTG6ycHp9TWfOsu7zNKxk+dvgGYFw=,tag:fgNtYkTRe0wiU5NnrOhYLA==,type:str]
 /bcda/${env}/bene-prefs/nonsensitive/bfd_bene_prefs_bucket: s3://bfd-test-bene-prefs-bcda20260225231712780400000002/bfdeft01/bcda/in/test
+/bcda/${env}/attribution-import/sensitive/delivery_roles: ENC[AES256_GCM,data:eZ86sAbWmaDmKR2SIGelpzPxwZp0bX9UKMLNelXzvZJAvA2j7Mwx1frEdbWThR0bzjs+s+Lws/Y=,iv:bBoe+L3ndFcasBE7zsKp1fDri2OJ6OXtmcdz10UzbSE=,tag:Bj+EvLU49/WjGZ3OJJRk0A==,type:str]
```
carlpartridge: Nit: are these arns?  Would `delivery_role_arns` be a more helpful description?  Right now it seems a bit nebulous.  Or `external_role_arns`?  `external_delivery_role_arns` or is this getting out of hand 😅

ops/services/30-attribution-import/iam.tf
carlpartridge: Unfortunately I think we need both:
`...bcda/in/${var.env}/*"` as well as `...bcda/in/*"`.  The former is an annoying artifact, the latter what prod uses.  Unless something has changed in which case I need to update some tickets 😅 
Actually maybe it would be best to very explicitly to just call out `bcda/in/test/*` and `bcda/in/*` since we dont have these set up to run in dev/sandbox anyways?  The ticket I referenced above is to remove this artifact so I can add cleaning this up as well to it if we want.

michaeljvaldes: Fixed! It would be great to clean this up as part of that other ticket.

ops/services/10-config/values/test.sopsw.yaml
michaeljvaldes: Renamed to `delivery_role_arns` 👍

ops/services/30-attribution-import/iam.tf
```diff
@@ -0,0 +1,68 @@
...
+
+    actions = [
+      "s3:PutObject",
+      "s3:AbortMultipartUpload",
+      "s3:ListMultipartUploadParts"
+    ]
+    resources = [
+      "${module.attribution-import_file_bucket.arn}/bfdeft01/bcda/in/*",
+      "${module.attribution-import_file_bucket.arn}/bfdeft01/bcda/in/test/*"
+    ]
```
carlpartridge, line 42: I think we also dont want the `bfdeft01` portion.  I dont see that path in our prod file bucket.  I do see it as part of our path in the test env and as part of our workflows but I think we can remove that to clean things up.

In prod I also see a `prod` subfolder (bcda/in/prod) that does have files in it.  I think those are some of our test files?  Or potentially files we gave external clients to test with?  But Im not sure if there is a good way to know what path we told them to use?  If we cant track that down and we want to be safe we should add {var.env} back in I think.

michaeljvaldes: I limited the path for the new delivery role to just files in the root "directory"
I figure we don't need all of the subdirectory stuff anymore, and when we ask EFT to move to the new role, we can also ask them to change the bucket directory. This should not impact their current operations.

optout/utils.go
```diff
@@ -101,16 +94,3 @@ func ConvertDt(s string) (time.Time, error) {
 	}
 	return t, nil
 }
-
-// Checks if the given S3 filePath is for the current environment; this is necessary for lower environments
-// since they share a single BFD S3 bucket and will upload files under a subdirectory for the given env.
-func IsForCurrentEnv(filePath string) bool {
```
michaeljvaldes, line 107: Remove code intended for a scenario in which different envs share the same attribution bucket.

.github/workflows/attribution-import-integration-test.yml
```diff
@@ -89,16 +91,16 @@ jobs:
           zip -jr $fname $new_dir
 
           aws s3 cp --no-progress $fname \
-            s3://$BUCKET/bfdeft01/bcda/in/test/$fname
+            s3://$BUCKET/$fname
```
carlpartridge: Super nit: We are doing $var-name here but below we do ${var-name}.  Same with earlier in the line as well.

carlpartridge (dismissed):
Will want to get Grant and/or Mias input as well here I think!

.github/workflows/attribution-import-integration-test.yml
michaeljvaldes: Fixed!

carlpartridge (dismissed):

gfreeman-navapbc (dismissed):
Infra changes LGTM

ops/services/30-attribution-import/iam.tf
```diff
@@ -0,0 +1,67 @@
+locals {
+  assume_role_arns = split(",", module.platform.ssm.attribution-import.delivery_role_arns.value)
```
bhagatparwinder: Should we wrap this in `compact`?

ops/services/30-attribution-import/main.tf
bhagatparwinder: If the resource hasn't been created, wouldn't it be represented as `(known after apply)` and not null? I believe OpenTofu handles this as arn are guaranteed to exist eventually.

This would be considered an unknown value and not null: https://developer.hashicorp.com/terraform/plugin/framework/functions/concepts#argument-handling

ops/services/30-attribution-import/iam.tf
michaeljvaldes: Great idea, fixed!

ops/services/30-attribution-import/main.tf
michaeljvaldes: I'm not familiar with exactly why tofu is not assuming the arn will be created, but tofu did throw an error in the planning stage if I didn't use the "join" wrapper.

bhagatparwinder (approved):

carlpartridge (approved):

PR #1423, opened 2026-07-02
BCDA-10197: Add config file sync workflows

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10197

## 🛠 Changes

Add a new workflow that syncs repo config file to our s3 bucket.  Slight modifications of existing conf files (syncing from bucket files).

## ℹ️ Context

Managing the config files in the s3 buckets is a pain and leads to things slowly getting out of sync.  This should remedy that while also giving us the ability to manually update a file in case of emergency.

## 🧪 Validation

Need to get this merged before I can even run the new workflow :(

### DIFF
--- a/.github/workflows/publish-and-deploy-non-prod.yml
+++ b/.github/workflows/publish-and-deploy-non-prod.yml
@@ -4,7 +4,7 @@ name: Publish and Deploy Non-Prod
 
 on:
   schedule:
-    - cron: 0 10 * * 1-5 # every workday at 10am UTC (6am EST)
+    - cron: 0 9 * * 1-5 # every workday at 9am UTC
   workflow_dispatch:
     inputs:
       ref:
--- a/.github/workflows/publish-and-deploy-non-prod.yml
+++ b/.github/workflows/publish-and-deploy-non-prod.yml
@@ -4,7 +4,7 @@ name: Publish and Deploy Non-Prod
 
 on:
   schedule:
-    - cron: 0 12 * * 1-5 # every workday at 8am EST
+    - cron: 0 10 * * 1-5 # every workday at 10am UTC (6am EST)
   workflow_dispatch:
     inputs:
       ref:
--- a/.github/workflows/sync-config.yml
+++ b/.github/workflows/sync-config.yml
@@ -60,6 +60,9 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.3.0
         with:
           ref: ${{ inputs.ref || 'main' }}
+      - name: Install yamllint
+        run: |
+          sudo dnf install -y yamllint
       - name: Verify yml file
         run: |
           yamllint ./conf/configs/${{ env.RELEASE_ENV }}.yml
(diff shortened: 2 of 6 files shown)


.github/workflows/sync-config.yml
```diff
@@ -0,0 +1,71 @@
...
+          ref: ${{ inputs.ref || 'main' }}
+      - name: Verify yml file
+        run: |
+          yamllint ./conf/configs/${{ env.RELEASE_ENV }}.yml
+      - name: Sync to API s3 bucket
+        run: |
+          aws s3 cp ./conf/configs/${{ env.RELEASE_ENV }}.yml s3://$CONFIG_BUCKET/api/config.yml --region ${{ vars.AWS_REGION }}
+      - name: Sync to Worker s3 bucket
+        run: |
+          aws s3 cp ./conf/configs/${{ env.RELEASE_ENV }}.yml s3://$CONFIG_BUCKET/worker/config.yml --region ${{ vars.AWS_REGION }}
```
michaeljvaldes, line 74: Do you think it's worth it to have a final step which forces a new ECS deployment in the case of this workflow being run via workflow_dispatch?

carlpartridge: Hmm my initial reaction was a strong no since that could be akin to a new 'release'/'deploy'.  But it would kind of make sense as well.  I think we arent going to manually run this very often since its now a part of the actual deploy?

.github/workflows/publish-and-deploy-non-prod.yml
```diff
@@ -4,7 +4,7 @@ name: Publish and Deploy Non-Prod
 
 on:
   schedule:
-    - cron: 0 12 * * 1-5 # every workday at 8am EST
+    - cron: 0 9 * * 1-5 # every workday at 9am UTC
```
carlpartridge, line 7: Updated this to be earlier, currently its running around 10am, but I think its most convenient if its run/deployed before we even start work.

.github/workflows/sync-config.yml
michaeljvaldes: Yeah, I think it would be fine either way. Maybe just a comment about needing to force a new deployment would be just as good, either here or in our deployment documentation or something.
For what it's worth this is already akin to a new "deploy" since I believe newly spun up containers would still grab the newest version of the config file in the bucket, which could differ from the version that older containers might be using.

michaeljvaldes (dismissed):
It might be a good idea to do a general sweep of the differences in the configs between our repo and the deployed environment files before merging or implementing the new workflow.

carlpartridge:
> It might be a good idea to do a general sweep of the differences in the configs between our repo and the deployed environment files before merging or implementing the new workflow.

Already done!

.github/workflows/sync-config.yml
carlpartridge: Added some comments in our README as well as updated a couple of conflunce pages.

michaeljvaldes (approved):

bhagatparwinder (approved):

PR #1424, opened 2026-07-06
Carl/bcda 10197 fix config workflow

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10197

## 🛠 Changes

Adjustments to the sync config workflow.

## ℹ️ Context

Previously stubbed out this workflow as you cant test workflows without them existing in the default (main) branch.

## 🧪 Validation

Passing workflow: https://github.com/CMSgov/bcda-app/actions/runs/28811286359
I also tested yq locally after adjusting a yml file to be invalid.  It throws an error as expected.

### DIFF
--- a/.github/workflows/sync-config.yml
+++ b/.github/workflows/sync-config.yml
@@ -60,14 +60,10 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.3.0
         with:
           ref: ${{ inputs.ref || 'main' }}
-      - name: Install yamllint
-        run: |
-          sudo dnf install epel-release -y
-          sudo dnf clean all && sudo dnf makecache
-          sudo dnf install yamllint -y
       - name: Verify yml file
-        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-config.yml
```diff
@@ -60,12 +60,12 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.3.0
         with:
           ref: ${{ inputs.ref || 'main' }}
-      - name: Install yamllint
+      - name: Install yq to lint yaml file
```
michaeljvaldes, line 63: I think the CDAP repo has an [action that does something similar](https://github.com/CMSgov/cdap/tree/main/actions/setup-yq). Would it be more convenient to use that?

bhagatparwinder (dismissed):

bhagatparwinder (approved):

michaeljvaldes (approved):

PR #1426, opened 2026-07-07
BCDA-10222: config updates for access

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10222

## 🛠 Changes

- Add lookback for 3 years
- deny v1 and v2
- Add data access to adjudicated

## ℹ️ Context

Add config for ACCESS model


conf/configs/sandbox.yml
```diff
@@ -96,6 +98,8 @@ rate_limit_config:
   all: false
   acos: ['SBXBM0002']
 v3_enabled_acos: []
-v1_v2_deny_regexes: []
+v1_v2_deny_regexes:
+  - '^ACCES\d{5}$'
+  - '^GUIDE-\d{5}$'
```
carlpartridge: This is different from prod (mission IOTA).  Also since this is sandbox do we actually want to restrict anyone from anything?

conf/configs/prod.yml
```diff
@@ -133,6 +146,9 @@ v3_enabled_acos:
   - 'D0315'
   - 'D0324'
   - 'D0347'
-v1_v2_deny_regexes: ['IOTA\d{3}']
+v1_v2_deny_regexes:
+  - 'IOTA\d{3}'
```
carlpartridge: To keep it up with how we have things elsewhere I think we should change `IOTA\d{3}` -> `^IOTA\d{3}$`

conf/configs/test.yml
```diff
@@ -89,6 +91,8 @@ rate_limit_config:
   all: true
   acos: []
 v3_enabled_acos: []
-v1_v2_deny_regexes: []
+v1_v2_deny_regexes:
+  - '^ACCES\d{5}$'
```
carlpartridge, line 97: Add Iota?

conf/configs/sandbox.yml
bhagatparwinder: Removed it altogether since this is sandbox.

conf/configs/prod.yml
bhagatparwinder: Done

conf/configs/test.yml
bhagatparwinder: Done

conf/configs/prod.yml
```diff
@@ -80,6 +80,19 @@ aco_config:
         file_performance_year: 6
         file_date: 7
         metadata_matches: 8
+    # ACCESS (Advancing Chronic Care with Effective, Scalable Solutions) Model
+    - model: 'ACCESS'
+      name_pattern: '^ACCES\d{5}$' # only one 's' due to character limits
+      lookback_period: 3
+      data:
+        - 'adjudicated'
```
greg-pf, line 89: Is this used in v3? If so, ACCESS should have `partially-adjudicated` too. If not, we could remove it altogether for ACCESS, IOTA, and GUIDE

greg-pf (dismissed):

conf/configs/prod.yml
bhagatparwinder: v3 still uses it to validate resource access

https://github.com/CMSgov/bcda-app/blob/c3d4c9497db0019cd325b90ef010bad33cf2a8c9/bcda/api/requests.go#L821-L828

If `adjudicated` is not present in `cfg.Data`, the API will return a `400 Bad Request`. Also, when a client performs an export request without specifying resources via `_type`, BCDA determines the default resource types

https://github.com/CMSgov/bcda-app/blob/c3d4c9497db0019cd325b90ef010bad33cf2a8c9/bcda/api/requests.go#L792-L794

Without `adjudicated` configured, the default export would not include any resource types.

bhagatparwinder: To answer the question about `partially-adjudicated`, we used it in v1/v2. For v3, we are relying on the property `v3_no_partial_claims_models`

michaeljvaldes (dismissed):

conf/configs/prod.yml
greg-pf: Thanks! It looks like we're missing `adjudicated` for GUIDE's config then. Could you sneak that addition into this PR so that GUIDE participants will be able to make requests once onboarded?

bhagatparwinder: Done!

greg-pf (approved):

carlpartridge (approved):

michaeljvaldes (approved):

PR #1429, opened 2026-07-08
BCDA-10213: sort benes

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10213

## 🛠 Changes

- Sort benes ascending, placing NULLS LAST
- Update DB with latest bbID
- Updates to mock repo

## ℹ️ Context

BFD handles massive amounts of data; the index it uses to find patients does not entirely fit in its memory. When BCDA requests patients in random order, BFD is forced to repeatedly fetch data from the source rather than its mem cache.
By sorting beneficiaries by their BFD ID (`bene_sk`) before bundling them into subjobs and querying BFD, we ensure that the data from consecutive requests is stored close together. This allows BFD to keep the relevant index in its memory cache across sequential requests, allowing some performance improvements.

## 🧪 Validation

Linter and unit tests


bcdaworker/worker/worker.go
```diff
@@ -405,7 +405,13 @@ func getBeneficiary(ctx context.Context, r repository.Repository, beneID uint, b
...
 		}
 
-		cclfBeneficiary.BlueButtonID = bbID
+		if cclfBeneficiary.BlueButtonID != bbID {
+			err = r.UpdateCCLFBeneficiaryBlueButtonID(ctx, beneID, bbID)
+			if err != nil {
+				log.GetCtxLogger(ctx).Warnf("failed to update blueButtonId in DB for cclfBeneficiaryId %d: %s", beneID, err.Error())
+			}
+			cclfBeneficiary.BlueButtonID = bbID
+		}
```
carlpartridge, line 414: So if Im following correctly what currently happens is:
- for each bene as part of a subjob we request from BFD the bene id using the MBI
- once we have this bene id we then use that to get the resource data

We are adding this to the workflow:
- once we have the bene id, save it into the db.
- on general bene queries sort by bene_id.

Im not sure this fixes the problem wholesale but I think would help it some.  Basically we generate a massive list of benes to query for, and we iterate through them one at a time, saving the results back into the DB.  Since we dont have any of this ahead of time then the new sorts dont really help on the first run, but would help on future runs.  At least until a new cclf_file happens.  

An alternative (and much more complex fix) would be to as a prior step to the current processing, go get all bene_ids beforehand, sort those into smaller subjobs, then get resource data.

colby-seyferth-nava (dismissed):
The code looks good. I am still unsure about the benefits. are running any baseline performance tests before and after the change? not sure how the extra DB write for us will compare to the faster reads on the BFD side.

If we update our DB with the bene_sk, do we need to fetch the bene_sk from BFD the next time? it feels safer to continue to fetch it each time, but it seems a lot faster to not fetch it every time.

Also, agree with Carl that there could be a larger re-write to gather the bene_sk list first (maybe at the same time as gathering the whole patient resource that we are creating files for), then sort, then get resources. but that seems like it's beyond the scope of this quick win

bhagatparwinder:
> The code looks good. I am still unsure about the benefits. are running any baseline performance tests before and after the change? not sure how the extra DB write for us will compare to the faster reads on the BFD side.

This is hard to benchmark without real-world traffic. However, it doesn't hurt to sort benes (especially if they are indexed on the BFD side, allowing for faster reads).

> If we update our DB with the bene_sk, do we need to fetch the bene_sk from BFD the next time?

This PR completes the work for the ticket, but it was the exact (additional) improvement I spoke about at stand-up. There isn't a need to fetch it from BFD. We do it for every single bene in worker (see here: https://github.com/CMSgov/bcda-app/blob/ee4d86a5b2844b533311f3b2f65015581fda43fe/bcdaworker/worker/worker.go#L401) and add it to the model, but do not persist it. I am going to skip that network call with this change (incoming commit).

bcdaworker/worker/worker.go
bhagatparwinder: > but would help on future runs

This is where I would also potentially see the benefit. There is an additional benefit I mentioned in replying to Colby. I am going to push that commit now.

carlpartridge (approved):

colby-seyferth-nava (approved):

PR #1431, opened 2026-07-09
Return after responding for an expired or archived job in RequireTokenJobMatch

arpitjain099:
The expired/archived branch of `RequireTokenJobMatch` writes a 404 but is missing a `return`:

```go
if job.Status == models.JobStatusExpired || job.Status == models.JobStatusArchived {
    ctx, _ = log.WriteWarnWithFields(...)
    rw.NotFound(..., http.StatusNotFound, responseutils.JobExpiredErr, "")
}   // no return

// ACO did not create the job
if !strings.EqualFold(ad.ACOID, job.ACOID.String()) { ... return }
next.ServeHTTP(w, r)
```

Every other error branch in this function returns after responding (no-auth-data, job-not-found, ACO-mismatch). This one does not, so for a caller requesting their **own** expired/archived job the ACO check passes (same ACO) and `next.ServeHTTP` runs the real handler after the 404 was already written. The result is a "superfluous response.WriteHeader" condition, a corrupted response body (the JobExpiredErr OperationOutcome followed by the handler output), and the intended expired/archived denial is defeated.

This is not a cross-tenant issue (the ACO check still runs, so another ACO's expired job is still rejected), which is why it is a plain correctness fix. Impact is confined to a caller's own expired/archived jobs.

Fix: add the missing `return`.

I also strengthened `TestRequireTokenJobMatchExpiredJob`: its `archivedJob` fixture was created with `JobStatusExpired` (copy-paste), so the "Archive Job" case never exercised the archived branch, and it used a no-op handler so the fall-through was invisible (`httptest.ResponseRecorder.Code` locks on the first `WriteHeader`). It now sets `JobStatusArchived` and uses a sentinel handler, asserting the downstream handler is never reached.

Note: this middleware test needs a Postgres test database (`postgrestest.CreateJobs`), which I could not stand up locally, so I verified the change compiles (`go vet ./auth/`) and relied on CI to run the DB-backed suite. Happy to open a JIRA ticket if the team prefers.

### DIFF
--- a/bcda/auth/middleware.go
+++ b/bcda/auth/middleware.go
@@ -270,6 +270,7 @@ func (m AuthMiddleware) RequireTokenJobMatch(db *sql.DB) func(next http.Handler)
 					logrus.Fields{"resp_status": http.StatusNotFound},
 				)
 				rw.NotFound(log.NewStructuredLoggerEntry(log.Auth, ctx), w, http.StatusNotFound, responseutils.JobExpiredErr, "")
+				return
 			}
 
 			// ACO did not create the job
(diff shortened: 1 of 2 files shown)


carlpartridge:
Hello, thank you for your interest in making BCDA better.  We have taken your feedback and created a ticket to look into this further (https://jira.cms.gov/browse/BCDA-10254).

PR #1439, opened 2026-07-15
BCDA-10268: configure worker graceful shutdowns

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10268

## 🛠 Changes

- upgrade river to v0.38.0
- add graceful shutdown handling for river client and workers
- set timeout for worker service in docker to 30s (to mimic ECS)
- (unrelated) docker ignore ops directory to speed up builds significantly

## ℹ️ Context

When ECS scales in, it issues a SIGTERM to worker containers. Presently, these worker are configured to intercept this signal and run os.Exit(), which immediately shuts down the running process and skips a graceful exit.
This PR leverages river's graceful exit timeout and a signal NotifyContext to understand when an exit signal is received and initiate a soft stop on the running workers, waiting 30 seconds for them to complete their work.

Some thing to consider: is 30 seconds a reasonable wait time to expect a running worker to complete its work? This is the default setting for ECS, and configuring a longer wait would involve a small infra change, too.

[Relevant river documentation](https://riverqueue.com/docs/graceful-shutdown)

## 🧪 Validation

Testing this is difficult, since it would involve issuing and handling exit signals in a test.
I tested this manually by configuring jobs locally to run long, and then shutting down the docker container.
Previous behavior: the code to rollback the transaction in `worker_prepare.go` was not reached
New behavior: the code to rollback the transaction in `worker_prepare.go` is reached


bcdaworker/cli/cli.go
```diff
@@ -126,39 +122,6 @@ func clearTempDirectory(tempDir string) error {
 	return nil
 }
 
-func waitForSig() {
```
michaeljvaldes, line 129: Remove function that intercepts shut down signals and runs os.Exit()

bcdaworker/queueing/river.go
```diff
@@ -70,35 +71,30 @@ func StartRiver(db *sql.DB, numWorkers int) *queue {
...
-		Logger:       logger,
-		MaxAttempts:  8, // This is roughly an hour of retries
-		Workers:      workers,
-		PeriodicJobs: periodicJobs,
+		JobTimeout:      10 * time.Minute,
+		Logger:          logger,
+		MaxAttempts:     8, // This is roughly an hour of retries
+		Workers:         workers,
+		PeriodicJobs:    periodicJobs,
+		SoftStopTimeout: 30 * time.Second,
```
michaeljvaldes: Configure a soft stop timeout, which is the amount of time that the river client will wait for workers to complete their jobs before shutting down. Thirty seconds is the corresponding default timeout of ECS.

bcdaworker/queueing/river.go
```diff
@@ -70,35 +71,30 @@ func StartRiver(db *sql.DB, numWorkers int) *queue {
...
 	})
 
 	if err != nil {
 		logger.Error("failed to init river client", "error", err)
 		panic(err)
 	}
 
-	if err := riverClient.Start(context.Background()); err != nil {
+	signalCtx, stop := signal.NotifyContext(ctx, syscall.SIGINT, syscall.SIGTERM)
```
michaeljvaldes: Create a notify context which is marked done when a shutdown signal is received. We can pass this context to riverClient.Start(), and then try to wait for workers to finish their current job before exiting.

bcdaworker/cli/cli.go
```diff
@@ -84,9 +86,21 @@ func startWorker() {
 	defer profiler.Stop()
 
 	createWorkerDirs()
-	queue := queueing.StartRiver(db, utils.GetEnvInt("WORKER_POOL_SIZE", 4))
-	defer queue.StopRiver()
-	waitForSig()
+
+	ctx := context.Background()
+	signalCtx, stop := signal.NotifyContext(ctx, syscall.SIGINT, syscall.SIGTERM)
```
michaeljvaldes: Create a notify context which is marked done when a shutdown signal is received. We can pass this context to riverClient.Start(), and then try to wait for workers to finish their current job before exiting.

.dockerignore
```diff
@@ -12,4 +12,6 @@
 !./bcdaworker/data/test/test.ndjson
 **/bcda/pending_delete_dir/*
 
-**/.DS_Store
\ No newline at end of file
+**/.DS_Store
+
+ops/
```
carlpartridge, line 17: Why does this speed up docker?  Are there other dirs we could be ignoring?

bcdaworker/cli/cli.go
```diff
@@ -84,9 +86,21 @@ func startWorker() {
...
+	riverClient := queueing.CreateRiverClient(logger, db, utils.GetEnvInt("WORKER_POOL_SIZE", 4))
+	if err := riverClient.Start(ctx); err != nil {
+		logger.Error("failed to start river client", "error", err)
+		panic(err)
+	}
+
+	<-signalCtx.Done()
+	stop()
+	logger.Info("Received exit signal; initiating soft stop (waiting for cancelled jobs to finish)")
+	<-riverClient.Stopped()
```
carlpartridge, line 102: Id imagine this would testable?  Start a river client process up, send a sigterm, verify river stopped?  Maybe we can make a fake job that just sleeps for 3s to test what happens?  I feel like Ive done that somewhere already?

carlpartridge (commented):
This is awesome! 🤞 

In `bcdaworker/queueing/enqueue.go` we also start up a river client.  This seems to be slightly different use case, as this seems to be an insert only client, not an actual worker.  Just wanted to double check that we dont need to adjust that as well?  Im thinking no...

.dockerignore
michaeljvaldes: The /ops directory can be huge, so it takes a lot of time to copy it over.
A better fix could involve specifying every directory that should be copied over in the Dockerfile instead of running `COPY . .` and for that it would be helpful to restructure our golang packages, which is a separate tech debt effort.

carlpartridge: Ooooh I wasnt considering all of the downloaded modules and hidden stuff... just the base tf files themselves which I would imagine to be quite light.

bcdaworker/cli/cli.go
michaeljvaldes: Yeah, I actually did manually test this doing something similar to that.
I suppose by "testable" I mean in a "testable in a unit test"
I don't immediately see a way to set up a test that is useful (i.e. doesn't just test river's features) but I'm open to suggestions.

bcdaworker/cli/cli.go
bhagatparwinder: I am confused; we created this context, but I don't see us passing it to `riverClient.Start()` on line 95.

I believe we either need to pass `signalCtx` to `riverClient.Start()` instead of `ctx` (`context.Background()`), or we need to call `riverClient.Stop()` right before line 103.

bcdaworker/cli/cli.go
```diff
@@ -84,9 +86,21 @@ func startWorker() {
...
-	queue := queueing.StartRiver(db, utils.GetEnvInt("WORKER_POOL_SIZE", 4))
-	defer queue.StopRiver()
-	waitForSig()
+
+	ctx := context.Background()
+	signalCtx, stop := signal.NotifyContext(ctx, syscall.SIGINT, syscall.SIGTERM)
+	defer stop()
+
+	riverClient := queueing.CreateRiverClient(logger, db, utils.GetEnvInt("WORKER_POOL_SIZE", 4))
+	if err := riverClient.Start(ctx); err != nil {
```
bhagatparwinder, line 94: Pass signalCtx here OR

bcdaworker/cli/cli.go
```diff
@@ -84,9 +86,21 @@ func startWorker() {
...
+		logger.Error("failed to start river client", "error", err)
+		panic(err)
+	}
+
+	<-signalCtx.Done()
+	stop()
+	logger.Info("Received exit signal; initiating soft stop (waiting for cancelled jobs to finish)")
```
bhagatparwinder, line 101: OR Add `riverClient.Stop(context.Background())` here.

bcdaworker/cli/cli.go
michaeljvaldes: Fixed!

bhagatparwinder (dismissed):

bcdaworker/queueing/river.go
carlpartridge: If ECS is 30s, does this need to be more like 25s?  Or ~29?  Im thinking we want river to have wrapped up everything plus a bit of time outside of that before ECS forces things?

michaeljvaldes: That sounds reasonable to me. Updated to 29 seconds to give the workers as long as possible to complete their work.

carlpartridge (approved):

bhagatparwinder (approved):

PR #1443, opened 2026-07-17
BCDA-10236: new v3 entities

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10236

## 🛠 Changes

Adding all the new v3 entities.

## ℹ️ Context

All model entities in the list have access to BCDA v3, which includes

- recent requests from SSP, REACH, and KCC entities
- all the IOTA participants
- all the ACCESS participants
- all the GUIDE participants


carlpartridge (commented):
Its a lot but I guess its not complicated!  To note we will need to run the sync_config workflow to update these in prod and when we want this to be live we will need to manually restart services.

michaeljvaldes (approved):

carlpartridge (approved):

PR #1446, opened 2026-07-22
BCDA-10210: Split BFD job fail metrics

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10210

## 🛠 Changes

- Fix jobs.benes_attributed_to_acos being overwritten
- Add subjob_id to subjob logging
- Split job failures into two categories:
  - Actual request failures (remains error)
  - Bene not found in BFD (turn into acceptable warning, but still commit error to file)

## ℹ️ Context

Benes attributed to ACOs was broken (being overwritten with a ~null value).  Job failures recently caused an issue with bene sorting where some subjobs ended up being entirely unfindable benes in BFD, thus failing the main job.  This is also just better and more accurate information.

## 🧪 Validation

Local linting and testing.


bcdaworker/worker/worker.go
```diff
@@ -296,8 +297,10 @@ func writeBBDataToFile(ctx context.Context, r repository.Repository, bb client.A
 
 	w := bufio.NewWriter(f)
 	defer w.Flush()
-	errorCount := 0
-	benesWithDataCount := 0
+
```
carlpartridge, line 300: The changes in this file are a bit gnarly but each time I tried a more "elegant" refactor I ended up breaking a bunch of stuff 😓.

bcdaworker/queueing/worker_prepare.go
```diff
@@ -157,7 +157,7 @@ func (p *PrepareJobWorker) prepareExportJobs(ctx context.Context, args worker_ty
 		return exports, args.Since, err
 	}
 
-	exports, err = p.svc.GetQueJobs(ctx, args)
+	exports, benesAttributed, err := p.svc.GetQueJobs(ctx, args)
```
carlpartridge, line 160: This is where the job.benes_attributed_to_acos was getting overwritten (technically in the defer statement on line 134).  BenesAttributedToACOs is figured in GetQueJobs and saved appropriately, but this parent function is not aware of that.  Our repo.UpdateJob function also overwrites all fields which is not ideal.  Also benes_attributed_to_acos has a default value which makes it hard to check for its presence.  There were a few different ways to approach this fix:
- change UpdateJob to save only passed in fields.  Probably a better change overall but with a much bigger impact radius.
- add a new repo ORM function to just save job count.  Felt a bit excessive.
- Have GetQueJobs return benes_attributed_to_acos so that we can update the job so that when we call UpdateJob later on it is saved properly.  This is what I ended up doing.  Its a bit widespread to update all function signatures but felt like the simplest approach.
- Get the updated job from the DB (repo.GetJobByID), update the args.Job to match each field, so that when its saved later its properly set.  Also felt a bit excessive.

bcdaworker/worker/worker.go
```diff
@@ -340,9 +343,16 @@ func writeBBDataToFile(ctx context.Context, r repository.Repository, bb client.A
 		}()
 
 		if err != nil {
-			logger.Error(err)
-			errorCount++
+			switch err.(type) {
```
bhagatparwinder: Hmm, I see Copilot's comment for this error switch statement. Is there anywhere in the call stack where we wrap the error, or do we not mind if the errorCount is incremented?

We could update it to `errors.As(err, &notFoundErr)`

bcdaworker/worker/worker.go
bhagatparwinder: Do we want to add a comment here that we are excluding system errors and missing benes in `benesRetrievedCount`?

bhagatparwinder (dismissed):

bcdaworker/worker/worker.go
carlpartridge: We dont currently, I imagine the comment is more to future proof things.  I was unable to get their suggestion to work but Im going to try again.  We have tests set up that should catch this in case we do modify this in the future.

bcdaworker/worker/worker.go
carlpartridge: Hmm maybe a different name would be better?  Or maybe just updating the comment...  Right now we report on, from top to bottom:
- total benes for this subjob
- total benes that did not have a request error (total benes - error count)
- total benes that we did get (total benes - error count - not found benes, this is benesRetrievedCount)
- total benes that we did get that actually had data (benesWithData)

bcdaworker/worker/worker.go
carlpartridge: Ok I retried it and it just worked... not sure what happened last time 😅

bhagatparwinder (approved):

michaeljvaldes (approved):

PR #1448, opened 2026-07-23
BCDA-10245: refactor attribute-import permissions

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10245

>[!WARNING]
>This PR should not be merged until EFT changes are made

## 🛠 Changes

- remove superfluous role permissions and enforce privilege role permissions
- reorganize lambda role permissions
- remove eft role from lambda management

## 🧪 Validation

Applied and evaluated with attribution-import integration test workflow


ops/services/30-attribution-import/iam.tf
```diff
@@ -25,9 +30,28 @@ data "aws_iam_policy_document" "delivery_assume_role" {
...
+    }
+    condition {
+      test     = "StringEquals"
+      variable = "${local.provider_domain}:aud"
+      values   = ["sts.amazonaws.com"]
+    }
+    condition {
+      test     = "StringLike"
+      variable = "${local.provider_domain}:sub"
+      values   = ["repo:CMSgov/bcda-app:*"]
```
carlpartridge, line 49: Quick check on the * here.  Can you walk me through this resource real quick?

michaeljvaldes: Absolutely! This part of the policy is intended to allow our github runners assume this role for running integration tests. I followed the pattern set by CDAP in the function module [here](https://github.com/CMSgov/cdap/blob/9cb3840542b06f72c3e2f64982e9610602798c46/terraform/modules/function/iam.tf#L31-L50).
I believe the asterisk covers the latter part of the provider domain, which I think can include additional stuff like an environment in github.

carlpartridge (dismissed):
Looks good to me but I think it would be good to get Grant or someone from CDAP approval as well.

.github/workflows/attribution-import-deploy.yml
```diff
@@ -60,7 +94,7 @@ jobs:
           AWS_REGION: ${{ vars.AWS_REGION }}
         with:
           params: |
-            BUCKET=/bcda/${{ env.DEPLOY_ENV }}/attribution-import-bucket
+            BUCKET=/bcda/${{ env.RELEASE_ENV }}/attribution-import-bucket
       - name: Upload and reload
         run: |
           aws s3 cp --no-progress function.zip \
```
bhagatparwinder, line 100: Out of curiosity, are we not appending the SHA to the zip file anymore `function-${{ github.sha }}.zip`?

bhagatparwinder (dismissed):

.github/workflows/attribution-import-deploy.yml
michaeljvaldes: Nice catch! I don't know why this lambda was different, but I updated it to match the other lambdas.

carlpartridge (approved):

bhagatparwinder (approved):

PR #1458, opened 2026-08-03
BCDA-10141: add opoutcome for v3 default filtering

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10141

## 🛠 Changes

- swap out hardcoded URL with const
- remove no longer needed v3 TODOs
- set up creation of a general warnings-and-info.ndjson file that can be used to set issues throughout a jobs processing
- quick extraction of some basic file and dir manipulation

## ℹ️ Context

When we kick off a v3 $export without the client specifying a system-type filter, we add our own default System-types in order to filter out SharedSystem claims.  This could be suprprising to a client so we are adding a new warnings file to hold this information.

## 🧪 Validation

Local linting and testing.


carlpartridge:
@colby-seyferth-nava Tagging you on this as 1) its your original ticket and 2) to verify the removal of all the v3 TODOs.  Let me know if any of them seem like they still actually need additional work or not.

bcda/models/postgres/pgx_repository.go
```diff
@@ -124,3 +125,37 @@ func (r *PgxRepository) UpdateCCLFFileImportStatus(ctx context.Context, fileID u
...
+
+	var id uint
+	err := r.pool.QueryRow(ctx, query, jobID, fname).Scan(&id)
+	if err != nil {
+		if err != pgxv5.ErrNoRows {
+			return fmt.Errorf("failed to check for existing job key for job %d, err: %w", jobID, err)
+		}
+
+		query = `
+			INSERT INTO job_keys (job_id, file_name)
```
bhagatparwinder: Are we setting `OperationOutcome` somewhere else in the code? This will set resource_type to "". When the client sends a request to get export status, we make the output array using `jobKey.ResourceType`. We will have type as "" in the output array. I think this should be

```
INSERT INTO job_keys (job_id, file_name, resource_type)
VALUES ($1, $2, 'OperationOutcome')
```

Edit: Also, we are doing a SELECT followed by INSERT. Do you see any race conditions when we have concurrent tasks (see TOCTOU)? Maybe having a condition where ON CONFLICT we do nothing?

bcda/service/warnings_and_info.go
```diff
@@ -0,0 +1,53 @@
...
+	"regexp"
+
+	"github.com/CMSgov/bcda-app/bcda/constants"
+	"github.com/CMSgov/bcda-app/bcda/models/fhir/r4"
+	"github.com/CMSgov/bcda-app/bcda/models/postgres"
+	"github.com/CMSgov/bcda-app/bcdaworker/worker"
+	"github.com/CMSgov/bcda-app/conf"
+)
+
+var HasSystemTypeRegex = regexp.MustCompile(`.*\?_typeFilter=.*\?_tag=.*\/System-Type.*`)
```
bhagatparwinder: Ok, I am thinking this might fail more than we want:
- This is assuming that `_typeFilter` is always the first query parameter
- If the client sends URL-encoded strings, this will fail (like %2FSystem-Type instead of /System-Type)

We probably want to decode `RequestURL` before checking the query parameters (and use contains for finding parameters).

bcda/service/warnings_and_info.go
```diff
@@ -0,0 +1,53 @@
...
+func SetupWarningsAndInfoFile(ctx context.Context, pgxRepo *postgres.PgxRepository, jobID uint) (string, error) {
+	payloadPath := fmt.Sprintf("%s/%d", conf.GetEnv("FHIR_PAYLOAD_DIR"), jobID)
+	filePath := payloadPath + "/" + constants.WarningsAndInfoFileName
+
+	err := worker.CreateDir(payloadPath)
+	if err != nil {
+		return "", fmt.Errorf("error creating payload directory: %w", err)
+	}
+
+	file, err := os.OpenFile(filePath, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0600) // #nosec G304
```
bhagatparwinder: Are we opening and closing the file immediately and then opening it again in the caller function? Could we use `*os.File` handle, or maybe add append functionality to this function (better encapsulated file operations)?

bcdaworker/queueing/worker_prepare_test.go
```diff
@@ -403,3 +417,102 @@ func (s *PrepareWorkerIntegrationTestSuite) TestQueueExportJobs() {
...
+			createsFile: false,
+		},
+	}
+
+	for _, tt := range tests {
+		s.Run(tt.name, func() {
+			err := handleDefaultSystemTypeWarningNeeded(s.ctx, s.pool, tt.job)
+			assert.NoError(s.T(), err)
+
+			// filePath := fmt.Sprintf("%s/%d/%s", conf.GetEnv("FHIR_PAYLOAD_DIR"), tt.job.Args.Job.ID, constants.WarningsAndInfoFileName)
```
bhagatparwinder: Delete?

bcda/models/postgres/pgx_repository.go
carlpartridge: Good call!  Added.  As far as I understand the `ON CONFLICT` only really works when you have unique constraints on the related columns.  Also a good call in double checking how we eventually serve these files up.  I will modify that to be a bit better as well.

bcda/service/warnings_and_info.go
carlpartridge: Good call!  I ran it against a few request urls I found in our prod db but they happened to always set typeFilter first.  I think we can just simplify it a bit by removing the initial ? and the / as well.  @colby-seyferth-nava Do you think we need to make this more specific?

bcda/service/warnings_and_info.go
carlpartridge: I started with creating a file pointer and passing that back but that can get pretty gnarly in golang and is generally preferred not to do, at least as I understand it.

bcda/api/requests.go
```diff
@@ -883,7 +891,7 @@ func (h *Handler) validateTypeFilterPACEligibility(ctx context.Context, typeFilt
 }
 
 // extractTagCodeFromValue extracts tag codes from either a short format (e.g., "SharedSystem")
-// or a full URL format (e.g., "https://bluebutton.cms.gov/fhir/CodeSystem/System-Type|SharedSystem").
+// or a full URL format (e.g., FHIRCodeSystemURL + "System-Type|SharedSystem").
```
michaeljvaldes: The old string is probably more readable in this comment

bcdaworker/queueing/worker_prepare.go
```diff
@@ -116,7 +118,18 @@ func (w *PrepareJobWorker) Work(ctx context.Context, rjob *river.Job[worker_type
 			}
 
 			err = tx.Commit(ctx)
-			return err
+			if err != nil {
+				logger.Errorf("prepare job tx failed for job id: %d, err: %v", rjob.Args.Job.ID, err)
+				return err
+			}
+
+			err = handleDefaultSystemTypeWarningNeeded(ctx, w.pool, rjob)
```
michaeljvaldes: Am I correct in assuming we don't have to worry about concurrent writes to the same error file because it's only being created/appended-to in the prepare worker, and there's only one prepare worker per main job?

bcda/service/warnings_and_info.go
```diff
@@ -0,0 +1,53 @@
...
+			},
+			Severity: "warning",
+		},
+	},
+}
+
+// SetupWarningsAndInfoFile finds or creates an info file to house generic warnings, issues, etc for a given job.
+// It will also find or create a jobkey.  There should only be 1 warnings and info file per job.
+func SetupWarningsAndInfoFile(ctx context.Context, pgxRepo *postgres.PgxRepository, jobID uint) (string, error) {
+	payloadPath := fmt.Sprintf("%s/%d", conf.GetEnv("FHIR_PAYLOAD_DIR"), jobID)
```
michaeljvaldes, line 29: Do you think it's worth creating a shared function for getting the path to a job directory?

bcda/api/requests.go
carlpartridge: Yeah 😞, however if you were to update this new var everywhere else at some point this would get lost in the shuffle.  Alternatively I could use a dummy URL?

bcdaworker/queueing/worker_prepare.go
carlpartridge: Hmm as of now yes...  However at some point the thought is that this could be written to for any errors that happen during the job process.  Not sure what the best way to handle this would be?  Do we need to figure that out now?

bcda/service/warnings_and_info.go
carlpartridge: Ummmm hmm, it would theoretically change:
```
payloadPath := fmt.Sprintf("%s/%d", conf.GetEnv("FHIR_PAYLOAD_DIR"), jobID)
```
to
```
payloadPath := getPayloadPath(jobID)

and

func getPayloadPath(str) str {
  return fmt.Sprintf("%s/%d", conf.GetEnv("FHIR_PAYLOAD_DIR"), str)
}
func getStagingPath(str) str {...}
func getTempPath(str) str {...}
```
Which doesnt feel like it gets us that much?  Especially since sometimes we want the dir path and sometimes we want the file path.  But if you feel strongly I can change it!

bcda/api/v3/api.go
```diff
@@ -340,7 +340,7 @@ func (a ApiV3) Metadata(w http.ResponseWriter, r *http.Request) {
 					{
 						Type: r4.ResourceTypeCodeExplanationOfBenefit,
 						SearchParam: []r4.SearchParam{
-							restResourceSearchParam("_tag", r4.SearchParamTypeToken, "Filter ExplanationOfBenefit by the meta.tag element. Pass full token as <system>|<code>. Supported codes in the 'https://bluebutton.cms.gov/fhir/CodeSystem/System-Type' system are: 'SharedSystem', 'NationalClaimsHistory', and 'DDPS'. By Default, only NationalClaimsHistory and DDPS claims will be returned."),
+							restResourceSearchParam("_tag", r4.SearchParamTypeToken, "Filter ExplanationOfBenefit by the meta.tag element. Pass full token as <system>|<code>. Supported codes in the '"+constants.FHIRCodeSystemURL+"System-Type' system are: 'SharedSystem', 'NationalClaimsHistory', and 'DDPS'. By Default, only NationalClaimsHistory and DDPS claims will be returned."),
```
colby-seyferth-nava: nit: to be consistent with style, should we put spaces around the + signs

bcda/constants/constants.go
```diff
@@ -63,3 +63,5 @@ const SoftwareDescription = "The Beneficiary Claims Data API (BCDA) enables heal
 
 const SmartOAuthURIsExtensionURL = "http://fhir-registry.smarthealthit.org/StructureDefinition/oauth-uris"
 const RestfulSecurityServiceSystem = "http://terminology.hl7.org/CodeSystem/restful-security-service"
+const FHIRCodeSystemURL = "https://bluebutton.cms.gov/fhir/CodeSystem/"
```
colby-seyferth-nava: Since this is specific to bluebutton/BFD, maybe we can call it `BluebuttonCodeSystemURL` or `BFDCodeSystemURL`

My concern is that there are other FHIR System URLs we may want to include in the future that are not the bluebutton ones.

related to this, it might also be good to include the full URL in the constant. so it'd be like:

`BfdSystemTypeURL = https://bluebutton.cms.gov/fhir/CodeSystem/System-Type`

and then we'd have a different constant for a different system URL like:

`BfdFinalActionURL = https://bluebutton.cms.gov/fhir/CodeSystem/Final-Action`

We _could_ also make constants for the codes. Or constants for the whole tokens (<system_url|code>). But maybe this is creeping the scope too much 🙃

bcda/api/v3/api.go
carlpartridge: Golangci lint seems to prefer the no space + here for whatever reason 😅 (maybe line length?).  I dont think its worth overriding.

bcda/service/warnings_and_info.go
colby-seyferth-nava: Yeah, the _typefilter param should be URL encoded, so agree with Parwinder's comment. 

Also, agree the first `?` doesn't need to be there. if we require `?_typefilter` then that means that _typefilter has to be the first param. Though it could be the 2nd param and it'd be `&_typefilter` (also the `?` before _tag implies that _tag is the first subquery param, but it could be something else [like service-date])

But also I am not sure if regex is the best way to validate this. in request.go, we already have the logic in place to identify when the default is applied. Is there a way to leverage that here? Maybe it'd be to messy to set a flag there and then just check the flag here. It'd have to be threaded thru the job args or context or something...

colby-seyferth-nava: also, because the API allows any random param to be passed, the regex would technically be true if someone sent a valid _typeFilter param for service-date or something, then, outside of the _typeFilter subquery, passed the param `_tag=https://bluebutton.cms.gov/fhir/CodeSystem/System-Type|NationalClaimsHistory`.

like this:

`/v3/Patient/$export?_type=ExplanationOfBenefit&_typeFilter=ExplanationOfBenefit%3Fservice-date%3Dgt2026-01-01&_tag%3Dhttps%3A%2F%2Fbluebutton.cms.gov%2Ffhir%2FCodeSystem%2FSystem-Type%7CNationalClaimsHistory`

this would end up applying the default filtering, because we would ignore _tag. But the OperationOutcome would not show up

carlpartridge: Ok, I think I found a seam with which to re-use existing logic, ty!  (Working on it now)

bcdaworker/queueing/worker_prepare_test.go
```diff
@@ -347,6 +353,14 @@ func (s *PrepareWorkerIntegrationTestSuite) TestPrepareWorkerWork_Integration()
 	assert.Equal(s.T(), result.EventKind, river.EventKindJobCompleted)
 	assert.Equal(s.T(), rivertype.JobStateCompleted, result.Job.State)
 
+	// verify file was created
```
colby-seyferth-nava, line 356: This test case looks like it might just be for the coverage resource?

Once you make the change to use the existing logic, I think it will ensure that it only includes the OperationOutcome when there is ExplanationOfBenefit in the ResourceTypes (which is good). But maybe this test case will fail?

bcda/service/warnings_and_info.go
colby-seyferth-nava: nice! also great since it actually depends on info not in the URL. if the ACO does not have access to SharedSystems claims, then there is no default filtering needed. that may come up as surprising test case behavior too if you arent aware of it!

carlpartridge: If the ACO does not have access to SharedSystems claims, then it doesnt need default filtering, but does it hurt it to have it?  Ie do we actively need to check on this condition or can we ignore it?

colby-seyferth-nava: Yeah good question. Looks like the code does this in two different spots...

First it makes sure if you are requesting SS, then your ACO mush have access to SS data.
Then, it checks to see if you passed any System-Type filters. If not, then only add NCH/DDPS (whether you have SS access or not).

But yeah, when applying the default:
* if you have SS access, then it should let you know that the response only includes NCH and DDPS
* if you dont have SS access, then it could be weird if the OperationOutcome says "hey this only includes NCH and DDPS", since that is what you already expect. However, it's not wrong. So we could return the OperationOutcome in both cases

bcda/api/v3/api.go
colby-seyferth-nava: weeird but agree we should trust the linter! i didn't even want to write the comment because of how nitpicky it was

bcda/service/warnings_and_info.go
carlpartridge: Ok Im going to just add it for now, but let me know if you change on your mind!

bcdaworker/queueing/worker_prepare.go
michaeljvaldes: I don't think we need to figure this out now, but we should definitely be careful since it's not necessarily obvious.

bcda/service/warnings_and_info.go
michaeljvaldes: Yeah, it doesn't get us much in any one instance, but it would at least align all of the places that do the same (tiny) logic.
I don't think it's a big deal either way, totally up to you.

bcda/api/requests.go
```diff
@@ -882,76 +890,19 @@ func (h *Handler) validateTypeFilterPACEligibility(ctx context.Context, typeFilt
 	return nil
 }
 
-// extractTagCodeFromValue extracts tag codes from either a short format (e.g., "SharedSystem")
```
carlpartridge, line 885: Moved these out of requests as I had to use some of these as shared logic and was hitting circular deps.  Moved them into the middleware package which seems like a reasonable place to be manipulating request params and such.  Might be better in a new query URL parsing package?

bcda/api/requests.go
```diff
@@ -882,76 +890,19 @@ func (h *Handler) validateTypeFilterPACEligibility(ctx context.Context, typeFilt
...
-	return systems
-}
-
 // omitSharedSystemByDefault ensures that all ACOs in v3 do not receive SharedSystem data by default
 // by adding a System-Type tag filter if no explicit filter is provided
 func (h *Handler) omitSharedSystemByDefault(ctx context.Context, typeFilter fhir.TypeFilterParameter, cmsID string) fhir.TypeFilterParameter {
-
-	// Check if there's already a _tag parameter that specifies a System-Type value
-	// If one is specified, no need to set a default
-	hasRelevantFilter := false
```
carlpartridge, line 928: Extracted this block out to its own function and moved it to middleware.

bcda/api/requests_test.go
```diff
@@ -1483,31 +1483,7 @@ func TestValidateTypeFilterPACEligibility(t *testing.T) {
 	}
 }
 
-func TestExtractTagCodeFromValue(t *testing.T) {
```
carlpartridge, line 1486: Moved this to middleware test.

bcda/web/middleware/validation.go
```diff
@@ -167,64 +165,54 @@ func validateResourceTypes(r *http.Request, rw fhirResponseWriter, w http.Respon
 	return resourceTypes, true
 }
 
-// validateTypeFilterParameter parses the _typeFilter subquery and validates its contents.
-// For _tag, it validates each comma-separated token to correctly resolve compound query filters.
+// validateTypeFilterParameter validates the contents of the typeFilter param.
```
carlpartridge, line 168: Split this into two functions one to validate and one to get the type filter.  Although it seems like some of the validation of the subquery params is still happening in the Get function.

bcdaworker/queueing/worker_prepare_test.go
```diff
@@ -289,13 +292,12 @@ func (s *PrepareWorkerIntegrationTestSuite) TestPrepareWorkerWork() {
...
 
 func (s *PrepareWorkerIntegrationTestSuite) TestPrepareWorkerWork_Integration() {
-
 	cfg, err := service.LoadConfig()
 	if err != nil {
 		log.API.Fatalf("Failed to load service config. Err: %v", err)
 	}
 
-	svc := service.NewService(s.r, cfg, "/v1/fhir")
+	svc := service.NewService(s.r, cfg, "/v3/fhir")
```
carlpartridge, line 300: Changed this to v3 so that we could test the new warnings file as well.

bcdaworker/queueing/worker_prepare_test.go
carlpartridge: Is checking that typeFilter resource type is ExplanationOfBenefit a requirement?  And for clarification if this is a requirement then we want to include the OpOutcome warning only when EOB and not other resource types?

bcda/web/middleware/validation.go
```diff
@@ -255,30 +236,75 @@ func validateTypeFilterParameter(r *http.Request, rw fhirResponseWriter, w http.
...
+			return typeFilterParam, fmt.Errorf("invalid _typeFilter subquery parameter: %s", paramName)
 		}
 	}
 
 	typeFilterParam = fhir.TypeFilterParameter{ResourceType: resourceType, QueryParameters: typeFilterSubqueryParams}
-	return typeFilterParam, true
+	return typeFilterParam, nil
+}
+
+func HasSharedSystemTag(typeFilter fhir.TypeFilterParameter) bool {
```
carlpartridge, line 252: These three functions were either moved or extracted (and moved) out of larger functions.  `ExtractTagCodeFromValue` has unit-tests while `HasSharedSystemTag` and `ExtractTagSystemFromValue` are pretty thoroughly tested via the `TestOmitSharedSystemByDefault_Integration` integration test.  I can try to break those up a bit if preferred.

bcda/web/middleware/validation.go
```diff
@@ -255,30 +236,75 @@ func validateTypeFilterParameter(r *http.Request, rw fhirResponseWriter, w http.
...
+	hasTag := false
+
+	for _, subqueryParam := range typeFilter.QueryParameters {
+		if subqueryParam.Name == "_tag" {
+			tagSystems := ExtractTagSystemFromValue(subqueryParam.Value)
+			for _, tagSystem := range tagSystems {
+				if tagSystem == constants.BFDSystemTypeURL {
+					hasTag = true
```
colby-seyferth-nava: nit: Niiice, now that this has been modularized, we can probably simplify the logic further and avoid using the `hasTag` flag altogether.

Instead, just return true instead of setting the flag, remove all the breaks, and then return false at the end if we make it through all the loops without returning true?

bcda/web/middleware/validation.go
```diff
@@ -404,8 +430,8 @@ func validateTagSubqueryParameter(tag string) error {
 
 	// Validate that the _tag system and code are supported values
 	validTagTokens := map[string][]string{
-		"https://bluebutton.cms.gov/fhir/CodeSystem/System-Type":  {"SharedSystem", "NationalClaimsHistory", "DDPS"},
-		"https://bluebutton.cms.gov/fhir/CodeSystem/Final-Action": {"FinalAction", "NotFinalAction"},
+		constants.BFDSystemTypeURL:  {"SharedSystem", "NationalClaimsHistory", "DDPS"},
+		constants.BFDFinalActionURL: {"FinalAction", "NotFinalAction"},
```
colby-seyferth-nava, line 428: everything is so much cleaner with these constants. Nice job 👍

bcda/service/warnings_and_info.go
```diff
@@ -0,0 +1,53 @@
...
+	"regexp"
+
+	"github.com/CMSgov/bcda-app/bcda/constants"
+	"github.com/CMSgov/bcda-app/bcda/models/fhir/r4"
+	"github.com/CMSgov/bcda-app/bcda/models/postgres"
+	"github.com/CMSgov/bcda-app/bcdaworker/worker"
+	"github.com/CMSgov/bcda-app/conf"
+)
+
+var HasSystemTypeRegex = regexp.MustCompile(`.*\_typeFilter=.*\?_tag=.*System-Type.*`)
```
colby-seyferth-nava: looks like we can remove this regex now that it's not used?

bcdaworker/queueing/worker_prepare_test.go
colby-seyferth-nava: sorry, to clarify: We only want to return the OperationOutcome when ExplanationOfBenefit is in the ResourceTypes list. 

So if the client requests `/Patient/$export?type=Coverage` then we are not returning any ExplanationOfBenefit resources. So if we said "we are only returning NCH/DDPS EOBs", it would be confusing.


in requests.go, this is where we check if the default filter is needed:
```
// For v3, ensure any ACOs requesting ExplanationOfBenefit default to NCH and DDPS data only
// This ensures they do not get SharedSystem data by default
if h.apiVersion == constants.V3Version {
	if utils.ContainsString(resourceTypes, "ExplanationOfBenefit") {
		rp.TypeFilter = h.omitSharedSystemByDefault(ctx, rp.TypeFilter, ad.CMSID)
	}
}
```

carlpartridge: To check on EOB do we want to compare against the requests ResourceTypes or do we need to grab it out of the typeFilter?  Would the two ever be at odds?
Edit: I see from your example that we get it out of the requests resource types so I will use that.

colby-seyferth-nava (dismissed):

bcda/api/requests.go
bhagatparwinder: The file `warnings-and-info.ndjson` does not contain `-error.ndjson`. Our warning file will be listed twice (once in the `output` and once in the `error`). `warnings-and-info.ndjson` is registered in `job_keys`.

You could exclude it in this loop:
```
for _, jobKey := range jobKeys {
    fileName := strings.TrimSpace(jobKey.FileName)
    if fileName == constants.WarningsAndInfoFileName {
        continue
    }
}
```

bcda/models/postgres/pgx_repository.go
bhagatparwinder: Ok, I know I suggested `ON CONFLICT DO NOTHING`, but after reading the unique constraints requirements, this is how I think it will work. If we do run into a conflict, the affected rows will be 0 because we asked it to do nothing (this is the behavior we want). But the error check below will treat it as a **failure** because the affected rows are 0. I don't think we want that 🫠

I don't know if I have a better solution, because adding the UNIQUE constraint will require some sort of DB migration.

bcda/service/warnings_and_info.go
bhagatparwinder: That's fair. I would still think that the second option to change this function into `AppendToWarningsAndInfoFile` where we encapsulate the file operations (directory creation, file opening, JSON writing, DB job key registration), would be better. Maybe not in the same PR, but in the follow-up ticket.

bcda/api/requests.go
```diff
@@ -882,76 +890,19 @@ func (h *Handler) validateTypeFilterPACEligibility(ctx context.Context, typeFilt
...
+	if middleware.HasSharedSystemTag(typeFilter) {
 		return typeFilter
 	}
 
 	// For all ACOs without a qualifying filter, add System-Type filter
 	// to ensure they do not receive SharedSystem data by default.
 	// This tag filters response data by System-Type=NCH OR System-Type=DDPS
 	// This function is only called when ExplanationOfBenefit is in the resource types
-	tagValue := "https://bluebutton.cms.gov/fhir/CodeSystem/System-Type|NationalClaimsHistory,https://bluebutton.cms.gov/fhir/CodeSystem/System-Type|DDPS"
+	tagValue := constants.BFDSystemTypeURL + "|NationalClaimsHistory," + constants.BFDSystemTypeURL + "|DDPS"
```
laurenkrugen-navapbc, line 905: nit: should `NationalClaimsHistory` and `NationalClaimsHistory` be made into constants?

bcda/service/warnings_and_info.go
```diff
@@ -0,0 +1,51 @@
...
+				Text: "Default System-Type behavior includes only claims from NCH and DDPS in this export.",
+			},
+			Severity: "warning",
+		},
+	},
+}
+
+// SetupWarningsAndInfoFile finds or creates an info file to house generic warnings, issues, etc for a given job.
+// It will also find or create a jobkey.  There should only be 1 warnings and info file per job.
+func SetupWarningsAndInfoFile(ctx context.Context, pgxRepo *postgres.PgxRepository, jobID uint) (string, error) {
```
laurenkrugen-navapbc: Question: it's my understanding that we pass errors into the NDJSON files so we can communicate potentially sensitive information. In this case of the parameter defaulting to an expected value that is _not_ sensitive in nature, does it make sense for us to write that informational/warning to a new file? Or is this something that we could just include in the HTTP response during a job kickoff?

If we did include it in the response during the export request, then folks would have the option to cancel the job and use different parameters if desired, instead of getting that error back when the job completes. I don't think that's the case at the moment, but just a thought!

bcda/web/middleware/validation.go
```diff
@@ -404,8 +424,8 @@ func validateTagSubqueryParameter(tag string) error {
 
 	// Validate that the _tag system and code are supported values
 	validTagTokens := map[string][]string{
-		"https://bluebutton.cms.gov/fhir/CodeSystem/System-Type":  {"SharedSystem", "NationalClaimsHistory", "DDPS"},
-		"https://bluebutton.cms.gov/fhir/CodeSystem/Final-Action": {"FinalAction", "NotFinalAction"},
+		constants.BFDSystemTypeURL:  {"SharedSystem", "NationalClaimsHistory", "DDPS"},
```
laurenkrugen-navapbc, line 427: Similar to my comment above, should these values be constants instead?

bcda/service/warnings_and_info.go
carlpartridge: My thoughts are that what I have now is:
Something, somewhere needs to add to the warnings-and-info file.  Therefore it needs to first set everything properly:
- directory where its stored
- file itself
- singular job_key reference
At this point we can then add whatever it is we need to the file.  The alternative would be that in every place we need to modify this file we need to do all of the steps above since you cant be sure the file already exists.

I can think of two options to adjust:
1.  Have the setup func handle the dir and the job_key but leave the file out.  The file open is find or create anyways so that seems fine.  This seems like a good path?
2. Larger refactor: always create this file on every job start.  Then only present it to the client if at the very end its not empty.  This could do away with the somewhat messy find or creates (file, db record, etc).  But not super clean by itself either.
3. Not try to future proof as much and just do it all in the main calling function for now.  Im not a fan of this as I got the impression from Colby that this file would (likely?) get used more often now that its set up.

bcda/api/requests.go
carlpartridge: Nice catch, fixed.  I still dont like how this section handles the error files (it attempts to find a file from a successful job_key by appending -error to the filename and looking for the file in the payload dir instead of just using the error job_key...).  Im also a bit hesitant to change this as of our last incident 😓.

bcda/api/requests.go
carlpartridge: Could, but I feel like those are options and so a bit less set in stone.  Also I feel like for whatever reason like thats going a bit too far with these.

bcda/service/warnings_and_info.go
carlpartridge: My understanding is that this is just a generally available, re-usable warnings, info, issues, FYIs kind of file, so nothing sensitive necessarily.  I imagine this approach is mostly dictated by FHIR standards but I could be wrong.  Including it up front does make sense but its also likely the case that once they see this issue they will adjust future requests anyways.  This is probably a better question for @colby-seyferth-nava

colby-seyferth-nava: Both are valid ways to handle it as far as the spec goes. We can return an OperationOutcome in the response to the $export or (inclusive) in the response file. There was a very limited discussion in slack:

> Colby Seyferth  [11:32 AM]
> My main uncertainty is: Should we return the error just in the OperationOutcome download file? Or should we return it in the initial $export response body? Today there is no response body unless there is a failure. But it's valid to return an OperationOutcome with a 202 Accepted status.
> 
> The ticket is written with the former approach, but I kind of like the latter more now.
> 
> But i was wondering if anyone had thoughts... Maybe Parwinder or Carl?
> 
> Colby Seyferth  [11:33 AM]
> Concern with the OperationOutcome in the $export response is that some clients may think "Oooh, operation outcome. this must have failed" or something. But they should be looking at the HTTPS status, so maybe I'm being overprotective of bad potential clients
> 
> Parwinder Bhagat  [11:38 AM]
> I'm indifferent to which call we add it to. The only minor concern I have is end users parsing the response. Since we never had OperationOutcome in $export, I don't think anyone will be watching for it (making it less likely to break things).

carlpartridge: I think one advantage to including it in the response file is that it can be a place to put any issues throughout the job process and have it available for the client.

laurenkrugen-navapbc: makes sense, thanks for giving more context as to why we went with a file!

laurenkrugen-navapbc:
should we add this case to our smoke tests either in this ticket, or in a follow up?

carlpartridge:
> should we add this case to our smoke tests either in this ticket, or in a follow up?

I dont think we want to get this granular in our smoke tests.  But I am planning on adding a better integration test for the request endpoint.  Just need to figure out my docker issues first...

bhagatparwinder (dismissed):

bcda/service/warnings_and_info.go
bhagatparwinder: I think option 1 is good, and I believe the scope of your existing ticket has been surpassed. Maybe we can add all feedback (non-critical) to a follow-up ticket for cleanup?

colby-seyferth-nava (dismissed):

michaeljvaldes (approved):

colby-seyferth-nava (approved):

PR #1460, opened 2026-08-05
BCDA-10185: update bene-prefs lambda

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10185

## 🛠 Changes

- bene-prefs:
  - refactored to use CDAP lambda module
  - organized security group and iam tf to separate files
  - restricted security group to limited ingress/egress rules
  - restricted iam roles to least-required permissions
- attribution-import:
  - added a 14-day cleanup rule for old files that failed to be processed in the file bucket
  - fixed an issue in which csv attribution files were not automatically cleaned up after successful processing

## 🧪 Validation

- successful runs of lambda integration tests in the test environment


.github/workflows/bene-prefs-import-integration-test.yml
```diff
@@ -40,7 +41,7 @@ jobs:
       - uses: aws-actions/configure-aws-credentials@ec61189d14ec14c8efccab744f656cffd0e33f37 # v6.1.0
         with:
           aws-region: ${{ vars.AWS_REGION }}
-          role-to-assume: arn:aws:iam::${{ secrets.NON_PROD_ACCOUNT_ID }}:role/delegatedadmin/developer/bcda-test-bene-prefs
+          role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/delegatedadmin/developer/${{ vars.AWS_ROLE_TO_ASSUME }}
```
carlpartridge: Using `secrets.AWS_ACCOUNT_ID` and `vars.AWS_ROLE_TO_ASSUME` here, but `secrets.NON_PROD_ACCOUNT_ID` and `bcda-test-bene-prefs-function` below.
Feels like the reason to use `secrets.AWS_ACCOUNT_ID` is if we need to deploy across prod and non-prod.  If we dont then I think `secrets.NON_PROD_ACCOUNT_ID` might be a bit clearer.  `vars.AWS_ROLE_TO_ASSUME` also feels preferable to the hardcode and works in both scenarios.  Or are these two different roles/use cases and need to be different?

carlpartridge (commented):
Great stuff!

.github/workflows/bene-prefs-import-integration-test.yml
michaeljvaldes: These are two different roles/use cases
`AWS_ROLE_TO_ASSUME` is the dev role with which we can fetch the params in the subsequent step
`bcda-test-bene-prefs-function` is the lambda function role with which we can upload the bene-prefs file and trigger the lambda

I updated `secrets.AWS_ACCOUNT_ID` to `secrets.NON_PROD_ACCOUNT_ID` for clarity, though.

carlpartridge (dismissed):

ops/services/30-bene-prefs/main.tf
```diff
@@ -22,285 +20,51 @@ module "platform" {
...
   cluster_identifier = "${local.app}-${local.env}-aurora"
 }
 
-data "aws_security_groups" "db" {
-  tags = {
-    Name = "bcda-${local.env}-db"
-  }
-}
+// the role generated for this function is depended on by BFD
+// any updates to the role must be coordinated with BFD before deployment
```
laurenkrugen-navapbc, line 32: thanks for the note here!

laurenkrugen-navapbc (dismissed):

ops/services/30-bene-prefs/iam.tf
```diff
@@ -0,0 +1,29 @@
...
+    resources = [module.platform.ssm.bene-prefs.iam_bucket_role_arn.value]
+  }
+}
+
+data "aws_iam_policy_document" "subscribe_bfd_topic" {
+  statement {
+    sid       = "AllowBFDSNSSubscribe"
+    effect    = "Allow"
+    actions   = ["sns:Subscribe"]
+    resources = ["arn:aws:sns:us-east-1:830858426211:bfd-test-bene-prefs-received-s3-bcda"]
```
gfreeman-navapbc: We should hide this account ID somehow, probably ssm

michaeljvaldes: Good catch, fixed!

gfreeman-navapbc (dismissed):

carlpartridge (dismissed):

carlpartridge (approved):

laurenkrugen-navapbc (approved):

PR #1461, opened 2026-08-05
BCDA-9638: smoke test rate limiting

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9638

## 🛠 Changes

- Added `IsTestACO` helper to service config
- Updated `shouldRateLimit` to use `IsTestACO` and skip rate limiting for test ACOs
- Added unit tests

## ℹ️ Context

Because production configuration sets `rate_limit_config.all: true`, the rate-limiting middleware previously enforced concurrent job rate-limiting on all requests, including smoke test ACOs.
These changes bypass duplicate job rate-limiting checks, so smoke test runs without intermittent 429 failures.

## 🧪 Validation

Unit tests

### DIFF
--- a/bcda/service/config.go
+++ b/bcda/service/config.go
@@ -153,6 +153,8 @@ func (cfg *Config) IsSupportedACO(cmsID string) bool {
 	return false
 }
 
+var testACORegex = regexp.MustCompile(`^(TEST\d{3}|A999\d|A888\d|D999\d|E999\d|V994|T8882)$`)
+
 // IsTestACO determines if the specified CMS ID belongs to a test ACO.
 func (cfg *Config) IsTestACO(cmsID string) bool {
 	if cmsID == "" {
@@ -172,7 +174,6 @@ func (cfg *Config) IsTestACO(cmsID string) bool {
 			}
 		}
 	}
-	testACORegex := regexp.MustCompile(`^(TEST\d{3}|A999\d|A888\d|D999\d|E999\d|V994|T8882)$`)
 	return testACORegex.MatchString(cmsID)
 }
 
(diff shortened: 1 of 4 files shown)


bcda/service/config.go
```diff
@@ -152,6 +153,30 @@ func (cfg *Config) IsSupportedACO(cmsID string) bool {
 	return false
 }
 
+var testACORegex = regexp.MustCompile(`^(TEST\d{3}|A999\d|A888\d|D999\d|E999\d|V994|T8882)$`)
```
michaeljvaldes, line 156: Another reference to test ACOs that jumps to mind is in the [refresh_attribution script](https://github.com/CMSgov/bcda-ops/blob/2be95f148b494bf6cf9746e96ae32bf8b1dfa5e7/scripts/refresh_test_attribution.sh#L45). Would it be possible to align these two lists?

michaeljvaldes: Another thought; is it worth just adding everything to the priorityACORegex and/or priorityACOIds list and just using that instead?

michaeljvaldes (approved):

bcda/service/config.go
laurenkrugen-navapbc: +1 - source of truth would be nice!

laurenkrugen-navapbc (approved):

bcda/service/config.go
bhagatparwinder: I am using the SSM param to get the values of PRIORITY_ACO_REG_EX and PRIORITY_ACO_IDS. This is a fallback in case the SSM param is not set.

PR #1465, opened 2026-08-11
BCDA-10337: disable sqs monitors in prod

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10337

## 🛠 Changes

- disable sqs monitors in prod

## ℹ️ Context

SQS monitors in prod were setting off false victorops alerts -- they should be temporarily disabled and evaluated in a lower environment.

## 🧪 Validation

Already applied in prod

### DIFF
--- a/ops/services/40-datadog-monitors/config/prod.yml
+++ b/ops/services/40-datadog-monitors/config/prod.yml
@@ -5,7 +5,7 @@ lambda:
 
 enabled:
   ecs: true
-  sqs: true
+  sqs: false
   sns: false
   lambda: true
   s3: true


carlpartridge (approved):

carlpartridge (commented):
Honestly we could probably turn this off at the default level as well, if we dont want these pinging us in prod then it seems unlikely we will want them pinging us in other envs.

michaeljvaldes:
> Honestly we could probably turn this off at the default level as well, if we dont want these pinging us in prod then it seems unlikely we will want them pinging us in other envs.

That makes sense to me, I'll leave it up for now though so that folks can debug in lower envs. I don't think we're getting on-call alerts for other envs.

bhagatparwinder (approved):

PR #1472, opened 2026-08-17
BCDA-10344: Use most restrictive service date params

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10344

## 🛠 Changes

Fix issue where we can pass multiple upper/lower bound servce-date params to BFD.  Slightly unrelated changes:
- merge package bluebutton with package bluebutton_test (we dont really use this pattern elsewhere in our code base)
- limit passable prefixes for service-date to what BFD allows

## ℹ️ Context

Recent incident exposed an edge case where we could send multiple upper/lower bend service-date params to BFD.  This was failing jobs.  The fixes implemented should address a very large percent of actual use cases.  There likely remain edge cases that this does not fully address but of which dont seem worth spending a lot of time trying to implement.

## 🧪 Validation

Local linting and testing.


bcda/client/bluebutton.go
```diff
@@ -414,6 +416,56 @@ func updateParamWithServiceDate(params *url.Values, claimsWindow ClaimsWindow) {
...
+
+	params.Del("service-date")
+	if len(lowerBoundDates) > 0 {
+		params.Add("service-date", lowerBoundDates[0])
+	}
+	if len(upperBoundDates) > 0 {
+		params.Add("service-date", upperBoundDates[0])
+	}
+}
+
```
carlpartridge, line 528: This is pretty verbose as is.  It should cover 95%+ of client use cases as, at least in my experience, 99% of what clients want is something along the lines of everything on/after August 1st.  There are at least a few edge cases this does not cover and Im sure plenty that I cant even think of ahead of time.  I dont currently think its worth:
- taking a lot of development time to code these out and add tests (much more complex and time consuming)
- pushing out the current fix deployment due to the above

bcda/client/bluebutton_test.go
```diff
@@ -1,4 +1,4 @@
-package client_test
+package client
```
carlpartridge, line 1: 99% of the changes in this file are just merging this test suite into the parent package (eg `client.x() -> x()`).  The actual additions are unit-tests at the end.  We only do this style of testing a few places elsewhere in our code base.  The goal of a separate test package is to treat the main package like an external client would and therefore all testing happens via public interfaces and are therefore more integration like tests.  Interested to hear if folks have strong preferences here.

bcda/web/middleware/validation.go
```diff
@@ -452,7 +452,8 @@ func validateServiceDateSubqueryParameter(dateParam string) error {
 	fhirDateTime := ""
 
 	// Check for the optional 2-character prefix
-	var validPrefixes = []string{"eq", "ne", "lt", "gt", "le", "ge", "sa", "eb", "ap"}
+	// BFD only supports eq, ge, gt, lt, le as of 2026-08-17. See: https://cmsgov.slack.com/archives/CMT1YS2KY/p1786716165942379
+	var validPrefixes = []string{"eq", "lt", "gt", "le", "ge"} // "ne", "sa", "eb", "ap"}
```
carlpartridge, line 491: Limit allowed prefixes based on what BFD supports.  Might make sense to add a separate error message if a client does pass in one of the unsupported?

bcda/client/bluebutton_test.go
```diff
@@ -1014,3 +1036,40 @@ func TestBBTestSuite(t *testing.T) {
...
+		params       url.Values
+		expectedVals []string
+	}{
+		{
+			name:         "No prior service date set",
+			params:       url.Values{},
+			expectedVals: []string(nil),
+		},
+		{
+			name:         "Multiple upper and lower bounds set",
```
bhagatparwinder: What about when there is no prefix (implicit `eq`, which is supported in validation.go below)?

bcda/client/bluebutton.go
```diff
@@ -414,6 +416,56 @@ func updateParamWithServiceDate(params *url.Values, claimsWindow ClaimsWindow) {
...
+	var lowerBoundDates, upperBoundDates []string
+	for _, date := range serviceDates {
+		if strings.HasPrefix(date, "ge") || strings.HasPrefix(date, "gt") {
+			lowerBoundDates = append(lowerBoundDates, date)
+		} else if strings.HasPrefix(date, "le") || strings.HasPrefix(date, "lt") {
+			upperBoundDates = append(upperBoundDates, date)
+		}
+	}
+
+	slices.SortFunc(lowerBoundDates, func(a, b string) int {
```
michaeljvaldes: Can we just extract this to a `sortDates()` function, and call that for both the lower bound and upper bound dates?

bcda/client/bluebutton.go
```diff
@@ -414,6 +416,56 @@ func updateParamWithServiceDate(params *url.Values, claimsWindow ClaimsWindow) {
...
+		bDate, err := time.Parse("2006-01-02", b[2:])
+		if err != nil {
+			return 0
+		}
+		return aDate.Compare(bDate)
+	})
+
+	params.Del("service-date")
+	if len(lowerBoundDates) > 0 {
+		params.Add("service-date", lowerBoundDates[0])
```
michaeljvaldes: Do we need to add the prefix back in so that BFD knows whether to start or end at this date?

bcda/client/bluebutton.go
laurenkrugen-navapbc: What are the edge cases we should be aware of/what are the other 5%? Can we maybe name those and put in the function comment?

carlpartridge: The big one was eq and no-prefix which Im adding support for after discussing with Colby more.  Another one is if there are contradictory statements eg `gt2024` and `lt2022`, which (after discussion) we dont care about (will return no data from BFD), as that is likely user error.

bcda/client/bluebutton.go
carlpartridge: Unfortunately we are sorting asc in one and desc in the other.  We could add in a function param to set the sort order but as we only use this in two places right now Im hesitant to try to build out a customized function.

bcda/client/bluebutton.go
carlpartridge: In the sort func we dont actually manipulate the ~"date" string, it just orders them.  So we should be able to just get the first in the array.

bcda/client/bluebutton.go
```diff
@@ -414,6 +416,56 @@ func updateParamWithServiceDate(params *url.Values, claimsWindow ClaimsWindow) {
...
 
+// BFD only allows for one lower bound and one upper bound service-data param
+// so we need to do some comparisons to make sure we apply the more restrictive option for each.
+func setAppropriateServiceDateParams(params *url.Values) {
+	serviceDates := (*params)["service-date"]
+	if len(serviceDates) <= 0 {
+		return
+	}
+
+	var lowerBoundDates, upperBoundDates []string
```
laurenkrugen-navapbc: we've been making a move to rename upper/lower with a clearer naming like oldest/newest -- can we rename here to make the date bounds easier to understand?

bcda/client/bluebutton.go
bhagatparwinder: I am unsure if we can use `SortFunc` for equality comparison. Like if we run this logic with the same date but different operators (`ge2006-01-02` and `gt2006-01-02`), it will return 0. `gt` is more restrictive than `ge`.

bcda/client/bluebutton.go
```diff
@@ -414,6 +416,56 @@ func updateParamWithServiceDate(params *url.Values, claimsWindow ClaimsWindow) {
...
+	for _, date := range serviceDates {
+		if strings.HasPrefix(date, "ge") || strings.HasPrefix(date, "gt") {
+			lowerBoundDates = append(lowerBoundDates, date)
+		} else if strings.HasPrefix(date, "le") || strings.HasPrefix(date, "lt") {
+			upperBoundDates = append(upperBoundDates, date)
+		}
+	}
+
+	slices.SortFunc(lowerBoundDates, func(a, b string) int {
+		aDate, err := time.Parse("2006-01-02", a[2:])
```
bhagatparwinder: Should this have a length check? `len(date) >= 2`

bcda/client/bluebutton.go
```diff
@@ -414,6 +416,56 @@ func updateParamWithServiceDate(params *url.Values, claimsWindow ClaimsWindow) {
 	}
 }
 
+// BFD only allows for one lower bound and one upper bound service-data param
+// so we need to do some comparisons to make sure we apply the more restrictive option for each.
```
laurenkrugen-navapbc: does restrictive mean using any params set by the app, like lookback and runout?

bcda/web/middleware/validation.go
bhagatparwinder, line 498: We allow all of these formats. Wouldn't `time.Parse("2006-01-02", a[2:])` fail on datetime strings or partial dates?

bcda/client/bluebutton.go
```diff
@@ -414,6 +416,56 @@ func updateParamWithServiceDate(params *url.Values, claimsWindow ClaimsWindow) {
...
+			upperBoundDates = append(upperBoundDates, date)
+		}
+	}
+
+	slices.SortFunc(lowerBoundDates, func(a, b string) int {
+		aDate, err := time.Parse("2006-01-02", a[2:])
+		if err != nil {
+			return 0
+		}
+		bDate, err := time.Parse("2006-01-02", b[2:])
```
bhagatparwinder: Nit but this time.Parse parses date strings on every comparison. It would be better as a struct or helper.

bcda/client/bluebutton_test.go
bhagatparwinder: I think we need to test the following cases:

- Identical dates with differing operators (["ge2022-01-01", "gt2022-01-01"] -> expecting ["gt2022-01-01"]).
- eq prefix or prefix-less dates.
- ISO datetime strings (gt2022-01-01T00:00:00Z).
- Malformed/short inputs.

bcda/client/bluebutton.go
laurenkrugen-navapbc: I believe it will do comparison, but it sounds like the comparison may or may not preserve order on equality, which would matter here if the dates were like the ones in your comment above.

bcda/client/bluebutton.go
carlpartridge: Ooof I think I have it pretty clear in my head now and would rather not mix that up 😅.  I think last time we talked about this your notion of oldest was not my notion as well so Im not sure that would clarify things for future us either 😅.

bcda/client/bluebutton.go
carlpartridge: Restrictive applies to the dates that we are ultimately passing to BFD in order to get data in specific time windows.  So most restrictive would be smallest window.  If we have multiple lower bound service-dates (eg "gt2022", "gt2023", and "gt2024") then we would use the closest one to present time (ie "gt2024"), thus being the most restrictive in the information we would be returning (smallest window).

bcda/client/bluebutton.go
michaeljvaldes: We could just grab the first or last in the slice depending on the case

bcda/client/bluebutton.go
michaeljvaldes: Earliest/latest?

bcda/client/bluebutton.go
michaeljvaldes: We could potentially convert them all to "equality dates" by adding or subtracting a day, for the purposes of comparison. This is getting a little gnarly, but I think the requirements are inherently a little gnarly.

bcda/client/bluebutton_test.go
carlpartridge: I think my latest commit addresses most of these concerns.  I think adding logic to handle "ge2022-01-01" + "gt2022-01-01" is a bit too edge casey for me, at least for now.  But if folks disagree Im happy to add it.  This could be a good candidate for a comment re: known edge cases.

bcda/web/middleware/validation.go
carlpartridge: It sure does 😓, see modifications made in my latest commit.

bcda/client/bluebutton.go
laurenkrugen-navapbc: I would be down for earliest/latest or oldest/newest. I think because we both had a different interpretation of those the last time we did a rename and it still feels really ambiguous, I would be in favor of a rename.

bcda/client/bluebutton.go
laurenkrugen-navapbc: gotcha - could we add that clarifier to this func name? (sorry for all the requests to add clarifiers; dates are confusing and there's multiple things that could be restricted, so having that documented will hopefully make life easier in the future!)

bcda/client/bluebutton.go
carlpartridge: I see earliest/latest elsewhere in the codebase.  When I think of these:
- earliest = lowerbound, the beginning time period for our window.  These are represented by "gt"/"ge".
- latest = upperbound, the end of our time period window.  These are represented by "lt"/"le".
Is that how other folks read it?

bcda/client/bluebutton.go
```diff
@@ -414,6 +416,98 @@ func updateParamWithServiceDate(params *url.Values, claimsWindow ClaimsWindow) {
...
+	var earliestDates, latestDates []serviceDateVal
+	for _, date := range serviceDates {
+		if strings.HasPrefix(date, "ge") || strings.HasPrefix(date, "gt") {
+			earliestDates = append(earliestDates, serviceDateVal{prefix: date[:2], date: parseDate(date[2:])})
+		} else if strings.HasPrefix(date, "le") || strings.HasPrefix(date, "lt") {
+			latestDates = append(latestDates, serviceDateVal{prefix: date[:2], date: parseDate(date[2:])})
+		} else if strings.HasPrefix(date, "eq") {
+			earliestDates = append(earliestDates, serviceDateVal{prefix: "ge", date: parseDate(date[2:])})
+			latestDates = append(latestDates, serviceDateVal{prefix: "lt", date: parseLatestDateFromEqualPrefix(date[2:])})
+		} else if strings.HasPrefix(date, "20") {
```
bhagatparwinder, line 445:
```suggestion
		} else if parsedDate := parseDate(date); !parsedDate.IsZero() {
```

Do you think that if we do a non-zero check on the parsed date, that would be better?

bcda/client/bluebutton.go
```diff
@@ -414,6 +416,98 @@ func updateParamWithServiceDate(params *url.Values, claimsWindow ClaimsWindow) {
...
+		}
+	}
+
+	// remove any zero dates to cleanse list of invalids
+	earliestDates = slices.DeleteFunc(earliestDates, func(s serviceDateVal) bool { return s.date.IsZero() })
+	latestDates = slices.DeleteFunc(latestDates, func(s serviceDateVal) bool { return s.date.IsZero() })
+
+	// sort lower bounds in descending, and upper bounds in ascending order so we can pick the most restrictive for each
+	slices.SortFunc(earliestDates, func(a, b serviceDateVal) int {
+		return b.date.Compare(a.date)
```
bhagatparwinder:
```suggestion
		if c := b.date.Compare(a.date); c != 0 {
          return c
        }
        if a.prefix == "gt" && b.prefix == "ge" { return -1 }
        if a.prefix == "ge" && b.prefix == "gt" { return 1 }
        return 0
```

I think if we do this (and the same for `lt` > `le`), it should™ solve the edge case. Definitely an edge case, so I will leave it to you.

bcda/client/bluebutton.go
```diff
@@ -414,6 +416,98 @@ func updateParamWithServiceDate(params *url.Values, claimsWindow ClaimsWindow) {
...
+		return parsedDate.AddDate(0, 1, 0)
+	}
+
+	parsedDate, err = time.Parse("2006", date)
+	if err == nil {
+		return parsedDate.AddDate(1, 0, 0)
+	}
+
+	return time.Time{}
+}
```
bhagatparwinder, line 527:
```suggestion
func parseDate(date string) time.Time {
	formats := []string{
		time.RFC3339,
		"2006-01-02T15:04:05",
		"2006-01-02",
		"2006-01",
		"2006",
	}

	for _, format := range formats {
		if parsedDate, err := time.Parse(format, date); err == nil {
			return parsedDate
		}
	}

	return time.Time{}
}
```

We can iterate over slices of time that we support. I have gone from most specific to least specific.

bcda/client/bluebutton_test.go
```diff
@@ -1014,3 +1036,211 @@ func TestBBTestSuite(t *testing.T) {
...
+		},
+		{
+			name: "Multiple eq set",
+			params: url.Values{"service-date": []string{
+				"eq2022-01-01",
+				"eq2023-02",
+				"eq2024",
+				"bad time",
+			}},
+			expectedVals: []string{"ge2024-01-01", "lt2022-01-02"},
```
bhagatparwinder, line 1185: Isn't this in reverse, where lower bound > upper bound?

bcda/client/bluebutton.go
carlpartridge: Hmm yeah I was going to mention that checking for 20 is a bit lazy (but will work for ~75 years 😅).  We do the actual date parsing and discarding of bad dates later on so it doesnt seem necessary here, and I think would need to refactor things a bit more as well.  Happy to dig into it if you feel strongly about it.

bcda/client/bluebutton.go
carlpartridge: I have applied that for parseDate but left it off for the latest version as that one gets a bit more complex.  I have also left off the hour, min, sec, tz and such checks as those are unnecessary (as per Colby).

bcda/client/bluebutton_test.go
carlpartridge: The eq ones are pretty weird.  We are taking each one and splitting it into an earlier and a later so:
- "eq2022-01-01" becomes:
- - "ge2022-01-01" and "lt2022-01-02"
- "eq2023-02":
- - "ge2023-02-01" and "lt2023-03-01"
- "eq2024":
- - "ge2024-01-01" and "lt2025-01-01"

Then they get sorted by their type.  Of all the earliest bounds the most restrictive is "ge2024-01-01".  Of all the latest bounds the most restrictive is "lt2022-01-02".  In this test case it results in a window that would find zero data.  Which (as per Colby) is an ok outcome as it likely means client error (ie passing in multiple at odds service-date params).

I guess you could argue that if you have multiple "eq" prefixes you would want to pick one first, then get the earliest and latest from that?  The issue with that is that how do we pick which one?  Also it should be very unlikely that we get multiple eq service dates in general.

colby-seyferth-nava: Yeah, regarding "How do you pick one?", i agree that it is an issue when multiple `eq` are passed in. If a client tries to pass multiple `eq`, then we will return an error (this will be updated in the upstream validation logic in a follow-up ticket). 

I think as far as this ticket goes, it will just need to prevent the scenario where we (BCDA) add an extra service-date to an existing service-date range that the client passes in.

But yeah, definitely passing in two `eq` will fail (appropriately)upstream, so i don't think we will run into some of these scenarios, but glad it won't explode in case we do

bcda/client/bluebutton.go
colby-seyferth-nava: I'm ok with the current approach. We can add reevaluating this to the follow-up scope in https://jira.cms.gov/browse/BCDA-10352 if we are concerned?

bcda/client/bluebutton.go
```diff
@@ -414,6 +416,116 @@ func updateParamWithServiceDate(params *url.Values, claimsWindow ClaimsWindow) {
...
+	slices.SortFunc(latestDates, func(a, b serviceDateVal) int {
+		if c := a.date.Compare(b.date); c != 0 {
+			return c
+		}
+
+		if a.prefix == "gt" {
+			return 1
+		}
+		if b.prefix == "gt" {
+			return -1
```
colby-seyferth-nava: I think these need to be `le` and `lt`

colby-seyferth-nava: or both `le`* ..? trying to wrap my head around the logic 😅

carlpartridge: We want to prefer "lt" as its the more restrictive of the two, and I dont think we care which one we use since the date is the same.  Im not 100% on the logic either but I think the test cases cover it thoroughly enough?

colby-seyferth-nava: Nice, thanks for the clarification!

colby-seyferth-nava (approved):

laurenkrugen-navapbc (approved):

bhagatparwinder (approved):

PR #1473, opened 2026-08-19
BCDA-10309: datadog dashboard and alert adjustments

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10309

## 🛠 Changes

- update health check monitor tick rate from 5 min > 3 min
- update APM tags to match pattern specified by CDAP (e.g. application:bcda, service:worker)

## ℹ️ Context

For the health check, the current configuration is tickrate: 5 min, failure time: 10 min. This _seems_ to trigger relatively easily, with alerts (which sent on-call pages) off-and-on every once in a while that are immediately resolved.
There are a few ways to approach solving this:
- adding a retry function: it seems this is not currently supported by the CDAP synthetics module
- increasing the failure time: it's already set to a pretty conservative 10 minutes
- increasing the max response time: the max response time was the cause of many of the intermittent "failures" but 1s also seems reasonable for this check
- increasing the number of ticks: the chose solution, increasing from every 5 minutes to every 3 minutes; the intent here is that, like a retry function, we'll moderately increase the number of checks, filter out false failures, increasing the number of "failures" before alert without increasing the time required to trigger an alert 

## 🧪 Validation

Validated APM changes in datadog dashboard

<details>
<summary>tf plan in prod</summary>

```

OpenTofu will perform the following actions:

  # module.datadog_synthetics[0].datadog_synthetics_test.this["Health Check"] will be updated in-place
  ~ resource "datadog_synthetics_test" "this" {
        id               = "f82-hvy-rbz"
        name             = "bcda-prod-Health Check"
        tags             = [
            "application:bcda",
            "environment:prod",
            "managed-by:tofu",
            "shadow-mode:false",
            "team:bcda",
            "service:bcda",
        ]
        # (10 unchanged attributes hidden)

      ~ options_list {
          ~ tick_every                        = 300 -> 180
            # (19 unchanged attributes hidden)
        }

        # (6 unchanged blocks hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.

```

</details>

### DIFF
--- a/bcda/bcdacli/cli.go
+++ b/bcda/bcdacli/cli.go
@@ -97,7 +97,8 @@ func setUpApp() *cli.App {
 				fmt.Fprintf(app.Writer, "%s\n", "Starting bcda...")
 
 				err := profiler.Start(
-					profiler.WithService("BCDA"),
+					profiler.WithTags("application:bcda"),
+					profiler.WithService("api"),
 					profiler.WithEnv(os.Getenv("ENV")),
 					profiler.WithVersion(constants.Version),
 				)
--- a/bcdaworker/cli/cli.go
+++ b/bcdaworker/cli/cli.go
@@ -76,7 +76,8 @@ func startWorker() {
 	logger.Info("Starting bcdaworker...")
 
 	err := profiler.Start(
-		profiler.WithService("BCDA"),
+		profiler.WithTags("application:bcda"),
+		profiler.WithService("worker"),
 		profiler.WithEnv(os.Getenv("ENV")),
 		profiler.WithVersion(constants.Version),
 	)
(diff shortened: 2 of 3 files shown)


laurenkrugen-navapbc (approved):

bcda/bcdacli/cli.go
```diff
@@ -97,7 +97,8 @@ func setUpApp() *cli.App {
 				fmt.Fprintf(app.Writer, "%s\n", "Starting bcda...")
 
 				err := profiler.Start(
-					profiler.WithService("BCDA"),
+					profiler.WithTags("application:bcda"),
+					profiler.WithService("api"),
```
carlpartridge, line 101: There was discussion on calling this `bcda-api` to distinguish services apart from other APIs.  Was that decided to be done further down the line?  Would it hurt to do it now?

michaeljvaldes: My understanding is that we decided (at least for now) to keep the naming "api" and make the distinction on the dashboard end, based on the application tag.

carlpartridge (approved):

PR #1480, opened 2026-08-27
BCDA-10182: migrate scripts to app repo

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10182:

## 🛠 Changes

- scripts from ops repo migrated to bcda-app `scripts/` folder.
- workflows updated to point to `bcda-ops/scripts` instead of `bcda-ops/scripts`.
- .gitignore updated for terrascan report

## ℹ️ Context

Part of the effort to deprecate the ops repo and move everything to bcda-app. Companion PR: https://github.com/CMSgov/bcda-ops/pull/1358

## 🧪 Validation

TBD - testing three workflows that now reference the scripts directory in the bcda-app repo instead of bcda-ops.


.github/workflows/tag-and-release.yml
bhagatparwinder: Let's remove values for `RELEASE_GITHUB_USER` and `RELEASE_GITHUB_EMAIL` or use dummy values?

bhagatparwinder: and probably this too?

.github/workflows/tag-and-release.yml
```diff
@@ -61,23 +61,11 @@ jobs:
         run: |
           echo "${{ env.BCDA_GPG_PUB_KEY }}" > ./${{ env.GPG_KEY_FILE_NAME }}
           GITHUB_REPO_PATH='/CMSgov/bcda-ops' GITHUB_ACCESS_TOKEN=${{ env.OPS_GITHUB_TOKEN }} GITHUB_USER='${{ env.RELEASE_GITHUB_USER }}' GITHUB_EMAIL='${{ env.RELEASE_GITHUB_EMAIL }}' GITHUB_GPG_KEY_FILE=${{ env.GPG_KEY_FILE_NAME }} bash scripts/tag_and_release/release.sh ${{ env.MANUAL_TAG_ARGS }}
-      - name: Upload release scripts
```
michaeljvaldes, line 64: These scripts are needed when we release bcda-ops in the previous step, but since these scripts have been deleted from bcda-ops we actually need to check out bcda-app and upload them first.

laurenkrugen-navapbc: Agreed! We checkout BCDA-app on line 71 and run the scripts on line 84 -- did I miss something there?

.github/workflows/tag-and-release.yml
laurenkrugen-navapbc: good catch - I saw these but thought they were elsewhere in the the repo workflows. We should also make this a service account. Will remove!

.github/workflows/tag-and-release.yml
michaeljvaldes: I think so -- we still need to use these scripts on line 63 (for the release of bcda-ops), but those files won't exist in that repo anymore.

laurenkrugen-navapbc: I see - I will update!

.github/workflows/refresh_test_attribution.yml
```diff
@@ -41,10 +41,10 @@ jobs:
       - name: Install psql
         run: |
           sudo dnf install postgresql16 -y
-      - name: checkout bcda-ops
+      - name: checkout bcda-app
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.3.0
         with:
-          repository: CMSgov/bcda-ops
+          repository: CMSgov/bcda-app
```
carlpartridge: Since we call this from a bcda-app workflow, I dont think we need to specify this?

scripts/tag_and_release/github_release.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+ 
+    parser.add_argument(
+        '--repo', dest='repo', type=verify_repo,
+        help='The repository of the release (i.e., /repos/CMSgov/bcda-app/releases)'
+    )
+
+    args = parser.parse_args()
+
+    main(args.release, args.release_file, args.repo)
```
carlpartridge, line 79: These two tag and release scripts seem a bit excessive in the sense that there is a working dpc workflow that does this.  These two scripts also dont even fully work right (they dont get the full git history properly).  Is it worth switching over to the dpc workflow?  Seems somewhat preferable?

scripts/refresh_test_attribution.sh
```diff
@@ -0,0 +1,82 @@
...
+if [ -z "$conn" ]; then
+  echo "Error: Missing required options." >&2
+  usage
+fi
+
+# Remove parsed options from the argument list
+shift "$((OPTIND - 1))"
+
+# list of test acos (same for all envs)
+ACOS=("A9990" "A9991" "A9992" "A9993" "A9994" "A9998" "A9999" "TEST993" "TEST994" "SBXPS002" "SBXPL002")
```
carlpartridge, line 45: Is there any chance that these are sensitive?  Or the DB queries below?  I think we have gone over this  numerous times from numerous people but just figured one last before we merged would be good.

scripts/tag_and_release/github_release.py
laurenkrugen-navapbc: I would be up for switching over to the DPC workflow; I will put that in a separate ticket/pr since that's outside the scope of this ticket.

scripts/refresh_test_attribution.sh
laurenkrugen-navapbc: I think these are fine to leave in!

.github/workflows/refresh_test_attribution.yml
laurenkrugen-navapbc: yep, I don't think we need to check out the repo that we are calling from, unless we've check out another repo.

bhagatparwinder (dismissed):

.github/workflows/refresh_test_attribution.yml
```diff
@@ -41,12 +41,6 @@ jobs:
       - name: Install psql
         run: |
           sudo dnf install postgresql16 -y
-      - name: checkout bcda-ops
```
michaeljvaldes, line 44: Do we need to check out bcda-app here instead?

laurenkrugen-navapbc: since the workflow is running from bcda-app, the checkout shouldn't be necessary! 

https://github.com/CMSgov/bcda-app/pull/1480#discussion_r3905937651

michaeljvaldes: I think we still need a checkout step since we need to access a file in the repository, but we don't need to specify anything more, e.g.
```
      - name: Checkout BCDA
        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.3.0
```

bhagatparwinder (dismissed):

carlpartridge (approved):

bhagatparwinder (approved):

PR #1481, opened 2026-08-27
BCDA-10167: refactor attr import lambda

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10167

## 🛠 Changes

- DI changes: Refactored attr import lambda to use `AttributionImportHandler` struct with injected dependencies
- Optimized cold start by extracting one-time initialization logic into a separate function (`initHandler`)
- Declared interfaces for CSV and CCLF importer to decouple the lambda from concrete importer structs
- SQ cleanup by removing duplicate test-only helper functions and removing dead code
- Enables S3 file processor tests using plain mocks (no localstack/ministack)
- Created a standalone `bcda/filehandler` package to eliminate cross-package dependencies between attribution import and beneficiary preferences.

## ℹ️ Context

This is phase 2 of the attr import refactor. We are not changing the logic of the attr import lambda in this phase. We are cleaning up the code to have better dependency injection and testability. Unit tests run without external dependencies. We also enable S3 file processor tests using plain mocks.

## 🧪 Validation

Executed lambda unit tests.


bcda/filehandler/s3_file_handler.go
laurenkrugen-navapbc, line 1: this looks much cleaner, thanks for the updates!

any benefit to having this in its own package vs putting it in the broader `attribution-import` package? I like packages that are based on their own responsibility, but I know BCDA has had some issues with circular dependencies due to granularity in the past. 

I don't know that there's necessarily any benefit of moving it under `attribution-import`, though, since this functionality won't ever be used by other packages.

laurenkrugen-navapbc: disregard my comment, I see the separation of this from bene-prefs! this is much cleaner in its own package.

laurenkrugen-navapbc:
this is really clean - lots of improvements on logging, error handling, and packages!

bcda/attribution-import/utils/cclfUtils.go
```diff
@@ -138,18 +135,8 @@ func ImportCCLFPackage(db *sql.DB, pgxPool *pgxv5Pool.Pool, acoSize, environment
 
 	_ = zipWriter.Close()
 
-	hours, err := safecast.ToUint(utils.GetEnvInt("FILE_ARCHIVE_THRESHOLD_HR", 72))
```
laurenkrugen-navapbc, line 141: This is needed for our file cleanup and I don't see it elsewhere -- any concerns cleaning up old files with this removed?

bhagatparwinder: For CCLF and CSV cleanup, the methods `checkDeliveryDate`, `CleanUpCCLF`, and `CleanUpCSV` directly look up the threshold in the environment variable (see line 297 for example). We were never using the `FileArchiveThresholdHr` variable.

laurenkrugen-navapbc (approved):

bcda/lambda/attribution-import/main.go
```diff
@@ -25,97 +25,139 @@ import (
...
 
 	err = os.Setenv("DATABASE_URL", dbURL)
 	if err != nil {
 		logger.Errorf("error setting dbURL env var: %+v", err)
-		return "", err
+		return nil, err
 	}
 
-	// Create pgx pool for bulk operations
 	pool := database.ConnectPool()
```
michaeljvaldes, line 85: If we're creating the database connection outside of the handler, do we need to be concerned about the connection dropping? I'm looking at the "use a keep-alive directive to maintain persistent connections" section of the [best practices docs](https://docs.aws.amazon.com/lambda/latest/dg/golang-handler.html#go-best-practices).
If this is a concern, in the case of the attr-import, the lambda runs in spurts of like 500 every few weeks. It might make sense to maintain a persistent connection through those 500 runs, and maybe double-check that it's still live in the handler and recreate-it if it's closed (for the first run in a block).

michaeljvaldes (commented):
Sweet -- the repackaging of s3_file_handler and moving lambda setup outside of the handler function are super slick.
I just left one comment about the latter.

bcda/lambda/attribution-import/main.go
bhagatparwinder: Totally valid concern, but I'll explain why it doesn't apply here:
- AWS itself recommends creating the database connection outside of the function handler multiple times. See [Function code](https://docs.aws.amazon.com/lambda/latest/dg/best-practices.html#function-code), [code best practices for Go lambda functions](https://docs.aws.amazon.com/lambda/latest/dg/golang-handler.html#go-best-practices), and [an example under lambda function accessing RDS](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-lambda-tutorial.html#vpc-rds-create-deployment-package).
- `pgxpool.Pool` handles connection dropping automatically. If a warm lambda function drops connections (a timeout happened or our NAT gateway timed out), it will establish a new connection. See the implementation of `Acquire`, `Exec`, and `Query` [here](https://github.com/jackc/pgx/blob/master/pgxpool/pool.go#L615). Its clean code and easy to read!
- Coming to Lambda itself. AWS doesn't keep Lambda available for more than 5-15 minutes before it is destroyed. If the lambda is warm, AWS reuses the initialized `AttributionImportHandler`. So this handles the burst of 500 runs. If a new batch arrives after some time, we create a whole new execution environment (cold start), so we also create a new pool.
- The document you linked (and I have used it in my first bullet point) mentions
> Initialize SDK clients and database connections outside of the function handler, and cache static assets locally in the /tmp directory.

The part where it says

>use a keep-alive directive to maintain persistent connections

is meant for HTTP/HTTPS clients to reuse TCP sessions for REST requests across warm invocations. Check the example it links to.

michaeljvaldes: Gotcha, thanks for the explanation!

michaeljvaldes (approved):

PR #1483, opened 2026-09-02
BCDA-10317: Use correct lastUpdated for v3

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10317

## 🛠 Changes

- Updated `GetBundleLastUpdated()` to query BFD and retrieve `Bundle.meta.LastUpdated` for v3 export jobs, matching v1 and v2 behavior.
- Added a safeguard fallback in `GetBundleLastUpdated()`.
  - If BFD returns an epoch (`1970-01-01`), zero, or pre-2000 timestamp (as currently seen with synthetic data in lower environments), it falls back to the job's `TransactionTime` (`time.Now()`) and logs a warning.
- Improved error handling for unknown/unsupported BFD base paths
- Added unit tests to validate all BFD paths, fallback conditions (epoch, zero, nil bundle), error cases, and unsupported paths.

## ℹ️ Context

When processing Bulk FHIR export requests, BCDA sets `Job.TransactionTime` as the `_lastUpdated` upper bound for BFD resource queries and returns it to API consumers in the export completion status. API consumers use this `transactionTime` as the `_since` lower bound in subsequent exports.

Querying `GET /Patient?_id=0` returns `Bundle.meta.LastUpdated`, which reflects the exact timestamp when the most recent ETL data load completed. Using this value avoids timestamp skew and prevents data gaps between exports.

BFD created a ticket to update the lower environment default to `NOW()`, the safeguard fallback ensures lower environments and sandbox synthetic data exports continue to function without interruption.

## 🧪 Validation

Successful unit test execution.


bcdaworker/queueing/worker_prepare.go
```diff
@@ -194,18 +195,34 @@ func (p *PrepareJobWorker) prepareExportJobs(ctx context.Context, args worker_ty
...
-		return time.Time{}, errors.New("no BFD base path")
+		return time.Time{}, fmt.Errorf("unsupported BFD base path: %s", basepath)
+	}
+
+	if err != nil {
+		return time.Time{}, err
 	}
+
+	// Safeguard: If BFD lower environments return 1970/epoch or an unpopulated timestamp, fallback to request transaction time. minValidYear guards against epoch (1970) or bogus default dates from mock/lower-environment BFD servers.
+	const minValidYear = 2000
```
carlpartridge, line 219: Did BFD come up with this year?  Wondering if we should pass it by product as well.

carlpartridge (commented):
This was dependent on a BFD change right?  Has that been released?  Do we want to have a buffer period of time before we deploy this just in the unlikely chance that BFD has to rollback or anything like that?

bhagatparwinder:
> This was dependent on a BFD change right? Has that been released? Do we want to have a buffer period of time before we deploy this just in the unlikely chance that BFD has to rollback or anything like that?

That is what the safeguard code is for. Also, the safeguard only applies to lower environments. Production does give us the correct lastUpdated for v3.

Please see this thread for additional details: https://cmsgov.slack.com/archives/CMT1YS2KY/p1788197054432209

bcdaworker/queueing/worker_prepare.go
bhagatparwinder: This piece of code is a safeguard for lower environments if BFD ever starts returning 1970/epoch (again). I can remove this piece of code if preferred or if it is confusing. This was done because:
- BFD is still making this change for lower environments.
- In case in the future BFD goes back to its previous behavior, we use `TransactionTime` as the lastUpdated.

carlpartridge (approved):

laurenkrugen-navapbc (approved):

colby-seyferth-nava (approved):

PR #1485, opened 2026-09-02
BCDA-10182: update refresh attribution and tag/release workflows

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10182

## 🛠 Changes

- reverted changes in tag/release python script
- dump errors to log file instead of suppressing (refresh_attribution), for easier troubleshooting

## ℹ️ Context

Follow up to https://github.com/CMSgov/bcda-app/pull/1480/changes; unable to test those changes until they were merged. After merge and test, needed to make some changes for the workflows to run successfully!

## 🧪 Validation

Ran and tested both workflows, verified successful runs.

### DIFF
--- a/.github/workflows/ci-checks.yml
+++ b/.github/workflows/ci-checks.yml
@@ -127,6 +127,7 @@ jobs:
             -Dsonar.sources=.
             -Dsonar.go.coverage.reportPaths=testcoverage.out
             -Dsonar.coverage.exclusions=**/*test.go,**/test/**/*,**/testUtils/*,**/scripts/**/*,**/ops/*,**/mock*.go,**/mock/**/*
+            -Dsonar.exclusions=**/scripts/**/*
             -Dsonar.branch.name=${{ env.REF }}
             -Dsonar.projectVersion=${{ github.sha }}
             -Dsonar.qualitygate.wait=true
--- a/.github/workflows/ci-checks.yml
+++ b/.github/workflows/ci-checks.yml
@@ -127,7 +127,7 @@ jobs:
             -Dsonar.sources=.
             -Dsonar.go.coverage.reportPaths=testcoverage.out
             -Dsonar.coverage.exclusions=**/*test.go,**/test/**/*,**/testUtils/*,**/scripts/**/*,**/ops/*,**/mock*.go,**/mock/**/*
-            -Dsonar.exclusions=**/scripts/**/*
+            -Dsonar.exclusions=/scripts/**/*
             -Dsonar.branch.name=${{ env.REF }}
             -Dsonar.projectVersion=${{ github.sha }}
             -Dsonar.qualitygate.wait=true
--- a/scripts/tag_and_release/github_release.py
+++ b/scripts/tag_and_release/github_release.py
@@ -6,18 +6,10 @@ import urllib.request
 from urllib.parse import urlparse
 
 
-def safe_path(path):
-    resolved = os.path.realpath(path)
-    base_dir = os.path.realpath(os.getcwd())
-    if resolved != base_dir and not resolved.startswith(base_dir + os.sep):
-        raise ValueError(f"path {path!r} is outside the allowed directory")
-    return resolved
-
 def main(release, release_file, repo):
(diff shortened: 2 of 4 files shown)


.github/workflows/ci-checks.yml
```diff
@@ -127,6 +127,7 @@ jobs:
             -Dsonar.sources=.
             -Dsonar.go.coverage.reportPaths=testcoverage.out
             -Dsonar.coverage.exclusions=**/*test.go,**/test/**/*,**/testUtils/*,**/scripts/**/*,**/ops/*,**/mock*.go,**/mock/**/*
+            -Dsonar.exclusions=**/scripts/**/*
```
carlpartridge: This is going to ignore all folders everywhere with scripts in the path.  I dont think this affects currently but maybe we want to be more specific?  (ie `/scripts/**/*`)

scripts/tag_and_release/github_release.py
```diff
@@ -3,21 +3,12 @@
...
-from urllib.parse import urlparse
 
 
-def safe_path(path):
-    resolved = os.path.realpath(path)
-    base_dir = os.path.realpath(os.getcwd())
-    if resolved != base_dir and not resolved.startswith(base_dir + os.sep):
-        raise ValueError(f"path {path!r} is outside the allowed directory")
-    return resolved
-
```
carlpartridge, line 15: Does reverting these changes open us up to security issues?  Looks like sonarqube passed but thats due to skipping the folder, is snyk checking these?

carlpartridge (dismissed):

scripts/tag_and_release/github_release.py
```diff
@@ -3,21 +3,12 @@
...
 def main(release, release_file, repo):
     access_token = os.environ['GITHUB_ACCESS_TOKEN']
 
-    resp = None
-    with open(safe_path(release_file), 'r') as f:
```
bhagatparwinder, line 20: Any reason we removed `safe_path`? Isn't this supposed to prevent path traversal attacks?

scripts/tag_and_release/github_release.py
```diff
@@ -26,41 +17,32 @@ def main(release, release_file, repo):
...
+        resp = urllib.request.urlopen(req)
 
-    if not resp or resp.status != 201:
+    if resp.status != 201:
         print("Could not create release: %s" % release)
         sys.exit(1)
     else:
         print("Successfully created release: %s" % release)
 
-def verify_repo(repo):
```
bhagatparwinder, line 50: Same question for `verify_repo` and `verify_release` below. Wouldn't this save us from typos?

scripts/tag_and_release/github_release.py
```diff
@@ -26,41 +17,32 @@ def main(release, release_file, repo):
...
         }
 
-        url = "https://api.github.com" + repo
+        base_url = "https://api.github.com"
+        path = repo
         headers = {
             "Authorization": "Bearer %s" % access_token
         }
 
-        if urlparse(url).hostname in 'https://api.github.com':
```
bhagatparwinder, line 34: Same question here, isn't this validating the URL for us? I think I might be missing some context.

.github/workflows/ci-checks.yml
laurenkrugen-navapbc: I'm fine to leave it, unless anyone feels strongly? we generally keep all of our scripts in the `scripts` folder, and I would assume that any additional scripts folders would also be excluded from sonar.

michaeljvaldes: I lean towards changing it -- I think if someone ever adds a new scripts directory, they should have to intentionally specify exclusions like this. It feels like .gitignore to me, where additions should be pretty deliberate.
But not a huge deal either way.

laurenkrugen-navapbc:
All of the values passed in an validated are generated by a script that we run -- after discussing with the team, there is not a security concern with removing some of these validation steps.

carlpartridge (approved):

michaeljvaldes (approved):

PR #1486, opened 2026-09-03
BCDA-10345: Core lambda refactors and cleanup

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10345

## 🛠 Changes

- Split s3_fileprocessor logic into its attr-import specific code (and moved to cclf.go) and into its s3 specific code (and moved to aws package)
- Simplified some interfaces and removed others.
- Removed local file processor as it is extraneous code that is not used in any actual env outside of local testing.
- Moved s3 file handler into bcda/aws package.  Merged the two versions (attr-import and bene-prefs had slight differences between their functions).
- Added an init handler to bene-prefs lambda (mimics attr-import and best practices).
- Converted some functions and data refs to unexported (private, lowercase starting letter).
- Remove references to pkg/errors in favor of built in golang version.
- Remove some unused cli functionality and related code.
- Temporarily comment out tests that need ministack implementation.

## ℹ️ Context

Refactoring and cleanup of core lambdas.  Reduce complexity and cruft, increase readability, and make future changes and improvements easier.

## 🧪 Validation

Local linting and testing.


bcda/attribution-import/utils/cclfUtils.go
carlpartridge, line 1: Testing specific utils that I dont think we use nor seem useful.  Easy to bring back if need be.

bcda/bcdacli/cli.go
```diff
@@ -305,41 +304,6 @@ func setUpApp() *cli.App {
 				return nil
 			},
 		},
-		{
-			Name:     "import-synthetic-cclf-package",
```
carlpartridge, line 309: I dont think ive ever seen us use this.

bcda/bene-prefs/importer.go
```diff
@@ -23,7 +24,7 @@ const (
 )
 
 type BenePrefsImporter struct {
-	FileHandler          BenePrefsFileHandler
+	FileClient           bcdaaws.CustomS3Client
```
carlpartridge, line 27: Ive moved away from some more generic file clients and just set s3 as the default.  I think thats fine as it simplifies many things and there are no plans to move away from s3 anytime in the distant future so we can approach this when necessary.  This CustomS3Client allows for mocks as well as actual s3 clients (live use as well as ministack).

bcda/attribution-import/cclf_test.go
```diff
@@ -79,52 +95,50 @@ func TestCCLFTestSuite(t *testing.T) {
 	suite.Run(t, new(CCLFTestSuite))
 }
 
-func (s *CCLFTestSuite) TestImportCCLF0() {
-	ctx := context.Background()
-
-	assert := assert.New(s.T())
+// func (s *CCLFTestSuite) TestImportCCLF0() {
```
bhagatparwinder, line 98: Now that I see this on the PR, I am a bit hesitant to comment out tests. These tests weren't just testing local file processor code, but they were also validating file header/trailer, CCLF0 and CCLF8 matching, pgx bulk imports, DB connections, and deduplication. We are basically losing the entire attribution and suppression pipeline.

Maybe instead of choosing between bringing back `LocalFileProcessor` or leaving tests commented out until ministack is ready, we can use the `CustomS3Client` interface that’s already in place.

We can wire the existing test fixtures by mocking GetObject and ListObjectsV2 to return the test file bytes/zips.

bcda/attribution-import/cclf.go
```diff
@@ -204,28 +254,28 @@ func (importer CclfImporter) importCCLF8(ctx context.Context, zipMetadata *cclfZ
 	// Step 4: Bulk insert using pgx transaction
 	importedCount, recordCount, err := CopyFrom(ctx, pgxTx, sc, cclfFile.ID, utils.GetEnvInt("CCLF_IMPORT_STATUS_RECORDS_INTERVAL", 10000), importer.logger, validator.maxRecordLength)
 	if err != nil {
-		return errors.Wrap(err, "failed to copy data to beneficiaries table")
+		return fmt.Errorf("failed to copy data to beneficiaries table: %w", err)
 	}
 
 	if recordCount > validator.totalRecordCount {
-		err := fmt.Errorf("unexpected number of records imported for file %s (expected: %d, actual: %d)", fileMetadata.name, validator.totalRecordCount, recordCount)
+		err := fmt.Errorf("unexpected number of records imported for file %s (expected: %d, actual: %d), err: %w", fileMetadata.name, validator.totalRecordCount, recordCount, err)
```
bhagatparwinder, line 261: `err` is `nil` here; we checked it above.

bcda/attribution-import/cclf.go
```diff
@@ -234,60 +284,147 @@ func (importer CclfImporter) importCCLF8(ctx context.Context, zipMetadata *cclfZ
...
+
+			if err != nil {
+				// skipping files with a bad name.  An unknown file in this dir isn't a blocker
+				importer.logger.Errorf("Issue parsing filename into metadata: %v", err)
+				continue
+			}
+
+			if metadata.cclfNum == 0 {
+				if cclf0Metadata != nil {
+					readError = fmt.Errorf("multiple CCLF0 files found in zip (%s/%s), err: %w", bucket, *obj.Key, err)
```
bhagatparwinder, line 343: same thing with `err` being `nil` here; we checked it above.

bcda/bene-prefs/importer.go
```diff
@@ -244,3 +245,59 @@ func (importer BenePrefsImporter) createBenePrefsRecord(ctx context.Context, met
...
+	var result []*models.BenePrefsFilenameMetadata
+
+	bucket, prefix := bcdaaws.ParseS3Uri(path)
+	s3Objects, err := bcdaaws.ListFiles(ctx, importer.FileClient, bucket, prefix)
+	if err != nil {
+		return &result, skipped, err
+	}
+
+	for _, obj := range s3Objects {
+		metadata, err := parseMetadata(*obj.Key)
```
bhagatparwinder, line 259: Are we not checking for `err` being `nil` here?

bcda/bene-prefs/importer.go
```diff
@@ -244,3 +245,59 @@ func (importer BenePrefsImporter) createBenePrefsRecord(ctx context.Context, met
...
+		metadata.FilePath = fmt.Sprintf("s3://%s/%s", bucket, *obj.Key)
```
bhagatparwinder, line 260: and mutation of `metadata` should be after a `err` check.

bcda/aws/s3.go
```diff
@@ -0,0 +1,108 @@
...
+	timeoutDuration := time.Duration(utils.GetEnvInt("S3_DELETE_TIMEOUT", 60)) * time.Second
+
+	_, err := client.DeleteObject(ctx, &s3.DeleteObjectInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(path),
+	})
+	if err != nil {
+		return fmt.Errorf("file %s failed to clean up properly, error occurred while deleting object: %w", filePath, err)
+	} else {
+		err = s3.NewObjectNotExistsWaiter(client).Wait(
```
bhagatparwinder, line 94: We don't need this for `PUT` and `DELETE` operations in S3. This is only adding latency to our Lambda execution. S3 has strong read-after-write consistency for both operations. If you check the `s3_file_handler.go` file deleted in this PR, you'll see that we don't do this.

bcda/attribution-import/csv.go
```diff
@@ -67,21 +59,21 @@ func (importer CSVImporter) ImportCSV(ctx context.Context, filepath string) erro
...
-	data, _, err := importer.FileProcessor.LoadCSV(ctx, filepath)
+	data, _, err := importer.loadCSV(ctx, filepath)
 	if err != nil {
 		return err
 	}
 
 	file.data = data
 
-	err = importer.ProcessCSV(file)
+	err = importer.processCSV(file)
```
bhagatparwinder, line 69: The parent function takes context from the caller, but we are not passing it to `processCSV`; instead, `processCSV` declares a new context. 

Edit: We are doing the same for `importCCLF0` but we are passing it to `importCCLF8`.

bcda/aws/s3.go
```diff
@@ -0,0 +1,108 @@
...
+		if resp.IsTruncated == nil || !*resp.IsTruncated {
+			break
+		}
+		continuationToken = resp.NextContinuationToken
+	}
+
+	return objects, nil
+}
+
+func OpenFileAsScanner(ctx context.Context, client CustomS3Client, filePath string) (*bufio.Scanner, func(), error) {
```
bhagatparwinder, line 42: Please let's make a ticket for changing this implementation. We have `validate` and `scanAndImport` using this method. This method then calls `OpenFileAsBytes` below. We are reading entire files into memory. I remember we've had to increase memory limits in the past for this (because of large files). We can get a better performance and memory usage by using a stream.

bcda/bcdacli/cli.go
bhagatparwinder: I see it being referenced in `scripts/bulk_import_synthetic_cclf_package.sh`. If we are removing this, that script could also be removed.

bcda/attribution-import/cclf_test.go
carlpartridge: Hmm, ministack would do this for us without the need for customizing mocks for specific tests but I think your concern is very valid.  Will dig into, ty for the push!

bcda/aws/s3.go
carlpartridge: Very valid concern.  Both of the uses of this function are part of bene-prefs which only deal with very small files.  Im also not happy with this implementation/functionality but didnt feel like it was part of the scope of this ticket to adjust this (and therefore the nitty gritty workings of `validate` and `scanAndImport`.  We have a ticket designed to address memory concerns for attr-import that we could maybe adjust to take a look at this as well.

bcda/aws/s3.go
carlpartridge: My understanding is that the s3 Delete method accepts the request (and returns success) but then async actually deletes the file.  This polls to make sure that the file was actually deleted and is the only way to make 100% sure that things are as expected.  I agree that this does cause extra delay and is likely unnecessary.  As you mentioned s3 has good consistency and we also have a lifecycle policy to clean up files after ~2 weeks.  I dont think adding (likely at most) a few seconds to each lambda invocation is the end of the world to keep things 100% but also happy to remove this if everyone agrees that it is excessive.  One last point is that bene-prefs is currently a bit awkwardly built in that it retries every file in a directory each time it is run.  Meaning that a file that fails to delete will be re-run, and if it has already been successfully imported but not deleted then it will throw an error on unique name check, which could send us an alert.

bcda/aws/s3.go
carlpartridge: Im going to add a comment to this function just mentioning this and to warn against using it, then we can maybe refactor its uses away as part of the aforementioned ticket(s).

bcda/attribution-import/utils/cclfUtils.go
laurenkrugen-navapbc, line 1: I think this is safe to remove but wanted to confirm: have we done a test of our makefile local setup (`bootstrap`, etc) with this gone/do any of the local setup steps rely on this?

bcda/attribution-import/csv_test.go
```diff
@@ -102,50 +100,49 @@ func TestCSVTestSuite(t *testing.T) {
...
-		cclfFileID  int
-		cclfBeneRec []string
-		err         error
-	}{
-		{"Import CSV attribution success", filepath.Join(s.basePath, "cclf/archives/csv/P.PCPB.M2411.D181120.T1000000"), 0, []string{"MBI000001", "MBI000002", "MBI000003", "MBI000004", "MBI000005"}, nil},
-		{"Import CSV attribution that already exists", filepath.Join(s.basePath, "cclf/archives/csv/P.PCPB.M2411.D181121.T1000000"), 0, []string{}, errors.New("already exists")},
-		{"Import CSV attribution invalid name", filepath.Join(s.basePath, "cclf/archives/csv/P.PC.M2411.D181120.T1000000"), 0, []string{}, errors.New("Invalid filename")},
-		{"Import bene-prefs failure", filepath.Join(s.basePath, "cclf/archives/csv/T#EFT.ON.ACO.NGD1800.DPRF.D181120.T1000010"), 0, []string{}, errors.New("Invalid filename for csv attribution file")},
-	}
+// func (s *CSVTestSuite) TestImportCSV_Integration() {
```
laurenkrugen-navapbc: are these commented out until we implement ministack?

bcda/attribution-import/csv_test.go
```diff
@@ -262,3 +259,67 @@ func (s *CSVTestSuite) TestPrepareCSVData() {
 	}
 
 }
+
+// func (s *CSVTestSuite) TestCleanupCSV() {
```
laurenkrugen-navapbc, line 293: same question as above about adding these once we implement ministack!

bcda/aws/s3.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	return objects, nil
+}
+
+// OpenFileAsScanner opens a file from S3 and returns a bufio.Scanner for reading its contents line by line.
+// It also returns a cleanup function and an error if any occurred.
+// Be warned this is not memory efficient for large files as it reads the entire file into memory.
+func OpenFileAsScanner(ctx context.Context, client CustomS3Client, filePath string) (*bufio.Scanner, func(), error) {
+	byte_arr, err := OpenFileAsBytes(ctx, client, filePath)
```
laurenkrugen-navapbc, line 46: any use in checking the file size here before opening? might be useful to log that here with the filename and fail a little more gracefully?

bcda/bcdacli/cli.go
laurenkrugen-navapbc, line 309: same question here as above -- I tried to remove this a while back and our `Makefile` utilized some of this functionality during local setup. Can we verify our Makefile commands to bootstrap still work from a completely clean setup without this?

bcda/bene-prefs/suppression_s3_test.go
```diff
@@ -228,7 +228,7 @@ package beneprefs
 // 	assert.Equal(2, skipped)
 // }
 
-// func (s *SuppressionS3TestSuite) TestCleanupSuppression() {
+// func (s *SuppressionS3TestSuite) TestCleanupBenePrefsFiles() {
```
laurenkrugen-navapbc, line 231: minor nit and maybe a todo comment so we don't add more to this PR: 

Can we update the TestSuite name to BenePrefs?

bcda/lambda/attribution-import/main.go
```diff
@@ -93,30 +82,30 @@ func initHandler(ctx context.Context) (*AttributionImportHandler, error) {
...
 	}
 
 	s3Event, err := bcdaaws.ParseSQSEvent(sqsEvent)
 	if err != nil {
-		h.Logger.Errorf("failed to parse S3 event: %v", err)
+		h.logger.Errorf("failed to parse S3 event: %v", err)
 		return "", err
 	} else if s3Event == nil {
-		h.Logger.Info("No S3 event found, skipping safely.")
+		h.logger.Info("No S3 event found, skipping safely.")
```
laurenkrugen-navapbc, line 104: this is more general discussion and no action required, but should we have a ticket to add some custom errors, constants, or formatting constraints for all of our log messages? we have a mix of capitalized vs not and lots of free text across the codebase; that seems like something we could discuss and clean up in the future.

bcda/aws/s3.go
carlpartridge: Keeping this in is making things a bit difficult to test as well 😓

bcda/lambda/attribution-import/main.go
carlpartridge: Go linting says dont capitalize errors (or end with punctuation).  However just logging something doesnt care about that.  Ive somewhat been shifting things over as I see them but not wholesale as it tends to break tests and could also break splunk alerts if we arent careful 😓.  But yeah clean up and standards would be great!

bcda/bene-prefs/suppression_s3_test.go
carlpartridge: This whole file is commented out and there is no corrollary `suppression_s3.go` file.  I believe these are all integration tests that are likely equally tested in the importer_test now.  I was thinking I would keep this file around til we get ministack up and running again, then do a more thorough dive on this file.

bcda/bcdacli/cli.go
carlpartridge: Just did a total wipe of all docker images, containers, and volumes and did a `make docker-bootstrap` and everything seems good.

bcda/attribution-import/csv_test.go
carlpartridge: I brought back as many as possible with some customized mocks but some of these would require much more complicated customizations and seems like its worth just waiting for ministack.

michaeljvaldes (dismissed):
Appreciate the removal of local file handlers and the reduction of complexity in moving a lot of stuff from the file handler to the importer. This is a lot more readable now.

bcda/attribution-import/cclf_test.go
```diff
@@ -399,3 +425,290 @@ func buildZipMetadata(t *testing.T, processor CclfFileProcessor, cmsID, zipName,
...
+
+	bucketName := uuid.NewRandom().String()
+
+	// failed import: stay put
+	fileTime, _ := time.Parse(time.RFC3339, constants.TestFileTime)
+	cclf0metadata := &cclfFileMetadata{
+		name:         "T.BCD.ACO.ZC0Y18.D181120.T0001000",
+		env:          "test",
+		acoID:        acoID,
+		cclfNum:      8,
```
bhagatparwinder: Should this be 0 for `cclf0metadata`?

bcda/bene-prefs/importer_test.go
```diff
@@ -342,7 +542,7 @@ func (s *BenePrefsTestSuite) TestCleanupSuppression() {
 	}
 
 	suppresslist = []*models.BenePrefsFilenameMetadata{metadata, metadata2, metadata3}
-	err := importer.FileHandler.CleanupBenePrefsFiles(ctx, suppresslist)
+	err := importer.cleanupBenePrefsFiles(ctx, suppresslist)
 	assert.Nil(err)
 
 	files, err := os.ReadDir(conf.GetEnv("PENDING_DELETION_DIR"))
```
bhagatparwinder: Is this valid now that we use `cleanupBenePrefsFiles` above that calls `bcdaaws.Delete` instead of moving the file?

bcda/bene-prefs/importer_test.go
```diff
@@ -389,16 +604,16 @@ func (s *BenePrefsTestSuite) TestCleanupSuppression_Bad() {
...
 
 	suppresslist = []*models.BenePrefsFilenameMetadata{metadata1, metadata2}
-	err := importer.FileHandler.CleanupBenePrefsFiles(ctx, suppresslist)
-	assert.EqualError(err, "2 files could not be cleaned up")
+	err := importer.cleanupBenePrefsFiles(ctx, suppresslist)
+	assert.ErrorContains(err, "files could not be cleaned up")
 }
 
-func (s *BenePrefsTestSuite) TestCleanupSuppression_RenameFileError() {
+func (s *BenePrefsTestSuite) TestCleanupBenePrefsFiles_RenameFileError() {
```
bhagatparwinder: nit: test name. We assert no error, and we are testing for success rather than error handling.

bcda/attribution-import/csv_test.go
```diff
@@ -262,3 +290,109 @@ func (s *CSVTestSuite) TestPrepareCSVData() {
...
+				FileClient: client,
+				PgxPool:    s.pool,
+			}
+
+			r, _, err := importer.loadCSV(ctx, test.filepath)
+			if test.err == nil {
+				assert.Nil(err)
+				assert.NotNil(r)
+			} else {
+				s.T().Log("FOO BAR")
```
bhagatparwinder: cleanup

bcda/aws/s3_test.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+
+		objects, err := ListFiles(context.Background(), client, "test-bucket", "test-prefix/")
+		require.ErrorIs(t, err, mockErr)
+		assert.Nil(t, objects)
+	})
+}
+
+func TestOpenFileAsScanner(t *testing.T) {
+	client := &ConfigurableMockS3Client{}
```
bhagatparwinder, line 109: We only test for error. Do we want to test for the happy path?

bcda/bene-prefs/importer.go
```diff
@@ -244,3 +245,58 @@ func (importer BenePrefsImporter) createBenePrefsRecord(ctx context.Context, met
...
+
+	for _, bpFile := range suppresslist {
+		if !bpFile.Imported {
+			// Don't do anything. The S3 bucket should have a retention policy that
+			// automatically cleans up files after a specified period of time,
+			importer.Logger.Warningf("File %s was not imported successfully. Skipping cleanup", bpFile)
+			continue
+		}
+
+		importer.Logger.Infof("Cleaning up file %s\n", bpFile)
```
bhagatparwinder: nit: I believe we can remove `\n` Logrus does this explicitly.

bcda/attribution-import/cclf_test.go
```diff
@@ -80,82 +83,83 @@ func TestCCLFTestSuite(t *testing.T) {
...
+// 	assert.EqualError(err, "Failed to import 15 files")
+
+// 	//Target bad file directory
+// 	cclfDirectory = filepath.Join(s.basePath, constants.CCLFDIR, "archives", "invalid_bcd")
+// 	imported, failed, skipped, err := s.importer.ImportCCLFDirectory(s.T().Context(), cclfDirectory)
+// 	assert.NoError(err)
+// 	assert.Equal(0, imported)
+// 	assert.Equal(4, failed)
+// 	assert.Equal(0, skipped)
+// }
```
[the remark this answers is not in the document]
carlpartridge: This is a known temporary removal of tests as we wait for the re-implementation of ministack.

bcda/bene-prefs/importer_test.go
```diff
@@ -173,142 +255,260 @@ func (s *BenePrefsTestSuite) TestImport_MissingData() {
...
-		assert.Equal(modtimeAfter.Format("010203040506"), f.DeliveryDate.Format("010203040506"))
-	}
-}
-
-func (s *BenePrefsTestSuite) TestLoadBenePrefsFiles_TimeChange() {
+// func (s *BenePrefsTestSuite) TestLoadBenePrefsFiles() {
+// 	assert := assert.New(s.T())
+// 	ctx := context.Background()
+// 	repo := &models.MockRepository{}
+// 	client := &bcdaaws.ConfigurableMockS3Client{}
```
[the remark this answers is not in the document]
carlpartridge: This is a known temporary removal of tests as we wait for the re-implementation of ministack.

bcda/bene-prefs/importer_test.go
carlpartridge: Going to leave this for the post ministack re-implementation work as all of these cleanup related tests are failing on the verification step of the s3 delete call.

bhagatparwinder (approved):

laurenkrugen-navapbc (approved):

carlpartridge:
Sonarqube is failing this PR due to code coverage.  This is a known issue as we have a number of tests that need to stay commented out until we re-implement ministack.

PR #1487, opened 2026-09-04
BCDA-9903: add ssas creds rotation lambda

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9903

## 🛠 Changes

- create infrastructure for rotate-ssas-creds lambda
- create function code for rotate-ssas-credslambda
- create workflows for rotate-ssas-creds lambda deployment and integration testing
- regenerate mocks with mockery and create mock for SSASHTTPClient to help with testing new lambda
- create sensible helper functions for slack msgr (post successes to operations channel, post failures to alerts channel)
- slightly refactor mock ssm client to store params and update affected tests (this will likely change again soon when ministack is integrated)

## ℹ️ Context

The rotate-ssas-creds lambda runs every Monday early in the morning, reads a param store value containing a list of systems to rotate and performs the following steps for each system:
- reset the SSAS credentials for the system
- create or update a paramstore param with the new credentials

This lambda can be easily configured to receive a list of test systems and rotate them weekly.
To integrate into BCDA testing (add one test system to automatic rotation): 
- add the relevant test system's system ID to the rotation-systems param in SOPS/paramstore
- run the lambda to generate new creds in param store
- update all references to the credentials in the repo with the new paramstore address

Additionally, this PR also includes updates to mocks generated by mockery and updates to the mock ssm client, which both helped with testing the new lambda. In particular, the mock ssm client now stores a map of param key-value pairs, which can be useful in testing updates to the param store. This may be changed in the near future when ministack is implemented, but it was simple enough to mock out and help test this PR in the meantime.

## 🧪 Validation

rotate-ssas-creds lambda integration test


bcda/aws/parameters_test.go
```diff
@@ -7,13 +7,18 @@ import (
 )
 
 func TestGetParameter(t *testing.T) {
-	value, err := GetParameter(t.Context(), &MockSSMClient{}, "name")
+	params := map[string]string{"name": "value"}
+	ssmClient := MockSSMClient{Params: params}
+
+	value, err := GetParameter(t.Context(), &ssmClient, "name")
```
carlpartridge, line 13: Are these testing our mocks, specifically our customizations of the mocks?  I guess its better than what we have.  Feels like we should probably tag this as part of the move to ministack?

bcda/lambda/admin_create_group/main_test.go
```diff
@@ -6,20 +6,22 @@ import (
...
-}
+// type mockSSASClient struct {
+// }
 
-func (s *mockSSASClient) CreateGroup(groupId string, name string, acoCMSID string) ([]byte, error) {
-	return []byte(`{"group_id":"00001"}`), nil
-}
+// func (s *mockSSASClient) CreateGroup(groupId string, name string, acoCMSID string) ([]byte, error) {
+// 	return []byte(`{"group_id":"00001"}`), nil
+// }
```
carlpartridge: Nit: cleanup

bcda/slackmessenger/slackmessenger.go
```diff
@@ -27,3 +28,17 @@ func SendSlackMessage(sc *slack.Client, channel string, msg string, color string
...
+	channel := OperationsChannel
+	SendSlackMessage(sc, channel, formattedMsg, color)
+}
+
+func SendFailureToAlerts(sc *slack.Client, msg string) {
+	formattedMsg := fmt.Sprintf("%s: [%s environment] %s ", FailureMsg, os.Getenv("ENV"), msg)
+	color := Danger
+	channel := AlertsChannel
+	SendSlackMessage(sc, channel, formattedMsg, color)
+}
```
carlpartridge, line 44: Nit: Two small tests for these?  Im not sure what we can check besides the message?  Hmm we dont seem to have a slack mock client, so maybe not?

Makefile
```diff
@@ -178,7 +178,7 @@ fhir_testing:
 	fhir_testing
 
 generate-mocks:
-	docker run -v "$PWD":/src -w /src vektra/mockery:v3.6.1
+	docker run -v ${PWD}:/src -w /src vektra/mockery:latest
```
carlpartridge, line 181: Do we not want to lock this to a version?

bcda/slackmessenger/slackmessenger.go
laurenkrugen-navapbc: I'm OK with not having a test for these, since we would have to mock a slack client or integrate with a live system.

laurenkrugen-navapbc:
This looks great! Excited to get some automation for this. 🎉 

How would it work getting those param store values in the relevant SOPS env files?

bcda/lambda/admin_create_aco/main_test.go
```diff
@@ -140,8 +142,15 @@ func (c *HandleCreateACOTestSuite) TestHandleCreateACOMissingCMSID() {
 }
 
 func TestGetAWSParams(t *testing.T) {
-	params, err := getAWSParams(context.Background(), &bcdaaws.MockSSMClient{})
+	env := conf.GetEnv("ENV")
```
bhagatparwinder, line 146: In both (`aws_test.go` and `mock_test.go`), should there be fallback logic so I don't see a generic error running the tests?

```
t.Setenv("ENV", "local")
```

bcda/lambda/rotate_ssas_creds/main.go
```diff
@@ -0,0 +1,157 @@
...
+)
+
+type rotationSystem struct {
+	SystemId   string `json:"system_id"`
+	CredsParam string `json:"creds_param"`
+}
+
+type shortCreds struct {
+	ClientID     string `json:"client_id"`
+	ClientSecret string `json:"client_secret"` // #nosec G117
```
bhagatparwinder, line 28: I think (G117) this needs to go down to [line 135](https://github.com/CMSgov/bcda-app/pull/1487/changes#diff-a5edfc8fe6659c853b4e758bf85676f6f971995368c96db9cd683065beec969cR135)

bcda/lambda/rotate_ssas_creds/main.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	err = setupEnv(params)
+	if err != nil {
+		logger.Errorf("unable to setupEnvironment properly: %+v", err)
+		return nil, err
+	}
+
+	slackClient := slack.New(params.slackToken)
+	if err != nil {
```
bhagatparwinder: Line 69 won't return an error, but `*slack.Client`.  This is dead code.

bcda/lambda/rotate_ssas_creds/main.go
```diff
@@ -0,0 +1,157 @@
...
+
+func (h RotateSSASCredsHandler) rotateCreds(ctx context.Context, rs rotationSystem) error {
+	if len(rs.SystemId) == 0 {
+		h.logger.Errorf("failed to get system id for system %s", rs.CredsParam)
+	}
+	if len(rs.CredsParam) == 0 {
+		h.logger.Errorf("failed to get creds param for a system")
+	}
+	var newCreds shortCreds
+	credsBytes, err := h.ssas.ResetCredentials(rs.SystemId)
```
bhagatparwinder, line 141: We should have a return statement (and not continue execution) if SystemId or CredsParam is empty (the first two if statements).

```
if rs.SystemId == "" || rs.CredsParam == "" {
    return fmt.Errorf("system_id and creds_param must not be empty (got system_id=%q, creds_param=%q)", rs.SystemId, rs.CredsParam)
}
```

bcda/lambda/rotate_ssas_creds/main.go
```diff
@@ -0,0 +1,157 @@
...
+		h.logger.Errorf("failed to unmarshal new creds for system %s", rs.CredsParam)
+		return err
+	}
+	newValueBytes, err := json.Marshal(newCreds)
+	if err != nil {
+		h.logger.Errorf("failed to re-marshal new creds for system %s", rs.CredsParam)
+		return err
+	}
+	newValue := string(newValueBytes)
+	return updateCredsParam(ctx, h.ssmClient, rs.CredsParam, newValue)
```
bhagatparwinder: This flow risks us resetting credentials first before we generate a new one and store it. If `updateCredsParam` fails for any reason, we will have a secret that we no longer have access to (it was in memory). I have pointed out the issue, but I don't know of a good way to fix it tbh.

Maybe a Slack alert on update failure? Or maybe test reset creds works before we execute the flow? Maybe I am trying to overthink it.

ops/services/30-rotate-ssas-creds/iam.tf
```diff
@@ -0,0 +1,11 @@
+data "aws_iam_policy_document" "put_param" {
+  statement {
+    sid    = "SSMPutParameter"
+    effect = "Allow"
+
+    actions = ["ssm:PutParameter"]
+    resources = [
+      "arn:aws:ssm:us-east-1:539247469933:parameter/${local.app}/${local.env}/rotate-ssas-creds/*"
```
bhagatparwinder: Umm, I don't think the account ID should be hardcoded (or the region)

bcda/lambda/rotate_ssas_creds/aws.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	return rotationSystems, nil
+}
+
+func updateCredsParam(ctx context.Context, ssmClient bcdaaws.CustomSSMClient, name string, value string) error {
+	fullCredsParam := fmt.Sprintf("/bcda/%s/rotate-ssas-creds/%s", conf.GetEnv("ENV"), name)
+	input := &ssm.PutParameterInput{
+		Name:        aws.String(fullCredsParam),
+		Value:       aws.String(value),
+		Type:        types.ParameterTypeSecureString,
```
bhagatparwinder, line 119: Do we not need to specify the `KeyId`? This will use the AWS managed default key rather than `data.aws_kms_alias.bcda_app_config_kms_key`

ops/services/30-rotate-ssas-creds/security_group_rules.tf
```diff
@@ -0,0 +1,13 @@
+data "aws_security_group" "db" {
```
bhagatparwinder: Are we referencing this anywhere in the lambda code?

ops/services/30-rotate-ssas-creds/main.tf
```diff
@@ -0,0 +1,62 @@
+locals {
+  app         = "bcda"
+  env         = terraform.workspace
+  full_name   = "${local.app}-${local.env}-${local.service}"
+  db_sg_name  = "bcda-${local.env}-db"
+  memory_size = 2048
```
bhagatparwinder: Do we need 2GB for a couple of HTTP requests or should this be reduced?

.github/workflows/rotate-ssas-creds-deploy.yml
```diff
@@ -0,0 +1,94 @@
...
+    inputs:
+      deploy_env:
+        description: "Environment you want to deploy to (dev, test, prod)"
+        required: true
+        default: test
+        type: choice
+        options:
+          - dev
+          - test
+          - prod
```
bhagatparwinder, line 14: We don't have `sandbox` as a choice for the `deploy_env` input? If we skipped it internally, we should probably remove it from the README.

.github/workflows/rotate-ssas-creds-integration-test.yml
```diff
@@ -0,0 +1,42 @@
+name: Rotate SSAS Creds Integration Test
+
+on:
+  pull_request:
```
bhagatparwinder: Wouldn't this result in a credential rotation every time there is a PR modifying `rotate_ssas_creds`?

bcda/aws/parameters_test.go
michaeljvaldes: Yep -- not necessary, but it doesn't hurt to have it.
If/when this PR gets merged, I can update the ministack ticket to consider this.

.github/workflows/rotate-ssas-creds-deploy.yml
```diff
@@ -0,0 +1,94 @@
...
+    runs-on: codebuild-bcda-app-${{ contains(fromJSON('["prod", "sandbox"]'), inputs.deploy_env) && 'prod' || 'non-prod'}}-${{github.run_id}}-${{github.run_attempt}}
+    steps:
+      - uses: aws-actions/configure-aws-credentials@ec61189d14ec14c8efccab744f656cffd0e33f37 # v6.1.0
+        with:
+          aws-region: ${{ vars.AWS_REGION }}
+          role-to-assume: arn:aws:iam::${{ secrets.AWS_ACCOUNT_ID }}:role/delegatedadmin/developer/${{ vars.AWS_ROLE_TO_ASSUME }}
+      - name: Checkout BCDA
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.3.0
+      - name: Install Cosign to verify tenv and tofu installs
+        uses: sigstore/cosign-installer@cad07c2e89fa2edd6e2d7bab4c1aa38e53f76003 #v4.1.1@d58896d6a1865668819e1d91763c7751a165e159 # v3.9.2
```
bhagatparwinder: Is the comment a typo with multiple versions and hashes?

Makefile
bhagatparwinder: Yep, this should go back to pinned version.

bcda/lambda/admin_aco_deny/main_test.go
```diff
@@ -25,8 +25,15 @@ func TestHandleACODenies(t *testing.T) {
...
+		dbURLName:      "db://url",
+		slackParamName: "test-slack-token",
+	}
+	params, err := getAWSParams(context.Background(), &bcdaaws.MockSSMClient{Params: storedParams})
 	assert.Nil(t, err)
-	assert.Equal(t, "value1", params.SlackToken)
-	assert.Equal(t, "value2", params.DBURL)
+	assert.Equal(t, "test-slack-token", params.SlackToken)
+	assert.Equal(t, "db://url", params.DBURL)
+	print(dbURLName)
```
bhagatparwinder, line 38: I think this is left over from debugging.

bcda/slackmessenger/slackmessenger.go
```diff
@@ -27,3 +28,17 @@ func SendSlackMessage(sc *slack.Client, channel string, msg string, color string
...
+
+func SendSuccessToOperations(sc *slack.Client, msg string) {
+	formattedMsg := fmt.Sprintf("%s: [%s environment]: %s ", SuccessMsg, os.Getenv("ENV"), msg)
+	color := Good
+	channel := OperationsChannel
+	SendSlackMessage(sc, channel, formattedMsg, color)
+}
+
+func SendFailureToAlerts(sc *slack.Client, msg string) {
+	formattedMsg := fmt.Sprintf("%s: [%s environment] %s ", FailureMsg, os.Getenv("ENV"), msg)
```
bhagatparwinder: Nit: Do we want to match the log formatting for line 33 and 40? Secondly, do we prefer using `conf.GetEnv` instead of `os.Getenv`?

bcda/lambda/rotate_ssas_creds/aws.go
```diff
@@ -0,0 +1,133 @@
...
+	err = json.Unmarshal([]byte(param), &rotationSystems)
+	if err != nil {
+		log.Errorf("failed to unmarshal cred systems: %+v", err)
+		return nil, err
+	}
+	return rotationSystems, nil
+}
+
+func updateCredsParam(ctx context.Context, ssmClient bcdaaws.CustomSSMClient, name string, value string) error {
+	fullCredsParam := fmt.Sprintf("/bcda/%s/rotate-ssas-creds/%s", conf.GetEnv("ENV"), name)
```
bhagatparwinder: What would happen if, as an input, the developer provides the full path in main.go? This would result in double prefixing of the path. We could either rename the field to `creds_param_suffix` / `system_name` or trim any leading slashes/prefixes in `updateCredsParam`. Just food for thought (to clear ambiguity in the process).

Makefile
michaeljvaldes: At this point, I don't think so. I think that the benefit here is that we don't have to keep updating this version every time we update go versions. The typical downside is that this would interfere with any automated processes that depend on it, but I don't believe we have any of those.
I propose that we _should_ have an automatic process for updating our mocks though, and I can create a ticket for that (and to revisit this as part of the ticket).

bcda/lambda/admin_create_group/main_test.go
michaeljvaldes: Fixed!

.github/workflows/rotate-ssas-creds-deploy.yml
michaeljvaldes: Fixed (across several workflows)

.github/workflows/rotate-ssas-creds-integration-test.yml
michaeljvaldes: This is a good point -- I had removed the pull_request trigger from the deploy workflow for that reason, and I just removed it from this one. If folks make changes to this lambda, they should run these workflows manually.

bcda/lambda/admin_aco_deny/main_test.go
michaeljvaldes: Fixed 👍

bcda/lambda/admin_create_aco/main_test.go
michaeljvaldes: Fixed 👍

bcda/lambda/rotate_ssas_creds/aws.go
michaeljvaldes: That's a good point. I renamed "CredsParam" to "CredsName" since it's the name of the credentials, not the full name of the paramstore param. I think unfortunately the design I chose does generate some ambiguity here regardless -- I'll aim to help alleviate that in our rotation docs as well.

.github/workflows/rotate-ssas-creds-deploy.yml
carlpartridge: We would want this for sandbox right?  But we would also need to work with a website update as well?

michaeljvaldes: The lambda is currently configured to only rotate single creds, not pairs, so it can't be used in sandbox.
If/when we extend it to handle cred pairs (which might be much more complex), then we should add it to sandbox too.

bcda/lambda/rotate_ssas_creds/aws.go
michaeljvaldes: This is a good point -- I updated it to pass the key (the same one used by our sops/config module) to the environment of the lambda. I also added validation for this value and the other values we require in the environment.

bcda/lambda/rotate_ssas_creds/main.go
michaeljvaldes: As far as I can tell, the linter requires it to be here. It's specific to the exported field of the struct, and follows the pattern we use for the Credentials struct we have in the auth package that we also marshal/unmarshal.

bcda/lambda/rotate_ssas_creds/main.go
michaeljvaldes: Oops -- fixed to check if the returned client is null.

bcda/lambda/rotate_ssas_creds/main.go
michaeljvaldes: Good point -- fixed!

bcda/slackmessenger/slackmessenger.go
michaeljvaldes: Fixed 👍

ops/services/30-rotate-ssas-creds/iam.tf
michaeljvaldes: Fixed 👍

ops/services/30-rotate-ssas-creds/main.tf
michaeljvaldes: Updated to match the value of our other admin lambdas (256MB)

ops/services/30-rotate-ssas-creds/security_group_rules.tf
michaeljvaldes: Fixed 👍

bhagatparwinder (dismissed):

carlpartridge (dismissed):

carlpartridge (dismissed):

bcda/lambda/rotate_ssas_creds/main.go
bhagatparwinder: I am not sure what's going on here 😭

Image

bhagatparwinder: But if this is working fine on your end and in CI, I am good with it.

bcda/lambda/rotate_ssas_creds/main.go
```diff
@@ -0,0 +1,180 @@
...
+
+	reportMsg := fmt.Sprintf("ssas creds rotation complete. successes: %d, failures: %d", successes, failures)
+	h.logger.Info(reportMsg)
+	if failures > 0 {
+		msgr.SendFailureToAlerts(h.slackClient, reportMsg)
+	} else {
+		msgr.SendSuccessToOperations(h.slackClient, reportMsg)
+	}
+
+	return nil
```
bhagatparwinder, line 130: This will always return `nil` so Lambda considers the execution a success and returns null to `test-result.txt`. The integration test's `Verify result` will report success every time.

bcda/lambda/rotate_ssas_creds/aws.go
```diff
@@ -0,0 +1,130 @@
...
+		return err
+	}
+	err = os.Setenv("SSAS_USE_TLS", "true")
+	if err != nil {
+		log.Errorf("error setting SSAS_USE_TLS env var: %+v", err)
+		return err
+	}
+	err = os.Setenv("BCDA_CA_FILE", pemFilePath)
+	if err != nil {
+		log.Errorf("error setting BCDA_CA_FILE env var: %+v", err)
```
bhagatparwinder, line 78: Should this have a `return err`?

ops/services/30-rotate-ssas-creds/main.tf
```diff
@@ -0,0 +1,64 @@
+locals {
+  app            = "bcda"
+  env            = terraform.workspace
+  full_name      = "${local.app}-${local.env}-${local.service}"
+  db_sg_name     = "bcda-${local.env}-db"
```
bhagatparwinder: We removed the source; should we remove this, too?

bcda/lambda/rotate_ssas_creds/main.go
michaeljvaldes: I'm not sure what's going on there, it might depend on a local installation of golangci-lint?
If you run `make lint` it should use the one installed in Dockerfile.tests which has a set version. It's possible we should update that version, too 👀

bcda/lambda/rotate_ssas_creds/main.go
michaeljvaldes: I couldn't decide if failing to rotate just one cred pair would count as a "failure" but that's a good point in that it's crucial to integration test. Fixed!

bcda/lambda/rotate_ssas_creds/main.go
carlpartridge: I get discrepencies in the linters occasionally between what is run on git commit and what is run as part of our workflows.  Would be nice to be consistent, but obviously out of scope.

bcda/lambda/rotate_ssas_creds/main.go
carlpartridge: So if 1 of the creds failed to rotate for whatever reason we would get a logger message for it.  I think thats enough for us to begin debugging/adjusting or would we want it more explicit (sent to slack?)?

carlpartridge (approved):

bcda/lambda/rotate_ssas_creds/main.go
michaeljvaldes: As of now, if there is at least one failure, a failure message is sent to slack and the lambda returns an error (which tells aws and the integration test that there was a failure)

bcda/lambda/rotate_ssas_creds/main.go
michaeljvaldes: I think `make lint` does only show new problems when compared to main, which should still catch this one since it hasn't been merged to main yet.

bhagatparwinder (approved):

PR #1492, opened 2026-09-10
BCDA-10415: Implement ministack

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10415

## 🛠 Changes

Implement ministack and one minor unit test suite to verify the functionality.

## ℹ️ Context

Bringing back proper aws mocking for local dev and testing use.

## 🧪 Validation

Local linting and testing.


bcda/aws/parameters_test.go
```diff
@@ -1,24 +1,106 @@
...
 	value, err := GetParameter(t.Context(), &ssmClient, "name")
 	assert.Nil(t, err)
 	assert.Equal(t, "value", value)
+
+	// test with ministack
+	key1 := "key1"
+	val1 := "val1"
+	cleanupParam1 := testUtils.SetParameter(t, key1, val1)
+	defer cleanupParam1()
+
```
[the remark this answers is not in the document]
carlpartridge: We dont really run tests outside of docker/compose so I believe this is moot.  It would also be a bit messy and likely a large undertaking to do `if os.Getenv("AWS_ENDPOINT_URL") != nil` throughout all of our various test suites.

compose.test.yml
```diff
@@ -42,6 +50,25 @@ services:
     build:
       context: .
       dockerfile: docker/Dockerfile.postman_test
+  ministack-unit-test:
+    image: ministackorg/ministack:latest
+    ports:
+      - "4567:4566"
+    environment:
```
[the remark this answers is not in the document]
carlpartridge: The container service by default is listening in on 4566.  We are creating a secondary ministack env/service in our testing compose to be used with our unit test suite.  This should hopefully keep any testing artifacts isolated.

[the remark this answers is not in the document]
michaeljvaldes: This concern might be valid though -- I think the health check should be mapped to the container port, or 4566 in this case.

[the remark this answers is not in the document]
michaeljvaldes: Ah I see the GATEWAY_PORT is also configured to be 4567, so the health check is hitting the right port. If you want to expose the right port, I think you might need to change the port mapping to `4567:4567`
However, I'd argue that we don't need to expose these ports to the host machine at all, and can remove the port mapping. Does that sound about right?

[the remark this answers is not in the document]
carlpartridge: Removed the port mapping as we dont actually need to access ministack (especially the unit-test version) externally.

bcda/testUtils/utils.go
```diff
@@ -140,26 +144,26 @@ func CopyToTemporaryDirectory(t *testing.T, src string) (string, func()) {
...
+		config.WithRegion(constants.DefaultRegion),
+	)
+	require.Nil(t, err)
 
-// 	return cfg
-// }
+	return cfg
+}
 
 // func TestS3Client(t *testing.T, cfg aws.Config) *s3.Client {
```
laurenkrugen-navapbc, line 158: are we leaving these commented out because of the cleanup discussion on [this PR](https://github.com/CMSgov/bcda-app/pull/1486)?

bcda/aws/parameters_test.go
```diff
@@ -1,24 +1,106 @@
...
+	"errors"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
+	"github.com/CMSgov/bcda-app/bcda/testUtils"
 )
 
 func TestGetParameter(t *testing.T) {
+	// test with mocks
```
laurenkrugen-navapbc, line 13: is it necessary to keep the mocks in or can we remove?

compose.yml
```diff
@@ -27,6 +27,11 @@ services:
       - ./shared_files/decrypted/local.env
     environment:
       SKIP_BOOTSTRAP: "true"
```
bhagatparwinder, line 29: Should this be removed/blanked out? Same on line 86.

bcda/aws/parameters_test.go
carlpartridge: I was torn, they dont really add a lot right now, but if we were to either expand on our mocks or move away from ministack they would be useful again.  I can go either way, so if anyone feels strongly let me know!

bcda/testUtils/utils.go
carlpartridge: Yeah, cleanup of tests, especially around aws uses (core lambdas) and associated test/aws helpers are another [ticket](https://jira.cms.gov/browse/BCDA-9950).

compose.test.yml
```diff
@@ -42,6 +50,23 @@ services:
...
+      - SERVICES=s3,ssm,cloudwatch,sts,iam
+      - LOG_LEVEL=DEBUG
+      - DOCKER_NETWORK=bcda-app-net
+      - GATEWAY_PORT=4567
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+      - "./shared_files/decrypted:/etc/config"
+    healthcheck:
+      test: ["CMD", "python", "-c", "import urllib.request; urllib.request.urlopen('http://localhost:4567/_ministack/health')"]
+      interval: 1m30s
```
bhagatparwinder, line 66: Do we want to add a 2-minute delay before every unit test run? Have we tried a lower value (I assume it wouldn't take ministack a couple of seconds, and we have 5 retries)?

compose.test.yml
```diff
@@ -42,6 +50,23 @@ services:
...
+  ministack-unit-test:
+    image: ministackorg/ministack:latest
+    environment:
+      - AWS_DEFAULT_REGION=us-east-1
+      - SERVICES=s3,ssm,cloudwatch,sts,iam
+      - LOG_LEVEL=DEBUG
+      - DOCKER_NETWORK=bcda-app-net
+      - GATEWAY_PORT=4567
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
```
bhagatparwinder, line 62: Do we need to give the container full root control? I see Ministack is configured with a couple of services. I think we can remove it from here and compose.test.yml (line 62).

compose.test.yml
```diff
@@ -42,6 +50,23 @@ services:
     build:
       context: .
       dockerfile: docker/Dockerfile.postman_test
+  ministack-unit-test:
+    image: ministackorg/ministack:latest
```
bhagatparwinder, line 54: We removed port mappings from here, leaving no port exposed on `localhost`. We use it in `.vscode\settings.json`.

compose.test.yml
```diff
@@ -8,6 +8,12 @@ services:
     env_file:
       - ./shared_files/decrypted/local.env
     environment:
+      # Set default values for ministack to work
+      - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID:-foobar}
+      - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY:-foobar}
+      - AWS_DEFAULT_REGION=us-east-1
+      - AWS_ENDPOINT_URL=http://ministack-unit-test:4567
```
bhagatparwinder, line 15: I'd say we make this port consistent (4566) with the other two in `compose.yml` and `vscode\settings.json`.

compose.test.yml
```diff
@@ -42,6 +50,23 @@ services:
...
+    image: ministackorg/ministack:latest
+    environment:
+      - AWS_DEFAULT_REGION=us-east-1
+      - SERVICES=s3,ssm,cloudwatch,sts,iam
+      - LOG_LEVEL=DEBUG
+      - DOCKER_NETWORK=bcda-app-net
+      - GATEWAY_PORT=4567
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+      - "./shared_files/decrypted:/etc/config"
```
bhagatparwinder, line 63: Are we consuming this volume anywhere?

bcda/testUtils/utils.go
bhagatparwinder, line 158: Are we not restoring additional tests in this file or `s3_fileprocessor_test.go`?

bcda/testUtils/utils.go
```diff
@@ -313,54 +317,54 @@ func CopyToTemporaryDirectory(t *testing.T, src string) (string, func()) {
 // 	return tempBucket, cleanup
 // }
 
-// // Inserts the provided parameter into localstack.
-// func putParameter(t *testing.T, input ssm.PutParameterInput) error {
-// 	ctx := context.Background()
+// Inserts the provided parameter into localstack.
```
bhagatparwinder: nit: should this say ministack? README.md also still has `make unit-test-localstack`

bcda/testUtils/utils.go
```diff
@@ -313,54 +317,54 @@ func CopyToTemporaryDirectory(t *testing.T, src string) (string, func()) {
...
+	})
+	assert.Nil(t, err)
 
-// 	cleanup := func() {
-// 		err := deleteParameters(t, ssm.DeleteParametersInput{Names: []string{name}})
-// 		assert.Nil(t, err)
-// 	}
+	cleanup := func() {
+		err := deleteParameters(t, ssm.DeleteParametersInput{Names: []string{name}})
+		assert.Nil(t, err)
```
bhagatparwinder, line 337: I think anywhere we are calling `putParameter` or `deleteParameters` we will never get a non-nil error back. Both of those functions use `require.Nil`. So this assertion is redundant.

bcda/testUtils/utils.go
```diff
@@ -313,54 +317,54 @@ func CopyToTemporaryDirectory(t *testing.T, src string) (string, func()) {
...
 
-// // Insert all given parameters into localstack and return a method for deferring cleanup.
-// func SetParameter(t *testing.T, name, value string) func() {
-// 	err := putParameter(t, ssm.PutParameterInput{
-// 		Name:  &name,
-// 		Value: &value,
-// 		Type:  "String",
-// 	})
-// 	assert.Nil(t, err)
+// Insert all given parameters into localstack and return a method for deferring cleanup.
```
bhagatparwinder: nit: ministack

bcda/testUtils/utils.go
```diff
@@ -313,54 +317,54 @@ func CopyToTemporaryDirectory(t *testing.T, src string) (string, func()) {
...
+func SetParameter(t *testing.T, name, value string) func() {
+	err := putParameter(t, ssm.PutParameterInput{
+		Name:  &name,
+		Value: &value,
+		Type:  "String",
```
bhagatparwinder: I think for test re-run safety, we should pass `Overwrite: aws.Bool(true)`.

bcda/testUtils/utils.go
```diff
@@ -313,54 +317,54 @@ func CopyToTemporaryDirectory(t *testing.T, src string) (string, func()) {
...
+// Inserts the provided parameter into localstack.
+func putParameter(t *testing.T, input ssm.PutParameterInput) error {
+	ctx := context.Background()
 
-// 	cfg, err := config.LoadDefaultConfig(ctx,
-// 		config.WithRegion(constants.DefaultRegion),
-// 	)
-// 	assert.Nil(t, err)
-// 	client := ssm.NewFromConfig(cfg)
+	cfg, err := config.LoadDefaultConfig(ctx,
```
bhagatparwinder: Can we use `TestAWSConfig(t)` here?

bcda/testUtils/utils.go
```diff
@@ -313,54 +317,54 @@ func CopyToTemporaryDirectory(t *testing.T, src string) (string, func()) {
...
+		config.WithRegion(constants.DefaultRegion),
+	)
+	require.Nil(t, err)
+	client := ssm.NewFromConfig(cfg)
```
bhagatparwinder: Same here by using `TestSSMClient(t, cfg)`.

bcda/testUtils/utils.go
```diff
@@ -313,54 +317,54 @@ func CopyToTemporaryDirectory(t *testing.T, src string) (string, func()) {
...
+		Value: &value,
+		Type:  "String",
+	})
+	assert.Nil(t, err)
 
-// 	cleanup := func() {
-// 		err := deleteParameters(t, ssm.DeleteParametersInput{Names: []string{name}})
-// 		assert.Nil(t, err)
-// 	}
+	cleanup := func() {
```
bhagatparwinder, line 334: Is it better to use `t.Cleanup(cleanup)` instead of returning a cleanup function, and then the caller must remember to `defer`?

bcda/aws/parameters_test.go
```diff
@@ -1,24 +1,106 @@
...
+		{
+			desc:          "Missing parameter",
+			keyname:       "asdf",
+			expectedValue: "",
+			expectedErr:   errors.New("error retrieving parameter asdf from parameter store"),
+		},
+	}
+
+	client := testUtils.TestSSMClient(t, testUtils.TestAWSConfig(t))
+	for _, test := range tests {
```
bhagatparwinder, line 49: This is just a minor improvement, but `t.Run` will allow us to run individual subtests.

compose.test.yml
carlpartridge: I started with a lower interval, but the compose would say that it failed to spin up (but would actually be up and running when you checked on it later).  Im not sure if this actually adds 2m to the start up time, it still seems to run pretty fast.  I will try to time it and see.  I took these numbers from our previous localstack set up.

compose.test.yml
carlpartridge: Its part of the basic config set up and according to google is necessary for ministack to be able to communicate with the docker daemon on the host machine.  I can look into it more if you are concerned.

compose.test.yml
carlpartridge: See below comment.

compose.test.yml
carlpartridge: We end up having two ministack services being spun up.  One as part of our main docker compose stack, then a second, more temporary one as part of our unit-test compose stack.  This unit-test stack is, I would consider, temporary and disposable, so I was hoping to keep it isolated from the main ministack.  Since both can be up at the same time we need different ports, thus the port customizations in the compose-test.yml.  If vscode does want to talk to ministack it can use the primary and more persistent ministack service.

compose.test.yml
carlpartridge: It was part of our localstack set up and I believe is part of the complex init needed for our services (I think both env vars as well as config ymls).  I took a bunch of config and set up from our previous localstack set up as looking at the [ministack docs on migrating](https://ministack.org/docs/migrating-from-localstack) the two are incredibly similar.  [Here is the PR with the removal changes](https://github.com/CMSgov/bcda-app/pull/1341/changes).

bcda/testUtils/utils.go
carlpartridge: See other comment but no, restoring aws needed tests and associated helpers is part of this [ticket](https://jira.cms.gov/browse/BCDA-9950).

bcda/testUtils/utils.go
carlpartridge: True but I believe the linter will complain about a non checked err return if we dont.

bcda/aws/parameters_test.go
carlpartridge: Do you prefer the individual runs as opposed to the for loop in general or in certain situations where setup/conditions are fairly different?

bcda/testUtils/utils.go
carlpartridge: Hmm unfortunately the testing context is cancelled before the t.Cleanup runs making the request fail.

bhagatparwinder (approved):

michaeljvaldes (approved):

PR #1495, opened 2026-09-14
BCDA-10137: allow cancellation of completed jobs

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10137

## 🛠 Changes

- update jobs DELETE endpoint to allow for cancellation of Completed jobs

## ℹ️ Context

The [FHIR bulk spec](https://hl7.org/fhir/uv/bulkdata/STU2/export.html#bulk-data-delete-request) specifies the following: 

> If the request has been completed, a server MAY use the request as a signal that a client is done retrieving files and that it is safe for the server to remove those from storage. Following the delete request, when subsequent requests are made to the polling location, the server SHALL return a 404 Not Found error and an associated FHIR OperationOutcome in JSON format.

**In other words, following a delete request for a job, the server must respond with a 404 to status requests for that job.**

Presently, when receiving a delete request for a Completed job, BCDA returns a 410 as it only allows such requests for jobs that are Pending or InProgress. This PR allows clients to send delete requests for Completed jobs, which changes the job status from Completed to Cancelled and returns a 202. Further requests for that job's status would then automatically return 404s since that the is the default behavior for a Cancelled job.

Side effects:
- job status /jobs/{jobID}: a job which has been Completed and subsequently deleted returns 404, since its status is now Cancelled
- file downloads /data/{jobID}/{fileName}: unaffected
- worker cleanup: for completed jobs, the cleanup works in two steps: first it archives the completed job after a time cutoff by updating its status to Archived and moving all files from its payload dir to an archive dir. In the second step, the archived job is updated to status Expired and its archive and staging dirs are removed. 
With this PR, a Completed job can be set to Cancelled. The cleanup worker is already designed to update the Cancelled job to CancelledExpired and remove its staging and payload directories. Consequently, the cleanup process should result in an identical outcome for the job's files (all relevant directories are removed), but the job's status will end up being CancelledExpired instead of Expired.
- alerts: there don't seem to be any alerts that would obviously be impacted, but the fact that a job can be completed and then cancelled should be a future consideration in weighing whether cancellations should be considered "failures" of any sort.

## 🧪 Validation

updated unit and integration tests


bhagatparwinder (dismissed):

bcda/service/service.go
```diff
@@ -259,7 +259,7 @@ func (s *service) CancelJob(ctx context.Context, jobID uint) (uint, error) {
 	}
 
 	// Check if the job is pending or in progress.
```
laurenkrugen-navapbc: minor nit: can we update or remove the comment? I think this line is pretty self-explanatory, so I'd be OK with removing it.

michaeljvaldes: Fixed 👍

laurenkrugen-navapbc:
Are cancelled jobs still available for download if they haven't been cleaned up yet?

If not, then we would need to update some [middleware](https://github.com/CMSgov/bcda-app/blob/main/bcda/auth/middleware.go#L266) that checks the job status before download. If they are still available, then like you said in the PR description, that should just clean those up automatically.

michaeljvaldes:
> Are cancelled jobs still available for download if they haven't been cleaned up yet?
> 
> If not, then we would need to update some [middleware](https://github.com/CMSgov/bcda-app/blob/main/bcda/auth/middleware.go#L266) that checks the job status before download. If they are still available, then like you said in the PR description, that should just clean those up automatically.

Technically the files are still available, and I think that's okay -- the spec doesn't specify that we have to delete the files immediately, just that we may. I appreciate the reassurance that they should be deleted automatically -- that's been the trickiest part of this analysis.

bhagatparwinder:
> > Are cancelled jobs still available for download if they haven't been cleaned up yet?
> > If not, then we would need to update some [middleware](https://github.com/CMSgov/bcda-app/blob/main/bcda/auth/middleware.go#L266) that checks the job status before download. If they are still available, then like you said in the PR description, that should just clean those up automatically.
> 
> Technically the files are still available, and I think that's okay -- the spec doesn't specify that we have to delete the files immediately, just that we may. I appreciate the reassurance that they should be deleted automatically -- that's been the trickiest part of this analysis.

1. We do need to update the middleware. If a client has sent a DELETE request for a  completed job, we are now conforming to the specs. However, they can still download the file by `/data/{jobID}/{fileName}`. Check [RequireTokenJobMatch](https://github.com/CMSgov/bcda-app/blob/a7b7813aaba54c39e838707ef2995fd36b19cc0c/bcda/auth/middleware.go#L224) (it should reject requests and send a 404).

2. Similarly, check the method [ServData](https://github.com/CMSgov/bcda-app/blob/a7b7813aaba54c39e838707ef2995fd36b19cc0c/bcda/api/v1/api.go#L270). It should reject requests with 404 if the job status is `JobStatusCancelled` or `JobStatusCancelledExpired`

> subsequent requests are made to the polling location, the server SHALL return a 404 Not Found error

3. We can also slightly expand the tests in [service_test.go](https://github.com/CMSgov/bcda-app/blob/a7b7813aaba54c39e838707ef2995fd36b19cc0c/bcda/service/service_test.go#L1106) to assert `job.Status == models.JobStatusCancelled` (currently we only validate the ID).

P.S. FHIR specs were themselves returning 202 instead of 404 until 2025: https://github.com/hapifhir/hapi-fhir/issues/5849 *insert homer hiding in the bushes GIF*

carlpartridge (dismissed):

colby-seyferth-nava (dismissed):

bhagatparwinder (dismissed):

laurenkrugen-navapbc (dismissed):

carlpartridge (approved):

bhagatparwinder (approved):

michaeljvaldes:
> > > Are cancelled jobs still available for download if they haven't been cleaned up yet?
> > > If not, then we would need to update some [middleware](https://github.com/CMSgov/bcda-app/blob/main/bcda/auth/middleware.go#L266) that checks the job status before download. If they are still available, then like you said in the PR description, that should just clean those up automatically.
> > 
> > 
> > Technically the files are still available, and I think that's okay -- the spec doesn't specify that we have to delete the files immediately, just that we may. I appreciate the reassurance that they should be deleted automatically -- that's been the trickiest part of this analysis.
> 
> 1. We do need to update the middleware. If a client has sent a DELETE request for a  completed job, we are now conforming to the specs. However, they can still download the file by `/data/{jobID}/{fileName}`. Check [RequireTokenJobMatch](https://github.com/CMSgov/bcda-app/blob/a7b7813aaba54c39e838707ef2995fd36b19cc0c/bcda/auth/middleware.go#L224) (it should reject requests and send a 404).
> 2. Similarly, check the method [ServData](https://github.com/CMSgov/bcda-app/blob/a7b7813aaba54c39e838707ef2995fd36b19cc0c/bcda/api/v1/api.go#L270). It should reject requests with 404 if the job status is `JobStatusCancelled` or `JobStatusCancelledExpired`
> 
> > subsequent requests are made to the polling location, the server SHALL return a 404 Not Found error
> 
> 3. We can also slightly expand the tests in [service_test.go](https://github.com/CMSgov/bcda-app/blob/a7b7813aaba54c39e838707ef2995fd36b19cc0c/bcda/service/service_test.go#L1106) to assert `job.Status == models.JobStatusCancelled` (currently we only validate the ID).
> 
> P.S. FHIR specs were themselves returning 202 instead of 404 until 2025: [hapifhir/hapi-fhir#5849](https://github.com/hapifhir/hapi-fhir/issues/5849) _insert homer hiding in the bushes GIF_

This was a great conversation that explored the grey area of BCDA behavior that isn't 100% prescribed by the FHIR specs. The conclusion we reached was to keep the /data endpoint behavior as-is, and include a response message (in an opoutcome) that notifies the user that their files will be deleted in the future.

PR #1496, opened 2026-09-14
BCDA-9957: ack cutoff date dynamically

bhagatparwinder:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9957

## 🛠 Changes

- Updated `ACO.Denylisted()` to dynamically evaluate `CutoffDate` against `time.Now()`. 
- If `CutoffDate` is unset/zero, falls back to checking `DenylistType`.
- Extended the `payload` struct to accept optional `cutoff_date` and `termination_date` parameters.
- Updated `denyACOs()` in lambda to persist `CutoffDate` and `TerminationDate`.
- Expanded unit tests
- Added validation to lambda when no ACOs are passed
- Added a check that the cutoff date is after the termination date (future-proofing)
- CLI cleanup

## ℹ️ Context

Previously, BCDA did not evaluate an ACO's cutoff date dynamically at request time. If an ACO was marked in the database with denylist details, they were immediately locked out regardless of any future dates. This required engineering/ops teams to wait until the exact final day/time of an ACO's termination before running the `admin_aco_deny` Lambda.

With these changes, we can schedule terminations in advance by passing a future `cutoff_date` to the Lambda. The ACO will continue to have access until that timestamp arrives, at which point BCDA automatically revokes access.

## 🧪 Validation

Local unit tests pass.


bcda/lambda/admin_aco_deny/README.md
```diff
@@ -1,6 +1,52 @@
...
 
+## Payload Specification
+
+The Lambda expects a JSON payload with the following fields:
+
+| Field | Type | Required | Description |
+|---|---|---|---|
+| `deny_aco_ids` | `[]string` | **Yes** | List of ACO CMS IDs to deny. Must not be empty. |
+| `cutoff_date` | `string` (RFC3339) | No | Date and time after which the ACO will have no access and requests will be blocked (denylisted). Defaults to current time (`time.Now()`) if omitted. |
+| `termination_date` | `string` (RFC3339) | No | Date and time when the ACO moved from full to limited access. Defaults to `cutoff_date` if omitted. Cannot be after `cutoff_date`. If `termination_date` is in the future, a `cutoff_date` is required. |
```
laurenkrugen-navapbc: I saw a comment in `termination.go` that  TerminationDate is set `When caller moved from full to limited access`;

1. should we default `TerminationDate` to time.Now() even when cutoff date _is_ set?
2. should TerminationDate` be set within the lambda instead of passing in the value?

bcda/models/termination.go
```diff
@@ -41,7 +41,7 @@ const (
 )
 
 type Termination struct {
-	TerminationDate time.Time // When caller moved from full to limited access
+	TerminationDate time.Time // When caller moved from full to limited access, currently not evaluated
```
laurenkrugen-navapbc: This is [checked in service.go](https://github.com/CMSgov/bcda-app/blob/main/bcda/service/service.go#L670) for some of the time constraints we use; do we mean it's just not evaluated within the lambda?

bcda/lambda/admin_aco_deny/README.md
```diff
@@ -1,6 +1,52 @@
...
+
+### Timestamp Format Requirement
+
+> **IMPORTANT:** Date fields (`cutoff_date` and `termination_date`) MUST be formatted in standard RFC3339 / ISO 8601 format including time and timezone (e.g., `"2026-12-31T23:59:59Z"`). Plain date strings such as `"2026-12-31"` are not supported and will result in a JSON unmarshaling error.
+
+### Example Payloads
+
+#### 1. Immediate Deny (default behavior)
```
laurenkrugen-navapbc, line 15: Thanks for adding examples to the readme, that is super helpful!

bcda/models/termination.go
carlpartridge: Yeah as far as I can tell we dont use it in the current deny logic nor in the lambda, but I do see it used elsewhere.

bcda/lambda/admin_aco_deny/README.md
```diff
@@ -1,6 +1,52 @@
...
+
+| Field | Type | Required | Description |
+|---|---|---|---|
+| `deny_aco_ids` | `[]string` | **Yes** | List of ACO CMS IDs to deny. Must not be empty. |
+| `cutoff_date` | `string` (RFC3339) | No | Date and time after which the ACO will have no access and requests will be blocked (denylisted). Defaults to current time (`time.Now()`) if omitted. |
+| `termination_date` | `string` (RFC3339) | No | Date and time when the ACO moved from full to limited access. Defaults to `cutoff_date` if omitted. Cannot be after `cutoff_date`. If `termination_date` is in the future, a `cutoff_date` is required. |
+
+### Timestamp Format Requirement
+
+> **IMPORTANT:** Date fields (`cutoff_date` and `termination_date`) MUST be formatted in standard RFC3339 / ISO 8601 format including time and timezone (e.g., `"2026-12-31T23:59:59Z"`). Plain date strings such as `"2026-12-31"` are not supported and will result in a JSON unmarshaling error.
```
carlpartridge: Nit:  We could maybe get rid of this block as its already stated above that it needs to be RFC3339.  We could add to the above description an example datetime string as well.

bcda/lambda/admin_aco_deny/README.md
bhagatparwinder: 1. I cannot set the TerminationDate to now. The whole purpose of this PR is to have the capability to set the CutoffDate in the future. If I set the CutoffDate to September 30th and the TerminationDate gets set to now, the ACO will have access to the data till September 30th, but their constraints will clamp down to now (see https://github.com/CMSgov/bcda-app/blob/main/bcda/service/service.go#L670). We would be providing them with limited data, even if they are allowed data till Sept 30th.
2. No. Currently, it is optional. We have the option to pass it. Setting it in the lambda means that in the future, when we want to support a termination date, we have to change the lambda. The way it is currently written, you wouldn't have to update the lambda.

bcda/lambda/admin_aco_deny/README.md
bhagatparwinder: Good idea, removed and added examples to the table above.

bcda/models/termination.go
bhagatparwinder: I updated the messaging to be a bit more clear. Hope it helps.

carlpartridge (dismissed):

bcda/lambda/admin_aco_deny/README.md
bhagatparwinder: The way code has been updated now (with this PR), we technically have all the pieces to support full access, limited access, and no access. No new updates will be required in the future.

If you pass the termination date and the cutoff date, the ACO will have full access now, limited access when they reach the termination date, and no access when they reach the cutoff date. If you don't provide a termination date and only a cutoff date, I match both, and my reasoning is in bullet point 1 above. If neither is provided, we set both to now.

laurenkrugen-navapbc (dismissed):

bcda/models/termination.go
```diff
@@ -41,7 +41,7 @@ const (
 )
 
 type Termination struct {
-	TerminationDate time.Time // When caller moved from full to limited access
+	TerminationDate time.Time // When caller moved from full to limited access, currently not evaluated for limited access as limited access is not implemented.
```
michaeljvaldes: It's so curious that TerminationDate really does seem to be used to calculate OptOutDate, ClaimsDate, and AttributionDate time constraints for terminated entities, but there doesn't seem to be any mechanism for setting the associated "strategies" when the termination is created 👀

michaeljvaldes (dismissed):

bcda/models/termination.go
bhagatparwinder: Updated messaging!

laurenkrugen-navapbc (approved):

michaeljvaldes (approved):

PR #1497, opened 2026-09-16
BCDA-9950: Reimplement ministack required tests

carlpartridge:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-9950

## 🛠 Changes

- Re-enabled all commented out tests that related to AWS.
- Refactored s3 manager to transfermanager as the former is now considered deprecated.

Test coverage of related packages (per file inside package):
- attribution-import: 78%, 80%, 97%, and 87%
- bene-prefs: 93% and 100%
- (bcda)aws: 100%, 100%, 61%, 100%, 91%, 77%, and 100%
- The attr-import and bene-prefs lambda coverage are both low as they contain a lot of untestable init code (53% and 33%).  They shuffle basically all of their work off to the above packages so no cause for concern there.
A majority of missing coverage for the rest of the pakcages are difficult to hit errors (eg pass in an int instead of a str) and less useful errors (eg json string marshalling err).

## ℹ️ Context

We lost the ability to mock aws services with the loss of localstack a bit ago.  With the recent introduction of ministack we can now run these tests again.

## 🧪 Validation

Local linting and testing.


bcda/aws/s3.go
```diff
@@ -55,56 +54,32 @@ func OpenFileAsScanner(ctx context.Context, client CustomS3Client, filePath stri
...
 		return fmt.Errorf("file %s failed to clean up properly, error occurred while deleting object: %w", filePath, err)
-	} else {
-		err = s3.NewObjectNotExistsWaiter(client).Wait(
-			ctx,
-			&s3.HeadObjectInput{
-				Bucket: aws.String(bucket),
-				Key:    aws.String(path),
-			},
-			timeoutDuration,
-		)
```
carlpartridge, line 104: Ended up removing this, after fighting for it last ticket 😅.  Further reading made it seem less needed than I originally thought and it was causing a lot of flaky issues with testing...

bcda/attribution-import/csv_test.go
```diff
@@ -105,74 +102,63 @@ func TestCSVTestSuite(t *testing.T) {
...
+
+			err = importer.ImportCSV(ctx, filepath.Join(bucketName, fpath))
+			if test.err == nil {
+				assert.Nil(s.T(), err)
+			} else {
+				assert.NotNil(s.T(), err)
+				assert.Contains(s.T(), err.Error(), test.err.Error())
+			}
+			r := postgres.NewRepository(s.db)
+			cclfRecords := postgrestest.GetCCLFFilesByName(s.T(), s.db, filepath.Clean(test.filepath))
```
bhagatparwinder: This is a false positive test. We will always go into the else block.

`ImportCSV` stores the file name into `cclf_files.name`. However, `test.filepath` is the full path. So when you query, you get 0 records every time, going into the else block.

Even if it was going to the if block, it would fail because `Name` does not include the full path (which it isn't now).

bcda/bene-prefs/importer_test.go
```diff
@@ -359,231 +358,182 @@ func (s *BenePrefsTestSuite) TestValidate() {
...
+		FilePath:     filepath.Join(fileNameBucket, s.basePath, "suppressionfile_BadFileNames/T#EFT.ON.ACO.NGD1800.DPRF.D190117.T9909420"),
+		Imported:     true,
+		DeliveryDate: time.Now(),
+	}
+
+	suppresslist = []*models.BenePrefsFilenameMetadata{metadata, metadata2, metadata3}
+	err := importer.cleanupBenePrefsFiles(ctx, suppresslist)
+	assert.Nil(err)
+
+	objs, err := bcdaaws.ListFiles(ctx, client, headerBucket, filepath.Join(s.basePath, "suppressionfile_BadHeader"))
```
bhagatparwinder: I believe this is a false positive test as well. If you add a breakpoint/log here, you will see `ListFiles` finds 0 items. `testUtils.CopyToS3` is stripping the leading slash, whereas `filepath.Join` is not, so those don't match, leading to the 0 items.

bcda/bene-prefs/importer_test.go
```diff
@@ -359,231 +358,182 @@ func (s *BenePrefsTestSuite) TestValidate() {
...
+	assert.NoError(err)
+	assert.Len(objs, 0)
```
bhagatparwinder: Furthermore, `cleanupBenePrefsFiles` explicitly skips unimported files. The unimported file was never deleted from S3, yet assert.Len(objs, 0) passes only because the prefix didn't match

bcda/bene-prefs/importer_test.go
```diff
@@ -359,231 +358,182 @@ func (s *BenePrefsTestSuite) TestValidate() {
...
+	for _, obj := range objs {
+		if *obj.Key == "T#EFT.ON.ACO.NGD1800.DPRF.D181120.T1000009" {
```
bhagatparwinder: Wouldn't this key be the full object path?

bcda/testUtils/utils.go
```diff
@@ -155,167 +160,168 @@ func TestAWSConfig(t *testing.T) (aws.Config, context.Context) {
...
+
+		fmt.Printf("Uploaded file in bucket %s, key %s\n", tempBucket, key)
+		return nil
+	})
+
+	if err != nil {
+		t.Fatalf("Failed to upload files to S3: %s", err.Error())
+	}
+
+	cleanup := func() {
```
bhagatparwinder, line 227: This cleanup cleans all objects inside the bucket; we never delete the bucket itself. Neither does `CreateBucket`. Do we want to add DeleteBucket in the cleanup callback?

bcda/testUtils/utils.go
```diff
@@ -155,167 +160,168 @@ func TestAWSConfig(t *testing.T) (aws.Config, context.Context) {
...
+			return err
+		}
+		defer f.Close()
+
+		key := path
+		parts := strings.Split(path, "shared_files/")
+		if len(parts) > 1 {
+			key = parts[1]
+		}
+		if key[0] == '/' {
```
bhagatparwinder: Nit: We could probably save ourselves the headache of `index out of range error` by using `strings.TrimPrefix(key, "/")` if `key` is ever empty.

bcda/bene-prefs/importer_test.go
```diff
@@ -359,231 +358,182 @@ func (s *BenePrefsTestSuite) TestValidate() {
...
+
+	//Utilize the other bad file, but set an invalid pending deletion directory.
+	filePath = filepath.Join(folderPath, constants.TestSuppressBadDeletePath)
+	_, err = os.Open(filePath)
+	assert.Nil(err)
+
+	timeChange = origTime.Add(-(time.Hour * 73)).Truncate(time.Second)
+	err = os.Chtimes(filePath, timeChange, timeChange)
+
+	if err != nil {
```
bhagatparwinder: Shouldn't there be an assertion here?

bcda/testUtils/utils.go
carlpartridge: In actual live use we just re-use the same bucket.  Only in testing do we create temp buckets.  Ministack does not persist state between runs so we should be fine.

bcda/bene-prefs/importer_test.go
carlpartridge: We used to very strictly check time, down to the ms.  According to copilot:
```
CopyToS3 sets each object's LastModified at upload time; it does not preserve the local mtime. This code also stats and changes the source directory rather than each file, so it compares unrelated times and is timing-dependent instead of validating S3 delivery-date handling.

This issue also appears on line 429 of the same file.
```
So this was being very flaky as it was and wasnt even working properly.  I removed that assertion as per copilots suggestion.  Im not even sure why we were checking and changing the time of the file (but not the file name itself).  Maybe in some previous iteration of the code we were using file time and not the filenames time?  I tried to keep as much of everything that was left as possible but that might just be misguided...  I will try and clean up this test a bit.

michaeljvaldes (approved):

laurenkrugen-navapbc (approved):

PR #1498, opened 2026-09-17
BCDA-10352: add complex service-date validation

michaeljvaldes:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10352

## 🛠 Changes

- moved most typefilter and search parameter logic to bcda/fhir/r4/search, including validations from validations middleware
- created standard logic for parsing different search parameter types, and a pattern for parsing and validating specific parameters
- created new validations for service-date parameter
- adjusted naming for some typefilter-related variables, structs, and functions

## ℹ️ Context

New validations for the service-date parameter are required, but adding to the existing typefilter validations is difficult because typefilter is complicated. New functions and structs were created to help clarify typefilter and further support adding to BCDA's typefilter code.

## 🧪 Validation

many updated unit tests


bcda/api/requests_test.go
```diff
@@ -1588,17 +1584,17 @@ func TestOmitSharedSystemByDefault_Integration(t *testing.T) {
 	tests := []struct {
 		name         string
 		cmsID        string
-		typeFilter   fhir.TypeFilterParameter
+		typeFilter   search.TypeFilterSubquery
```
carlpartridge, line 1587: Nit:  Would these search package references be more readable if the package was custom named, eg `fhirSearch.TypeFilterSubquery`?

bcda/client/bluebutton.go
```diff
@@ -20,7 +20,8 @@ import (
 
 	"github.com/CMSgov/bcda-app/bcda/client/fhir"
 	"github.com/CMSgov/bcda-app/bcda/constants"
-	fhirModels "github.com/CMSgov/bcda-app/bcda/models/fhir"
+	fhirModels "github.com/CMSgov/bcda-app/bcda/fhir"
```
carlpartridge: Nit: Does the custom name kind of imply that fhir models should be a separate models package?  I think you had said you didnt like that idea.  Would it be reasonable to try to break down the models as they are into separate more distinct packages like you have done with the search structs?

bcda/fhir/r4/search/date.go
```diff
@@ -0,0 +1,90 @@
...
+			}
+		}
+		if !valid {
+			return d, ParameterParsingError{Details: fmt.Sprintf("invalid date parameter value: %s", datetime)}
+		}
+	}
+	return d, nil
+}
+
+func ValidateServiceDates(serviceDateParams []DateParam) error {
```
carlpartridge, line 71: Nit:  We talked about this a bit but I am still a bit undecided.  These files contain two mostly distinct concerns:
- parsing fhir strings into usable objects
- business logic (validating/constructing for BFD)

Feels like they would be best split up.  However this might be a bit excessive and future proofing...

bcda/fhir/r4/search/date.go
```diff
@@ -0,0 +1,90 @@
...
+	if len(d.Datetimes) == 0 {
+		return d, ParameterParsingError{Details: fmt.Sprintf("value must include a valid FHIR date for date parameter %s", d.Name)}
+	}
+
+	var datetimeFormats = []string{
+		"2006",
+		"2006-01",
+		"2006-01-02",
+		"2006-01-02T15:04:05",
+		"2006-01-02T15:04:05+07:00",
```
bhagatparwinder, line 59: This will break (remove the service-date completely) if the user supplies `service-date=ge2024-01-15T10:30:00Z`.

The method `ValidateTypeFilterSubquery` delegates to this function for date validation, and this will accept it as valid. Downstream, it will make it to [parseDate](https://github.com/CMSgov/bcda-app/blob/c284737ca0485edff1b9b8ce2b58a5f67151b844/bcda/client/bluebutton.go#L496), which only accepts dates, will fail to parse it and return `time.Time{}` (no date). And look what we do with [zero dates](https://github.com/CMSgov/bcda-app/blob/c284737ca0485edff1b9b8ce2b58a5f67151b844/bcda/client/bluebutton.go#L453).

Now, when we build the query, we [delete the existing](https://github.com/CMSgov/bcda-app/blob/c284737ca0485edff1b9b8ce2b58a5f67151b844/bcda/client/bluebutton.go#L487) `service-date` and add a new one (which there is none of).

If you want to test this, make a request to:

```
GET /api/v3/Group/all/$export?_type=ExplanationOfBenefit&_typeFilter=ExplanationOfBenefit%3Fservice-date%3Dge2024-01-15T10:30:00Z
```

We will accept this with 202 Accepted, but return all data instead of honoring the `service-date`.

We either need to allow date and time in the BFD client or restrict validation here to match the BFD client.

bcda/fhir/r4/search/typefilter.go
```diff
@@ -0,0 +1,122 @@
...
+	Details string
+}
+
+func (e ParameterValidationError) Error() string {
+	return fmt.Sprintf("invalid parameter: %s", e.Details)
+}
+
+func ParseTypeFilterSubquery(s string) (TypeFilterSubquery, error) {
+	// The subquery is url-encoded. So we will first decode so we can parse it
+	decodedQuery, err := url.QueryUnescape(s)
```
bhagatparwinder, line 45: We are double-decoding at this point. Earlier in the flow, we have run [Query](https://github.com/CMSgov/bcda-app/blob/c284737ca0485edff1b9b8ce2b58a5f67151b844/bcda/web/middleware/validation.go#L171) on the URL.

Under the hood, it is using [parseQuery](https://cs.opensource.google/go/go/+/refs/tags/go1.27.1:src/net/url/url.go;l=994) that already does QueryUnescape.

Why would it be a proble to double-decode? Imagine this `date%3Dge2024-01-15T10%3A30%3A00%2B05%3A00`. Decoded once we get `date=ge2024-01-15T10:30:00+05:00` (we got our + sign) but decoded again it turns into a space `date=ge2024-01-15T10:30:00 05:00`

bcda/fhir/r4/search/date.go
```diff
@@ -0,0 +1,90 @@
...
+func ValidateServiceDates(serviceDateParams []DateParam) error {
+	lowerBoundCount, upperBoundCount, equalCount := 0, 0, 0
+	for _, sd := range serviceDateParams {
+		if sd.Name != string(TypeFilterParamServiceDate) {
+			return ParameterValidationError{Details: fmt.Sprintf("invalid key for service-date parameter: %s", sd.Name)}
+		}
+		if len(sd.Datetimes) > 1 {
+			return ParameterValidationError{Details: "invalid service-date parameter value: comma-separated values are not supported"}
+		}
+		switch sd.Prefix {
```
bhagatparwinder, line 80: This is missing the check for `""`. For example, `service-date=2024-01-15`. No prefix implies `eq`. A request like `service-date=2024-01-15&service-date=2024-01-16` has equalCount == 0 and completely avoids this check.

I would treat `""` as `eq`. in prefix counting (or default d.Prefix = "eq").

bcda/fhir/r4/search/date.go
```diff
@@ -0,0 +1,90 @@
...
+		case "lt", "le":
+			upperBoundCount = upperBoundCount + 1
+		case "gt", "ge":
+			lowerBoundCount = lowerBoundCount + 1
+		case "eq":
+			equalCount = equalCount + 1
+		}
+	}
+	if lowerBoundCount > 1 || upperBoundCount > 1 || equalCount > 1 {
```
bhagatparwinder, line 89: This only checks if an individual counter is > 1. A query like `service-date=eq2024-01-15&service-date=gt2024-01-01` has equalCount == 1 and lowerBoundCount == 1, which evaluates to false and passes validation, even though eq and gt are conflicting.

I would check for conflicting conditions when equalCount > 0 and lowerBoundCount + upperBoundCount > 0.

bcda/fhir/r4/search/token.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+
+		validTagTokens := map[string][]string{
+			constants.BFDSystemTypeURL:  {"SharedSystem", "NationalClaimsHistory", "DDPS"},
+			constants.BFDFinalActionURL: {"FinalAction", "NotFinalAction"},
+		}
+
+		validTagCodes, ok := validTagTokens[value.System]
+		if !ok || !slices.Contains(validTagCodes, value.Code) {
+			return ParameterValidationError{Details: "invalid _tag parameter. Searching by tag requires a token (system|code) to be specified"}
```
bhagatparwinder: Nit: I think this is a copy-paste message from line 50, but a bit misleading. In this case, they did supply the system and/or code, but it is just unsupported.

Maybe:
```
fmt.Sprintf("unsupported tag token system %s or code %s", value.System, value.Code)
```

bcda/fhir/r4/search/token.go
```diff
@@ -0,0 +1,64 @@
...
+		if found {
+			t.Values = append(t.Values, TokenValue{System: before, Code: after})
+		} else {
+			t.Values = append(t.Values, TokenValue{Code: before})
+		}
+	}
+	return t, nil
+}
+
+func ValidateTag(t TokenParam) error {
```
bhagatparwinder, line 49: This function needs the same check we do in [string.go](https://github.com/CMSgov/bcda-app/pull/1498/changes#diff-68a337ea0f88b51902fc561a6ca715d605e6783947165aea9456b34c5cd88c6eR32) for validating the modifiers. If this function runs with `_tag:not=` it will have no problem.

We will end up doing:

```
params.Add("_tag:not", "https://bluebutton.cms.gov/fhir/CodeSystem/Final-Action|FinalAction")
```

in BFD [client](https://github.com/CMSgov/bcda-app/blob/78db99d7aebbb95789d86aff9da4ab265fec5ecd/bcda/client/bluebutton.go#L541).

bcda/fhir/r4/search/string.go
```diff
@@ -0,0 +1,41 @@
...
+	Modifier string
+	Values   []string
+}
+
+func ParseStringParam(subqueryParam TypeFilterSubqueryParam) (StringParam, error) {
+	s := StringParam{}
+	if len(subqueryParam.Name) == 0 {
+		return s, ParameterParsingError{Details: "string parameter missing key"}
+	}
+	s.Name, s.Modifier, _ = strings.Cut(subqueryParam.Name, ":")
```
bhagatparwinder, line 17: This `Cut` on line 20 needs to move above the `if`. If `subqueryParam.Name` is ":exact", `len(...) == 0` is false, and `s.Name` becomes `""` while `s.Modifier becomes` "exact". No parsing error is generated.

bcda/fhir/r4/search/string.go
```diff
@@ -0,0 +1,41 @@
+package search
+
+import (
+	"fmt"
+	"strings"
+)
+
+// https://hl7.org/fhir/R4/search.html#string
```
bhagatparwinder, line 8: It should not be string per FHIR R4 specs, it should be code: http://hl7.org/fhir/R4/datatypes.html#code

Specs: http://hl7.org/fhir/R4/explanationofbenefit-definitions.html (check outcome)

We even advertise it as a token/code in the metadata: https://github.com/CMSgov/bcda-app/blob/78db99d7aebbb95789d86aff9da4ab265fec5ecd/bcda/api/v3/api.go#L344

bcda/fhir/r4/search/date.go
```diff
@@ -0,0 +1,90 @@
...
+		return ParameterParsingError{Details: "invalid service-date parameter value: conflicting prefix conditions"}
```
bhagatparwinder: Should this return a `ParameterParsingError` when our parent function is a validator and we do return `ParameterValidationError` on lines 72 and 75 earlier?

bcda/fhir/r4/search/typefilter.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return TypeFilterSubquery{}, fmt.Errorf("failed to unescape %s", s)
+	}
+
+	// Expected format is: <resourceType>?<paramList>
+	resourceType, params, ok := strings.Cut(decodedQuery, "?")
+	if !ok {
+		return TypeFilterSubquery{}, fmt.Errorf("missing question mark %s", decodedQuery)
```
bhagatparwinder: Any reason we are not using `ParameterValidationError` and `ParameterParsingError` in this file?

bcda/fhir/r4/search/token.go
```diff
@@ -0,0 +1,64 @@
...
+	if t.Name != string(TypeFilterParamTag) {
+		return ParameterValidationError{Details: fmt.Sprintf("invalid key for tag parameter: %s", t.Name)}
+	}
+	for _, value := range t.Values {
+		if len(value.System) == 0 || len(value.Code) == 0 {
+			return ParameterValidationError{Details: "invalid _tag parameter value. Searching by tag requires a token (system|code) to be specified"}
+		}
+
+		validTagTokens := map[string][]string{
```
bhagatparwinder: Maybe this map should be a package-level variable; otherwise, we are re-allocating it every iteration.

bcda/fhir/r4/search/date.go
```diff
@@ -0,0 +1,90 @@
...
+	afterPrefix := ""
+if len(subqueryParam.Value) >= 2 && slices.Contains(prefixes, subqueryParam.Value[:2]) {
+		d.Prefix = subqueryParam.Value[:2]
+		afterPrefix = subqueryParam.Value[2:]
+	} else {
+		afterPrefix = subqueryParam.Value
+	}
+
+	d.Datetimes = strings.Split(afterPrefix, ",")
+	if len(d.Datetimes) == 0 {
```
bhagatparwinder: `strings.Split` returns a slice of 1, so this will become dead code. I'd check if `len(afterPrefix) == 0` before calling strings.Split.

bcda/fhir/r4/search/typefilter.go
```diff
@@ -0,0 +1,122 @@
+package search
```
bhagatparwinder, line 1: There is no test file for this?

bcda/fhir/r4/search/date.go
michaeljvaldes: Fixed!

bcda/fhir/r4/search/token.go
michaeljvaldes: Fixed!

bcda/fhir/r4/search/typefilter.go
michaeljvaldes: Fixed!

bcda/fhir/r4/search/date.go
michaeljvaldes: Fixed!

bcda/fhir/r4/search/string.go
michaeljvaldes: Fixed!

bcda/fhir/r4/search/token.go
michaeljvaldes: Fixed -- added a validation that ensures that a modifier is not present.

bcda/fhir/r4/search/token.go
michaeljvaldes: Fixed!

bcda/fhir/r4/search/date.go
michaeljvaldes: I think this might be even more complicated than that since that condition doesn't necessarily mean a conflict. I might push this to a stretch goal or another ticket.

bcda/fhir/r4/search/date.go
michaeljvaldes: Fixed!

bcda/fhir/r4/search/typefilter.go
michaeljvaldes: I think I left this as it was before (we were decoding this in validation.go, and this logic was just moved to typefilter.go). Was it broken before?

bcda/fhir/r4/search/date.go
michaeljvaldes: Good catch; I updated the service=date manipulation in bluebutton.go to account for service-dates with time info. We still format it without the time information just before we sent it to BFD -- I need to double-check whether that's necessary.

bcda/client/bluebutton.go
michaeljvaldes: I think domain-centered approach (models are in packages together with the services that are specific to them) makes sense here (especially in we're looking at a future where we further split the api and worker) and I think that that can pretty organically grow out of new work in BCDA without creating any conflict. 
I updated these imports to rename fhir --> fhirClient, and  fhirModels --> fhir

bcda/fhir/r4/search/date.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+		if !valid {
+			return d, ParameterParsingError{Details: fmt.Sprintf("invalid date parameter value: %s", datetime)}
+		}
+	}
+	return d, nil
+}
+
+func ValidateServiceDates(serviceDateParams []DateParam) error {
+	lowerBoundCount, upperBoundCount, equalCount := 0, 0, 0
```
laurenkrugen-navapbc, line 69: nit: can we use naming like olderDateCount, newerDateCount or something similar instead of upper/lower?

bcda/fhir/r4/search/typefilter.go
bhagatparwinder: It more than likely is broken, and we have just never encountered it because  service-date parsing with time was stripping out the date altogether (the above parseDate bug)

bcda/fhir/r4/search/date.go
carlpartridge: As per https://jira.cms.gov/browse/BCDA-10377 we seem to be trying to standardize on earliest/latest.

bcda/fhir/r4/search/date_test.go
```diff
@@ -0,0 +1,197 @@
...
+		// 		{Name: "service-date", Prefix: "lt", Datetimes: []string{"2005"}},
+		// 		{Name: "service-date", Prefix: "gt", Datetimes: []string{"2004"}},
+		// 	},
+		// 	expectedErr: nil,
+		// },
+		// {
+		// 	name:        "invalid multiple OR date times",
+		// 	dateParams:  []DateParam{{Name: "service-date", Datetimes: []string{"2004", "2003"}}},
+		// 	expectedErr: ParameterValidationError{},
+		// },
```
carlpartridge: Are these tests no longer useful or were they just temporarily commented out?

bcda/fhir/r4/search/string.go
michaeljvaldes: Nice catch, I updated it to a token parameter that only has a code (e.g. `_outcome=partial` like `[parameter]=[code]`

bcda/fhir/r4/search/typefilter.go
michaeljvaldes: Updated with some tests that are more general to typefilter, and added a couple more small validations to match.

bcda/fhir/r4/search/date.go
michaeljvaldes: Updated, but also noting that the spec it's based on and links to uses lower/upper bound language.

bcda/fhir/r4/search/date_test.go
michaeljvaldes: Just temporarily commented out!

bcda/api/requests_test.go
michaeljvaldes: I moved all the fhir param logic to a param.go file in the fhir package. I think the references are much more readable now (e.g. fhir.ParseTokenParam).

bcda/fhir/r4/search/typefilter.go
michaeljvaldes: Removed the extra decoding 👍

bcda/client/bluebutton.go
```diff
@@ -436,15 +437,27 @@ func setRestrictiveServiceDateWindow(params *url.Values) {
...
+			parsed, err := fhir.ParseDateString(date[2:])
+			if err == nil {
+				latestDates = append(latestDates, serviceDateVal{prefix: date[:2], date: parsed})
+			}
 		} else if strings.HasPrefix(date, "eq") {
-			earliestDates = append(earliestDates, serviceDateVal{prefix: "ge", date: parseDate(date[2:])})
-			latestDates = append(latestDates, serviceDateVal{prefix: "lt", date: parseLatestDateFromEqualPrefix(date[2:])})
+			parsed, err := fhir.ParseDateString(date[2:])
+			if err == nil {
+				earliestDates = append(earliestDates, serviceDateVal{prefix: "ge", date: parsed})
```
bhagatparwinder, line 452: This is getting so complicated (cause dates are hard 😭), but I think it will return no data when the client passes equality (eq).

When the client passes `service-date=eq2022-01-01T01:01:01Z` or without prefix `service-date=2022-01-01T01:01:01Z`, it will return earliestDates as `ge2022-01-01`

latestDates gets `lt` and the result of  `parseLatestDateFromEqualPrefix` (`lt2022-01-01`)

This is what we end up sending to BFD:

```
service-date=ge2022-01-01&service-date=lt2022-01-01
```

I think on [line 527](https://github.com/CMSgov/bcda-app/pull/1498/changes#diff-93f5a217a0e2685428bb817c4fa6943f235cd0fa61b7a5830c75b7d029d47460R527) we need to add a day.

bcda/client/bluebutton.go
```diff
@@ -523,6 +521,12 @@ func parseLatestDateFromEqualPrefix(date string) time.Time {
 		return parsedDate.AddDate(1, 0, 0)
 	}
 
+	// if an "interval" date format is not specified, default to standard parsing
+	parsedDate, err = fhir.ParseDateString(date)
+	if err == nil {
+		return parsedDate
```
bhagatparwinder:
```suggestion
		return parsedDate.AddDate(0, 0, 1)
```

bcda/fhir/param.go
```diff
@@ -0,0 +1,300 @@
...
+		switch sd.Prefix {
+		case "lt", "le":
+			lateBoundCount = lateBoundCount + 1
+		case "gt", "ge":
+			earlyBoundCount = earlyBoundCount + 1
+		case "eq", "":
+			equalCount = equalCount + 1
+		}
+	}
+	if earlyBoundCount > 1 || lateBoundCount > 1 || equalCount > 1 {
```
bhagatparwinder, line 224: I believe we spoke about tackling this complication in a different ticket.

bcda/fhir/param.go
```diff
@@ -0,0 +1,300 @@
...
+			return TypeFilterSubquery{}, ParameterParsingError{Details: fmt.Sprintf("_typeFilter value missing equals sign: %s", paramPair)}
+		}
+		subqueryParams = append(subqueryParams, TypeFilterSubqueryParam{Name: name, Value: value})
+	}
+	return TypeFilterSubquery{ResourceType: resourceType, QueryParameters: subqueryParams}, nil
+}
+
+func ValidateTypeFilterSubquery(subquery TypeFilterSubquery) error {
+	if subquery.ResourceType != "ExplanationOfBenefit" {
+		return fmt.Errorf("invalid _typeFilter Resource Type (Only EOBs valid): %s", subquery.ResourceType)
```
bhagatparwinder:
```suggestion
		return ParameterValidationError{Details: fmt.Sprintf("invalid _typeFilter Resource Type (Only EOBs valid): %s", subquery.ResourceType)}
```

carlpartridge (dismissed):

bcda/fhir/r4/search/date.go
michaeljvaldes: I'm happy to leave the business logic where it is for now (at the moment, fhir/params.go) and move them when we feel it organically makes sense. I could also see us creating a different validation file, perhaps in the same package, but I'd like to avoid moving stuff back to our middleware if there's a risk of needing to import it in other places like the worker.

bcda/fhir/param_test.go
```diff
@@ -0,0 +1,491 @@
...
+		},
+		{
+			name: "invalid resource type",
+			subquery: TypeFilterSubquery{
+				ResourceType: "Patient",
+				QueryParameters: []TypeFilterSubqueryParam{
+					{Name: "_tag", Value: "https://bluebutton.cms.gov/fhir/CodeSystem/System-Type|SharedSystem"},
+				},
+			},
+			expectedErr: ParameterValidationError{},
```
bhagatparwinder, line 96: In param.go [line 71](https://github.com/CMSgov/bcda-app/pull/1498/changes#diff-3b7bf9f89f3faa9f6a01bf26d7bb3f1ddf6599478abb6c558e3ccb532f6a1d96R71), I requested to change `fmt.Errorf` and use `ParameterValidationError`. However, this unit test is still succeeding without that change.

I think everywhere in this file where we do `assert.ErrorAs(t, err, &tt.expectedErr)`, we need to change it to `assert.IsType(t, tt.expectedErr, err)`

bcda/fhir/param_test.go
```diff
@@ -0,0 +1,491 @@
...
+			if tt.expectedErr != nil {
+				assert.ErrorAs(t, err, &tt.expectedErr)
+			} else {
+				assert.Nil(t, err)
+			}
+		})
+	}
+}
+
+func TestParseStringParam(t *testing.T) {
```
bhagatparwinder, line 122: I think this test suite, `StringParam`, and `ParseStringParam` can now be removed since we changed `outcome` to token.

michaeljvaldes: I'm down to leave them in since they might be helpful in the future. We can always delete them if they become more trouble than they're worth, but they shouldn't break anything in the meantime.

bcda/fhir/param_test.go
michaeljvaldes: Fixed!

bcda/fhir/param.go
michaeljvaldes: Fixed!

bcda/fhir/param.go
michaeljvaldes: Yep, I think further validations for situations where we have, for example, conflicting eq and lt prefixes are out of scope for this ticket.

bcda/client/bluebutton.go
michaeljvaldes: Nice catch! Updated, and I hope we can find a way to simplify this date logic in the future.

bhagatparwinder (dismissed):

bcda/fhir/param.go
```diff
@@ -0,0 +1,300 @@
...
+
+type ParameterValidationError struct {
+	Details string
+}
+
+func (e ParameterValidationError) Error() string {
+	return fmt.Sprintf("invalid parameter: %s", e.Details)
+}
+
+func ParseTypeFilterSubquery(s string) (TypeFilterSubquery, error) {
```
michaeljvaldes, line 46: This file contains the core of the changes made in this PR. Here are structs that represent different params including typefilter and its subquery param _types_ (strings, tokens, and dates) and validations for implementations of those types (_tag, outcome, and service-date)

carlpartridge (dismissed):

bcda/web/middleware/validation.go
```diff
@@ -188,119 +186,41 @@ func validateTypeFilterParameter(r *http.Request, rw fhirResponseWriter, w http.
...
 
-func HasSharedSystemTag(typeFilter fhir.TypeFilterParameter) bool {
-	for _, subqueryParam := range typeFilter.QueryParameters {
-		if subqueryParam.Name == "_tag" {
-			tagSystems := ExtractTagSystemFromValue(subqueryParam.Value)
-			for _, tagSystem := range tagSystems {
-				if tagSystem == constants.BFDSystemTypeURL {
-					return true
-				}
+func HasSharedSystemTag(subquery fhir.TypeFilterSubquery) bool {
```
colby-seyferth-nava, line 212: nit: I think it was already named this, but i think this is not quite the right name. 

it looks like this function checks to see if there is a _tag param value (token) specified in _typeFilter that has the BFD System-Type url as the token.system.

Should it be named `HasSystemTypeFilter` or something?

bcda/fhir/param.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+var validTagTokens = map[string][]string{
+	constants.BFDSystemTypeURL:  {"SharedSystem", "NationalClaimsHistory", "DDPS"},
+	constants.BFDFinalActionURL: {"FinalAction", "NotFinalAction"},
+}
+
+func ValidateTag(t TokenParam) error {
+	if t.Name != string(TypeFilterParamTag) {
+		return ParameterValidationError{Details: fmt.Sprintf("invalid key for tag parameter: %s", t.Name)}
```
colby-seyferth-nava, line 270: nit: should be `_tag` instead of `tag`?

bcda/fhir/param.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	if len(t.Modifier) > 0 {
+		return ParameterValidationError{Details: fmt.Sprintf("invalid _tag parameter; modifier %s not supported", t.Modifier)}
+	}
+	for _, value := range t.Values {
+		if len(value.System) == 0 || len(value.Code) == 0 {
+			return ParameterValidationError{Details: "invalid _tag parameter value. Searching by tag requires a token (system|code) to be specified"}
```
colby-seyferth-nava, line 277: nit: should be `_tag` for both instead of `tag` on the second one?

bcda/fhir/param.go
```diff
@@ -0,0 +1,300 @@
...
+		if len(sd.Datetimes) > 1 {
+			return ParameterValidationError{Details: "invalid service-date parameter value: comma-separated values are not supported"}
+		}
+		switch sd.Prefix {
+		case "lt", "le":
+			lateBoundCount = lateBoundCount + 1
+		case "gt", "ge":
+			earlyBoundCount = earlyBoundCount + 1
+		case "eq", "":
+			equalCount = equalCount + 1
```
colby-seyferth-nava, line 221: equal prefix is a range, that would add +1 to both the early bound and late bound counts

colby-seyferth-nava: ope, i see in the below resolved comment that this is out of scope for this ticket? feel free to disregard

colby-seyferth-nava (dismissed):
A couple small nit-pick comments:
 - function name is confusing/misleading
 - a couple error texts with potential typo 

And one that looks like it is just a follow-up for next time. (complex `eq` + `lt` query validation)

Otherwise LGTM, so I'd be fine with merge as is or with the couple small nits addressed

bcda/fhir/param.go
michaeljvaldes: Fixed!

bcda/fhir/param.go
michaeljvaldes: Fixed!

bcda/web/middleware/validation.go
michaeljvaldes: Updated to HasSystemTypeTag 👍

bcda/fhir/r4/search/date.go
michaeljvaldes: I've talked myself back in to adding this. The limitation of this condition is that if there is something like "eq2005&lt2006" that technically is not conflicting, but I supposed it doesn't make sense for a user to send that anyway since simply sending "eq2005" would suffice.

bcda/fhir/param.go
michaeljvaldes: I've talked myself back in to adding this. The limitation of this condition is that if there is something like "eq2005&lt2006" that technically is not conflicting, but I supposed it doesn't make sense for a user to send that anyway since simply sending "eq2005" would suffice.
If we think that there should be a more nuanced approach to this validation, I'd push it to another ticket.

bhagatparwinder (approved):

carlpartridge (approved):

bcda/fhir/param.go
colby-seyferth-nava: Yeah, the HAPI FHIR server would fails on this condition, so I think it's fair for us to fail.

colby-seyferth-nava (approved):

PR #1500, opened 2026-09-21
BCDA-10416: add check for jobs that have been in a pending state more than 24 hours

laurenkrugen-navapbc:
## 🎫 Ticket

https://jira.cms.gov/browse/BCDA-10416

## 🛠 Changes

- mockery cli tool auto updates to existing mocks (required to add new mocks)
- updated some error messages in `health.go` to use lowercase on starting character
- added an extra check to our existing health check to poll the database for jobs that have been in a pending state for more than 24 hours

## ℹ️ Context

Historically, there have been times when job processing has halted or failed, but not met particular monitoring conditions to alert us to failure. This is very rare but still possible, so a health check to see if we have jobs in the `pending` state that are more than 24 hours old will notify us if a silent failure occurs. Jobs are only valid for 24 hours, so querying for `pending` jobs that are 24 hours old will alert us if they aren't processing fast enough (or at all) without falsely notifying us of jobs that are just processing slowly.

This is also part of deprecating the existing worker health alert.

## 🧪 Validation

New integration test added; tests passing.


bcda/models/postgres/repository.go
```diff
@@ -435,18 +435,40 @@ func (r *Repository) GetJobs(ctx context.Context, acoID uuid.UUID, statuses ...m
 
 }
 
-func (r *Repository) GetJobsByUpdateTimeAndStatus(ctx context.Context, lowerBound, upperBound time.Time, statuses ...models.JobStatus) ([]*models.Job, error) {
+func (r *Repository) GetJobsByUpdateTimeAndStatus(ctx context.Context, olderTime, newerTime time.Time, statuses ...models.JobStatus) ([]*models.Job, error) {
```
carlpartridge: Can we go with earliest/latest as per https://jira.cms.gov/browse/BCDA-10377?

bcda/health/health.go
```diff
@@ -90,12 +96,26 @@ func (h healthCheck) IsSsasOK() (result string, ok bool) {
...
-		log.API.Error("Health check: ssas health check error: ", err.Error())
+		log.API.Error("health check: ssas health check error: ", err.Error())
 		return "Cannot connect to SSAS", false
 	}
 	return "ok", true
 }
 
+// jobs in a pending state for more than 24hrs are an indication of a silent failure of job processing
+func (h healthCheck) IsJobQueueOK() (ok bool) {
+	jobs, err := h.r.GetJobsByCreateTimeAndStatus(context.Background(), time.Time{}, time.Now().Add(-24*time.Hour), models.JobStatusPending)
```
carlpartridge: What happens at the 24 hour mark of a pending job?  Does it get cleaned up like the others or does it just sit there?

bcda/health/health.go
```diff
@@ -90,12 +96,26 @@ func (h healthCheck) IsSsasOK() (result string, ok bool) {
...
+	if err != nil {
+		log.API.Errorf("health check: pending jobs query returned errors. err: %s", err)
+		return false
+	}
+	if len(jobs) > 0 {
+		return false
+	}
+
+	return true
```
carlpartridge: Nit: I think we could do `return len(jobs) < 1`, would save a few lines, and I think its as readable.

bcdaworker/cli/cli.go
```diff
@@ -161,6 +161,13 @@ func checkHealth(healthChecker health.HealthChecker) bool {
...
+	jobQueueOK := healthChecker.IsJobQueueOK()
+	if jobQueueOK {
+		logFields["jobqueue"] = "ok"
+	} else {
+		logFields["jobqueue"] = "error"
+	}
+
 	entry.WithFields(logFields).Info()
-	return dbOk && bbOk
+	return dbOk && bbOk && jobQueueOK
```
carlpartridge: Im wondering how useful this is in the worker health check.  Like if this started to fail it would constantly rotate worker instances until something happened to this record right?  Since there is nothing that actually updates this record it would just permacycle, or even worse, permafail worker instances, preventing any work at all?  Workers also already scale in and out quite often which if there happens to be just a singular wonky instance gets cleaned up pretty quickly.  I kind of feel like this check is overall out of the scope of the health check as such and therefore should happen elsewhere.  At the same time the health check is a pretty convenient place to put a regular check.  Im thinking we could either create a secondary ping check of some kind that runs maybe less often than the health check or have the health check report this data (as success and failures all report to splunk), then use splunk to pick up on it if this specific check is failing.

I think if we are to go with this approach we could also add in some information about failed jobs, or other DB info that could be useful to ping devs about.  We could also expand the response from just true/false (ok/not ok) to include the job ID, aco id, other info?  Thinking about this some more, but a lot of the goal here is to come up with a way to get data that we can then monitor that we cant easily get out of splunk/aws/datadog.  Another option could be to log this data to cloudwatch (?) like we do with the JobQueueCount metrics, then set up alarms based on that.

bcda/health/testdata/pending_jobs_single.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO jobs (aco_id, request_url, status, transaction_time, job_count, created_at, updated_at, benes_attributed_to_aco)
+VALUES ('A40404F7-1EF2-485A-9B71-40FE7ACDCBC2', 'integration test', 'Pending', NOW(), 1, NOW() - INTERVAL '24 HOURS', NOW(), 1)
```
carlpartridge: Whats the advantage of the two different test sql files?

bcdaworker/cli/cli.go
laurenkrugen-navapbc: I had similar thoughts as you but ultimately included it in our existing health check for a few reasons:

- the job processing check fits the same setup as the SSAS/db/worker check, (poll periodically and send a log if something is wrong). to keep things consistent, I included it there instead of making an entire new check.
- I also wanted to keep this more frequent because if jobs are stuck in pending and then expire, this would hopefully keep us right in front the issue and we can address quickly.
- if jobs aren't being processed, rotating the instances/checking the instances is something I would do first, troubleshooting-wise, especially if this was occurring after a deployment had been out for a while.

Ultimately, the goal was to use the existing health check that gets logged to keep the code consisten, and then create a new splunk alert that notifies us of jobs stuck in a pending state. 

I wasn't planning on putting any details in this check because having things like IDs, etc, won't give us much context until we look in the database, which will just give us that information anyway.

bcda/health/testdata/pending_jobs_single.sql
laurenkrugen-navapbc: the test container loads data into the data during the runs; test data files can be created on a per-test basis.

bcda/health/health.go
laurenkrugen-navapbc: Nothing - the job will remain in the pending status until is it cancelled or completed.

bcdaworker/cli/cli.go
carlpartridge: I guess my main thought/concern is that restarting instances is very unlikely to fix the issue, at least after 24hrs.  And I think a lot of that comes down to the fact that we would have 24hrs of restarts already likely happening, and if it hasnt fixed itself by then, then its very unlikely to fix it now.  And going into perma-cycle/failure seems like a bad fix.  I think another part of the problem is that even if we were to cycle the service, and it fixes itself, the health check might ping the job status before the worker can even actually pick up the job and start working on it.  This would definitely lead to a perma-dead cycle.  

I think if we want to keep this as part of the health check I would want a way to allow the system to fix itself (besides just processing the job again).  And by that I mean update the jobs status to something besides pending, at least temporarily.

Im thinking the scenarios where this approach would be useful are:
- very, very little traffic and therefore no other instance cycling happening over a long period of time.
- a smaller timeframe (maybe if a job is stuck in pending for 6 hours?)
- more reliably self correcting (to prevent death cycle)

Overall though I think this check is more useful as a way to alert a dev to look into it and as such, would be best sent off as a metric to be picked up and alerted off from splunk, aws, or datadog.

But to be fair my opinion is also very set by my thought that this is unlikely to resolve itself so curious how others feel.

bcda/models/postgres/repository.go
laurenkrugen-navapbc: earliest = older and latest = newer, correct?

bcdaworker/cli/cli.go
laurenkrugen-navapbc: Yeah, I also wouldn't expect the rotation to fix the issue but similar to the database and SSAS check, I wouldn't expect a rotation to fix those either, so maybe we need to have a larger conversation about rotating the instance when those health check fail.

In the case of SSAS, the database, or the job processing to not be working, the API is now in a level of degradation. To your point, it could get to a place where it's constantly restarting until the job is taken out of the pending state - I do not think we should put those jobs in another state temporarily. I think we should leave them and allow them to process without modifying our job cycle.

> But to be fair my opinion is also very set by my thought that this is unlikely to resolve itself so curious how others feel.

I agree with this - I think all three health checks won't be resolved by a rotation, but it is helpful to be notified. Can we just update the rotation aspect of this?

My hope by adding this to the existing health check was to not have almost duplicate code running a similar check for a different aspect of the service, but if we want to keep the rotation of the instances on failure, maybe that's the best path forward.

laurenkrugen-navapbc: Or like you said, we can just modify the existing check and not return `false` if pending fails and just log it instead. that way, we can use a splunk alert to notify without rotation.

carlpartridge: Hmmm fair point.  I think I disagree to some degree.  If the DB goes down for a bit this will continue to fail until the DB comes back up, at which point it will start working again.  This assumes the DB just comes back up, but I think thats a lot of AWSs job 😅.  Somewhat similar with SSAS, SSAS can go down for a bit, health starts to fail, then at some point they both correct and come back up automatically.  Can even see there being a cred rotation, worker cant reach SSAS/BFD, goes down, restarts with new creds, and we are back up and running.

I think the big thing with all of these failures is that A) we want the system to try to correct itself if possible and without intervention and B) we need to alert an engineer about it (at least after a bit of time to allow it try to autocorrect).

I also agree that modifying job work flow is not a good approach.  But I do like maybe checking after 6 hours of pending, as that already seems like a pretty big indication that something is wrong.  I dont think we need to wait 24 full hours.

I think we can keep this as part of the health check in the sense that it is logging useful metrics to splunk about pending jobs, and we continue to alert devs based off these splunk logs, but that the health check doesnt actually fail due to just the job pending check.

laurenkrugen-navapbc: > Hmmm fair point. I think I disagree to some degree. If the DB goes down for a bit this will continue to fail until the DB comes back up, at which point it will start working again. This assumes the DB just comes back up, but I think thats a lot of AWSs job 😅. Somewhat similar with SSAS, SSAS can go down for a bit, health starts to fail, then at some point they both correct and come back up automatically. Can even see there being a cred rotation, worker cant reach SSAS/BFD, goes down, restarts with new creds, and we are back up and running.

It's also possible for the database and ssas to fully be unavailable due to something other than an intermittent loss of connectivity, including our upsteam partner. For temporary things like changing env variables or rotations, that makes sense but I think counting on an auto rotation can put us in a precarious position if something is really wrong?

> I also agree that modifying job work flow is not a good approach. But I do like maybe checking after 6 hours of pending, as that already seems like a pretty big indication that something is wrong. I dont think we need to wait 24 full hours.

Re: checking at 6 hours instead of 24; I picked 24 because our job processing for larger jobs can take some time and the goal of this new worker alert was to eliminate excess noise/false positives. I would push in favor of something more lenient, like 12 hours. That reduces false positives and allows flexibility in extra traffic, especially in the near future.

> I think we can keep this as part of the health check in the sense that it is logging useful metrics to splunk about pending jobs, and we continue to alert devs based off these splunk logs, but that the health check doesnt actually fail due to just the job pending check.

I updated the code so it only logs the status, but doesn't include it in the overall check that would cause a rotation.

carlpartridge: > Re: checking at 6 hours instead of 24; I picked 24 because our job processing for larger jobs can take some time and the goal of this new worker alert was to eliminate excess noise/false positives. I would push in favor of something more lenient, like 12 hours. That reduces false positives and allows flexibility in extra traffic, especially in the near future.

Job status in pending means it hasnt even started yet right?  Once it has started it gets moved to inProgress.  So I think a job that hasnt even started after 6 hours is a pretty conservative indicator?

bcda/health/health.go
```diff
@@ -90,12 +96,22 @@ func (h healthCheck) IsSsasOK() (result string, ok bool) {
...
 }
 
+// jobs in a pending state for more than 6hrs are an indication of a silent failure of job processing
+func (h healthCheck) IsJobQueueOK() bool {
+	jobs, err := h.r.GetJobsByCreateTimeAndStatus(context.Background(), time.Time{}, time.Now().Add(-6*time.Hour), models.JobStatusPending)
+	if err != nil {
+		log.API.Errorf("health check: pending jobs query returned errors. err: %s", err)
+		return false
+	}
+	return len(jobs) < 1
```
carlpartridge: I think I would still like this to give more info but that is out of scope of this ticket.  Just thinking out loud I think it would be useful to maybe report a few metrics:
- jobs stuck in pending, useful data: job id, aco id, time in pending?
- failed jobs, data: job id, aco id, when failed?

carlpartridge (dismissed):

bcda/health/health.go
```diff
@@ -90,12 +96,22 @@ func (h healthCheck) IsSsasOK() (result string, ok bool) {
...
-		log.API.Error("Health check: ssas health check error: ", err.Error())
+		log.API.Error("health check: ssas health check error: ", err.Error())
 		return "Cannot connect to SSAS", false
 	}
 	return "ok", true
 }
 
+// jobs in a pending state for more than 6hrs are an indication of a silent failure of job processing
+func (h healthCheck) IsJobQueueOK() bool {
+	jobs, err := h.r.GetJobsByCreateTimeAndStatus(context.Background(), time.Time{}, time.Now().Add(-6*time.Hour), models.JobStatusPending)
```
bhagatparwinder, line 110: I'm slightly confused because the PR description mentioned 24 hours, and our tests also say

> multiple pending jobs returned older than 24 hours
> one pending job older than 24 hours

But I see 6 hours in the code.

bcda/health/health.go
```diff
@@ -90,12 +96,22 @@ func (h healthCheck) IsSsasOK() (result string, ok bool) {
...
+	if err != nil {
+		log.API.Errorf("health check: pending jobs query returned errors. err: %s", err)
```
bhagatparwinder: Should this be `log.Worker`?

bcda/models/postgres/repository.go
```diff
@@ -435,18 +435,41 @@ func (r *Repository) GetJobs(ctx context.Context, acoID uuid.UUID, statuses ...m
...
+
+	if len(s) > 0 {
+		sb.Where(sb.In("status", s...))
+	}
+
+	query, args := sb.Build()
+	return r.getJobs(ctx, query, args...)
+}
+
+func (r *Repository) GetJobsByCreateTimeAndStatus(ctx context.Context, olderTime, newerTime time.Time, statuses ...models.JobStatus) ([]*models.Job, error) {
```
bhagatparwinder, line 461: Curious if we need all the jobs since the check only cares whether at least one job exists. Could we probably add `LIMIT 1` or get `COUNT`?

bcda/health/health_test.go
```diff
@@ -36,6 +44,18 @@ func (s *HealthCheckerTestSuite) SetupSuite() {
 	origSSASUseTLS = conf.GetEnv("SSAS_USE_TLS")
 	origSSASClientID = conf.GetEnv("BCDA_SSAS_CLIENT_ID")
 	origSSASSecret = conf.GetEnv("BCDA_SSAS_SECRET")
+
+	var err error
+	s.dbContainer, err = db.NewTestDatabaseContainer()
```
bhagatparwinder, line 49: I'm thinking this from a load on unit tests perspective. We had tests without a Docker container earlier. Since healthCheck accepts Repository, we could use `mock_repository.go` to test IsJobQueueOK without Docker dependencies?

bcda/models/postgres/repository.go
bhagatparwinder: Also, do we have any tests for this?

bcdaworker/cli/cli.go
```diff
@@ -161,6 +161,13 @@ func checkHealth(healthChecker health.HealthChecker) bool {
...
 
+	jobQueueOK := healthChecker.IsJobQueueOK()
+	if jobQueueOK {
+		logFields["jobqueue"] = "ok"
+	} else {
+		logFields["jobqueue"] = "error"
+	}
+
 	entry.WithFields(logFields).Info()
 	return dbOk && bbOk
```
bhagatparwinder, line 176: I see the extensive conversation between you and @carlpartridge about removing the jobQueueOK check. I do think we should:

1. Add a comment here that excluding `jobQueueOK` is intentional. We don't want future devs to think this was a miss on our part (and accidentally add it back in).
2. Test that the health check does not fail when `jobQueueOK` is false.

bcda/health/health.go
bhagatparwinder: Ignore this, please, because I see the conversation between you and Carl. However, please update the PR description and the test cases.

bcdaworker/cli/cli.go
bhagatparwinder: This is a great discussion! There is only one place where I think differently.

> I wasn't planning on putting any details in this check because having things like IDs, etc, won't give us much context until we look in the database

Logging `len(jobs)` and the oldest `jobs[0].ID / jobs[0].CreatedAt` takes no extra overhead and immediately tells the engineer on call whether 1 job is stuck or 5,000 jobs are stuck, and which job ID to inspect.

bcda/health/health.go
laurenkrugen-navapbc: test cases should be updated - I will double check and also update the PR description.

bcda/models/postgres/repository.go
laurenkrugen-navapbc: not for just this query, but we do have an integration test for it. I can add a test for this function, if desired? 

> Curious if we need all the jobs since the check only cares whether at least one job exists. Could we probably add LIMIT 1 or get COUNT?

and 

> Logging len(jobs) and the oldest jobs[0].ID / jobs[0].CreatedAt takes no extra overhead and immediately tells the engineer on call whether 1 job is stuck or 5,000 jobs are stuck, and which job ID to inspect.

contradict each other, should I disregard your first comment?

bcda/health/health_test.go
laurenkrugen-navapbc: I would prefer to leave this one in, if that's OK? I initially had a mock test in there and upon adding this one, discovered a small issue I had originally missed. The overhead on this container is pretty small, maybe an extra 1-2 seconds to initially configure.

bcdaworker/cli/cli.go
laurenkrugen-navapbc: yup, can do - originally had a test with jobqueueOK and removed it but should have kept it in.

bcda/health/health.go
laurenkrugen-navapbc: job failures I was going to address separately in splunk, since we log those failures. 

do you want the details of _all_ the jobs in pending after six hours? that might be a very lengthy message, so I would be more in favor of implementing what Parwinder mentioned of just a count and the oldest pending job.

bcda/models/postgres/repository.go
bhagatparwinder: If we end up logging the job length and job ID, then I would suggest not doing this (`LIMIT 1` or getting `COUNT`).

I am fine if there is an integration test, but I know SQ usually has a threshold for unit tests.

bcda/health/health.go
bhagatparwinder: These tests: https://github.com/CMSgov/bcda-app/pull/1500/changes#diff-735495e43a6e5eadd7cfc0c4f1e42aee469d9df9fd8ba15ee04e3625e22b3f13R351

bhagatparwinder (approved):

laurenkrugen-navapbc:
SQ is failing due to duplicate code in a generated file from `mockery`. I will not be addressing this since it's a generated file, and the duplicate code is necessary for the mocks. No risk associated with this.

carlpartridge (approved):

